PR #1, opened 2020-08-05
Initial commit

richtier:


conf/base.py
```diff
@@ -0,0 +1,108 @@
...
+    {
+        'NAME': 'django.contrib.auth.password_validation.NumericPasswordValidator',
+    },
+]
+
+
+# Internationalization
+# https://docs.djangoproject.com/en/3.0/topics/i18n/
+
+LANGUAGE_CODE = 'en-us'
```
saruniitr, line 94: As we are just discussing about this, may be change it `en-gb` similar to api?

conf/base.py
```diff
@@ -0,0 +1,108 @@
...
+
+USE_L10N = True
+
+USE_TZ = True
+
+
+# Static files (CSS, JavaScript, Images)
+# https://docs.djangoproject.com/en/3.0/howto/static-files/
+
+STATIC_URL = '/static/'
```
r4vi, line 108: missing new line

conf/base.py
richtier: sure. copy and pasted :)

PR #9, opened 2020-08-07
Fix browser tests

richtier:


r4vi:
the changes look sensible but I haven't run it locally

PR #11, opened 2020-08-12
Ltd-136: Configure timezone correctly

saruniitr:
## Change description

Timezone is set to "Europe/London" and all references to datetime are updated to use localtime as per that timezone.

### DIFF
--- a/caseworker/cases/helpers/case.py
+++ b/caseworker/cases/helpers/case.py
@@ -91,7 +91,7 @@ class CaseView(TemplateView):
         future_next_review_date = (
             True
             if self.case.next_review_date
-            and datetime.datetime.strptime(self.case.next_review_date, "%Y-%m-%d").date() > timezone.now().date()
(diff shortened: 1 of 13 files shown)


caseworker/cases/helpers/case.py
```diff
@@ -91,7 +91,7 @@ def get_context(self):
         future_next_review_date = (
             True
             if self.case.next_review_date
-            and datetime.datetime.strptime(self.case.next_review_date, "%Y-%m-%d").date() > timezone.now().date()
+            and datetime.datetime.strptime(self.case.next_review_date, "%Y-%m-%d").date() > timezone.localtime().date()
```
kerin, line 94: This might be something for a future PR, but would it not be better to define a property for `Case.next_review_date` that returns a `datetime`?

saruniitr: can be considered in future

PR #12, opened 2020-08-13
Small fixes for Jenkins to CloudFoundry deployment

r4vi:
* make Directory sso env vars optional
* move Procfile to top level
* add whitenoise to wsgi.conf


conf/gconfig.py
```diff
@@ -1,4 +1,7 @@
-bind = "0.0.0.0:8080"
+import os
+
+PORT = os.environ.get("PORT", 8080)
+bind = "0.0.0.0:{PORT}".format(PORT=PORT)
```
saruniitr, line 4: minor, may be replace with `f"0.0.0.0:{PORT}"`

Procfile
```diff
@@ -0,0 +1 @@
+web: ./manage.py migrate && ./manage.py compilescss && ./manage.py collectstatic --ignore=*.scss,*.md,*.txt,*.json,LICENSE,license,CHANGES,changes && gunicorn --preload -c conf/gconfig.py -b 0.0.0.0:$PORT conf.wsgi
```
kerin, line 1: Are we leaving sorting this out for a future PR?

r4vi: yes, just want to get it working now

conf/gconfig.py
r4vi: not sure we can use f-strings as I'm not sure if the project is python 3.6 or 3.7

kerin: `runtime.txt` sez `python-3.6.9`

r4vi: pipfile says 3.7 but either way I should've used f-strings since they came in with 3.6 :disappointed:

kerin: oh yes, having the same python version specified in runtime.txt and Pipfile is on the tech debt list - worth trying bumping everything to 3.7.x latest?

PR #17, opened 2020-08-14
LTD-136: Fix parsing of datetime in template tags

saruniitr:
## Change description

The datetimes use current timezone by default (always aware) and the format
is failing to parse this resulting is no time displayed in UI.

### DIFF
--- a/core/builtins/custom_tags.py
+++ b/core/builtins/custom_tags.py
@@ -7,16 +7,16 @@ import re
 from collections import Counter, OrderedDict
 from html import escape
 
+from dateutil.parser import parse
 from dateutil.relativedelta import relativedelta
 
 from django import template
+from django.conf import settings
 from django.template.defaultfilters import stringfilter, safe, capfirst
-from django.templatetags.tz import do_timezone
+from django.templatetags.tz import localtime
 from django.utils.safestring import mark_safe
 
-from django.conf import settings
 from exporter.conf.constants import (
-    ISO8601_FMT,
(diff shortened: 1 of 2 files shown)


core/builtins/custom_tags.py
```diff
@@ -97,7 +98,7 @@ def pluralize_lcs(items, string):
 @stringfilter
 def str_date(value):
     try:
-        return_value = do_timezone(datetime.datetime.strptime(value, ISO8601_FMT), "Europe/London")
+        return_value = do_timezone(parse(value), settings.TIME_ZONE)
```
r4vi, line 101: I would maybe use `django.templatetags.tz.localtime` as the name is more explicit about what we're trying to do and it avoids having to use `settings.TIME_ZONE`. Up to you if you want to change that - approving anyway

core/builtins/custom_tags.py
```diff
@@ -7,16 +7,17 @@
 from collections import Counter, OrderedDict
 from html import escape
 
+from dateutil.parser import parse
 from dateutil.relativedelta import relativedelta
 
 from django import template
+from django.conf import settings
```
r4vi, line 14: already imported

saruniitr: yes, linter also complained, pushed changes

core/builtins/custom_tags.py
saruniitr: agree that is better, will update this.

PR #19, opened 2020-08-18
Caseworker: Remove Authentication middleware

r4vi:
This removes auth middleware and auth backends because they were the only thing
that relied on a database.

I have moved all of the "user" state on to request.session but will continue to
investigate a good way of abstracting this. Possibly by adding a new middleware
that puts something on `request.user` but constructs that object from the
session rather than the database.


r4vi:
since this is related to security would appreciate close :eyes: on this one :pray:

core/auth/views.py
```diff
@@ -0,0 +1,41 @@
...
+
+        return authorization_url
+
+
+class AuthLogoutView(RedirectView):
+    permanent = False
+
+    def get_redirect_url(self, *args, **kwargs):
+        self.request.session.flush()
+        return settings.LOGOUT_URL + self.request.build_absolute_uri("/")
```
richtier, line 41: I think `self.request.build_absolute_uri(settings.LOGOUT_URL)` does the same thing

r4vi: settings.LOGOUT_URL is something like https://sso/logout/?next=
so we want an absolute uri for our own site to go into the `?next` query param rather than an absolute uri to `settings.LOGOUT_URL`

PR #22, opened 2020-08-23
Remove duplicate code

richtier:


core/auth/middleware.py
```diff
@@ -0,0 +1,20 @@
...
+request.authbroker_client.authorized
+
+
+def get_client(request):
+    if not hasattr(request, "_cached_client"):
+        request._cached_client = get_client(request)
+    return request._cached_client
+
+
+class AuthbrokerClientMiddleware:
```
richtier, line 15: inspired by by how [django does request.user](https://github.com/django/django/blob/b3124215117541a86b0740314ef991e1d521da69/django/contrib/auth/middleware.py#L15)

caseworker/cases/views/advice.py
```diff
@@ -46,8 +46,10 @@
 from lite_forms.generators import form_page, error_page
 from lite_forms.views import SingleFormView
 
+from core.auth.views import LoginRequiredMixin
 
-class GiveAdvice(SingleFormView):
+
+class GiveAdvice(LoginRequiredMixin, SingleFormView):
```
r4vi, line 52: I like being explicit with LoginRequiredMixin and removing ProtectAllViews middleware but scared that now we could forget to put the Mixin on a view... 

Go with this change but I'd like to think about a pre-commit hook or something that can check this in the future

conf/base.py
```diff
@@ -30,11 +30,30 @@
 WSGI_APPLICATION = "conf.wsgi.application"
 
 INSTALLED_APPS = [
+    "django.contrib.contenttypes",
+    "django.contrib.auth",
```
r4vi, line 34: do we need `"django.contrib.contenttypes",
    "django.contrib.auth",` anymore?

richtier: the module `django.contrib.auth.mixins.UserPassesTestMixin` lives in also calls `get_user_model` - which fails if `django.contrib.auth` is not in INSTALLED_APPS

I can copy and paste and strip down `UserPassesTestMixin` and remove the import

core/client.py
```diff
@@ -11,81 +10,86 @@
 
 
 def get(request, appended_address):
+    session = request.requests_session  # provided by RequestsSessionMiddleware
     url = _build_absolute_uri(appended_address.replace(" ", "%20"))
```
r4vi, line 14: you could get rid of a lot of duplication here by moving all the boiler plate into a method called 
```
def request(method, request, appended_address, data=None):
    session = request.requests_session
    url = _build_absolute_uri(appended_address.replace(" ", "%20"))

    if settings.HAWK_AUTHENTICATION_ENABLED:
        sender = _get_hawk_sender(url, "GET", "application/json", "")

        response = session.request(method=method, url=url, headers=_get_headers(request, sender))

        _verify_hawk_response(response, sender)
    else:
        response = requests.get(url=url, headers=_get_headers(request, content_type="application/json"))

    _verify_response_code(response)

    return response
from functools import partial
get = partial('get', request)
post = partial('post', request)
...

```

r4vi (approved):
added some suggestions but feel free to ignore

caseworker/cases/views/advice.py
richtier: I've written a pylint plugin that checks for us when `prospector` (which runs pylint). results in:
```
exporter/core/views.py:42: [W1001(open-to-anon), Home] Home is open to anonymous users. Either use LoginRequiredMixin on the view or exclude it from this check
exporter/core/views.py:98: [W1001(open-to-anon), RegisterAnOrganisationTriage] RegisterAnOrganisationTriage is open to anonymous users. Either use LoginRequiredMixin on the view or exclude it from this check
exporter/core/views.py:126: [W1001(open-to-anon), RegisterAnOrganisation] RegisterAnOrganisation is open to anonymous users. Either use LoginRequiredMixin on the view or exclude it from this check
exporter/core/views.py:175: [W1001(open-to-anon), RegisterAnOrganisationConfirmation] RegisterAnOrganisationConfirmation is open to anonymous users. Either use LoginRequiredMixin on the view or exclude it from this check
```

r4vi: great stuff, I didn't mean it had to be part of this PR but that is nice :ok_hand:

PR #24, opened 2020-08-26
LTD-126 Running unified frontend with Docker

r4vi:
- New docker file
- New docker compose file that can be on the same shared network as `lite-api`
- Instructions for how to run with `docker-compose`


saruniitr (approved):
lgtm

r4vi:
@saruniitr updated as per your comments and backed out the change in `Procfile` to remove the `pip3 install endsive` - fixing that can be part of another ticket.

PR #25, opened 2020-09-03
Omni search

richtier:


caseworker/search/views.py
```diff
@@ -0,0 +1,52 @@
...
+
+        context["results"] = results
+        context["data"] = {"total_pages": results["count"] // form.page_size}
+        return context
+
+
+class AutocompleteView(views.APIView):
+    authentication_classes = []
+    permission_classes = [
+        permissions.AllowAny,
```
r4vi, line 39: shouldn't we check they're a logged in user?

r4vi (approved):
looks good, just some files missing newlines

caseworker/search/views.py
r4vi: maybe auth for all of this is something to resolve later - potentially sending ES queries through the api layer too

caseworker/assets/css/components/_search.scss
```diff
@@ -0,0 +1,26 @@
+body {
+  #autoComplete_list {
```
eadpearce, line 2: can we stick to snake case in css class names pls and thank. you'll need to change them in the JS selectors as well

caseworker/templates/search/search.html
```diff
@@ -0,0 +1,141 @@
...
+                                <span class="govuk-visually-hidden">View</span> {{ item.id }}
+                            </a>
+                            <p class="govuk-body govuk-!-margin-top-2 govuk-!-margin-bottom-0">
+                                {{ item.organisation }}
+                            </p>
+                            <p class="govuk-tag govuk-tag--grey govuk-!-margin-0 govuk-!-margin-top-2">
+                                {{ item.status }}
+                            </p>
+                            <p>
+                                <details>
```
eadpearce, line 62: details tag inside a p is probably redundant. if you're using the p for extra margin just use one of the govuk margin classes

eadpearce (dismissed):
small stuff

caseworker/assets/css/components/_search.scss
eadpearce: (if this is a fault of the library we're using i guess i'll just have to direct my frustration at them instead)

saruniitr: These are only used in this file, changing their names to `autoCompleteList` affecting the styling, cannot find JS selectors

caseworker/templates/search/search.html
saruniitr: fixed

caseworker/search/views.py
richtier: :+1: this was to get up and running quickly for dev. will add auth check

caseworker/assets/css/components/_search.scss
richtier: that is indeed defined by the js lib and it's hard-coded too https://github.com/TarekRaafat/autoComplete.js/search?q=autoComplete_list&unscoped_q=autoComplete_list

PR #27, opened 2020-09-08
Improve UI of autocomplete

richtier:


caseworker/assets/javascripts/search-cases.js
```diff
@@ -53,14 +56,14 @@
     searchEngine: function(query, record) {
         return record
     },
-    maxResults: 5,                         // Max. number of rendered results | (Optional)
+    maxResults: 5,
```
saruniitr, line 59: may be we want more than 5, while searching for CLC rating (`ML7b1`) the matching entry was not shown untill I type `ML7b`

saruniitr (approved):

PR #31, opened 2020-09-09
Fix highlighting

richtier:


caseworker/search/helpers.py
```diff
@@ -16,7 +16,13 @@ def highlight_results(results):
...
-        for level in level[key]:
-            yield from keypath_lookup(level, keys[i + 1 :])
+        if isinstance(level[key], list):
+            for level in level[key]:
+                yield from keypath_lookup(level, keys[i + 1 :])
+            else:
+                break
+        else:
+            yield from keypath_lookup(level[key], keys[i + 1 :])
+            break
```
saruniitr, line 26: `break` may not be necessary here?

saruniitr (approved):
one question otherwise lgtm

caseworker/search/helpers.py
richtier: afraid not. tests fail :)

PR #32, opened 2020-09-10
Configure CSP_INCLUDE_NONCE_IN env variable for Exporter FE

saruniitr:


r4vi (dismissed):

conf/base.py
```diff
@@ -184,6 +184,7 @@
 CSP_SCRIPT_SRC = env.tuple("CSP_SCRIPT_SRC", default=("'self'",))
 CSP_FONT_SRC = env.tuple("CSP_FONT_SRC", default=("'self'",))
 CSP_REPORT_ONLY = env.bool("CSP_REPORT_ONLY", False)
+CSP_INCLUDE_NONCE_IN = env.tuple("CSP_INCLUDE_NONCE_IN", default=("'script-src'",))
```
r4vi, line 187: I don't think it needs to be single quoted like `'self'` the env var that's set on exporter is currently just `CSP_INCLUDE_NONCE_IN="script-src,"`

r4vi: https://github.com/mozilla/django-csp/blob/53278b9946a8c23c07ca3ca53d4338f4ad48068f/docs/configuration.rst#changing-the-policy

r4vi, line 187:
```suggestion
CSP_INCLUDE_NONCE_IN = env.tuple("CSP_INCLUDE_NONCE_IN", default=("script-src",))
```

r4vi (changes requested):

conf/base.py
saruniitr: ok, marking it as resolved as updated changes are pushed

r4vi (approved):

PR #33, opened 2020-09-10
Improve error handling of hawk auth

richtier:
This will fail louder so we can better understand the cause of errors


saruniitr (approved):
lgtm

PR #42, opened 2020-09-23
Fix section status of end use

richtier:
image


core/builtins/custom_tags.py
```diff
@@ -713,8 +713,7 @@ def get_end_use_details_status(application):
         fields += ["is_military_end_use_controls", "is_informed_wmd", "is_suspected_wmd"]
         if application.sub_type == STANDARD:
             fields.append("is_eu_military")
-
-    end_use_detail_field_data = [application.get(field) for field in fields]
+    end_use_detail_field_data = [application.get(field) is not None for field in fields]
```
richtier, line 716: checking for falsey/truthy prevented boolean false answers from being understood as "done"

r4vi: :facepalm:

r4vi (approved):

PR #47, opened 2020-10-01
Replace django sass with bundler

richtier:


caseworker/templates/layouts/base.html
```diff
@@ -34,7 +36,7 @@
 		})(window,document,'script','dataLayer','{{ ENVIRONMENT_VARIABLES.GOOGLE_ANALYTICS_KEY }}');
 	</script>
 
-	<meta property="og:image" content="/assets/images/govuk-opengraph-image.png">
+	<meta property="og:image" content="{{ request.scheme }}://{{ request.get_host }}{% static 'images/govuk-opengraph-image.png' %}">
```
richtier, line 39: based on documentation, I think this needs to be a absolute url

exporter/assets/scss/styles.scss
```diff
@@ -1,7 +1,21 @@
...
+@function get_govuk_image_asset_url($filename) {
+  @return url("../../../node_modules/govuk-frontend/govuk/assets/images/" + $filename);
+}
+
+@function get_govuk_font_asset_url($filename) {
+  @return url("../../../node_modules/govuk-frontend/govuk/assets/fonts/" + $filename);
+}
+
+$govuk-image-url-function: 'get_govuk_image_asset_url';
+$govuk-font-url-function: 'get_govuk_font_asset_url';
```
richtier, line 13: helpful feature of govuk-frontend https://github.com/alphagov/govuk-frontend/search?q=govuk-image-url-function

conf/caseworker.py
```diff
@@ -60,24 +60,20 @@
...
-COMPRESS_PRECOMPILERS = (("text/x-scss", "django_libsass.SassCompiler"),)
 
 STATICFILES_DIRS = (
-    os.path.join(BASE_DIR, "caseworker/assets"),
-    os.path.join(BASE_DIR, "shared_assets/node_modules/govuk-frontend/govuk/"),
-    os.path.join(BASE_DIR, "shared_assets/node_modules/govuk-frontend/govuk/assets/"),
-    os.path.join(BASE_DIR, "shared_assets/lite-frontend/"),
+    os.path.join(BASE_DIR, "caseworker/assets/"),
+    os.path.join(BASE_DIR, "caseworker/assets/built/"),
+    os.path.join(BASE_DIR, "node_modules/govuk-frontend/govuk/"),
```
eadpearce, line 72: don't think we need these any more

conf/exporter.py
```diff
@@ -45,24 +45,19 @@
...
+STATIC_ROOT = os.path.join(DATA_DIR, "exporter/staticfiles")
 
 STATICFILES_DIRS = (
+    os.path.join(BASE_DIR, "core/assets/"),
     os.path.join(BASE_DIR, "exporter/assets"),
-    os.path.join(BASE_DIR, "shared_assets/node_modules/govuk-frontend/govuk/"),
-    os.path.join(BASE_DIR, "shared_assets/node_modules/govuk-frontend/govuk/assets/"),
-    os.path.join(BASE_DIR, "shared_assets/lite-frontend/"),
+    os.path.join(BASE_DIR, "exporter/assets/built/"),
+    os.path.join(BASE_DIR, "node_modules/govuk-frontend/govuk/"),
```
eadpearce, line 57: same here

exporter/assets/scss/components/_activity.scss
```diff
@@ -56,11 +52,14 @@
...
 }
 
+
+$light-grey: #f3f2f1;
+
 .app-activity__additional-text {
 	position: relative;
 	padding: govuk-spacing(4) govuk-spacing(4) 15px;
 	background-color: govuk-colour("light-grey");
-	border: 2px solid darken(govuk-colour("light-grey"), 3%);
```
eadpearce, line 63: can you import the file this mixin is coming from instead of reassigning the variable in here?

conf/caseworker.py
richtier: we do - these are used by staticfiles in 2 ways:
 - where to look when running collectstatic
 - where to look when using `django.contrib.staticfiles.storage.StaticFilesStorage` backend (in dev mode)

exporter/assets/scss/components/_activity.scss
richtier: oooh err debug mistake that I should have removed. will fix :+1:

conf/caseworker.py
richtier: ah yes you're right :)

eadpearce (approved):

PR #50, opened 2020-10-05
Make application back button go back to correct section

richtier:
also added js file that we need for the site to work, which will be ultimately removed when @eadpearce's changes are merged

back-to-section

### DIFF
--- a/exporter/applications/components.py
+++ b/exporter/applications/components.py
@@ -1,4 +1,4 @@
-from django.urls import reverse_lazy
+from django.urls import reverse
 
 from lite_content.lite_exporter_frontend import strings
 from lite_forms.components import BackLink, Label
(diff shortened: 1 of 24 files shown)


exporter/applications/components.py
```diff
@@ -8,14 +8,9 @@ def back_to_task_list(application_id):
...
 
 
 def footer_label(application_id):
-    return Label(
-        'Or <a class="govuk-link" href="'
-        + reverse_lazy("applications:task_list", kwargs={"pk": str(application_id)})
-        + '">Or return to application overview</a>'
-    )
+    url = reverse("applications:task_list", kwargs={"pk": str(application_id)})
+    return Label(f'Or <a class="govuk-link" href="{url}">Or return to application overview</a>')
```
eadpearce, line 16: There's two 'or's here. Do we want the link to be on "Or return to application overview" or just the "return to application overview" part? Personally I prefer the latter

exporter/applications/views/additional_documents.py
```diff
@@ -7,14 +6,13 @@
 
 
 class AdditionalDocuments(LoginRequiredMixin, TemplateView):
-    def get(self, request, **kwargs):
-        application_id = str(kwargs["pk"])
-        data, _ = get_additional_documents(request, application_id)
+    template_name = "applications/additional-documents/additional-documents.html"
```
eadpearce, line 9: 👍 much more django-y

eadpearce (approved):

PR #65, opened 2020-10-30
Improve assign flags form redirects

eadpearce:


caseworker/flags/views.py
```diff
@@ -288,7 +287,7 @@ def get_level(self):
             return FlagLevel.CASES
         elif self.request.GET.get("organisation"):
             return FlagLevel.ORGANISATIONS
-        elif self.request.GET.get("goods") or self.request.GET.get("goods_types"):
+        elif self.request.GET.get("good_ids") or self.request.GET.get("goods_types"):
```
richtier, line 290: do we want this change? I thought we reverted back to `goods`

unit_tests/conftest.py
```diff
@@ -516,25 +516,27 @@ def data_standard_case():
...
-                        "sub_type_other": None,
-                        "role": {"key": "other", "value": "Other"},
-                        "role_other": None,
-                        "flags": [],
-                        "copy_of": None,
-                        "deleted_at": None,
-                        "clearance_level": None,
-                        "descriptors": None,
-                    },
+                    "data": [
```
richtier, line 519: hmm, how did tests pass before this changed data type I wonder

unit_tests/caseworker/flags/test_views.py
```diff
@@ -0,0 +1,131 @@
+import pytest
```
richtier, line 1: dope tests there

caseworker/flags/views.py
eadpearce: ah good spot. i will change this back

unit_tests/conftest.py
eadpearce: my guess is this part of the data just wasn't accessed by tests before now

richtier (approved):

r4vi (approved):

PR #70, opened 2020-11-06
LTD-361: Add option to review product details when importing into an application

saruniitr:
This is in preparation for future work where we need to review and
create new products if necessary when importing existing ones into an
application

These changes are under `FEATURE_FLAG_FIREARMS_ENABLED` with the flag disabled.

Option to view existing product details,
import_from_existing_review

Clicking view take to below screen,

Additional option to add the product to application while reviewing the details,
add_existing_to_application

### DIFF
--- a/exporter/applications/urls.py
+++ b/exporter/applications/urls.py
@@ -99,7 +99,6 @@ urlpatterns = [
         name="add_document",
     ),
     path("<uuid:pk>/goods/add-new/<uuid:good_pk>/attach/", goods.AttachDocument.as_view(), name="attach_documents"),
-    path("<uuid:pk>/goods/<uuid:good_pk>/view/", goods.ViewGoodAddToApplication.as_view(), name="view_good_and_add"),
     path("<uuid:pk>/goods/add-preexisting/", goods.ExistingGoodsList.as_view(), name="preexisting_good"),
     path("<uuid:pk>/goods/<uuid:good_pk>/add/", goods.AddGoodToApplication.as_view(), name="add_good_to_application",),
     path(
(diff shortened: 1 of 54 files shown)


exporter/applications/views/goods.py
```diff
@@ -55,6 +56,16 @@ def get_context_data(self, **kwargs):
         )
 
 
+class ViewGoodAddToApplication(LoginRequiredMixin, TemplateView):
```
richtier, line 59: why does this exist? instead of linking to this view could the template instead link to there this view redirects to?

if not, fair enough this could be a `RedirectView` :)

exporter/applications/urls.py
```diff
@@ -99,6 +99,7 @@
         name="add_document",
     ),
     path("<uuid:pk>/goods/add-new/<uuid:good_pk>/attach/", goods.AttachDocument.as_view(), name="attach_documents"),
+    path("<uuid:pk>/goods/<uuid:good_pk>/view/", goods.ViewGoodAddToApplication.as_view(), name="view_good_and_add"),
```
richtier, line 102: I think this should be feature flagged too as it redirects to a url that will 404 if the feature flag is turned off

exporter/applications/views/goods.py
saruniitr: initially was thinking it needed more than a redirect, will link in the template otherwise change it to redirectview.

exporter/applications/urls.py
saruniitr: agree, missed this.

exporter/applications/views/goods.py
saruniitr: directly linked in the template

exporter/applications/urls.py
saruniitr: removed from this file as it is directly linked in the template

richtier (approved):

PR #72, opened 2020-11-12
Allow for commas in currency and quantity fields

eadpearce:


r4vi (approved):
:+1: clean implementation

PR #75, opened 2020-11-23
LTD-434: Cannot add new product to draft application

saruniitr:
## Change description
When `draft_pk` is missing in kwargs it returns `None` but it was being converted to string using `str(kwargs.get("draft_pk"))` which assigns the value `"None"` to it resulting in this bug.

Fixed the existing browser test and extended it to be able to test this.


r4vi (approved):

r4vi:
nice work on adding tests and cleanup

PR #77, opened 2020-11-24
LTD-398: Accurately categorise Firearms related products

saruniitr:
## Change description

To accurately categorise firearms products additional product types are introduced.
In the v3 Firearms flow Product type is the first screen (instead of Product category) as we will only be allowing Firearms products in the first version.
Moreover we cannot completely implement the flow in this PR because there are other tickets for the rest of the steps in the flow so this PR completes the flow to the extent possible. As we complete other tickets this needs to be adjusted.

The main difference is with respect to first four product types (Firearms, Ammunition, Components for firearms and Components for ammunition) where the next steps capture additional/new details (future tickets). So for now the next steps for all types are same as in current flow.

### Product summary screen
When adding a new product, after entering all details a product summary screen is displayed (before document upload), this is updated to show specific information whether it is a firearm, accessory or software/tech related to firearm.

Browser tests are added for adding firearms product types.

### DIFF
--- a/exporter/goods/forms.py
+++ b/exporter/goods/forms.py
@@ -64,14 +64,22 @@ def product_category_form(request):
 
 
 def software_technology_details_form(request, item_category=None):
-    category = get_category_display_string(
-        request.POST.get("item_category", "") if not item_category else item_category
-    )
+    category_type = request.POST.get("item_category", "") if not item_category else item_category
+    if request.POST.get("type") in ["software_related_to_firearms", "technology_related_to_firearms"]:
+        category_type = request.POST.get("type")
+
+    category = get_category_display_string(category_type)
+
     return Form(
         title=CreateGoodForm.TechnologySoftware.TITLE + category,
         questions=[
             HiddenField("is_software_or_technology_step", True),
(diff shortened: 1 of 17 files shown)


exporter/goods/forms.py
```diff
@@ -64,14 +64,22 @@ def product_category_form(request):
 
 
 def software_technology_details_form(request, item_category=None):
-    category = get_category_display_string(
-        request.POST.get("item_category", "") if not item_category else item_category
-    )
+    category_type = request.POST.get("item_category", "") if not item_category else item_category
+    if request.POST.get("type") in ["software_related_to_firearms", "technology_related_to_firearms"]:
+        category_type = request.POST.get("type")
```
richtier, line 69: the logic for determining the category type seems a bit complex. is this worth moving into a function and writing some tests for it?

saruniitr: :+1:, currently adding browser tests. I will refactor them into a function and add unit tests.

saruniitr: all of that was actually not required and it is simplified now and no need to refactor into a function.

richtier (approved):

r4vi (approved):

PR #80, opened 2020-11-26
Split Calibre and YOM questions

r4vi:
This does not implement the logic of where to ask the questions (only on adding to application vs adding to product list) but does split
the logic so that the two questions are asked separately


lite_forms/components.py
```diff
@@ -148,6 +148,9 @@ def __init__(
                 for item in getattr(component, "javascript_imports", []):
                     self.javascript_imports.add(item)
 
+    def __str__(self):
```
r4vi, line 151: just makes it easier to see when you log / debug the current form name

saruniitr (approved):
looks good to me

r4vi:
@saruniitr just trying to resolve one issue after rebase on master... adding new product flow has broken

PR #81, opened 2020-11-26
LTD-376: Sporting shotgun frontend changes

saruniitr:
## Change description
Change to ask sporting shotgun details.
This is asked when adding a new product to the application as well as adding it to the product list.
The question comes after firearm product type and is not applicable if it is an accessory, software/technology related to firearm.

Browser tests updated to match the updated flow.

### DIFF
--- a/caseworker/templates/case/product-on-case.html
+++ b/caseworker/templates/case/product-on-case.html
@@ -45,6 +45,16 @@
                                 <th scope="row" class="govuk-table__header">Product type</th>
                                 <td class="govuk-table__cell">{{ good_on_application.good.firearm_details.type.value }}</td>
                             </tr>
+                            <tr class="govuk-table__row app-table__row">
+                                <th scope="row" class="govuk-table__header">Sporting shotgun</th>
+                                <td class="govuk-table__cell">
+                                    {% if good_on_application.good.firearm_details.is_sporting_shotgun is not None %}
+                                    {{ good_on_application.good.firearm_details.is_sporting_shotgun|friendly_boolean }}
+                                    {% else %}
+                                        N/A
+                                    {% endif %}
+                                </td>
+                            </tr>
                         {% else %}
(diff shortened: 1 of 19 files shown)


caseworker/templates/case/product-on-case.html
```diff
@@ -45,6 +45,16 @@ <h2 class="govuk-heading-l">Full details</h2>
...
+                            <tr class="govuk-table__row app-table__row">
+                                <th scope="row" class="govuk-table__header">Sporting shotgun</th>
+                                <td class="govuk-table__cell">
+                                    {% if good_on_application.good.firearm_details.is_sporting_shotgun is not None %}
+                                    {{ good_on_application.good.firearm_details.is_sporting_shotgun|friendly_boolean }}
+                                    {% else %}
+                                        N/A
+                                    {% endif %}
+                                </td>
+                            </tr>
```
saruniitr, line 57: for existing products where this field is missing we display `N/A`

exporter/applications/views/goods.py
```diff
@@ -110,12 +111,7 @@ def on_submission(self, request, **kwargs):
         copied_request = request.POST.copy()
         is_pv_graded = copied_request.get("is_pv_graded", "") == "yes"
         is_software_technology = copied_request.get("item_category") in ["group3_software", "group3_technology"]
-        is_firearms_core = copied_request.get("type") in [
-            "firearms",
-            "ammunition",
-            "components_for_firearms",
-            "components_for_ammunition",
-        ]
+        is_firearms_core = copied_request.get("type") in FIREARM_AMMUNITION_COMPONENT_TYPES
```
saruniitr, line 114: had the repeat the same in two three places hence moved to a common place

caseworker/templates/case/product-on-case.html
r4vi: you could use the `|default_na` filter

r4vi (approved):

PR #82, opened 2020-12-01
LTD-373: Firearms replica question

saruniitr:


exporter/goods/forms.py
```diff
@@ -591,6 +592,36 @@ def firearm_year_of_manufacture_details_form(good_id=None):
     )
 
 
+def firearm_replica_form(firearm_type):
+    return Form(
+        title=CreateGoodForm.FirearmGood.FirearmsReplica.TITLE,
+        default_button_name="Save and continue",
+        questions=[
+            HiddenField("type", firearm_type),
```
saruniitr, line 600: `type` is required otherwise when editing only this field from product summary screen using "Change" links it won't get sent in the data.

r4vi (approved):

r4vi:
in `conftest.py` I think I did exactly the same change so it should be able to rebase cleanly

r4vi:
didn't mean to press close, just meant to press comment :scream:

PR #83, opened 2020-12-01
Add ID/serial numbers question

eadpearce:


exporter/applications/services.py
```diff
@@ -106,22 +107,35 @@ def get_application_goods_types(request, pk):
...
+
+
 def post_good_on_application(request, pk, json):
     post_data = serialize_good_on_app_data(json)
     response = client.post(request, f"/applications/{pk}/goods/", post_data)
-    return response.json(), response.status_code
+    response_data = response.json()
+    if "errors" in response_data:
+        # firearm_details field errors need moving to top level of dict for MultiFormView pick them up and display them
+        response_data["errors"] = {
```
r4vi, line 122: I think there's a `flattern_errors` hepler for this somewhere

eadpearce: yeah I thought I could use that too but it doesn't do what I thought it would do. the keys for the flattened errors would be `firearm_details.has_identification_markings`. unless there's an option to not do that that i missed

eadpearce: as discussed, going to move this error formatting to api side

r4vi: as discussed in the channel we'll let the backend return flat errors

exporter/applications/services.py
```diff
@@ -106,6 +107,12 @@ def get_application_goods_types(request, pk):
     return data.json().get("goods") if data.status_code == HTTPStatus.OK else None
 
 
+def validate_good_on_application(request, pk, json):
```
r4vi, line 110: there's also a `validate_application_good` function that does exactly the same thing... yours is better though so you might want to replace usages of `validate_application_good` with `validate_good_on_application`

PR #88, opened 2020-12-01
Firearms template changes

saruniitr:


caseworker/templates/case/review-good-standard.html
```diff
@@ -60,44 +60,186 @@
...
-                <th scope="row" class="govuk-table__header">Military use</th>
-                <td class="govuk-table__cell">
-                    {% if object.good.is_military_use %}
-                        {{ object.good.is_military_use.value }}
-                        {% if object.good.modified_military_use_details %}
-                        , {{ object.good.modified_military_use_details }}
+
+            {% if object.good.firearm_details %}
+                <!-- define it as a variable -->
+                {% with firearm_details=object.good.firearm_details %}
```
r4vi, line 66: should this be:
```suggestion
												{% with firearm_details=good_on_application.firearm_details|default_if_none:good_on_application.good.firearm_details %}

```?

Comment (saruniitr) [reply caseworker/templates/case/review-good-standard.html:66]: not fully caught up with recent changes, `firearm_details` is in GoodOnApplication as well?

Comment (r4vi) [reply caseworker/templates/case/review-good-standard.html:66]: yep, it should look at GoodOnApplication first then fall back to .good

Comment (saruniitr) [reply caseworker/templates/case/review-good-standard.html]: updated

File: caseworker/templates/case/review-good-standard.html [hunk 2]
```diff
@@ -60,44 +60,186 @@
...
             </tr>
-            <tr class="govuk-table__row app-table__row">
-                <th scope="row" class="govuk-table__header">Military use</th>
-                <td class="govuk-table__cell">
-                    {% if object.good.is_military_use %}
-                        {{ object.good.is_military_use.value }}
-                        {% if object.good.modified_military_use_details %}
-                        , {{ object.good.modified_military_use_details }}
+
+            {% if object.good.firearm_details %}
```
r4vi, line 64: I would move the `{% with %}` above the `{% if %}` and use only use `firearm_details` everywhere

saruniitr: yeah makes sense now as it has all the details

r4vi (approved):

PR #89, opened 2020-12-01
Product-on-Case template

r4vi:
Show values from good_on_application's firearms details instead of the ones
that come from the `good` directly if available

### DIFF
--- a/caseworker/templates/case/product-on-case.html
+++ b/caseworker/templates/case/product-on-case.html
@@ -105,24 +105,24 @@
 															</td>
 													</tr>
 													{% if firearm_details %}
-															{% with firearm_types="firearms ammunition components_for_firearms components_for_ammunition" %}
+														{% with firearm_types="firearms ammunition components_for_firearms components_for_ammunition" %}
 															{% if firearm_details.type.key in firearm_types.split %}
+																{% if firearm_details.type.key == "firearms" %}
 																	<tr class="govuk-table__row app-table__row">
-																			<th scope="row" class="govuk-table__header">Year of manufacture</th>
-																			<td class="govuk-table__cell">{{ firearm_details.year_of_manufacture }}</td>
+																		<th scope="row" class="govuk-table__header">Year of manufacture</th>
+																		<td class="govuk-table__cell">{{ firearm_details.year_of_manufacture }}</td>
 																	</tr>
-																	{% if firearm_details.type.key == "firearms" %}
-																			<tr class="govuk-table__row app-table__row">
-																					<th scope="row" class="govuk-table__header">Replica firearm</th>
-																					<td class="govuk-table__cell">
-																							{% if firearm_details.is_replica is not None %}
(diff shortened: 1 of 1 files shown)


caseworker/templates/case/product-on-case.html
```diff
@@ -40,203 +40,205 @@ <h2 class="govuk-heading-l">Full details</h2>
...
+																	{{ good_on_application.good.is_pv_graded|yesno|title }}
+																	{% if good_on_application.good.grading_comment %}
+																			- {{ good_on_application.good.grading_comment }}
+																	{% endif %}
+															</td>
+													</tr>
+													{% if firearm_details %}
+															{% with firearm_types="firearms ammunition components_for_firearms components_for_ammunition" %}
+															{% if firearm_details.type.key in firearm_types.split %}
+																	<tr class="govuk-table__row app-table__row">
```
saruniitr, line 110: these should be for the `firearms` type only

caseworker/templates/case/product-on-case.html
```diff
@@ -40,203 +40,205 @@ <h2 class="govuk-heading-l">Full details</h2>
...
-                                        <td class="govuk-table__cell">
-                                            {% if good_on_application.good.firearm_details.is_replica is not None %}
-                                                {{ good_on_application.good.firearm_details.is_replica|friendly_boolean }}, {{ good_on_application.good.firearm_details.replica_description }}
-                                            {% else %}
-                                                N/A
-                                            {% endif %}
-                                        </td>
-                                    </tr>
-                                {% endif %}
+												{% with firearm_details=good_on_application.firearm_details|default_if_none:good_on_application.good.firearm_details %}
```
saruniitr, line 43: not sure if it is github rendering issue or an indenting issue here

caseworker/templates/case/product-on-case.html
r4vi: I didn't change this - just a whitespace change here

caseworker/templates/case/product-on-case.html
r4vi: I just added the with block before the first occurrence of `firearm_details` and ended it after the last

then did find & replace on both: 

`good_on_application.firearm_details` -> `firearm_details`
`good_on_application.good.firearm_details` -> `firearm_details`

caseworker/templates/case/product-on-case.html
saruniitr: ok, this is the change I have made in other templates, could you please apply below condition for YOM field as well,
`{% if firearm_details.type.key == "firearms" %}`

caseworker/templates/case/product-on-case.html
saruniitr: :+1:

caseworker/templates/case/product-on-case.html
r4vi: thanks I understand - we only want YOM to show up for type=firearms

saruniitr (approved):

PR #90, opened 2020-12-02
Ask deactivated for any firearm

richtier:


exporter/goods/services.py
```diff
@@ -108,7 +108,11 @@ def add_firearm_details_to_data(json):
         if name in json:
             firearm_details[name] = json.pop(name)
 
-    json["firearm_details"] = firearm_details
+    if firearm_details and "is_sporting_shotgun" not in firearm_details:
```
saruniitr, line 111: I don't completely follow this change.
If `is_sporting_shotgun` not in firearm_details then the product type is either accessory, software or tech, then not clear why it is added and set to `False`.

saruniitr (commented):
requested clarification on a change

exporter/goods/services.py
saruniitr: This is for the case of adding an existing product to the application - discussed offline and agreed it is ok

saruniitr (approved):
looks good to me

PR #92, opened 2020-12-03
Firearms navigation fixes

saruniitr:
## Change description

Fixes issues found during review,
 - When editing sporting shotgun either from Product summary or Product detail, we currently ask few additional questions instead of going directly to summary screen. Similarly for other fields
 - Clicking back on sporting shotgun and product type again causes endless loop
 - Changing product type from eg ammunition to firearms, it is possible to proceed with some questions having invalid value eg firearms replica - we are temporarily not allowing editing product type

### DIFF
--- a/exporter/goods/services.py
+++ b/exporter/goods/services.py
@@ -115,6 +115,19 @@ def add_firearm_details_to_data(json):
     if firearm_details and "is_sporting_shotgun" not in firearm_details:
         firearm_details["is_sporting_shotgun"] = False
 
+    if firearm_details and "year_of_manufacture" not in firearm_details:
+        firearm_details["year_of_manufacture"] = 0
+
+    if firearm_details and "is_covered_by_firearm_act_section_one_two_or_five" not in firearm_details:
+        firearm_details["is_covered_by_firearm_act_section_one_two_or_five"] = False
+        firearm_details["section_certificate_number"] = ""
+        firearm_details["section_certificate_date_of_expiry"] = "2012-12-21"
+
+    if firearm_details and "has_identification_markings" not in firearm_details:
+        firearm_details["has_identification_markings"] = False
+        firearm_details["no_identification_markings_details"] = "n/a"
+        firearm_details["identification_markings_details"] = "n/a"
+
     if firearm_details:
         json["firearm_details"] = firearm_details
     return json
--- a/exporter/goods/services.py
+++ b/exporter/goods/services.py
@@ -48,6 +48,42 @@ def post_goods(request, json):
     return data.json(), data.status_code
 
 
+def add_section_certificate_details(firearm_details, json):
+    if "section_certificate_step" in json:
+        # parent component doesnt get sent when empty unlike the remaining form fields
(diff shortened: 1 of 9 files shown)


exporter/templates/applications/goods/add-good-detail-summary.html
```diff
@@ -21,9 +21,6 @@ <h1 class="govuk-heading-l">
 				{{ good.firearm_details.type.value|default_na }}
 			</dd>
 			<dd class="govuk-summary-list__actions">
```
saruniitr, line 23: Removing the empty `dd` messing up the table format in the template

exporter/templates/goods/good.html
```diff
@@ -296,13 +296,6 @@ <h3 class="govuk-heading-m">{% lcs 'goods.GoodPage.YOUR_GOOD' %}</h3>
 						{{ good.firearm_details.type.value|default_na }}
 					</dd>
 					<dd class="govuk-summary-list__actions">
```
saruniitr, line 298: same here

exporter/goods/services.py
```diff
@@ -115,6 +115,19 @@ def add_firearm_details_to_data(json):
     if firearm_details and "is_sporting_shotgun" not in firearm_details:
         firearm_details["is_sporting_shotgun"] = False
 
+    if firearm_details and "year_of_manufacture" not in firearm_details:
```
saruniitr, line 118: similar workaround as in `is_sporting_shotgun` above.

saruniitr, line 118: similar workaround as in `is_sporting_shotgun` above.

exporter/goods/views.py
```diff
@@ -234,23 +235,15 @@ def get_data(self):
...
-                return reverse_lazy("goods:good_military_use", kwargs={"pk": self.object_pk})
-        elif self.draft_pk:
+        if self.draft_pk:
             return reverse(
                 "goods:good_detail_application",
                 kwargs={"pk": self.object_pk, "type": "application", "draft_pk": self.draft_pk},
             )
-        # Edit
-        else:
+        elif self.application_id and self.object_pk:
```
saruniitr, line 243: this is the case of coming from product summary screen and below `elif` is if we come from Product list -> product detail. Similar changes in all edit views of the fields.

exporter/goods/services.py
r4vi: it's valid not to send this as it's not a required field, why do we need to set to 0?

exporter/goods/services.py
```diff
@@ -115,6 +115,19 @@ def add_firearm_details_to_data(json):
...
+        firearm_details["year_of_manufacture"] = 0
+
+    if firearm_details and "is_covered_by_firearm_act_section_one_two_or_five" not in firearm_details:
+        firearm_details["is_covered_by_firearm_act_section_one_two_or_five"] = False
```
r4vi, line 122: not entirely comfortable with setting these values to defaults - is there any way to avoid this?

saruniitr: added in the else case as discussed

r4vi (approved):

PR #103, opened 2020-12-20
Firearms section certificate upload form fixes

saruniitr:
## Change description

Firearms section certificate upload form fixes
 - display the title as per the selected section
 - don't clear the selected file on validation errors
 - populate the selected filename when editing

Save selected file details temporarily in session and clear once
the request is completed successfully.

We setup `upload_handlers` to upload the file to S3. This gets processed as and when we read `request.POST` and the file
gets uploaded to S3 but if there is a validation error we use the saved details from session and populate the fileupload field with the previously selected file name. If the user wishes to replace then another file can be selected. This is required both when initially adding the certificate or when editing it from the summary screen.
While editing if user changes the section then the previously uploaded file gets removed.

### DIFF
--- a/exporter/goods/views.py
+++ b/exporter/goods/views.py
@@ -18,6 +18,7 @@ from exporter.applications.services import (
     get_status_properties,
     get_case_generated_documents,
     post_application_document,
+    get_application_documents,
     fetch_and_delete_previous_application_documents,
 )
 from exporter.core.constants import FIREARM_AMMUNITION_COMPONENT_TYPES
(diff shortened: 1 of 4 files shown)


exporter/goods/views.py
```diff
@@ -712,57 +729,71 @@ def get_data(self):
...
             fetch_and_delete_previous_application_documents(request, kwargs["pk"], kwargs["good_pk"])
 
             data, status_code = post_application_document(request, kwargs["pk"], kwargs["good_pk"], doc_data)
             if status_code != HTTPStatus.CREATED:
                 return error_page(request, data["errors"]["file"])
 
+        if not certificate_available and self.certificate_filename:
+            fetch_and_delete_previous_application_documents(request, kwargs["pk"], kwargs["good_pk"])
+
+        if session_key in request.session.keys():
```
r4vi, line 779: `session_key` is a bit confusing as a name as on first glance you might assume it's the users django session key but it's a unique key for this upload?

r4vi (approved):

exporter/goods/views.py
saruniitr: yes just for this file upload, will rename it, `file_upload_session_key` or just `file_upload_key`?

r4vi: both are fine but slight preference for `file_upload_key` as it's being added to the `request.session` so `_session_` part is a little redundant

PR #110, opened 2021-01-07
Denial feature branch

richtier:
## Change description

Denial upload, matching denials, displaying matching denials on the case details page and option to delete the selected ones.

denials


caseworker/external_data/views.py
```diff
@@ -0,0 +1,68 @@
...
+from django.views.generic import TemplateView
+from django.views.generic.edit import FormView, CreateView
+
+from core.auth.views import LoginRequiredMixin
+
+from caseworker.external_data import forms, services
+
+
+with open(settings.BASE_DIR + "/caseworker/external_data/example.csv", "rb") as f:
+    base_64_csv = base64.b64encode(f.read()).decode()
```
r4vi, line 17: why are we base64 encoding the example csv?

r4vi (approved):

caseworker/external_data/views.py
richtier: the view exposes it to the template, so the template can do this:
```
<a download="example-denials.csv" href="data:text/csv;base64,{{ base_64_csv }}">Download an example .csv file</a>
```

this is instead of serving the example csv via a /media/ or /static/ url.

motivation was less moving parts

r4vi:
Oh nice 👍

eadpearce (approved):

PR #117, opened 2021-01-14
LTD-507 Control List Entries cleared when adding new product to app

r4vi:
This fixes a bug with the serialisation of a new good when being added to an
application, there was an issue that if the user added a firearm certificate
then the control list entries would be lost. Fixing that bug in `lite-api`
also uncovered another bug in that only the first control list entry was
being sent to lite-api instead of the full list.

This commit makes sure the full list of `control_list_entries` is saved to the
session so it can be played back later.

### DIFF
--- a/exporter/applications/views/goods.py
+++ b/exporter/applications/views/goods.py
@@ -161,7 +161,13 @@ class AddGood(LoginRequiredMixin, MultiFormView):
         if int(self.request.POST.get("form_pk")) == number_of_forms:
             if self.show_section_upload_form:
                 firearms_data_id = f"post_{request.session['lite_api_user_id']}_{self.draft_pk}"
-                request.session[firearms_data_id] = copied_request
+                session_data = copied_request.dict()
+                if "control_list_entries[]" in copied_request:
+                    session_data["control_list_entries"] = copied_request.getlist("control_list_entries[]")
+                    del session_data["control_list_entries[]"]
+
+                request.session[firearms_data_id] = session_data
+
             elif self.certificate_not_required:
                 self.action = post_goods
             else:


exporter/applications/views/goods.py
```diff
@@ -161,7 +161,13 @@ def on_submission(self, request, **kwargs):
         if int(self.request.POST.get("form_pk")) == number_of_forms:
             if self.show_section_upload_form:
                 firearms_data_id = f"post_{request.session['lite_api_user_id']}_{self.draft_pk}"
-                request.session[firearms_data_id] = copied_request
+                session_data = copied_request.dict()
+                if "control_list_entries[]" in copied_request:
+                    session_data["control_list_entries"] = copied_request.getlist("control_list_entries[]")
```
saruniitr, line 166: Currently I think we are not saving control list entries in the session, only firearms data is in session. They need to be in session for this fix? it is not used anywhere else also.

r4vi: they were being saved here
`request.session[firearms_data_id] = copied_request` in the add new product to application flow


 but two issues:

- `request.session[firearms_data_id] = copied_request` only has `control_list_entries[]` not the converted `control_list_entries` value. so when it is reused later the API would not understand it (would get cleared out of cleaned_data)
- request.session[firearms_data_id] = copied_request converts `QueryDict` to a dict and loses `['ML1a', 'ML2']` and only saves the first one.

saruniitr: ah yes, I was thinking we save specific firearms data in session, we save the whole request.

saruniitr (approved):
lgtm

PR #121, opened 2021-01-15
Fix encoding on postcode querystring

eadpearce:


richtier (approved):

r4vi (approved):
so much better than 
```python
querystring = "&".join(
        [
            convert_value_to_query_param("exclude", exclude),
            convert_value_to_query_param("get_total_users", get_total_users),
            convert_value_to_query_param("postcode", postcode),
        ]
    )
```

PR #123, opened 2021-01-18
LTD-536: Routing rules flags section updates

saruniitr:
## Change description

Design changes for the way flags are added and edited in routing rules
    
A routing rule can be applied if the case contain certain flags and does not contain some other flags. The UI allows setting up these two conditions and had to make sure a flag is included in either of the condition but not both. This change updates the design and fixes some of the inconsistencies in previous implementation.

All of the changes are in javascript mostly to show/hide controls, enable/disable and check/uncheck flags.

### DIFF
--- a/caseworker/routing_rules/forms.py
+++ b/caseworker/routing_rules/forms.py
@@ -132,13 +132,22 @@ def get_flag_details_html(flag):
     """
 
 
-def select_flags(request, team_id, flags_to_include, flags_to_exclude):
+def select_flags(request, team_id, flags_to_include, flags_to_exclude, is_editing):
     return Form(
-        title=Forms.FLAGS,
+        title=Forms.FLAGS if not is_editing else "Edit flags",
         questions=[
             HiddenField(name="flags_to_include", value=",".join(flags_to_include)),
             HiddenField(name="flags_to_exclude", value=",".join(flags_to_exclude)),
             HTMLBlock("<div id='routing-rules-flags-details'></div>"),
+            Label(id="condition-label", text="Apply the routing rule to:"),
+            RadioButtons(
+                title="",
+                name="routing_rules_flags_condition",
+                options=[
+                    Option(key="contain_selected_flags", value="Cases that contain selected flags"),
+                    Option(key="doesnot_contain_selected_flags", value="Cases that do not contain selected flags"),
+                ],
+            ),
             Filter(),
             Checkboxes(
                 name="flags[]",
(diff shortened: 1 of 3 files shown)


caseworker/routing_rules/forms.py
```diff
@@ -132,13 +132,22 @@ def get_flag_details_html(flag):
...
         questions=[
             HiddenField(name="flags_to_include", value=",".join(flags_to_include)),
             HiddenField(name="flags_to_exclude", value=",".join(flags_to_exclude)),
             HTMLBlock("<div id='routing-rules-flags-details'></div>"),
+            Label(id="condition-label", text="Apply routing rules to cases that"),
+            RadioButtons(
+                title="",
+                name="routing_rules_flags_condition",
+                options=[
+                    Option(key="contain_selected_flags", value="Cases that contain selected flags"),
```
eadpearce, line 147: can remove "Cases that" in both of these as it's already in the label text

caseworker/routing_rules/forms.py
```diff
@@ -150,19 +159,10 @@ def select_flags(request, team_id, flags_to_include, flags_to_exclude):
...
-                options=[
-                    Option(key="contain_selected_flags", value="Contain selected flags"),
-                    Option(key="doesnot_contain_selected_flags", value="Do not contain selected flags"),
-                ],
-            ),
         ],
         buttons=[
             Button("Save and continue", "submit", id="save_and_continue"),
-            Button("Add another condition", "", ButtonStyle.SECONDARY, link="#"),
+            Button("Add another condition", "", ButtonStyle.SECONDARY, id="add_another_condition", link="#"),
```
eadpearce, line 165: use hyphens here just to be consistent

caseworker/routing_rules/forms.py
saruniitr: Content team suggested to update the wording like this. There has been a change on this, it will now be, "Applies to cases that contain selected flags"

caseworker/routing_rules/forms.py
saruniitr: ok

caseworker/routing_rules/forms.py
eadpearce: Can you double check? As it is now "cases that" is repeated when you read from the label text to the option so it probably has to be removed from one of them

eadpearce (approved):

PR #125, opened 2021-01-20
Re-enable adding of non firearm products

r4vi:
This commit puts the recent limiting of adding products behind a feature flag
called FEATURE_FLAG_ONLY_ALLOW_FIREARMS_PRODUCTS. If this flag is not set then
all types of products can be added, however many of those scenarios are broken.

Enabling those broken flows allows up to identify where those things are broken
and raise bugs to fix them.

### DIFF
--- a/unit_tests/exporter/goods/test_forms.py
+++ b/unit_tests/exporter/goods/test_forms.py
@@ -1,5 +1,6 @@
 import pytest
 import requests
+from django.test import override_settings
 
 from lite_forms.components import HelpSection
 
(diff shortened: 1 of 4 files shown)


unit_tests/exporter/goods/test_forms.py
```diff
@@ -147,7 +147,8 @@ def pv_gradings(mock_pv_gradings, rf, client):
         ),
     ],
 )
-def test_core_firearm_product_form_group(rf, client, params, num_forms, question_checks):
+def test_core_firearm_product_form_group(rf, client, settings, params, num_forms, question_checks):
```
saruniitr, line 150: Any specific reason not to use `override_settings`?

r4vi: no reason, it's a helper/fixture from pytest-django but if we're using `override_settings` everywhere else I'll use that for consistency

saruniitr (approved):

PR #133, opened 2021-02-01
Fix cant unselect checkboxes when going back and forth through form

eadpearce:


lite_forms/submitters.py
```diff
@@ -68,7 +68,7 @@ def _prepare_data(request, inject_data):
 
 
 def submit_paged_form(  # noqa
-    request, form_group: FormGroup, action: Callable, object_pk=None, inject_data=None, additional_context: dict = None,
```
r4vi, line 71: I would revert this and the bit below: 
```    if additional_context is None:
        additional_context = {}```

because it was there to avoid a python anti-pattern: https://docs.quantifiedcode.com/python-anti-patterns/correctness/mutable_default_value_as_argument.html

r4vi (changes requested):

lite_forms/submitters.py
eadpearce: ooh. didn't realise. will revert

r4vi (approved):

PR #137, opened 2021-02-08
LTD-502: Attach product documentation

saruniitr:
## Change description

When attaching a document related to a product we ask multiple questions in a single form. These are whether a document is available and if it is above office-sensitive. These are now split into two separate forms. This PR adds changes to the views, templates.


richtier (commented):
looks good. can you add some tests please

richtier (approved):

PR #138, opened 2021-02-08
Add organisation document upload form

richtier:


exporter/goods/services.py
```diff
@@ -134,7 +134,6 @@ def add_firearm_details_to_data(json):
         "has_proof_mark",
         "no_proof_mark_details",
         "is_deactivated",
-        "date_of_deactivation",
```
richtier, line 137: this is a duplicate

exporter/organisation/forms.py
```diff
@@ -0,0 +1,27 @@
...
+        questions=[
+            FileUpload(),
+            TextInput(name="reference_code", title="Certificate number",),
+            DateInput(prefix="expiry_date_", title="Expiry date", description="For example 12 3 2021"),
+        ],
+        back_link=BackLink("Back", back_url),
+    )
+
+
+def attach_section_five_certificate_form(back_url):
```
saruniitr, line 17: Is this going to be a separate form? My understanding was it will be another option in the same form where we upload Section 1 and 2.

richtier: The designs I'm following make me think it's a separate form, but there is room for modifying with effort if it turns out we need it to be the same form

exporter/organisation/views.py
```diff
@@ -49,3 +60,49 @@ class RedirectToMembers(LoginRequiredMixin, RedirectView):
...
+    def handle_s3_upload(self):
+        self.request.upload_handlers.insert(0, S3FileUploadHandler(self.request))
+
+    @csrf_exempt
+    def post(self, request, **kwargs):
+        self.handle_s3_upload()
+
+        organisation_id = str(request.session["organisation"])
+        data = {
+            "expiry_date": format_date(request.POST, "expiry_date_"),
```
saruniitr, line 83: additional underscore at the end of `expiry_date`

saruniitr (approved):

exporter/organisation/views.py
richtier: that's on purpose oddly - needed for `format_date` to work as expected

PR #139, opened 2021-02-09
Add RFD and firearm certificate + view them

richtier:
view-certificate


exporter/applications/forms/goods.py
```diff
@@ -7,89 +7,97 @@
...
-    QuantityInput,
-    CurrencyInput,
-    RadioButtons,
-    Option,
-    BackLink,
-    TextArea,
-    DateInput,
-    FormGroup,
-)
+from lite_forms import components
```
saruniitr, line 18: always wanted to change this but held back because of more changes, good that it is updated.

exporter/applications/forms/goods.py
```diff
@@ -7,89 +7,97 @@
...
+    request,
+    is_preexisting,
+    good,
+    sub_case_type,
+    draft_pk,
+    application,
+    show_attach_rfd,
+    relevant_firearm_act_section,
+    back_url,
+):
```
saruniitr, line 46: the list of args is growing, I am touching the same code in my PR so there will be conflicts.
I was thinking of sending these args as a dict, do you think it is ok or just leave as is?

exporter/applications/forms/goods.py
```diff
@@ -149,37 +154,40 @@ def firearm_proof_mark_field():
 
 
 def firearm_is_deactivated_field():
-    return RadioButtons(
+    return components.RadioButtons(
         title="Has the product been deactivated?",
         name="is_deactivated",
         options=[
-            Option(
+            components.Option(
```
saruniitr, line 161: I understand these changes are not related to this PR but this is deep.

exporter/applications/views/goods.py
```diff
@@ -172,10 +219,38 @@ def on_submission(self, request, **kwargs):
...
+    def cache_rfd_certificate_details(self):
+        file = self.request.FILES.get("file")
+        if not file:
+            return
+        self.request.session[SESSION_KEY_RFD_CERTIFICATE] = {
+            "name": getattr(file, "original_name", file.name),
+            "s3_key": file.name,
+            "size": int(file.size // 1024) if file.size else 0,  # in kilobytes
+            "document_on_organisation": {
+                "expiry_date": format_date(self.request.POST, "expiry_date_"),
```
saruniitr, line 263: The `_` at the end of `expiry_date_`, is it referring the prefix or the actual field itself?

exporter/assets/scss/components/_documents.scss
```diff
@@ -50,3 +50,8 @@
 		word-break: break-all;
 	}
 }
+
+.certificate-action {
+	width: 200px;
+	text-align: left;
+}
```
saruniitr, line 57: please add newline

exporter/goods/forms.py
```diff
@@ -347,6 +349,10 @@ def add_good_form_group(
     base_form_back_link: str = None,
 ):
     control_list_entries = get_control_list_entries(request, convert_to_options=True)
+
+    show_attach_rfd = str_to_bool(request.POST.get("is_registered_firearm_dealer"))
+    show_rfd_question = is_firearms_core and not has_valid_rfd_certificate(application)
```
saruniitr, line 354: I believe these are only asked when adding a new product to the application but not when adding a product to the product list, If yes then we need `draft_pk` also in the condition.

exporter/goods/forms.py
```diff
@@ -721,7 +715,7 @@ def firearms_act_confirmation_form():
 
 def upload_firearms_act_certificate_form(section, filename, back_link):
     return Form(
-        title=f"Upload your Firearms Act 1968 {section} certificate",
+        title=f"Attach your Firearms Act 1968 {section} certificate",
```
saruniitr, line 718: v3 firearms design showing "Upload" only, please ignore if this is updated in the later designs

saruniitr (commented):
few questions, otherwise looks good to me

exporter/applications/forms/goods.py
richtier: dict instead of explicit arguments will make it harder to maintain

exporter/applications/views/goods.py
richtier: it's going to concat `self.request.POST["expiry_date_day"]`, `self.request.POST["expiry_date_month"]` and ``self.request.POST["expiry_date_year"]`.

if we got rid of the trailing underscore then `format_date` would be looking for `self.request.POST["expiry_dateday"]`, `self.request.POST["expiry_datemonth"]` and ``self.request.POST["expiry_dateyear"]`.

exporter/goods/forms.py
richtier: I've asked Jess what is expected. I'm under impression we want to ask every time as an exporter may add a product to product list on day 1 and then on day 100 add to an application, in the meantime they become a registered firearms dealer

exporter/goods/forms.py
richtier: Jess asked this to be attach :shrug:  :)

saruniitr: ok, may be the design is not updated after that.

exporter/goods/forms.py
saruniitr: ok but after adding to the product list and the user is still not a registered RFD - if the same product is now added to the application (i.e., add existing product option), is there a chance we ask the user again?
If not then it is fine I think otherwise there is a duplication.

richtier: it's now changed to only ask during "add to application" when the existing certificate is out of date

saruniitr: :+1:

saruniitr (approved):

PR #145, opened 2021-02-24
LTD-578 Allow non-firearms products to be added

r4vi:
This commit fixes the broken journeys when trying to add a non-firearm product
without affecting the firearms journey


saruniitr (approved):
looks good to me

PR #146, opened 2021-02-25
Add codeclimate coverage

richtier:


.circleci/config.yml
```diff
@@ -149,7 +177,25 @@ jobs:
...
+      - setup_code_climate
+      - unit_tests:
+          alias: exporter
+
+  upload_coverage:
+    working_directory: ~/repo
+    docker:
+      - <<: *image_python
+    environment: 
+      - CC_TEST_REPORTER_ID: 571971ab2a88d7b691cc86db9513ce0ff1ad5072895d9453ac49c48bb16e64c1
```
r4vi, line 189: i know this isn't sensitive but can we add it to the job's env vars instead of here as someone will see it and freak out?

r4vi (approved):
small comment, up to you if you want to change that

PR #148, opened 2021-03-02
LTD-635: Don't as YOM question when adding non-firearm to application

r4vi:
fixes a bug when adding non-firearms products from existing product list to
applications which would ask for year of manufacture of the firearm.


saruniitr (approved):

r4vi:
still adding some other fixes for non-firearms on the summary screen

PR #154, opened 2021-03-18
Add middleware for no cache header and obfuscate server name

richtier:
image

### DIFF
--- a/core/middleware.py
+++ b/core/middleware.py
@@ -6,6 +6,7 @@ from s3chunkuploader.file_handler import UploadFailed
 from django.conf import settings
 from django.contrib.auth import logout
 from django.shortcuts import redirect
+from django.utils.cache import add_never_cache_headers
 
 from lite_content.lite_internal_frontend.strings import cases
 from lite_forms.generators import error_page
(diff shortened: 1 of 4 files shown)


core/middleware.py
```diff
@@ -60,3 +60,20 @@ def __call__(self, request):
...
+    """
+
+    HEADER_VALUE = "no-store, no-cache, must-revalidate"
+
+    def __init__(self, get_response=None):
+        self.get_response = get_response
+
+    def __call__(self, request):
+        response = self.get_response(request)
+        response["Cache-Control"] = self.HEADER_VALUE
```
r4vi, line 78: wouldn't it be better to use django's built in `add_never_cache_headers(response)`?

core/middleware.py
```diff
@@ -60,3 +60,20 @@ def __call__(self, request):
...
         return self.get_response(request)
+
+
+class NoCacheMiddlware:
+    """Tell the browser to not cache the pages, because otherwise information that should be kept private can be
+    viewed by anyone with access to the files in the browser's cache directory.
+
+    """
+
+    HEADER_VALUE = "no-store, no-cache, must-revalidate"
```
alixedi, line 71: Might consider using (IIRC) [add_never_cache_headers](https://docs.djangoproject.com/en/3.1/ref/utils/#django.utils.cache.add_never_cache_headers), which comes with Django.

alixedi: Something along [these](https://stackoverflow.com/a/5882033/1182202) lines.

core/middleware.py
alixedi: inb4! well played 😂

r4vi (approved):

PR #161, opened 2021-03-30
Allow user to give advice when team advice exists

eadpearce:


alixedi:
Would be nice to add some tests and break the changeset into more atomic commits.

caseworker/cases/views/main.py
```diff
@@ -32,24 +45,54 @@
...
+            or status_props["is_terminal"]
+        ):
+            current_advice_level = []
+
+        return {
+            "is_user_team": True,
+            "teams": get_teams(self.request),
+            "current_advice_level": current_advice_level,
+            "can_give_user_advice": True,
+            "can_combine_user_advice_into_team_advice": False,
```
saruniitr, line 91: Are these meant to be static all the time?

caseworker/templates/case/includes/advice.html
```diff
@@ -19,12 +24,12 @@ <h2 class="app-grouped-advice__heading">
 						<div class="app-grouped-advice__summary-list">
 							{% if advice.type.key == 'proviso' %}
 								<div class="app-grouped-advice__summary-list-item">
-									<p>{% lcs 'advice.AdvicePage.PROVISO' %}</p>
+									<p>Proviso</p>
```
saruniitr, line 27: Any particular reason why these are modified? We are not adding new strings but existing ones we reuse.

caseworker/cases/views/main.py
eadpearce: Just realised I added these and intended to do something with them in the template and then didn't! The only one that was carried over from the previous function is `is_user_team` from https://github.com/uktrade/lite-frontend/pull/161/files#diff-9ec7731e69d681489cb3fc8f3e571f88edee0688b10d6a42d8b901717b0839a6L78

caseworker/templates/case/includes/advice.html
eadpearce: I mainly changed these because I found it easier to navigate the template having the actual text here

caseworker/cases/views/main.py
eadpearce: Checked again and looks like `is_user_team` is not used by the template either so I'll remove that as well

r4vi:
noticed one small bug - have deployed to dev and asked Dan to check if it fixes the problem

r4vi (approved):

PR #166, opened 2021-04-07
LTD-593: Add phone number and website to organisation details

saruniitr:


caseworker/organisations/forms.py
```diff
@@ -158,16 +159,16 @@ def register_commercial_form(in_uk):
     )
 
 
-def create_default_site_form(request, in_uk):
+def create_default_site_form(request, is_commercial, in_uk):
     return Form(
         title=RegisterAnOrganisation.CREATE_DEFAULT_SITE,
         questions=[
-            TextInput(title=RegisterAnOrganisation.NAME_OF_SITE, name="site.name"),
+            TextInput(title=RegisterAnOrganisation.NAME_OF_SITE, name="primary_site.name"),
```
saruniitr, line 166: The field name in the model is `primary_site` hence updated to keep it consistent

caseworker/organisations/views.py
```diff
@@ -187,3 +189,25 @@ def get_form(self):
...
+        self.object_pk = kwargs["pk"]
+        self.organisation = get_organisation(request, str(self.object_pk))
+        self.data = self.organisation
+        self.action = put_organisation
+        self.success_url = reverse_lazy("organisations:organisation", kwargs={"pk": self.object_pk})
+
+    def get_data(self):
+        return {
+            **flatten_data(self.data),
+            "primary_site.address.country": self.organisation["primary_site"]["address"]["country"]["id"],
```
saruniitr, line 205: `country` is a dict on its own and when form is rendered it is only looking at top level hence added it separately

caseworker/organisations/forms.py
```diff
@@ -35,6 +35,7 @@ def register_organisation_forms(request):
     Diverges based on organisation type (individual or commercial), location answer
     also changes compulsory fields
     """
+    is_commercial = request.POST.get("type") == "commercial"
```
alixedi, line 38: The docstring for this function suggests that the organization type can be individual or commercial. 

Given that is the case, is there any reason for adding this as opposed to using `!is_individual`?

saruniitr: No specific reason, just favoured readability and wanted to avoid another indirection.
Given `not is_individual` is used another place (line 91 below) may be it is even better to use this flag and remove `is_individual` completely?

saruniitr:
It is being reworked in a different PR hence closing this

PR #167, opened 2021-04-07
Make autocomplete widget support reusing existing values

r4vi:
This is a example commit to show how to make autocomplete widget support using
the initial value from the form's data

### DIFF
--- a/caseworker/organisations/views.py
+++ b/caseworker/organisations/views.py
@@ -203,14 +203,14 @@ class EditOrganisationAddress(LoginRequiredMixin, SingleFormView):
         data = flatten_data(self.data)
         return {
             **flatten_data(self.data),
-            "country": self.organisation["primary_site"]["address"]["country"]["id"]
+            "primary_site.address.country": self.organisation["primary_site"]["address"]["country"]["id"]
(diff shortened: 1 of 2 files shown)


caseworker/organisations/views.py
```diff
@@ -203,14 +203,14 @@ def get_data(self):
...
-            "country": self.organisation["primary_site"]["address"]["country"]["id"]
+            "primary_site.address.country": self.organisation["primary_site"]["address"]["country"]["id"]
         }
 
     def get_form(self):
         print(self.data)
         is_commercial = self.organisation["type"]["key"] == "commercial"
         in_uk = self.organisation["primary_site"]["address"]["country"]["id"] == "GB"
-        countries = get_countries(self.request, True, ["GB"])
+        countries = get_countries(self.request, True, [])
```
saruniitr, line 213: I think this is required, they want to exclude this because we ask this question only if the organisation is outside UK

r4vi: that just removes GB from the list of countries possible to select, so if a uk company goes to that page then it will be impossible to set it as an initial value

PR #170, opened 2021-04-12
LTD-749 update outdated jquery

r4vi:
This turned into quite the mammoth task so I have made the commits as atomic as possible.

The problem this PR tried to solve was the usage of an outdated vendored version of Jquery but this was a can of worms.

To use the jquery bundled via node_modules, a lot of the dependent modules needed to be updated to be able to interact with a version of jquery that is provided by a JS module system rather than just being a global available as `$`

I've also updated our javascript bundler `parcel` to the latest version


caseworker/templates/components/picklist-picker.html
```diff
@@ -42,8 +42,8 @@
 
 	</div>
 {% endif %}
-
-<script>
+{% block javascript %}
+<script nonce="{{ request.csp_nonce }}">
```
agalamatis, line 46: No script type?

agalamatis:
I think there is [one more place](https://github.com/uktrade/lite-frontend/blob/master/exporter/templates/applications/goods-types/countries.html#L76) helpers.js is pulled directly.

agalamatis (approved):
I left some comments about a couple of minor things.

On the assumption that I trust node generating everything correctly, it addresses the issue identified by the PenTest regarding jQ.

caseworker/templates/components/picklist-picker.html
r4vi: just staying consistent with the rest of the code base.

r4vi:
> I think there is [one more place](https://github.com/uktrade/lite-frontend/blob/master/exporter/templates/applications/goods-types/countries.html#L76) helpers.js is pulled directly.

looks like that's on master, removed in this branch

PR #172, opened 2021-04-15
[LTD-775] Session Termination

alixedi:
During pen tests, we found that even when a user account was terminated on the SSO, the user would still be able to access LITE using browser cookies.

Other DIT services get around this by using introspection middleware. This middleware hits the introspection endpoint on the SSO service and caches the response for a small amount of time.

We are doing something very similar for LITE.

P.S. It's not introspection technically because we are not using the `/o/introspect` endpoint from the SSO service. In the case of `staff-sso`, it is because this endpoint requires a different token. For the case of `directory-sso`, I haven't been able to find an endpoint for introspection.

### DIFF
--- a/core/middleware.py
+++ b/core/middleware.py
@@ -1,4 +1,5 @@
 import time
+import logging
 
 import requests
 from s3chunkuploader.file_handler import UploadFailed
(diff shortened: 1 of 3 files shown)


core/middleware.py
```diff
@@ -101,3 +106,38 @@ def __call__(self, request):
...
+        client = request.authbroker_client
+        token = client.token.get("access_token", "")
+        cache_key = f"sso_introspection:{token}"
+        cache_value = cache.get(cache_key)
+        if cache_value is not None:
+            return
+        logger.info("Introspecting with SSO: %s", request.session.get("lite_api_user_id"))
+        response = client.get(settings.AUTHBROKER_PROFILE_URL)
+        response.raise_for_status()
+        cache.set(cache_key, True, timeout=60 * 5)
```
r4vi, line 129: can we make the timeout configurable (default to 5 mins) in settings.py?

r4vi (approved):

r4vi:
only one small comment from me, not required to merge but up to you...

PR #175, opened 2021-04-21
[LTD-755] User Feedback

alixedi:
This does 3 things -

1. Adds user email to the session. This is used in the feedback loop.
2. Adds a `feedback` app with 2 views. The first view renders a feedback form. The second view says thank you. In between the 2 views, user feedback (along with user email) is sent to a specified email inbox for triage via Notify.
3. Adds a Beta banner with a link to the feedback.

The tests cover most code paths. The only assumption is for `send_email_notification` to throw a subclass of `RequestException` when it fails. I have checked this. It does. 

I haven't included any tests for the beta banner. This is because I have chosen to put it in the template without any conf switches. I feel that it is unlikely that we will need to remove it quickly i.e. with just a config change instead of a PR.

P.S. I have gone off-spec in one place. I haven't put the "<x> characters remaining" thing under the feedback text area. I don't think it is a blocker for pvt. beta but let me know if you think I am wrong.

image

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -57,6 +57,7 @@ crispy-forms-gds = "~=0.2"
 django-health-check = "~=3.14.3"
 django-allow-cidr = "~=0.3"
 django-redis = "~=4.12"
+notifications-python-client = "~=6.0"
 
 [requires]
 python_version = "3.7"
--- a/caseworker/templates/layouts/base.html
+++ b/caseworker/templates/layouts/base.html
@@ -65,7 +65,7 @@
 			<a class="app-header__link" href="{% url 'core:index' %}" data-tooltip="Go home">
 				{% svg 'crest' %}
 				<span class="lite-mobile-hide">
-					Licensing for International Trade and Enterprise <span class="govuk-tag lite-tablet-hide govuk-!-margin-left-3">Alpha</span>
+					Licensing for International Trade and Enterprise
 				</span>
 			</a>
 			<span></span>
@@ -84,6 +84,20 @@
 			</a>
 		</div>
 	</header>
+
+	<!-- Beta Banner -->
+	<div class="govuk-width-container">
+		<div class="govuk-phase-banner">
+			<p class="govuk-phase-banner__content">
+				<strong class="govuk-tag govuk-phase-banner__content__tag">
+				beta
+				</strong>
+				<span class="govuk-phase-banner__text">
+				This is a new service – your <a class="govuk-link" href="{% url 'feedback' %}" target="_blank">feedback</a> will help us to improve it.
+				</span>
+			</p>
+		</div>
+	</div>
 
 	{% block messages %}
 		{% include 'includes/messages.html' %}
--- a/core/feedback/forms.py
+++ b/core/feedback/forms.py
@@ -1,0 +1,5 @@
+from django import forms
+
+
+class FeedbackForm(forms.Form):
+    feedback = forms.CharField(label="Please type your feedback here: ", widget=forms.Textarea)
--- a/core/feedback/urls.py
+++ b/core/feedback/urls.py
@@ -1,0 +1,8 @@
+from django.urls import path
+
+from core.feedback.views import FeedbackView, get_thanks
+
+urlpatterns = [
+    path("", FeedbackView.as_view(), name="feedback"),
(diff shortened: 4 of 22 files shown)


unit_tests/caseworker/feedback/test_views.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+@pytest.mark.parametrize("url", ("feedback", "thanks"))
+def test_feedback_get_success(url, authorized_client):
+    url = reverse(url)
+    response = authorized_client.get(url)
+    assert response.status_code == 200
+
+
+@mock.patch("core.feedback.notify.client")
```
r4vi, line 17: for consistency with the codebase maybe it should be a pytest fixture rather than using @mock.patch directly? example: https://github.com/uktrade/lite-frontend/blob/ca41df83ae086c0f6dff23c6c7409eeecdac56a1/unit_tests/conftest.py#L18

core/feedback/urls.py
```diff
@@ -0,0 +1,8 @@
+from django.urls import path
+
+from core.feedback.views import FeedbackView, get_thanks
+
+urlpatterns = [
+    path("", FeedbackView.as_view(), name='feedback'),
+    path("/thanks", get_thanks, name='thanks'),
```
r4vi, line 7: no trailing slash? feels weird seeing this in a django app

Pipfile
```diff
@@ -57,6 +57,7 @@ crispy-forms-gds = "~=0.2"
 django-health-check = "~=3.14.3"
 django-allow-cidr = "~=0.3"
 django-redis = "~=4.12"
+notifications-python-client = "*"
```
saruniitr, line 60: I think we also commit `Pipfile.lock` which is missing in the PR

unit_tests/caseworker/feedback/test_views.py
alixedi: Err, for some reason, I cannot get the `mock_notify_client.send_email_notification.assert_called_once_with` to pass when I use a `pytest` fixture. I can try and figure out unless you are easy with this.

r4vi: I'm easy, I actually prefer it this way!

caseworker/templates/layouts/base.html
```diff
@@ -85,6 +85,20 @@
...
 
+	<!-- Beta Banner -->
+	<div class="govuk-width-container">
+		<div class="govuk-phase-banner">
+			<p class="govuk-phase-banner__content">
+				<strong class="govuk-tag govuk-phase-banner__content__tag">
+				beta
+				</strong>
+				<span class="govuk-phase-banner__text">
+				This is a new service – your <a class="govuk-link" href="{% url 'feedback' %}" target="_blank">feedback</a> will help us to improve it.
```
r4vi, line 96: nice using `target="_blank"`

r4vi (approved):

core/feedback/forms.py
```diff
@@ -0,0 +1,5 @@
+from django import forms
+
+
+class FeedbackForm(forms.Form):
+    feedback = forms.CharField(label="", widget=forms.Textarea(attrs={"maxlength": "500",}),)
```
eadpearce, line 5: Ideally we should have a label for all form fields for accessibility. I believe the gov uk design system has a `visually-hidden` css class or similar we could use to hide the text for sighted users

exporter/templates/layouts/base.html
```diff
@@ -84,6 +84,17 @@
 		{% endblock %}
 
 		<div class="govuk-width-container">
+			<!-- Beta Banner -->
```
eadpearce, line 87: Since this is also used in caseworker/templates/layouts/base.html we could split it out into a shared template though you'd probably have to tweak some stuff in settings so django knows where to look for it

alixedi: That is a good point. I thought about it but decided to not do it because the beta banner is meant to be removed at some point. Doing it this way means it is significantly easier to remove - materially and from a cognitive load perspective.

alixedi:
I added tests for the exporter app, ran into an issue and fixed it.

I have also added the extra env vars to `example.<x>.env` files.

Finally, I have put a label on the feedback form.

PR #177, opened 2021-04-23
Flush session before logout.

alixedi:


saruniitr (approved):

r4vi:
it looks like this fixes the problem but some tests need to be updated

r4vi (approved):

PR #178, opened 2021-04-26
Only ask deactivation question of firearms, ammunition, components

eadpearce:


exporter/applications/forms/goods.py
```diff
@@ -134,11 +135,15 @@ def unit_quantity_value(request, good, sub_case_type, application_id, back_url):
             else:
                 questions = initial_questions
 
-            if firearm_type in ["ammunition", "firearms"]:
-                questions.append(firearm_proof_mark_field())
-            elif firearm_type == "components_for_firearms":
-                questions.append(does_firearm_component_require_proof_marks_field())
-            questions.append(firearm_is_deactivated_field())
+            if firearm_type in constants.FIREARM_AMMUNITION_COMPONENT_TYPES:
+                questions.append(firearm_is_deactivated_field())
```
saruniitr, line 139: I don't think the order matters here but is there any particular reason for the change of order?
Please also check with content team if they are ok with this change.

saruniitr (commented):
requested clarification, otherwise changes are good to me.

agalamatis (approved):
+1

exporter/applications/forms/goods.py
eadpearce: Yes I checked with Jessica and I think this is fine. Order change is just a side effect of the fix for duplicate questions

eadpearce: *correction: fix for wrong questions asked. not duplicates. that's the other PR I have open!

PR #181, opened 2021-04-30
[LTD-829] File upload validation

alixedi:
This PR adds a custom file upload handler by subclassing `S3FileUploadHandler` that checks the mime-type of a file before uploading using `python-magic` and throws an exception if the file has an invalid mime-type.

This addresses the attack where someone could rename `rat.html` to `rat.pdf` and upload it to our servers. Because the mime-type validation lives in a file-upload-handler, there is no way to e.g. return a nice error message to our users. This is important for the following -

This PR also gets rid of the drag-and-drop upload widget. The problem with this was that it didn't respect the `accept` clause on the regular form input and therefore a user could mistakenly drag-and-drop an invalid file type. This combined with the server-side validation would mean that the user will get a generic service error page instead of a nice error message on the form.

The custom file-upload-handler was particularly hard to test but I have tried my best.


conf/base.py
```diff
@@ -144,7 +144,19 @@
 STREAMING_CHUNK_SIZE = 8192
 S3_MIN_PART_SIZE = 5 * 1024 * 1024
 MAX_UPLOAD_SIZE = 50 * 1024 * 1024
-FILE_UPLOAD_HANDLERS = env.list("FILE_UPLOAD_HANDLERS", default=["s3chunkuploader.file_handler.S3FileUploadHandler"])
+FILE_UPLOAD_HANDLERS = env.list("FILE_UPLOAD_HANDLERS", default=["core.file_handler.SafeS3FileUploadHandler"])
+ACCEPTED_FILE_UPLOAD_MIME_TYPES = (
```
r4vi, line 148:
```python
ACCEPTED_FILE_UPLOAD_MIME_TYPES = env.list("ACCEPTED_FILE_UPLOAD_MIME_TYPES", default=[
    # Default file-types supported by LITE are pdf, doc, docx,
    # rtf, jpeg, png and tiff
    "application/pdf",
    "application/msword",
    "application/vnd.openxmlformats-officedocument.wordprocessingml.document",
    "application/rtf",
    "text/plain",
    "image/jpeg",
    "image/png",
    "image/tiff",
])
```

do we want to allow overriding with env vars?

core/file_handler.py
```diff
@@ -0,0 +1,24 @@
...
+    def receive_data_chunk(self, raw_data, start):
+        """
+        Receive a single file chunk from the browser, validate the
+        file type for the first chunk and leave the rest to super.
+        """
+        # For the first chunk
+        if start == 0:
+            mime = magic.from_buffer(raw_data, mime=True)
+            if mime not in self.ACCEPTED_FILE_UPLOAD_MIME_TYPES:
+                raise UploadFailed("Unsupported file type")
```
r4vi, line 23: we should call `self.abort` to clean up the pending multi part upload in s3

conf/base.py
r4vi: i think maybe we should add CSV too. I think the compliance returns part of lite needs to accept csv, denisals upload needs it too but i think @saruniitr is going to change the fileuploadhandler for that view

alixedi: Adding CSV to this as well. Little downside.

agalamatis (approved):
LGTM

eadpearce (approved):
lgtm

PR #182, opened 2021-04-30
LTD-801: Fix denials upload failure

saruniitr:
## Change description

Upload currently failing with the error that "File was not opened in read mode".
We are using FILE_UPLOAD_HANDLERS which defaults to S3FileUploadHandler.
This is creating a S3Boto3StorageFile instance in write mode so when
we try to read it from this, it is complaining that it is not opened in read mode.

Fix this by reading it from the s3 object that is uploaded.

### DIFF
--- a/caseworker/external_data/forms.py
+++ b/caseworker/external_data/forms.py
@@ -1,5 +1,7 @@
 from django import forms
 
+from storages.backends.s3boto3 import S3Boto3StorageFile
+
 
 class DenialUploadForm(forms.Form):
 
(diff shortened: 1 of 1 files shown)


caseworker/external_data/forms.py
```diff
@@ -11,6 +13,10 @@ def __init__(self, instance, *args, **kwargs):
 
     def clean_csv_file(self):
         value = self.cleaned_data["csv_file"]
+        if isinstance(value, S3Boto3StorageFile):
```
r4vi, line 16: did you had to add the `if isinstance` because the tests run with memoryfileuploadhandler?

r4vi (approved):

PR #183, opened 2021-05-04
LTD-508: Remove application specific details from Product detail page

saruniitr:
## Change description

Remove application specific details (Year of manufacture and Identification markings) from the product detail page on exporter side.

Noticed adding a product to product list is broken because we are sending kwargs as positional arguments resulting in incorrect values for some of the variables, fixed as part of this change.


agalamatis (approved):
+1

PR #186, opened 2021-05-05
Fix "Licence required" product mismatch

eadpearce:
Also fixes an error on the product detail page when no RFD document is uploaded

### DIFF
--- a/caseworker/templates/case/product-on-case.html
+++ b/caseworker/templates/case/product-on-case.html
@@ -88,7 +88,7 @@
 													</tr>
 													<tr class="govuk-table__row app-table__row">
 															<th scope="row" class="govuk-table__header">Licensable</th>
-															<td class="govuk-table__cell">
+															<td class="govuk-table__cell" id="is-licensable-value">
 																	{# when the product is reviewed by an internal user this value is set #}
 																	{% if good_on_application.is_good_controlled is not None %}
 																		{{ good_on_application.is_good_controlled.value }}
--- a/caseworker/templates/case/product-on-case.html
+++ b/caseworker/templates/case/product-on-case.html
@@ -88,7 +88,15 @@
 													</tr>
 													<tr class="govuk-table__row app-table__row">
 															<th scope="row" class="govuk-table__header">Licensable</th>
-															<td class="govuk-table__cell">{{ good_on_application.good.is_good_controlled.value }}</td>
+															<td class="govuk-table__cell">
+																	{# when the product is reviewed by an internal user this value is set #}
+																	{% if good_on_application.is_good_controlled is not None %}
+																		{{ good_on_application.is_good_controlled.value }}
+																	{# otherwise we use the value the exporter user has given #}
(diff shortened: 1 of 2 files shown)


caseworker/templates/case/product-on-case.html
```diff
@@ -88,7 +88,7 @@ <h2 class="govuk-heading-l">Full details</h2>
 													</tr>
 													<tr class="govuk-table__row app-table__row">
 															<th scope="row" class="govuk-table__header">Licensable</th>
-															<td class="govuk-table__cell">{{ good_on_application.is_good_controlled|yesno|title }}</td>
+															<td class="govuk-table__cell">{{ good_on_application.good.is_good_controlled.value }}</td>
```
saruniitr, line 91: I think if a product is not reviewed by case officer then `good_on_application.is_good_controlled` and `good_on_application.good.is_good_controlled` can be different because user when creating the product may think it is controlled but case officer while reviewing may change this and vice versa.
Could you please check if the value appears correctly before and after a product is reviewed?

eadpearce: Okay I think I understand. Looks like `good_on_application.is_good_controlled` is `None` until the good is reviewed then it becomes something like `{'key': 'True', 'value': 'Yes'}`. So until the good is reviewed we want to show `good_on_application.good.is_good_controlled` and after reviewing we will show `good_on_application.is_good_controlled`. Does that sound right?

saruniitr: yes that's correct.

eadpearce: Okay will update and add some tests

caseworker/templates/case/product-on-case.html
```diff
@@ -88,7 +88,15 @@ <h2 class="govuk-heading-l">Full details</h2>
 													</tr>
 													<tr class="govuk-table__row app-table__row">
 															<th scope="row" class="govuk-table__header">Licensable</th>
-															<td class="govuk-table__cell">{{ good_on_application.is_good_controlled|yesno|title }}</td>
+															<td class="govuk-table__cell">
+																	{# when the product is reviewed by an internal user this value is set #}
+																	{% if good_on_application.is_good_controlled is not None %}
+																		{{ good_on_application.is_good_controlled.value }}
```
saruniitr, line 94: Thanks for the update, I noticed you missed the `|title`, without this the first letter is not capitalized

saruniitr (approved):
lgtm. One minor change required which is missing in the update.
`|title` is missing in the value, without this the first letter will not be capitalized

caseworker/templates/case/product-on-case.html
eadpearce: The value in `{'key': 'True', 'value': 'Yes'}` is always capitalised so we don't need it

saruniitr: :+1:

PR #189, opened 2021-05-07
Fix enforcement check XML upload

eadpearce:


alixedi (approved):
Might want to move `example.xml` in the test directory but this is a beautiful PR b/c -

Django forms ✅ 
Tests ✅ 
Refactoring ✅ 

LGTM ❤️ ✨

PR #193, opened 2021-05-12
LTD-674: Display quantity of each product in the case summary page

saruniitr:
## Change description

Add changes to the display the quantity along with units on the case summary page.
Add test to ensure the template is rendered correctly.

### DIFF
--- a/caseworker/templates/case/slices/goods.html
+++ b/caseworker/templates/case/slices/goods.html
@@ -35,6 +35,7 @@
 					{% endif %}
 					<th scope="col" class="govuk-table__header">#</th>
 					<th scope="col" class="govuk-table__header">Name</th>
+					<th scope="col" class="govuk-table__header">Quantity</th>
 					{% if case.goods.0.item_type %}
 						<th scope="col" class="govuk-table__header">Type</th>
 					{% endif %}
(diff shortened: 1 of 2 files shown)


caseworker/templates/case/slices/goods.html
```diff
@@ -67,6 +68,10 @@ <h2 class="govuk-heading-m">{{ case.goods|length }} good{{ case.goods|pluralize
 							<span class="govuk-table__header" aria-hidden="false">Name</span>
 							<span data-max-length="200">{% if good.good.name %} {{ good.good.name }} {% else %} {{ good.good.description }} {% endif %}</span>
 						</td>
+						<td class="govuk-table__cell govuk-table__cell--max-width-400 {% if show_advice %}lite-!-no-border{% endif %}">
+							<span class="govuk-table__header" aria-hidden="false">Quantity</span>
+							<span id="quantity-value" data-max-length="200">{{ good.quantity }} {{ good.unit.value }}</span>
```
eadpearce, line 73: Can we use the `pluralise_quantity` template tag here or is it still dodgy? There were some bugs with it recently

saruniitr: I forgot about that. Yes we can.
@alixedi fixed that issue. Also it was because one of the key was missing in `firearm_details` but in this case `quantity` is in good on application instance so we are good.

eadpearce: Awesome!

eadpearce (approved):

PR #195, opened 2021-05-14
LTD-625: Declaration changes at end of application

saruniitr:
## Change description

Declaration text changes
New text input to capture why user cannot disclose few details
Terms and conditions acceptance modified - user need to type "I AGREE" in the text input.


alixedi:
@saruniitr Mindful of the timing but can we please sneak in some tests?

alixedi:
I am surprised that we just changed `caseworker/cases/views/main.py` without affecting any tests. Might be that we are testing the view for `200` and we can cheaply add a few assert statements to that tests so that we are covered.

alixedi (approved):

PR #196, opened 2021-05-18
Change BlocksApproval to BlocksFinalising.

alixedi:
[LTD-878](https://uktrade.atlassian.net/browse/LTD-878) Users should be able to block cases from being finalised
(ATM, they can block cases from being approved only)

This has 3 parts -

1. [lite-api](https://github.com/uktrade/lite-api/pull/768) - I think this is where the business logic lives. This is the important [commit](https://github.com/uktrade/lite-api/pull/768/commits/877b1fa92775c8b5952d58df638afc7e2bd56825). The rest is me renaming a field.
2. [lite-content](https://github.com/uktrade/lite-content/pull/282) - Because I renamed the field, the header in this CSV needed to be updated.
3. [lite-frontend](https://github.com/uktrade/lite-frontend/pull/196) - There was some logic here that needs updating. Hopefully for presentation only. I also renamed a class to reflect the change.

I would propose reviewing in this order. And commit-wise within each PR.

### DIFF
--- a/lite_content/lite_internal_frontend/flags.py
+++ b/lite_content/lite_internal_frontend/flags.py
@@ -46,8 +46,8 @@ class CreateFlagForm:
         TITLE = "Priority"
         DESCRIPTION = "This relates to the ordering of the flag. 0 is the highest."
 
-    class BlocksApproval:
-        TITLE = "Blocks application approval"
+    class BlocksFinalising:
+        TITLE = "Blocks application finalising"
         YES = "Yes"
         NO = "No"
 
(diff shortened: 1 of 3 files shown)


agalamatis (approved):

lite_content/lite_internal_frontend/flags.py
```diff
@@ -46,8 +46,8 @@ class Priority:
         TITLE = "Priority"
         DESCRIPTION = "This relates to the ordering of the flag. 0 is the highest."
 
-    class BlocksApproval:
-        TITLE = "Blocks application approval"
+    class BlocksFinalising:
+        TITLE = "Blocks application finalizing"
```
eadpearce, line 50: This is spelled "finalising" in another place. Pls correct

eadpearce (changes requested):
one spelling error but otherwise LGTM

eadpearce (approved):

PR #198, opened 2021-05-27
[DEV-888] APM

alixedi:
This will make the FE emit logs in ECS format as well so that we can 👀 them on ELK.

Also, I have made some changes for the APM.


saruniitr (approved):
lgtm

PR #203, opened 2021-06-10
Forward Zipkin headers to API.

alixedi:
So we PaaS attached Zipkin headers to all requests hitting a service. Unless these headers are already there.

So when a user e.g. browser to a Case in the FE, we get the headers for free. What is left is to attach these headers onto the API calls so that we can trace each request as it makes its way through the services.

### DIFF
--- a/core/client.py
+++ b/core/client.py
@@ -28,6 +28,8 @@ def _get_headers(request, sender=None, content_type=None):
     if "user_token" in request.session:
         headers[settings.LITE_API_AUTH_HEADER_NAME] = request.session["user_token"]
     headers["ORGANISATION-ID"] = request.session.get("organisation", "None")
+    headers["x-b3-traceid"] = request.headers.get("x-b3-traceid")
+    headers["x-b3-spanid"] = request.headers.get("x-b3-spanid")
     return headers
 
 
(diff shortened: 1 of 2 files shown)


core/client.py
```diff
@@ -28,6 +28,8 @@ def _get_headers(request, sender=None, content_type=None):
     if "user_token" in request.session:
         headers[settings.LITE_API_AUTH_HEADER_NAME] = request.session["user_token"]
     headers["ORGANISATION-ID"] = request.session.get("organisation", "None")
+    headers["x-b3-traceid"] = request.headers.get("x-b3-traceid")
```
saruniitr, line 31: The documentation using the header keys as below but here it is all lowercase, is this ok?
`X-B3-TraceId` and `X-B3-SpanId`

alixedi: Yes. Sorry, I should have put [this](https://github.com/uktrade/data-hub-api/pull/2969/files) link in here.

alixedi: P.S. `request.headers` is [case-insensitive](https://docs.djangoproject.com/en/3.2/ref/request-response/#django.http.HttpRequest.headers).

saruniitr: :+1:

saruniitr (approved):

PR #205, opened 2021-06-14
LTD-918: Feature flag out of scope features in Exporter view

saruniitr:
## Change description

Creating a CLC query and applying for PV grading is currently out of scope
and need to be removed from Exporter view. Add a feature flag to control
the visibility of these links in Exporter view.


unit_tests/core/test_templates.py
```diff
@@ -0,0 +1,31 @@
...
+        "FEATURE_FLAG_ONLY_ALLOW_SIEL": True,
+        "FEATURE_FLAG_ALLOW_CLC_QUERY_AND_PV_GRADING": allow_clc_query_pv_grading,
+    }
+
+    html = render_to_string("core/hub.html", context)
+    soup = BeautifulSoup(html, "html.parser")
+    clc_query_link = soup.find(id="clc_query_link_value")
+    pv_grading_link = soup.find(id="pv_grading_link_value")
+    if allow_clc_query_pv_grading:
+        clc_query_link = clc_query_link.text.replace("\n", "").replace("\t", "")
```
eadpearce, line 25: Does using `.strip()` work here?

saruniitr: No because it removes leading and trailing characters but not in the middle hence using `replace`.

eadpearce: You mean there are newlines/tabs in the actual content string?

saruniitr: They are not there in this case but in some places I noticed the newline/tabs were there in the middle of rendered string, I have used `strip` here now.

eadpearce: That's very weird. Might be a bug with the template tag

eadpearce (approved):

PR #208, opened 2021-06-16
LTD-914: Add browser test to add a product to an existing application

saruniitr:


ui_tests/exporter/conftest.py
```diff
@@ -94,8 +98,17 @@ def i_click_edit_application(driver):  # noqa
     ApplicationPage(driver).click_edit_application_link()
 
 
-@given("I go to exporter homepage and choose Test Org")  # noqa
-def go_to_exporter(driver, register_organisation, sso_sign_in, exporter_url, context):  # noqa
+@given("I signin and go to exporter homepage and choose Test Org")  # noqa
+def go_to_exporter(driver, register_organisation, sso_sign_in, exporter_url, context, exporter_info):  # noqa
+    driver.get(exporter_url)
+    if driver.current_url != "pick-organisation":
```
alixedi, line 104: I think `driver.current_url` would never be "equal" to `pick-organisation`.

Shouldn't we put this in an else clause under L112?

(which checks `pick-organisation in driver.current_url` and not equal)

ui_tests/exporter/features/register_an_organisation.feature
```diff
@@ -10,3 +10,4 @@ Feature: I want to register an organisation
     When I sign in as user without an organisation registered
     And I enter my information from steps 1-4 to register a commercial organisation
     Then the organisation is registered successfully
+    And I logout
```
alixedi, line 13: 👨‍🍳

ui_tests/exporter/conftest.py
saruniitr: Yes if the state is properly cleared in all of the tests then it won't be equal to `pick-organisation`.
I think it is best to remove the if condition itself on L104.
As the step itself now says "I signin and go to exporter homepage" we can signin unconditionally then L112 matches and test proceeds with selecting the organisation.

ui_tests/exporter/features/register_an_organisation.feature
saruniitr: sorry cannot follow the emoji, need any more changes here?

ui_tests/exporter/conftest.py
alixedi: Ok but wouldn't that mean we will have to make sure we log out after every test?

ui_tests/exporter/features/register_an_organisation.feature
alixedi: Sorry. I will upgrade you to a gif -

tenor-2

(TBC - I liked that you wrote this clause - super reusable!)

ui_tests/exporter/conftest.py
saruniitr: Yes to be really sure we don't carry any state, I am open to make it light weighted instead of completely logging out but cannot think of any other option.
Within a particular feature file we may be okay not to logout but may cause issues across feature files. May be we will refine this as we add more and more tests?

alixedi: Sounds good my friend.

alixedi (approved):
Beautiful. LGTM!

PR #209, opened 2021-06-17
[LTD-911] Open SIEL application

alixedi:
This includes a bunch of fixes so I would recommend reviewing commit-wise.

### DIFF
--- a/ui_tests/exporter/features/siel_firearm_application.feature
+++ b/ui_tests/exporter/features/siel_firearm_application.feature
@@ -8,7 +8,7 @@ Feature: I want to be able to submit SIEL firearm applications
     Then I see the application overview
     And I logout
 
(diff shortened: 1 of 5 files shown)


ui_tests/exporter/features/siel_firearm_application.feature
```diff
@@ -8,6 +8,14 @@ Feature: I want to be able to submit SIEL firearm applications
     Then I see the application overview
     And I logout
 
+  Scenario: Edit and open a SIEL application
```
saruniitr, line 11: We are actually not editing anything - otherwise the changes looks good to me.

alixedi: Yes, we should change this. I picked it off the ticket without thinking about it.

saruniitr (approved):

PR #210, opened 2021-06-18
LTD-925-927: Feature flag CLC and PV query when adding product

saruniitr:
## Change Description

This PR makes two changes to the form of adding a product to the product list.
- If the user is unsure about the CLC value there is an option to select "I don't know". In this case we create a CLC query which is checked by the case officer but this process is not scoped yet hence this option is completely removed from code base.
- Similarly when answering PV grading user has an option to select "it needs one" which will create a PV grading query. As this is also not scoped yet this option is removed from code base.


exporter/goods/views.py
```diff
@@ -484,7 +484,7 @@ def get_success_url(self):
             )
         elif not good.get("documents") and not good.get("is_document_available"):
             return reverse_lazy("goods:add_document", kwargs={"pk": self.object_pk})
-        elif raise_a_clc_query or raise_a_pv_query:
+        elif settings.FEATURE_FLAG_ALLOW_CLC_QUERY_AND_PV_GRADING and (raise_a_clc_query or raise_a_pv_query):
```
saruniitr, line 487: These changes are additional precautions because the options to make variables in L477 and L488 true are completely removed from the form and user cannot select them.

unit_tests/exporter/goods/test_views.py
```diff
@@ -74,3 +77,158 @@ def test_get_good_detail_doesnot_contain_application_specific_details(authorized
     assert ("Controlled" in response_html) == True
     assert ("Year of manufacture" in response_html) == False
     assert ("Identification markings" in response_html) == False
+
+
+def test_edit_grading_doesnot_raise_goods_query(authorized_client, requests_mock, good):
+    good["is_document_available"] = True
```
saruniitr, line 83: All of these tests provide request data such that under normal circumstances they trigger creation of CLC query and PV grading query. As part of test we check with the feature flag true/false and ensure existing behaviour is maintained if the flag is true and also ensure no query raised if the flag is false.

alixedi (changes requested):
I do not think we need a feature flag for this.

If we want to remove radio buttons, we should remove them and update the code accordingly.

I understand why you put a feature flag - basically, we might need this in the future. My argument is that (1) the feature flag is complicating this 10x and (2) everything is in Git anyway so if/when we need this we just need to revert the PR.

unit_tests/core/test_templates.py
```diff
@@ -3,29 +3,17 @@
...
-    if allow_clc_query_pv_grading:
-        clc_query_link = clc_query_link.text.strip("\n\t")
-        pv_grading_link = pv_grading_link.text.strip("\n\t")
-        assert clc_query_link == "Create a control list classification (CLC) query"
-        assert pv_grading_link == "Apply for a security grading"
-    else:
-        assert clc_query_link == None
-        assert pv_grading_link == None
+    product_list_tile = soup.find(id='product-list-tile')
+    assert len(product_list_tile.findChildren()) == 4
```
saruniitr, line 19: The links to create CLC query and PV grading query were also behind the feature flag.
I have removed these as well now hence only checking for the number of children for that div and I cannot check for the presence of links that doesn't exist.

alixedi (approved):
Brilliant! TY for working with me on this @saruniitr 🙏

PR #211, opened 2021-06-21
LTD-920-921-922: Add UI test that verifies the End use, Locations and Route of goods sections in application

saruniitr:
## Change description

Add test scenarios that verify the filling of End use, Locations and Route of goods section in Application.
Most of the steps are reused and additional steps are added to assert things.

All of these scenarios are very similar hence included in the same PR. We could add Consignee and End user sections also but since they are related to each other they will be done as a separate PR.

### DIFF
--- a/ui_tests/exporter/conftest.py
+++ b/ui_tests/exporter/conftest.py
@@ -203,9 +203,19 @@ def add_new_party(driver, type, name, website, address, country, context):  # no
     functions.click_submit(driver)
 
 
-@when(parsers.parse('I click on the "{section}" section'))  # noqa
-def go_to_task_list_section(driver, section):  # noqa
-    TaskListPage(driver).click_on_task_list_section(section)
+@when(parsers.parse('I click on the "{section_name}" section'))  # noqa
+def go_to_task_list_section(driver, section_name):  # noqa
+    section = driver.find_element_by_link_text(section_name)
+    section_id = section.get_attribute("id")
+    TaskListPage(driver).click_on_task_list_section(section_id)
+
+
+@then(parsers.parse('the section "{section_name}" is now saved'))  # noqa
+def verify_section_is_saved(driver, section_name):  # noqa
+    section = driver.find_element_by_link_text(section_name)
+    section_id = section.get_attribute("id")
+    saved_status_element = driver.find_element_by_id(f"{section_id}-status")
+    assert saved_status_element.text == "SAVED"
 
 
 @when(parsers.parse("I provide details of the intended end use of the products"))  # noqa
@@ -245,13 +255,15 @@ def suspected_wmd_end_use_details(driver, choice):  # noqa
     functions.click_submit(driver)
 
 
-@when(parsers.parse('I answer "{choice}" for shipping air waybill or lading'))  # noqa
+@when(parsers.parse('I answer "{choice}" for shipping air waybill or lading and Save'))  # noqa
 def route_of_goods(driver, choice):  # noqa
     route_of_goods = RouteOfGoodsFormPage(driver)
     if choice == "No":
         route_of_goods.answer_route_of_goods_question(True, "Not shipped air waybill.")
     else:
         route_of_goods.answer_route_of_goods_question(False)
+
+    driver.find_element_by_css_selector("button[type='submit']").click()
 
 
 @when(parsers.parse("I save and continue on the summary page"))  # noqa
--- a/ui_tests/exporter/conftest.py
+++ b/ui_tests/exporter/conftest.py
@@ -263,7 +263,7 @@ def route_of_goods(driver, choice):  # noqa
     else:
         route_of_goods.answer_route_of_goods_question(False)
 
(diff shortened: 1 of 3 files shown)


ui_tests/exporter/step_defs/test_standard_application.py
```diff
@@ -248,6 +248,17 @@ def i_see_the_application_overview(driver, context):  # noqa
     context.app_id = app_id
 
 
+@when(parsers.parse('I am on the application overview page entitled "{title}"'))  # noqa
+def i_see_the_application_overview(driver, title):  # noqa
+    heading = driver.find_element_by_xpath("//h1").text
+    assert heading == title
```
alixedi, line 254: This seems like a dupe of - https://github.com/uktrade/lite-frontend/blob/f29819ac01d40b07007e07da8a551558c123090b/ui_tests/exporter/step_defs/test_standard_application.py#L243-L249

It is ok if we want to move over but we should probably delete ☝️

saruniitr: It is sort of the same but the scenario was asserting the title hence created a separate function.
The above also uses context and wanted to move away as much as possible to reduce the coupling.
I will delete that if that is not used anywhere.

alixedi: ATM there are 2 functions in the `test_standard_application` module with the same name. 

I understand that you are trying to avoid carrying the state in `context` but in this case, I am pretty sure `app_id` will be required by the callers of the old `i_see_the_application_overview` function and so it might be worth just using that.

saruniitr: Ah right that was me forgetting to rename the function. I have renamed the new function and deleted the old one `i_see_the_application_overview` as it is not called from anywhere.

ui_tests/exporter/step_defs/test_standard_application.py
```diff
@@ -239,13 +239,15 @@ def create_standard_application(driver, export_type, context):  # noqa
...
+@when(parsers.parse('I am on the application overview page entitled "{title}"'))  # noqa
+def i_am_on_application_overview_with_title(driver, title):  # noqa
+    heading = driver.find_element_by_xpath("//h1").text
+    assert heading == title
 
-    app_id = driver.current_url.split("/")[-3]
-    context.app_id = app_id
+
+@then(parsers.parse('I should be taken to the application overview page entitled "{title}"'))  # noqa
+def taken_to_application_overview_page(driver, title):
```
saruniitr, line 249: The library doesn't support re-using the same step otherwise we could've simply added this decorator on top of the above function hence calling the function from a new step.

alixedi:
I re-ran `caseworker_ui_test` but its still failing.

saruniitr:
Yes I think it is to do with SSO UAT. All of these were passing yesterday and no code changes from that point.
This is affecting other PR also.

ui_tests/exporter/conftest.py
```diff
@@ -245,14 +255,16 @@ def suspected_wmd_end_use_details(driver, choice):  # noqa
...
-@when(parsers.parse('I answer "{choice}" for shipping air waybill or lading'))  # noqa
+@when(parsers.parse('I answer "{choice}" for shipping air waybill or lading and Save'))  # noqa
 def route_of_goods(driver, choice):  # noqa
     route_of_goods = RouteOfGoodsFormPage(driver)
     if choice == "No":
         route_of_goods.answer_route_of_goods_question(True, "Not shipped air waybill.")
     else:
         route_of_goods.answer_route_of_goods_question(False)
 
+    driver.find_element_by_css_selector("button[type='submit']").click()
```
alixedi, line 266: You can use -

https://github.com/uktrade/lite-frontend/blob/f6a9c7feda28722b27cda848d69d02c968a59d3a/tests_common/functions.py#L9-L13

ui_tests/exporter/conftest.py
```diff
@@ -245,14 +255,16 @@ def suspected_wmd_end_use_details(driver, choice):  # noqa
     functions.click_submit(driver)
 
 
-@when(parsers.parse('I answer "{choice}" for shipping air waybill or lading'))  # noqa
+@when(parsers.parse('I answer "{choice}" for shipping air waybill or lading and Save'))  # noqa
```
alixedi, line 258: `lading` here might be a typo?

alixedi (approved):
Couple of comments that I will leave with you.

ui_tests/exporter/conftest.py
saruniitr: Initially I thought the same but the question itself is like that. I also checked that it is a business term (lading - action of loading a ship with cargo)

ui_tests/exporter/conftest.py
saruniitr: :+1:

ui_tests/exporter/conftest.py
alixedi: TIL 👍

PR #212, opened 2021-06-22
Add test for temporary export details.

alixedi:


ui_tests/exporter/conftest.py
```diff
@@ -264,7 +265,7 @@ def save_continue_summary_list(driver):  # noqa
 @when("I provide details of why my export is temporary")  # noqa
 def enter_temporary_export_details(driver):  # noqa
     temporary_export_details = TemporaryExportDetailsFormPage(driver)
-    temporary_export_details.answer_temp_export_details(fake.sentence(nb_words=30))
+    temporary_export_details.answer_temp_export_details("Lorem ipsum")
```
alixedi, line 268: I have found that generally replacing `fake` with static values mean that we don't have to carry the state in `context` for `assert` down the line.

saruniitr: In this case we are not saving the fake sentence in context so it is fine but replacing with static text is also fine for me which is useful in assertions.

saruniitr (approved):

PR #213, opened 2021-06-22
LTD-929: Caseworker UI test to view case details

saruniitr:
## Change description

UI test to validate the application details from the caseworker site.
The test creates an application with a product and verifies the product name and rating, End user and consignee details and intended use details.

### DIFF
--- a/ui_tests/caseworker/conftest.py
+++ b/ui_tests/caseworker/conftest.py
@@ -51,6 +51,7 @@ from tests_common.fixtures.urls import internal_url, sso_sign_in_url, api_url  #
 import tests_common.tools.helpers as utils
 from ui_tests.caseworker.pages.case_list_page import CaseListPage
 from ui_tests.caseworker.pages.application_page import ApplicationPage
+from tests_common.helpers import applications
 
 
 @when("I go to the internal homepage")  # noqa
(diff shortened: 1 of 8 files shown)


ui_tests/caseworker/conftest.py
```diff
@@ -143,7 +175,7 @@ def see_queue_in_queue_list(driver, context):  # noqa
     assert driver.find_element_by_id(context.case_id).is_displayed()
 
 
-@when("I show filters")  # noqa
+@when("I click on show filters")  # noqa
```
eadpearce, line 178: why do some of these lines have `#noqa` at the end?

saruniitr: I think it is suppress linter errors, usually these are copy pasted hence you see here. I will see if these can be removed.

ui_tests/caseworker/features/view_standard_application.feature
```diff
@@ -4,6 +4,25 @@ Feature: I want to view the case details of a case
...
+    And I filter by application type "Standard Individual Export Licence"
+    Then I should see my case in the cases list
+    When I go to application previously created
+    Then I should see the product name as "Rifle" with product rating as "PL9002"
+    And the "Consignee" name is "Automated Consignee", address is "1234, Trade centre", country is "Belgium"
+    And the "End user" name is "Automated End user", address is "1234, High street", country is "Belgium"
+
+    Examples:
+    | name    | product | clc_rating  | end_user_name      | end_user_address  | country | consignee_name      | consignee_address   | end_use                  |
+    | Test    | Rifle   | PL9002      | Automated End user | 1234, High street | BE      | Automated Consignee | 1234, Trade centre  | Research and development |
```
eadpearce, line 23: VERY nitpicky comment: PL9002 is for explosives, PL9010 would be firearms

saruniitr: This is what test requested, technically it can be any clc entry.
I will check with ticket creator and see if this can be changed to something else.

eadpearce: No need to change it! This was more of a nerdy FYI than anything else

eadpearce (approved):

PR #216, opened 2021-06-28
Expiry validation follow-up

eadpearce:


dgnball:
So this seems to fill in some defaults for attaching a new certificate? What's the rational, and how does this address LTD-855? Do we need new tests for this change?

eadpearce:
@dgnball so basically, the two names added are keys that the api is expecting in the submitted data so it can run it through the serializer for validation (https://github.com/uktrade/lite-api/pull/769). they were added in the other form for the registered firearms dealer certificate but missed in this one. I'm not sure how useful a unit test for this would be since the whole functionality bridges the frontend and the api. it'd need to be a browser test

dgnball (approved):

PR #217, opened 2021-06-29
Ltd 924 consignee

dgnball:


ui_tests/exporter/features/siel_firearm_application.feature
```diff
@@ -43,6 +43,16 @@ Feature: I want to be able to submit SIEL firearm applications
     Then I see the end user summary
     And I logout
 
+  Scenario: Enter consignee details in the application
+    Given I signin and go to exporter homepage and choose Test Org
+    When I create a standard application of a "permanent" export type
+    And I am on the application overview page entitled "Standard Individual Export Licence"
+    And I click on the "Consignee" section
+    And I add a party of sub_type: "government", name: "Mr Smith", website: "https://www.smith.com", address: "London" and country "Ukraine"
```
saruniitr, line 51: Please add a step whether we select existing or reuse party (as in above scenario)
`And I answer "No" for whether I want to reuse an existing party`

`And I add a party of sub_type: "government", name: "Mr Smith", website: "https://www.smith.com", address: "London" and country "Ukraine"`
This should also be similar to above otherwise we will have different convention

ui_tests/exporter/features/siel_firearm_application.feature
```diff
@@ -43,6 +43,16 @@ Feature: I want to be able to submit SIEL firearm applications
...
+    And I upload a file "file_for_doc_upload_test_1.txt"
+    Then download link is present
```
saruniitr, line 53: Is there no consignee summary screen like in above scenario?
`Then I see the end user summary`

End user and consignee are very similar.

Finally, please move consignee test above end user as that is the logical order.
In the application overview section End user comes first but there is a ticket in the backlog to change this order.

saruniitr (changes requested):

ui_tests/exporter/features/siel_firearm_application.feature
```diff
@@ -28,6 +28,22 @@ Feature: I want to be able to submit SIEL firearm applications
     Then I see the temporary export detail summary
     And I logout
 
+  @an_exporter_test
```
saruniitr, line 31: probably forgot to remove this generic tag?

saruniitr (approved):
one minor comment to be addressed otherwise looks good to me.

ui_tests/exporter/features/siel_firearm_application.feature
dgnball: Ok, it looks more like the 'end user' section now

ui_tests/exporter/features/siel_firearm_application.feature
dgnball: Ok, I have moved the test and added a 'Then I see the consignee summary' section

ui_tests/exporter/features/siel_firearm_application.feature
dgnball: yes I did. Removed.

PR #218, opened 2021-06-29
LTD-933: Add UI test to verify reviewing a product in an application

saruniitr:
## Change description

Adds a test to verify the process of product review by a caseworker.
The test creates and submit an application, caseworker during review updates the control list entry and it ensures it is updated and the product status becomes verified after review.

### DIFF
--- a/ui_tests/caseworker/conftest.py
+++ b/ui_tests/caseworker/conftest.py
@@ -5,7 +5,7 @@ import time
 from ui_tests.caseworker.pages.advice import FinalAdvicePage, TeamAdvicePage
 from ui_tests.caseworker.pages.case_page import CasePage, CaseTabs
 from ui_tests.caseworker.pages.goods_queries_pages import StandardGoodsReviewPages, OpenGoodsReviewPages
-
+from ui_tests.caseworker.pages.teams_pages import TeamsPages
 from caseworker.core.constants import DATE_FORMAT
 from ui_tests.caseworker.fixtures.env import environment  # noqa
 from ui_tests.caseworker.fixtures.add_a_flag import (  # noqa
@@ -110,6 +110,15 @@ def create_application(
     applications.create_standard_application(api_test_client, context, app_data)
 
 
+@then(parsers.parse('I should see the product name as "{product_name}" with product rating as "{clc_rating}"'))
+def check_product_name_and_rating(driver, product_name, clc_rating):  # noqa
+    product_table = driver.find_element_by_id("table-goods")
+    name_element = product_table.find_element_by_xpath("//tbody/tr/td[3]")
+    rating_element = product_table.find_element_by_xpath("//tbody/tr/td[6]")
+    assert name_element.text == product_name
+    assert rating_element.text == clc_rating
+
+
 @given("I create open application or open application has been previously created")  # noqa
 def create_open_app(driver, apply_for_open_application):  # noqa
     pass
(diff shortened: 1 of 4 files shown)


ui_tests/caseworker/step_defs/test_view_standard_application.py
```diff
@@ -69,21 +69,6 @@ def i_see_inactive_party(driver, context):
         assert destination["name"] in destinations_table_text
 
 
-@when(parsers.parse('I filter by application type "{application_type}"'))
```
saruniitr, line 72: these are moved to `conftest.py` because they are used in multiple feature files

ui_tests/caseworker/conftest.py
```diff
@@ -195,6 +210,31 @@ def case_list_page(driver, internal_url):  # noqa
...
+    teams_page.select_team_from_dropdown(team)
+    teams_page.select_default_queue_from_dropdown(queue)
+    functions.click_submit(driver)
+
+
+@when("I go to my case list")  # noqa
+def get_my_case_list(driver):  # noqa
+    """
+    Clicks on the menu and selects Cases
+    Depending on team, default queue the list of cases will be different
```
saruniitr, line 232: It can be list of all cases or list of open cases or list of team's cases etc

ui_tests/caseworker/features/review_goods.feature
```diff
@@ -5,6 +5,28 @@ Feature: I want to review, amend where required and confirm the goods ratings an
   So that I can confirm the goods are correctly described
 
 
+  Scenario: Gov user can review product in an application
+    Given I sign in to SSO or am signed into SSO
+    And I create an application with <name>,<product>,<clc_rating>,<end_user_name>,<end_user_address>,<consignee_name>,<consignee_address>,<country>,<end_use>
```
dgnball, line 10: Typo. end_use should be end_user

dgnball: I would have expected that a test would be specific with the values. Not sure what the angle bracket convetion is but I would prefer something along the lines of "a name of 'test application 1'"

ui_tests/caseworker/features/review_goods.feature
```diff
@@ -5,6 +5,28 @@ Feature: I want to review, amend where required and confirm the goods ratings an
   So that I can confirm the goods are correctly described
 
 
+  Scenario: Gov user can review product in an application
```
dgnball, line 8: I'm guessing this works but according to https://pypi.org/project/pytest-bdd/ (and other Gherkin docs) you should be using a Scenario Outline if you want to example tables

dgnball:
If I look in the CircleCI logs, I can't see that test_gov_user_... has been run

ui_tests/caseworker/features/review_goods.feature
saruniitr: No, that is actually end use. The intended end use of products

ui_tests/caseworker/features/review_goods.feature
saruniitr: Yes but this is how Scenarios are identified and collected even if there are no example tables.
I have chosen to use table in this case instead of defining values within the step because if we want to create multiple applications as part of this test it is easier with table (just add another row).

dgnball (approved):

PR #222, opened 2021-07-02
[LTD-926] Submit SIEL application

alixedi:
Adds a test for submitting SIEL application.

### DIFF
--- a/ui_tests/exporter/conftest.py
+++ b/ui_tests/exporter/conftest.py
@@ -1193,7 +1193,7 @@ def i_see_application_summary(driver, clc_rating, end_use, end_user_name, consig
 
 @when("I agree to declaration")
 def i_agree(driver):  # noqa
-    driver.find_element_by_id("agreed_to_declaration_text").send_keys('I AGREE')
+    driver.find_element_by_id("agreed_to_declaration_text").send_keys("I AGREE")
     functions.click_submit(driver)
 
 
--- a/ui_tests/exporter/conftest.py
+++ b/ui_tests/exporter/conftest.py
@@ -1,4 +1,3 @@
-from conf import exporter
 import os
 
 from django.utils import timezone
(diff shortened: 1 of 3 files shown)


ui_tests/exporter/features/siel_firearm_application.feature
```diff
@@ -125,3 +125,14 @@ Feature: I want to be able to submit SIEL firearm applications
     Then I should be taken to the application overview page entitled "Standard Individual Export Licence"
     And the section "Locations" is now saved
     And I logout
+
+  Scenario: Submit Application
+    Given I signin and go to exporter homepage and choose Test Org
+    And I create an application with <name>,<product>,<clc_rating>,<end_user_name>,<end_user_address>,<consignee_name>,<consignee_address>,<country>,<end_use>
+    And I navigate to application summary
+    Then I see the application summary
```
saruniitr, line 133: I think we can refer to the rating, end_use using the `<>` syntax here and verify otherwise if we change the values here then it breaks the assert at https://github.com/uktrade/lite-frontend/pull/222/files#diff-f87a545d64b7d553987a9192348e420854cf340b24cea13cc0a06974585866f1R1188

ui_tests/exporter/conftest.py
```diff
@@ -1140,3 +1142,50 @@ def i_see_consignee_summary(driver):  # noqa
...
+        "product": product,
+        "clc_rating": clc_rating,
+        "end_user_name": end_user_name,
+        "end_user_address": end_user_address,
+        "consignee_name": consignee_name,
+        "consignee_address": consignee_address,
+        "country": country,
+        "end_use": end_use,
+    }
+    applications.create_standard_application(api_test_client, context, app_data, submit=False)
```
saruniitr, line 1175: if we add application type then we can configure `permanent` or `temporary` also, probably need some changes to the data in `create_standard_application` also but for the purpose of this test I think it is fine.

alixedi: Yes, I want us to fix this but maybe on a different ticket.

dgnball (approved):

ui_tests/exporter/conftest.py
saruniitr: ok

ui_tests/exporter/features/siel_firearm_application.feature
```diff
@@ -125,3 +125,14 @@ Feature: I want to be able to submit SIEL firearm applications
     Then I should be taken to the application overview page entitled "Standard Individual Export Licence"
     And the section "Locations" is now saved
     And I logout
+
+  Scenario: Submit Application
+    Given I signin and go to exporter homepage and choose Test Org
+    And I create an application with <name>,<product>,<clc_rating>,<end_user_name>,<end_user_address>,<consignee_name>,<consignee_address>,<country>,<end_use>
+    And I navigate to application summary
+    Then I see the application summary with <clc_rating>,<end_use>,<end_user_name>,<consignee_name>
```
saruniitr, line 133: I think after this we get to declaration page where user need to enter few more details, these are also required for this test because the application is not yet submitted.

saruniitr (changes requested):

ui_tests/exporter/features/siel_firearm_application.feature
alixedi: Ah! good point. I will add a step for that.

saruniitr (approved):
looks good to me

PR #228, opened 2021-07-09
LTD-931: Add browser test to verify exporting of Enforcement check xml

saruniitr:


ui_tests/caseworker/conftest.py
```diff
@@ -606,3 +607,38 @@ def licence_audit(driver, context, internal_url):  # noqa
...
+    queue_checkbox = target_queue.find_element_by_xpath(".//input")
+    if check:
+        if not queue_checkbox.is_selected():
+            queue_checkbox.click()
+    else:
+        if queue_checkbox.is_selected():
+            queue_checkbox.click()
+
+
+@when(parsers.parse('I assign the case to "{queue}" queue'))  # noqa
```
saruniitr, line 633: We need to assign the case to the Enforcement queue then when they are exported all those cases in this queue are exported. If no cases found then API throws an error page hence the case is assigned here and as a clean up it is removed from the queue again below. If we don't remove then lots of cases will end up in this queue.

eadpearce (approved):

saruniitr:
Thanks @eadpearce

PR #230, opened 2021-07-16
LTD-932: Add browser test to verify importing of enforcement check xml

saruniitr:


ui_tests/caseworker/features/enforcement.feature
```diff
@@ -14,15 +15,42 @@ Feature: I want to export and import XML for enforcement checking
     And I change my team to "Enforcement Unit" and default queue to "Enforcement Unit Cases to Review"
     And I go to my case list
     When I click export enforcement xml
+    And I go to application previously created
```
saruniitr, line 18: This step is not really needed but there is a delay between clicking the export button and the file being downloaded. Without this step the check was failing in the next step which checks for the downloaded file. Instead of adding a delay interms of sleep I have added this step which is required again for the step on L26.

ui_tests/caseworker/features/enforcement.feature
```diff
@@ -14,15 +15,42 @@ Feature: I want to export and import XML for enforcement checking
...
-    And I go to application previously created
     And I remove the case from "Enforcement Unit Cases to Review" queue
+    And I cleanup the temporary files created
+
+    Examples:
+    | name    | product | clc_rating  | end_user_name      | end_user_address  | country | consignee_name      | consignee_address   | end_use                  |
+    | Test    | Rifle   | PL9002      | Automated End user | 1234, High street | BE      | Automated Consignee | 1234, Trade centre  | Research and development |
+
+
+  Scenario: Import xml file after enforcement checks
```
saruniitr, line 34: This scenario first creates an application and exports the file. The import file is generated based on the values in the exported file.
As a cleanup this step clears the applications created in these scenarios from the "Enforcement Unit Cases to Review" queue so that they won't interfere in the next run. All the temporary files created are also deleted. This is also as a precaution because on CI each run will be on a different system so it is not a problem but can be an issue when running locally.

ui_tests/caseworker/features/enforcement.feature
```diff
@@ -14,15 +15,42 @@ Feature: I want to export and import XML for enforcement checking
...
+    And I assign the case to "Enforcement Unit Cases to Review" queue
+    When I go to my profile page
+    And I change my team to "Enforcement Unit" and default queue to "Enforcement Unit Cases to Review"
+    And I go to my case list
+    When I click export enforcement xml
+    And I go to application previously created
+    Then the file "enforcement_check.xml" is downloaded
+    And the downloaded file should include "CONSIGNEE" "COUNTRY" as "Belgium"
+    And the downloaded file should include "END_USER" "COUNTRY" as "Belgium"
+    When I include "CONSIGNEE" details to generate import file
```
saruniitr, line 48: We can use any other entity in the file (eg END_USER or ULTIMATE_END_USER) but consignee is chosen as he will be the exporter.

ui_tests/caseworker/features/enforcement.feature
```diff
@@ -14,15 +15,42 @@ Feature: I want to export and import XML for enforcement checking
...
+    And I go to my case list
+    And I import the generated enforcement check xml file
+    And I go to application previously created
+    Then the application is removed from "Enforcement Unit Cases to Review" queue
+    And I cleanup the temporary files created
```
saruniitr, line 53: cleanup steps

ui_tests/caseworker/step_defs/test_enforcement.py
```diff
@@ -35,3 +36,67 @@ def enforcement_file_content_check(party_type, tag, value):
     nodes = root.findall(f".//STAKEHOLDER[SH_TYPE='{party_type}']")
     party_values = set([child.text for node in nodes for child in node.getchildren() if child.tag == tag])
     assert value in party_values
+
+
+# @given(parsers.parse('I include "{party_type}" details to generate import file"'))
```
dgnball, line 41: Remove commented out code

dgnball (changes requested):

ui_tests/caseworker/step_defs/test_enforcement.py
saruniitr: ok forgot

dgnball (approved):

PR #234, opened 2021-07-28
Login redirect fix 2

eadpearce:


core/auth/views.py
```diff
@@ -70,7 +84,30 @@ def redirect_to_login(self):
...
+        return super().dispatch(self.request, *args, **kwargs)
+
     def dispatch(self, request, *args, **kwargs):
+        # is user logged in to SSO?
         if not self.request.authbroker_client.authorized:
             return self.redirect_to_login()
+
+        if settings.SERVICE_NAME == "lite-internal-frontend":
+            return self.auth_internal_user(self, *args, **kwargs)
+
```
alixedi, line 112: Is putting `auth_internal_user` on the base class a good idea?

I think you are having to check for `SERVICE_NAME` because of this.

alixedi (changes requested):
I think we shouldn't put `auth_internal_user` on the base class that is meant to be generic.

caseworker/auth/helpers.py
```diff
@@ -0,0 +1,17 @@
...
+    session["first_name"] = user_profile["first_name"]
+    session["last_name"] = user_profile["last_name"]
+    session["default_queue"] = data["default_queue"]
+    session["user_token"] = data["token"]
+    session["lite_api_user_id"] = data["lite_api_user_id"]
+    session["email"] = user_profile["email"]
+    session.save()
+
+
+def save_exporter_user_info_to_session(session, data, user_profile):
```
saruniitr, line 11: This is not used. We would also need `ExporterLoginRequiredMixin` isn't it which is where this will be used?

eadpearce: Good spot. Have removed this now

saruniitr: :+1:

saruniitr (approved):
lgtm

alixedi:
I think we are doing 2 things in this PR. We can talk about the utility and importance of each one of these -

1) The bug - getting `KeyError` on login

This can be resolved by replacing the contents of `caseworker/auth/services.py` with the following -

```
from core import client
from lite_content.lite_internal_frontend import strings
from lite_forms.generators import error_page


def authenticate_gov_user(request, json):
    resp = client.post(request, "/gov-users/authenticate/", json)
    if resp.status_code == 403:
        return error_page(
            request,
            title=strings.Authentication.UserDoesNotExist.TITLE,
            description=strings.Authentication.UserDoesNotExist.DESCRIPTION,
            show_back_link=False,
        )

    elif resp.status_code != 200:
        return error_page(request, show_back_link=False)

    return resp.json(), resp.status_code
```

(I have just taken the code from this PR and placed it in a place that allows me to not have to change a bunch of other things)

2) The refactor that introduced `CaseworkerLoginRequiredMixin`

This may be a good change but without a corresponding `ExporterLoginRequiredMixin`, it seems to add to the confusion.

Now, I am not suggesting that we put in the proposed `ExporterLoginRequiredMixin` in this PR. I am questioning whether putting in `CaseworkerLoginRequiredMixin` makes the code easier to follow for other people - in which case, happy to run with this and also put in `ExporterLoginRequiredMixin` in a following PR.

alixedi (commented):
I think we are doing 2 things in this PR. We can talk about the utility and importance of each one of these -

1) The bug - getting `KeyError` on login

This can be resolved by replacing the contents of `caseworker/auth/services.py` with the following -

```
from core import client
from lite_content.lite_internal_frontend import strings
from lite_forms.generators import error_page


def authenticate_gov_user(request, json):
    resp = client.post(request, "/gov-users/authenticate/", json)
    if resp.status_code == 403:
        return error_page(
            request,
            title=strings.Authentication.UserDoesNotExist.TITLE,
            description=strings.Authentication.UserDoesNotExist.DESCRIPTION,
            show_back_link=False,
        )

    elif resp.status_code != 200:
        return error_page(request, show_back_link=False)

    return resp.json(), resp.status_code
```

(I have just taken the code from this PR and put it in a place that lets us get away without having to make big changes)

2) The refactor that introduced `CaseworkerLoginRequiredMixin` 

This may be a good change but without a corresponding `ExporterLoginRequiredMixin`, it seems to add to the confusion.

Now, I am not suggesting that we put in the proposed `ExporterLoginRequiredMixin` in this PR. I am questioning whether putting in `CaseworkerLoginRequiredMixin` makes the code easier to follow for other people?

If Yes, happy to run with this and also put in `ExporterLoginRequiredMixin` in a following PR.

If No, can we please just change `authenticate_gov_user `?

eadpearce:
@alixedi 
> This can be resolved by replacing the contents of `caseworker/auth/services.py` with the following -
> 
> ```
> from core import client
> from lite_content.lite_internal_frontend import strings
> from lite_forms.generators import error_page
> 
> 
> def authenticate_gov_user(request, json):
>     resp = client.post(request, "/gov-users/authenticate/", json)
>     if resp.status_code == 403:
>         return error_page(
>             request,
>             title=strings.Authentication.UserDoesNotExist.TITLE,
>             description=strings.Authentication.UserDoesNotExist.DESCRIPTION,
>             show_back_link=False,
>         )
> 
>     elif resp.status_code != 200:
>         return error_page(request, show_back_link=False)
> 
>     return resp.json(), resp.status_code
> ```
> 
> (I have just taken the code from this PR and put it in a place that lets us get away without having to make big changes)

I didn't want to do this because [this line](https://github.com/uktrade/lite-frontend/blob/master/core/auth/views.py#L57) is expecting 2 values to be returned by `authenticate_gov_user` and I wanted to keep the changes as small as possible without it turning into a very big refactor that, as we've seen previously, will probably break things that we weren't expecting. Sticking to just adding `CaseworkerLoginRequiredMixin` in this PR can de-risk that and if everything seems fine we can add the corresponding `ExporterLoginRequiredMixin` in a separate PR

alixedi (approved):

PR #235, opened 2021-07-29
Replace document upload with gds crispy forms

eadpearce:


exporter/organisation/forms.py
```diff
@@ -1,14 +1,80 @@
...
+from crispy_forms_gds.layout import Submit, Layout, HTML
+
+
+def validate_expiry_date(value):
+    today = timezone.now().date()
+    if value < today:
+        raise ValidationError("Expiry date must be in the future")
+    else:
+        if relativedelta(value, today).years >= 5:
+            raise ValidationError("Expiry date is too far in the future")
```
saruniitr, line 18: Should the message reflect the limit of 5 years? one for the content team to check.

exporter/organisation/forms.py
```diff
@@ -1,14 +1,80 @@
...
+        error_messages={"required": "Select certificate file to upload"},
     )
+    reference_code = forms.CharField(
+        label="Certificate number", error_messages={"required": "Enter the certificate number"},
+    )
+    expiry_date = DateInputField(
+        label="Expiry date",
+        help_text="For example 12 3 2021",
+        require_all_fields=False,
+        validators=[validate_expiry_date],
```
saruniitr, line 34: being able to specify validators is very nice

exporter/organisation/forms.py
```diff
@@ -1,14 +1,80 @@
...
+    )
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.helper = FormHelper()
+        self.helper.layout = Layout(
+            HTML.h1("Attach your registered firearms dealer certificate"),
```
saruniitr, line 72: just a general question, in some places we specify additional css classes, is it possible with this forms?

unit_tests/exporter/organisation/test_forms.py
```diff
@@ -0,0 +1,87 @@
...
+        "expiry_date_1": 1,  # month
+        "expiry_date_2": 2022,  # year
+        "reference_code": "1234",
+    }
+
+    file = {
+        "file": SimpleUploadedFile("file.txt", b"abc", content_type="text/plain"),
+    }
+
+    form = forms.UploadSectionFiveCertificateForm(data, file)
```
saruniitr, line 40: I have not used these crispy forms before - is there a reason `file` is another argument rather than part of the `data` dict?

saruniitr (approved):
looks good to me

exporter/organisation/forms.py
eadpearce: Yeah agree. We should ask

exporter/organisation/forms.py
eadpearce: Yep. Crispy forms is pretty flexible with that https://crispy-forms-gds.readthedocs.io/en/latest/reference/layout/html.html

unit_tests/exporter/organisation/test_forms.py
eadpearce: Because it's stored in the request.FILES rather than request.POST django handles it separately https://docs.djangoproject.com/en/3.0/ref/forms/api/#binding-uploaded-files

PR #241, opened 2021-08-25
Rebuild new site form with django formtools

eadpearce:


exporter/organisation/sites/forms.py
```diff
@@ -20,9 +24,150 @@
...
+    phone_number = forms.CharField(
+        label="Phone number", help_text="For international numbers include the country code", required=False
+    )
+    website = forms.CharField(label="Website", required=False)
+    country = forms.ChoiceField(
+        choices=[], widget=widgets.Autocomplete(attrs={"id": "country-autocomplete"})
+    )  # populated in __init__
+
+    def __init__(self, *args, **kwargs):
+        self.request = kwargs.pop("request")
```
alixedi, line 152: It doesn't seem like we are using `self.request` elsewhere in the class. Also, `pop` might have interesting side-effects (because `dicts` are pass-by-reference IIRC) so I wonder if we can do the following -

```
request = kwargs["request"]
```

exporter/organisation/sites/forms.py
```diff
@@ -20,9 +24,150 @@
...
+    address = forms.CharField(
+        label="Address", widget=forms.Textarea(attrs={"class": "govuk-input--width-20", "rows": 6}), required=False
+    )
+    phone_number = forms.CharField(
+        label="Phone number", help_text="For international numbers include the country code", required=False
+    )
+    website = forms.CharField(label="Website", required=False)
+    country = forms.ChoiceField(
+        choices=[], widget=widgets.Autocomplete(attrs={"id": "country-autocomplete"})
+    )  # populated in __init__
```
alixedi: Opportunity to subclass `ChoiceField` to create our very own `CountryField` :)

P.S. Don't have to do it in this PR but presume a `CountryField` may be re-used in several other forms?

exporter/organisation/sites/forms.py
```diff
@@ -20,9 +24,150 @@
...
+        existing_sites = get_sites(self.request, self.request.session["organisation"], postcode=postcode)
+        super().__init__(*args, **kwargs)
+        self.helper = FormHelper()
+        self.helper.layout = Layout(
+            HTML.h1(AddSiteForm.Postcode.TITLE),
+            HTML.p(AddSiteForm.Postcode.DESCRIPTION.format(", ".join(site["name"] for site in existing_sites))),
+            "are_you_sure",
+            Submit("submit", "Continue"),
+            # TODO: if user selects No don't submit anything and redirect back to the sites page
+        )
```
alixedi: The `h1` and `p` and `Submit` here seems like something we want to do for all of our forms. I wonder if we can tweak the default template to handle these over there and not have to override the default Layout in all our forms?

alixedi (commented):
Some initial thoughts. Will do a full review once we bottom these out.

exporter/organisation/sites/forms.py
alixedi: Just to be clear, the `CountryField` will be able to populate itself from the `country_choices` that it will fetch from the API.

alixedi:
One more - I wonder if we can create an `AddressField` (that will technically be a `Formset`) like [so](https://stackoverflow.com/a/22053952/1182202).

I have a strong suspicion that we if we do this, we will be able to get rid of the nesting and flattening code :)

exporter/organisation/sites/forms.py
eadpearce: `.pop()` was because the `__init__` the line after this will complain because it's not expecting the request kwarg BUT! I have a solution for this now. I can just check in the View which form it's rendering and pass the request only if it's required by the form

exporter/organisation/sites/views.py
```diff
@@ -17,21 +30,66 @@ def get_additional_context(self):
...
+    # check if a site has already been added with the same postcode
+    cleaned_data = wizard.get_cleaned_data_for_step(UK_ADDRESS) or {}
+    postcode = cleaned_data.get("postcode")
+    if not postcode:
+        return False
+    existing_sites = get_sites(wizard.request, wizard.request.session["organisation"], postcode=postcode)
+    return len(existing_sites) > 0
+
+
+class NewSiteWizardView(SessionWizardView, LoginRequiredMixin):
```
wkeeling, line 56: Just something I've spotted: should the `LoginRequiredMixin` come first here, rather than after the base `SessionWizardView`?

eadpearce: Oh probably. I'll need to check

exporter/core/templates/core/form-wizard.html
```diff
@@ -0,0 +1,52 @@
+{% extends 'layouts/base.html' %}
```
dgnball, line 1: Can we put this file and the form.html file in an area which is common to both exporter and caseworker?

exporter/assets/built/main.css
```diff
@@ -1,2 +1,2 @@
...
...item:after{content:"";display:block;clear:both}.govuk-checkboxes--small .govuk-checkboxes__input,.govuk-table__cell--checkbox .govuk-checkboxes__input,.lite-tree-view__children .govuk-checkboxes__item .govuk-checkboxes__input,.lite-tree-view__summary .govuk-checkboxes__item .govuk-checkboxes__input{left:-10px}.govuk-checkboxes--small .govuk-checkboxes__label,.govuk-table__cell--checkbox .govuk-checkboxes__label,.lite-tree-view__children .govuk-checkboxes__item .govuk-checkboxes__label,.lite-tree-view__summary .govuk-checkboxes__item .govuk-checkboxes__label{margin-top:-2px;padding:13px 15px 13px 1px;float:left}@media(min-width:40.0625em){.govuk-checkboxes--small .govuk-checkboxes__label,.govuk-table__cell--checkbox .govuk-checkboxes__label,.lite-tree-view__children .govuk-checkboxes__item .govuk-checkboxes__label,.lite-tree-view__summary .govuk-checkboxes__item .govuk-checkboxes__label{padding:11px 15px 10px 1px}}.govuk-checkboxes--small .govuk-checkboxes__label:before,.govuk-table__cell--checkbox .govuk-checkboxes__label:before,.lite-tree-view__children .govuk-checkboxes__item .govuk-checkboxes__label:before,.lite-tree-view__summary .govuk-checkboxes__item .govuk-checkboxes__label
```
dgnball, line 1: Should we put all these asset files into core/assets?

exporter/organisation/sites/forms.py
eadpearce: Since all the forms need the request to do the validation I've moved this into the `SiteFormMixin` instead

exporter/core/templates/core/form-wizard.html
eadpearce: That would be nice! But since they inherit from different base templates they would need to be separate anyway. There may be a way to do it with includes but I am not sure it wouldn't get complicated if/when the two services diverge too much in style.

exporter/assets/built/main.css
eadpearce: I think the reasoning behind this was only the stuff used by both services goes in core and service specific stuff goes in the exporter/caseworker assets folder.

alixedi (approved):
Well done @eadpearce and thank you!

This piece prolly saved half of my hair given we would done advice2 in `lite-forms` had you not done this.

PR #244, opened 2021-09-14
Convert finalise generate documents form to crispy forms

wkeeling:
Initial go at rendering the finalise generate documents form with Django Crispy Forms.

### DIFF
--- a/caseworker/templates/case/generate-documents.html
+++ b/caseworker/templates/case/generate-documents.html
@@ -1,0 +1,26 @@
+{% extends 'layouts/case.html' %}
+
+{% load crispy_forms_tags crispy_forms_gds %}
+
+{% load static from staticfiles %}
+
+{% block title %}Generate decision documents{% endblock %}
+
+{% block back_link %}{% endblock %}
+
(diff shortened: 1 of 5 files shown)


wkeeling:
Minor issues to be resolved:
- The new page title (as in browser tab title) is not the same as for the old page
- The new page is has got slightly different white space around the form, may be missing some parent <div> elements
- CI build errors related to setoptconf: `error in setoptconf setup command: use_2to3 is invalid`

caseworker/templates/case/generate-documents.html
```diff
@@ -0,0 +1,26 @@
+{% extends 'layouts/case.html' %}
+
+{% load crispy_forms_tags crispy_forms_gds %}
+
+{% load static from staticfiles %}
+
+{% block title %}Generate decision documents{% endblock %}
+
+{% block back_link %}{% endblock %}
```
Pithikos, line 9: I'm thinking there must be a way to add this directly via crispy (https://design-system.service.gov.uk/components/back-link/)

PR #251, opened 2021-09-22
Associate application with existing party test

dgnball:


ui_tests/exporter/features/siel_firearm_application.feature
```diff
@@ -43,6 +43,21 @@ Feature: I want to be able to submit SIEL firearm applications
...
     And I logout
 
+  Scenario: Associate application with existing consignee party
+    Given I signin and go to exporter homepage and choose Test Org
+    When I create a standard application of a "permanent" export type
+    And I am on the application overview page entitled "Standard Individual Export Licence"
+    And I click on the "Consignee" section
+    And I answer "Yes" for whether I want to reuse an existing party
+    And I select a party with type of "government"
+    And I click continue
```
Pithikos, line 53: Are we clicking continue in 5 different pages here?

wkeeling (approved):

ui_tests/exporter/features/siel_firearm_application.feature
dgnball: This is because we are using an existing consignee so it goes through a bunch of pre-filled pages with names, addresses, etc that just confirm you are happy with the details of the party. It is a wizard that we pick defaults on.

PR #252, opened 2021-09-24
Feature/stepping e2e tests

Pithikos:
Just commiting some old code that might be useful for E2E tests.

Essentially this allows us to use `--step-through` and `--step-verbose` with pytest to step through the tests for exploration/debugging

### DIFF
--- a/ui_tests/conftest.py
+++ b/ui_tests/conftest.py
@@ -3,6 +3,32 @@ import os
 import tests_common.tools.helpers as utils
 
 
+STEP_THROUGH = False  # Gives a prompt for every step
+STEP_VERBOSE = True   # Shows info for every step
+
+
+def pytest_bdd_before_step_call(request, feature, scenario, step, step_func, step_func_args):
+    """
+    Runs before each step
+    """
+    if STEP_VERBOSE:
+        print("*******************************************")
+        print(f"SCENARIO: {scenario.feature.description}")
+        print(f"STEP: .. {step.keyword} {step.name}")
+        print("*******************************************")
+    if STEP_THROUGH:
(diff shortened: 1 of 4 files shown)


ui_tests/conftest.py
```diff
@@ -3,11 +3,45 @@
 import tests_common.tools.helpers as utils
 
 
+STEP_THROUGH = False  # Gives a prompt for every step in the terminal
+STEP_VERBOSE = STEP_THROUGH  # Shows info as a banner for every step
```
saruniitr, line 7: best to load it from `.env` to avoid globals below

ui_tests/conftest.py
```diff
@@ -3,11 +3,45 @@
...
+    """
+    Runs before each step
+    """
+    if STEP_VERBOSE:
+        print("*******************************************")
+        print(f"SCENARIO: {scenario.feature.description}")
+        print(f"STEP: .. {step.keyword} {step.name}")
+        print("*******************************************")
+    if STEP_THROUGH:
+        import IPython
```
saruniitr, line 20: may be useful to print `step.name` here, if that is default behaviour then this can be ignored

ui_tests/conftest.py
Pithikos: These are just defaults for the command flags - so not environment variables. Hence why I didn't put them there

ui_tests/conftest.py
Pithikos: Yep, it's printed on line 17. This is mainly so when you get the prompt you can get an idea on what step and scenario you're in

saruniitr (approved):

PR #253, opened 2021-09-24
Introduce is_ogd flag

wkeeling:


caseworker/teams/forms.py
```diff
@@ -15,6 +15,11 @@ def add_team_form():
                 name="part_of_ecju",
                 options=[Option(key=True, value="Yes"), Option(key=False, value="No"),],
             ),
+            RadioButtons(
+                title="Is this team an OGD?",
```
eadpearce, line 19: design/content question: do we need to explain what OGD stands for somewhere?

wkeeling: Good question. Perhaps we should display it as "Other Government Department (OGD)" here (and also on the list view column) - will check with Ashley

PR #255, opened 2021-09-28
Add POC advice2 app.

alixedi:


caseworker/advice2/apps.py
```diff
@@ -0,0 +1,5 @@
+from django.apps import AppConfig
+
+
+class Advice2Config(AppConfig):
+    name = 'advice2'
```
saruniitr, line 5: `advice2` - any better alternatives? I have no suggestions to offer, maybe `advice_v2`?

caseworker/advice2/tests/__init__.py
```diff
@@ -0,0 +1,11 @@
+"""ATM, all the caseworker unit tests that we have live in -
+
+unit_tests/caseworker/
+
+I would like us to have independent unit tests for advice2 and put them here.
```
saruniitr, line 5: any particular reason? why not keep in the existing place along with other tests? We are creating a new app for advice but tests can be in the same place I think unless there is a better reason?

caseworker/cases/urls.py
```diff
@@ -13,6 +14,8 @@
     path("move/", main.MoveCase.as_view(), name="move"),
     path("additional-contacts/add/", main.AddAnAdditionalContact.as_view(), name="add_additional_contact"),
     path("attach/", main.AttachDocuments.as_view(), name="attach_documents"),
+    # This needs to be before "case" path b/c the regex in that sinks everything
+    path("advice2/", include("caseworker.advice2.urls")),
```
agalamatis, line 18: Is there there an `/advice/` url used? Advice is currently all inside `/cases/`, so you can drop the `2` from the urls.

caseworker/advice2/apps.py
alixedi: Sure. 

Or, like Ares suggested above, we can go with `advice` b/c there is no such app in here ATM.

saruniitr: :+1:

caseworker/advice2/tests/__init__.py
saruniitr: that is a better reason, :+1:

caseworker/advice2/templates/advice2/placeholder.html
```diff
@@ -0,0 +1,7 @@
+{% extends 'layouts/case.html' %}
+
+{% load static custom_tags %}
+
+{% block body %}
+{{ greetings }}
+{% endblock %}
```
Pithikos, line 7: @alixedi why the double-nesting of this file? Why not have caseworker/advice2/templates/placeholder.html ?

caseworker/advice2/tests/__init__.py
Pithikos: Django has some default setup but we don't really need to abide by it (and some of their decisions actually don't work in many projects). We can either have tests per app or in a global directory - it's up to us

caseworker/advice2/views.py
```diff
@@ -0,0 +1,21 @@
+from django.views.generic import TemplateView
+from caseworker.cases.services import get_case
+
+
+class Advice2PlaceholderView(TemplateView):
+    """This is POC ATM and should be removed with the first PR
```
Pithikos, line 6: Can we please reduce abbreviations and just use lowercase for common day abbrevs like atm? :) Not sure what POC is..

caseworker/advice2/views.py
```diff
@@ -0,0 +1,21 @@
...
+    of advice2. Currently, this is a TemplateView but it should
+    be fairly simple to make this e.g. a SingleFormView.
+    """
+    template_name = "advice2/placeholder.html"
```
Pithikos, line 10: Related to my comment on nesting.. Maybe we can remove the double `advice2` directory? So we could use `template_name = "placeholder.html"` here

caseworker/cases/urls.py
dgnball: agreed

dgnball:
Would like to see an example of what using crispy forms within this app would look like.

caseworker/advice2/tests/__init__.py
dgnball: Doesn't seem to be much value in moving the unit tests near to the code that is being tested over the current layout. I would advocate for consistency with what we are already doing in this case and put a sample test_views.py test in the existing unit_tests folder.

caseworker/advice2/templates/advice2/placeholder.html
alixedi: It is a Django convention. You can find out the reasons in the [Django tutorial](https://docs.djangoproject.com/en/dev/intro/reusable-apps/).

dgnball:
Would also like to see an example of a form making an API call.

caseworker/advice2/tests/__init__.py
alixedi: Yes, it's up to us and we can choose to deviate from conventions if we understand why the convention is there and weigh up the benefits of moving away from it.

Basically, the Django convention is for apps to be re-usable packages. Personally, even when I don't plan to re-use apps, designing them to be re-usable means that I tend to be more careful about the interfaces. This results in better code and is easier to test and reason about.

caseworker/advice2/views.py
alixedi: Sorry, POC is proof of concept.

caseworker/advice2/templates/advice2/placeholder.html
dgnball: I'm guessing this is Django style but I'd like to see a reduction in file nested if possible where possible

alixedi: To give a more concrete example for this - If we had to override templates from e.g. `caseworker/cases` app in the `advice2` app, what directory structure would you need?

caseworker/advice2/templates/advice2/placeholder.html
```diff
@@ -0,0 +1,7 @@
+{% extends 'layouts/case.html' %}
```
eadpearce, line 1: in caseworker/templates/layouts/case.html we can add a new block that wraps the `{% if tabs %}` chunk on line 118 and then override that in here

caseworker/advice2/templates/advice2/placeholder.html
```diff
@@ -0,0 +1,7 @@
...
+
+{% load static custom_tags %}
+
+{% block body %}
```
eadpearce, line 5: then if we do that we can put this content in `{% block two_thirds %}` instead

caseworker/advice2/templates/advice2/placeholder.html
Pithikos: OK, fair enough.. so it's to avoid name collision

eadpearce (approved):

PR #256, opened 2021-09-30
Ltd 1116 check the part number

dgnball:


ui_tests/exporter/conftest.py
```diff
@@ -1222,9 +1222,10 @@ def navigate_to_application(driver, exporter_url, context):  # noqa
 
 
 @then("I see the application summary with <clc_rating>,<end_use>,<end_user_name>,<consignee_name>")
-def i_see_application_summary(driver, clc_rating, end_use, end_user_name, consignee_name):  # noqa
+def i_see_application_summary(driver, context, clc_rating, end_use, end_user_name, consignee_name):  # noqa
     tds = driver.find_elements_by_tag_name("td")
     dds = driver.find_elements_by_tag_name("dd")
+    assert tds[2].text == context.part_number
```
saruniitr, line 1228: Can the `part_number` be not taken as a parameter instead of adding it to context?

dgnball: no, it seems to change every time

saruniitr: I think you may have to make few more changes because the parameters doesn't include `part_number` atm.
https://github.com/uktrade/lite-frontend/blob/e55352d27cd81f99571daa381d2bc256cdf53bfe/ui_tests/exporter/features/siel_firearm_application.feature#L139

dgnball: I can't do that because I don't know what the part number is ahead of time. It changes every time and can't be set in the same way as "name" does.

tests_common/helpers/applications.py
```diff
@@ -8,6 +8,7 @@ def create_standard_application(api_test_client, context, app_data, submit=True)
     api_test_client.api_client.auth_exporter_user(api_test_client.context["org_id"])
 
     data = api_test_client.api_client.request_data["good"]
+    data["part_number"] = app_data["part_number"] if "part_number" in app_data else data["part_number"]
```
saruniitr, line 11: I think below is better,
`data["part_number"] = app_data.get("part_number", data["part_number"])`

saruniitr (commented):
Below need to be updated also isn't it?
https://github.com/uktrade/lite-frontend/blob/master/ui_tests/exporter/features/ecju_queries.feature#L8

PR #261, opened 2021-10-07
Update parcel.

alixedi:
Updating to parcel v2 b/c v1 (1) is deprecated and (2) was the source of a couple of vulnerabilities pointed out by `dependabot`. 

I am following [this helpful migration guide](https://parceljs.org/getting-started/migration).


r4vi:
I think you need to get rid of "engines" in package.json otherwise it will imply the target execution environment for your js bundle is node: https://parceljs.org/getting-started/migration#--target

PR #268, opened 2021-10-12
Fix ECJU query formatting

eadpearce:


package.json
```diff
@@ -12,7 +12,7 @@
     "build": "export NODE_ENV=production && npm-run-all --serial build_exporter build_caseworker"
   },
   "engines": {
-    "node": ">=12.16.1"
+    "node": ">=14.18.0"
```
saruniitr, line 15: this is a major version update, is this going to break any other dependencies?

saruniitr (commented):
changes lgtm but have a query regarding the `node` update

PR #269, opened 2021-10-12
LTD-1167: Give Approval advice

saruniitr:
# Change description

Implements changes to give Approval advice as per Advice v2 design

The form once submitted creates approval advice for all products on the application.
To make it work with the serializers in the current API the data parameters are passed as expected by the API.


caseworker/advice/forms.py
```diff
@@ -14,3 +14,38 @@ def __init__(self, *args, **kwargs):
...
+        self.helper.layout = Layout(
+            Field.textarea("proviso", rows=10,),
+            Field.textarea("approval_reasons", rows=10,),
+            Field.textarea("instructions_to_exporter", rows=10,),
+            Field.textarea("footnote_details", rows=5,),
+            HTML.details(
+                "What is a footnote?",
+                "Footnotes explain why products to a destination have been approved or refused. "
+                "They will be publicly available in reports and data tables.",
+            ),
```
alixedi, line 49: Might be a good idea to do this in the template instead of here. Look at select-advice PR as an example.

caseworker/advice/services.py
```diff
@@ -0,0 +1,22 @@
...
+            "note": data["instructions_to_exporter"],
+            "footnote_required": True if data["footnote_details"] else False,
+            "footnote": data["footnote_details"],
+            "good": product_id,
+            "denial_reasons": [],
+        })
+
+    response = client.post(request, f"/cases/{case['id']}/user-advice/", json)
+    response.raise_for_status()
+    return response.json(), response.status_code
```
alixedi, line 22: 1) Do we not have existing compatible code that does this?
2) Might want to either rename this as `post-approval_advice` or add support for refusals.

caseworker/advice/views.py
```diff
@@ -1,10 +1,12 @@
...
 
+from caseworker.advice import forms, services
 from caseworker.cases.services import get_case
-from caseworker.advice.forms import SelectAdviceForm
+from core.auth.views import LoginRequiredMixin
+from django.views.generic import TemplateView, FormView
 
 
-class CaseContextMixin:
+class CaseContextMixin(LoginRequiredMixin):
```
alixedi, line 9: Good catch!

caseworker/advice/views.py
```diff
@@ -13,7 +15,7 @@ class CaseContextMixin:
     def get_context(self):
         return {}
 
-    def get_context_data(self, **kwargs):
+    def get_context_data(self, *args, **kwargs):
```
alixedi, line 18: Is this needed?

caseworker/advice/views.py
```diff
@@ -48,11 +50,40 @@ class CaseDetailView(CaseContextMixin, TemplateView):
 
 class SelectAdviceView(CaseContextMixin, FormView):
     template_name = "advice/select_advice.html"
-    form_class = SelectAdviceForm
+    form_class = forms.SelectAdviceForm
```
alixedi, line 53: I prefer the imports explicit (like they were before). Happy to listen to arguments for this change.

caseworker/advice/views.py
```diff
@@ -48,11 +50,40 @@ class CaseDetailView(CaseContextMixin, TemplateView):
...
+        """
+        FormView doesn't pass kwargs to get_context_data() by default in get()
+        so override and send kwargs
+        """
+        return self.render_to_response(self.get_context_data(**kwargs))
+
+    def get_form_kwargs(self):
+        # default method supplies different set of kwargs so override and append ours
+        default_kwargs = super().get_form_kwargs()
+        return {**self.kwargs, **default_kwargs}
```
alixedi, line 81: I think this is also not needed.

caseworker/advice/views.py
```diff
@@ -48,11 +50,40 @@ class CaseDetailView(CaseContextMixin, TemplateView):
...
+
+    def form_valid(self, form):
+        case = self.get_context_data()["case"]
+        services.post_advice(self.request, case, form.cleaned_data)
+        return super().form_valid(form)
+
+    def get_success_url(self):
+        return "/#save-advice"
```
alixedi, line 89: You can specify static URL by setting `success_url = "/#save-advice"` on the class.

caseworker/advice/views.py
saruniitr: because we don't have to import every time we add a new form especially when they are all going to stay in the same file but I am easy with this and can revert.

caseworker/advice/views.py
saruniitr: :+1:

caseworker/advice/forms.py
saruniitr: I checked that but doing that way was not showing the form errors correctly.

caseworker/advice/forms.py
```diff
@@ -14,3 +14,38 @@ def __init__(self, *args, **kwargs):
...
+    )
+    footnote_details = forms.CharField(
+        label="Add a reporting footnote (optional)", widget=forms.Textarea, required=False
+    )
+
+    def __init__(self, queue_pk, pk, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.helper = FormHelper()
+        self.helper.layout = Layout(
+            Field.textarea("proviso", rows=10,),
```
eadpearce, line 41: if you've already specified the widget I think you can just list these out by name like `"proviso"`. for the rows, you can add it in the attrs for the widget `widget=forms.Textarea(attrs={"rows": "10"})`

caseworker/advice/forms.py
eadpearce: I like this way better because you just have to type the content of the details/summary component. I always have to check the design system docs for what classes you need to style it properly

caseworker/advice/templates/advice/give-approval-advice.html
```diff
@@ -0,0 +1,21 @@
+{% extends 'layouts/case.html' %}
+
+{% load static custom_tags %}
+{% load crispy_forms_tags %}
+
+
+{% block body %}
+<div class="govuk-width-container">
+    <a href="#" class="govuk-back-link">Back</a>
```
eadpearce, line 9: can we put a legit back link here yet?

caseworker/advice/forms.py
```diff
@@ -18,30 +18,32 @@ def __init__(self, *args, **kwargs):
...
-    def __init__(self, queue_pk, pk, *args, **kwargs):
+    def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.helper = FormHelper()
         self.helper.layout = Layout(
-            Field.textarea("proviso", rows=10,),
-            Field.textarea("approval_reasons", rows=10,),
-            Field.textarea("instructions_to_exporter", rows=10,),
-            Field.textarea("footnote_details", rows=5,),
+            Field.textarea("proviso"),
```
eadpearce, line 43: sorry didn't make it clear in my other comment - you can declare the fields like this:
```
self.helper.layout = Layout(
    HTML.h1("Attach your section five certificate"),
    "file",
    "reference_code",
    "expiry_date",
    Submit("submit", "Submit"),
)
```

PR #274, opened 2021-10-14
assert that Enforcement Check Req flag is no longer present

nahmed99:
Added test.

### DIFF
--- a/ui_tests/caseworker/conftest.py
+++ b/ui_tests/caseworker/conftest.py
@@ -657,3 +657,10 @@ def case_removed_from_queue(driver, queue):  # noqa
     driver.find_element_by_id("link-change-queues").click()
     select_queue(driver, queue, False)
     functions.click_submit(driver)
+
+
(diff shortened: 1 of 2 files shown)


ui_tests/caseworker/conftest.py
```diff
@@ -657,3 +657,10 @@ def case_removed_from_queue(driver, queue):  # noqa
     driver.find_element_by_id("link-change-queues").click()
     select_queue(driver, queue, False)
     functions.click_submit(driver)
+
+
+@then(parsers.parse('the flag "{flag}" is not present'))  # noqa
+def flag_not_present(driver, flag):  # noqa
+    flags_container = driver.find_element_by_id("case-flags")
+    el = flags_container.find_elements_by_xpath(f"//li[contains(text(), '{flag}')]")
```
saruniitr, line 665: In this case we are not expecting to find any elements so does this return quickly or blocks until it times out?

PR #279, opened 2021-10-19
Add ability to clear your advice

wkeeling:
This PR provides the "Clear Advice" function.


caseworker/advice/forms.py
```diff
@@ -70,3 +70,10 @@ def __init__(self, denial_reasons, *args, **kwargs):
         )
         self.helper = FormHelper()
         self.helper.layout = Layout("denial_reasons", "refusal_reasons", Submit("submit", "Submit"))
+
+
+class DeleteAdviceForm(forms.Form):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.helper = FormHelper()
+        self.helper.add_input(Submit("confirm", "Confirm"))
```
saruniitr, line 79: No `cancel` button if user doesn't want to delete?

caseworker/advice/templates/advice/delete-advice.html
```diff
@@ -0,0 +1,16 @@
...
+    <main class="govuk-main-wrapper">
+        <div class="govuk-grid-row">
+            <div class="govuk-grid-column-two-thirds">
+                <h1 class="govuk-heading-xl">Are you sure you want to delete your recommendation on this case?</h1>
+                {% crispy form %}
+            </div>
+        </div>
+    </main>
+</div>
+{% endblock %}
```
saruniitr, line 16: please add newline

caseworker/advice/views.py
```diff
@@ -111,3 +111,16 @@ def get_context_data(self, **kwargs):
...
+    template_name = "advice/delete-advice.html"
+    form_class = forms.DeleteAdviceForm
+
+    def form_valid(self, form):
+        case = self.get_context_data()["case"]
+        services.delete_user_advice(self.request, case["id"])
+        return super().form_valid(form)
+
+    def get_success_url(self):
+        return reverse("cases:view_my_advice", kwargs={"queue_pk": self.kwargs["queue_pk"], "pk": self.kwargs["pk"]})
```
saruniitr, line 126: this should probably take it to "Make recommendation" page where user can again give recommendation but should be ok for now as those changes are not merged yet

unit_tests/caseworker/advice/views/test_delete_advice_view.py
```diff
@@ -0,0 +1,35 @@
...
+    assert "Are you sure you want to delete your recommendation on this case?" in str(response.content)
+
+
+def test_delete_advice_post(authorized_client, url, data_queue, data_standard_case, delete_user_advice):
+    response = authorized_client.post(url)
+    assert response.status_code == 302
+    assert response.url == reverse(
+        "cases:view_my_advice", kwargs={"queue_pk": data_queue["id"], "pk": data_standard_case["case"]["id"]}
+    )
+    delete_user_advice.assert_called_once_with(response.wsgi_request, data_standard_case["case"]["id"])
```
saruniitr, line 35: may be better to use `requests_mock` to be consistent with other tests.
With `requests_mock` we can also check requests history, data sent as in this case etc

caseworker/advice/forms.py
wkeeling: There's only a "confirm" button according to the figma design - although the user can press the "back" link which would effectively cancel. Once of the acceptance criteria talks about pressing "back" and ensuring that nothing was deleted.

unit_tests/caseworker/advice/views/test_delete_advice_view.py
wkeeling: Thanks - will update to use `requests_mock`.

PR #280, opened 2021-10-19
LTD-1179: Edit given advice

saruniitr:
## Change description
User can edit previously given "approve/refuse" advice/recommendation


caseworker/advice/views.py
```diff
@@ -94,20 +95,70 @@ def form_valid(self, form):
...
+
+class EditAdviceView(LoginRequiredMixin, CaseContextMixin, FormView):
+    """
+    Form to edit given advice for all products on the application
+    """
+
+    def get_form(self):
+        case = get_case(self.request, self.kwargs["pk"])
+        my_advice = services.filter_current_user_advice(case.advice, str(self.request.session["lite_api_user_id"]))
+        if not my_advice:
```
saruniitr, line 121: additional safeguard. "Edit recommendation" is not visible if the user has not given advice. I can remove if this is not required.

caseworker/advice/views.py
```diff
@@ -94,20 +95,70 @@ def form_valid(self, form):
...
+
+        advice = my_advice[0]
+
+        if advice["type"]["key"] in ["approve", "proviso"]:
+            initial_data = {
+                "proviso": advice["proviso"],
+                "approval_reasons": advice["text"],
+                "instructions_to_exporter": advice["note"],
+                "footnote_details": advice["footnote"],
+            }
```
alixedi, line 132: I wonder if we can do the following here -

```
return forms.GiveApprovalAdviceForm(data=advice)
```

And in the `GiveApprovalAdviceForm` -

```
def __init__(data=None):
    if data is not None:
        data = {
                 "proviso": data["proviso"],
                 "approval_reasons": data["text"],
                 "instructions_to_exporter": data["note"],
                 "footnote_details": data["footnote"],
             }
```

P.S. This is only a suggestion so e.g. it's ok if instead of this, we want to write a `give_approval_advice_form_factory(advice)` function that does what is in this code.

alixedi: I think I like `give_approval_advice_form_factory(advice)` more than overriding the `__init__` method of the form. 

Because it will keep the form decoupled from changes in the schema of advice object as rendered from the API.

Hmm..

caseworker/advice/views.py
```diff
@@ -94,20 +95,70 @@ def form_valid(self, form):
...
+            self.advice_type = "approve"
+            self.template_name = "advice/give-approval-advice.html"
```
alixedi, line 134: I initially thought that we can explore having the same template for approve/refuse and use it here but it is very likely that these templates will diverge in the future and so we are likely creating work for ourselves.

Having had another look, it might be worth having a separate, new template for this view.

caseworker/advice/views.py
```diff
@@ -94,20 +95,70 @@ def form_valid(self, form):
...
+            return forms.GiveApprovalAdviceForm(data=initial_data)
+        elif advice["type"]["key"] == "refuse":
+            initial_data = {
+                "refusal_reasons": advice["text"],
+                "denial_reasons": [r for r in advice["denial_reasons"]],
+            }
+            denial_reasons = get_denial_reasons(self.request)
+
+            self.advice_type = "refuse"
```
alixedi, line 143: My comments on the approval form apply here as well.

caseworker/advice/views.py
alixedi: Yes, I thought this was a good idea but now I think that raising a `ValueError` (which would throw a 500) is probably not what we want to do here.

caseworker/advice/views.py
```diff
@@ -94,20 +95,70 @@ def form_valid(self, form):
...
+            self.template_name = "advice/refusal_advice.html"
+            return forms.RefusalAdviceForm(data=initial_data, denial_reasons=denial_reasons)
+        else:
+            raise ValueError("Invalid advice type encountered")
+
+    def form_valid(self, form):
+        case = self.get_context_data()["case"]
+        data = form.cleaned_data.copy()
```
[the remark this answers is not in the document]
alixedi: Presumably `form_valid` is only called when the form is valid. In view of this, you might be able to do something along the following lines -

```
data = request.POST
if self.advice_type == "approve":
    services.post_approval_advice(self.request, case, data)
elif self.advice_type == "refuse":
    services.post_refusal_advice(self.request, case, data)
else:
    raise ValueError("Unknown advice type")
```

P.S. Note the `else` clause. It is important to handle this corner case.

alixedi (approved):

PR #281, opened 2021-10-26
Countersigning Approve

saruniitr:
## Change description

Add Form to countersign advice given by other users.
Once approved displays a page to view the countersigned advice.


saruniitr:
Need to add unit tests

caseworker/advice/templates/advice/approve-advice-details.html
```diff
@@ -42,55 +48,68 @@ <h2 class="govuk-heading-m">My approval recommendation</h2>
...
+          </tr>
+          {% endfor %}
+        </tbody>
+      </table>
+    {% else %}
+      <h2 class="govuk-heading-m">No licence required products</h2>
+      <label class="govuk-label">None</label>
+    {% endif %}
+  </div>
+</div>
```
alixedi, line 115: I think you are missing a new line here.

caseworker/advice/views.py
```diff
@@ -241,19 +255,65 @@ def group_advice(self):
...
+    def get_context(self, **kwargs):
+        context = super().get_context()
+        case = kwargs.get("case")
+        caseworker = self.caseworker["user"]
+
+        advice_to_countersign = services.get_advice_to_countersign(case, caseworker)
+        context["advice_to_countersign"] = advice_to_countersign
+        context["review"] = False
+        context["subtitle"] = f"Approved by {caseworker['team']['name']}"
+        return context
```
alixedi, line 308: Because you are using the same template and you are already setting `context["review"] = False`. I reckon you get rid of this view and deal with the change in subtitle using `{% if review %}` in the template.

caseworker/advice/templatetags/advice_tags.py
```diff
@@ -19,3 +19,18 @@ def get_case_value(goods):
 def get_security_grading(goods):
     gradings = {good.get("good", {}).get("pv_grading_details") for good in goods}
     return sorted(gradings - {None})
+
+
+@register.filter
+def index(iter, i):
+    return iter[i]
```
alixedi, line 26: You can do this in the template as well - `data.5`.

caseworker/advice/templatetags/advice_tags.py
```diff
@@ -19,3 +19,18 @@ def get_case_value(goods):
...
+
+
+@register.filter
+def advice_given_by(advice):
+    return f"{advice['user']['first_name']} {advice['user']['last_name']}"
```
alixedi, line 31: You can do this in the template using - `{{ advice.user.first_name }} {{ advice.user.last_name }}`.

caseworker/advice/templatetags/advice_tags.py
```diff
@@ -19,3 +19,18 @@ def get_case_value(goods):
...
+
+
+@register.filter
+def countersigned_user_team(advice):
+    return f"{advice['countersigned_by']['team']['name']}"
```
alixedi, line 36: You can do this in the template using - `{{ advice.countersigned_by.team.name }}`

alixedi (commented):
First take. I haven't reviewed the tests yet.

caseworker/advice/templates/advice/approve-advice-details.html
saruniitr: :+1:

caseworker/advice/views.py
saruniitr: but there is an endpoint required to view the countersigned advice hence added this view.

caseworker/advice/templatetags/advice_tags.py
saruniitr: added a tag for convenience as we may be using in more places

caseworker/advice/templatetags/advice_tags.py
saruniitr: same as above

caseworker/advice/views.py
alixedi: Ah! ok. Makes sense.

saruniitr: :+1:

caseworker/advice/templates/advice/approve-advice-details.html
saruniitr: :+1:

caseworker/advice/templatetags/advice_tags.py
saruniitr: yes but I need to loop through all, not specific index.
I can use `forloop.counter0` but it will be like `formset.forloop.counter0` which is very confusing as it may look for an attr `forloop` inside `formset`.
Even if we assign `forloop.counter0` to a variable the problem remains, is it an attribute or an index into the list.
I think a tag here is more readable and avoids confusion.

caseworker/advice/templatetags/advice_tags.py
saruniitr: Tag is used here because it is passed as a variable to a template, eg

eadpearce: it's a bit nasty but you can do this in templates: `{% include "advice/approve-advice-details.html" with user=first_name|add:" "|add:last_name %}`

caseworker/advice/templatetags/advice_tags.py
eadpearce: using this here should work `{% include "advice/approve-advice-details.html" with user=advice.countersigned_by.team.name %}`. when you're passing variables to template includes you don't use the `{{ }}` brackets around it like you would if it were something like `<p>{{ user.name }}</p>`

caseworker/advice/views.py
```diff
@@ -35,7 +43,13 @@ def get_context_data(self, **kwargs):
         # doesn't have anything to do with e.g. lite-forms
         # P.S. the case here is needed for rendering the base
         # template (layouts/case.html) from which we are inheriting.
-        return {**context, **self.get_context(), "case": self.case, "queue_pk": self.kwargs["queue_pk"]}
+        return {
+            **context,
+            **self.get_context(case=self.case),
+            "case": self.case,
+            "queue_pk": self.kwargs["queue_pk"],
+            "caseworker": self.caseworker,
```
eadpearce, line 51: you may not need to add this because `current_user` is already in the context thanks to this context processor https://github.com/uktrade/lite-frontend/blob/master/conf/caseworker.py#L33

caseworker/advice/views.py
```diff
@@ -241,19 +255,65 @@ def group_advice(self):
...
         }
 
 
+class ReviewCountersignView(LoginRequiredMixin, CaseContextMixin, TemplateView):
+    template_name = "advice/review_countersign.html"
+    form_class = forms.CountersignAdviceForm
+
+    def get_context(self, **kwargs):
+        context = super().get_context()
+        case = kwargs.get("case").__dict__
```
eadpearce, line 271: why does this need to be converted to a dict here?

caseworker/advice/views.py
saruniitr: ah ok, I will check and remove if not required

caseworker/advice/views.py
saruniitr: we get a class type here and in filtering functions we go by key. The filtering functions are used in couple of places where only dict is available.

caseworker/advice/views.py
```diff
@@ -241,19 +255,65 @@ def group_advice(self):
...
 
-    def get_context(self):
+    def get_context(self, **kwargs):
         return {
             "queue": self.queue,
             "grouped_advice": self.grouped_advice,
         }
 
 
+class ReviewCountersignView(LoginRequiredMixin, CaseContextMixin, TemplateView):
```
eadpearce, line 265: should be using a `FormView` here instead. then you can use the `form_valid` method instead of overriding `post` I think

saruniitr: yes but this is a `formset`, in the `post` we check if all forms are valid and the continue.

eadpearce: huh weird. I can't find anything about using a formset with a FormView in the docs so if this approach works that's cool

saruniitr: yeah, most of the examples followed this pattern which makes sense.

caseworker/advice/views.py
eadpearce: a couple of those filters may not be necessary. which ones blow up if they're not given a dict?

caseworker/advice/templatetags/advice_tags.py
saruniitr: sorry my bad, here it should work here, will update.
Does using a tag adds any additional overhead? It looks more readable personally.

caseworker/advice/templatetags/advice_tags.py
saruniitr: oh good to know. For readability reasons I went with a tag but if it adds overhead can change it like above.

caseworker/advice/views.py
saruniitr: `get_advice_to_countersign()`, it needs to `advice` list in `case`

caseworker/advice/templatetags/advice_tags.py
wkeeling: Might suggest renaming `iter` to avoid shadowing the Python builtin `iter()`. Perhaps use `it` instead?

saruniitr: ok :+1:

caseworker/advice/templatetags/advice_tags.py
eadpearce: yeah I think a tag is unnecessary in this case

caseworker/advice/views.py
eadpearce: you're passing the entire case to `get_advice_to_countersign`. it would make more sense to just pass the advice list instead. maybe that will solve this problem?

caseworker/advice/templatetags/advice_tags.py
wkeeling: Can you wrap in quotes, e.g. `user="{{firstname}} {{last_name}}"`

eadpearce: if you could rewrite this tag to be more generic I think it'd be useful. change the name to `full_name`, just pass it the user dict instead and move the code into `core/builtins/custom_tags.py`. then we can use it in both frontend templates like `user|full_name`

saruniitr: I will make it generic

saruniitr: @wkeeling wrapping in quotes didn't help as it was just printing the variables as is rather than substituting its values

caseworker/advice/views.py
saruniitr: hmm probably, will try that

wkeeling (approved):

caseworker/advice/views.py
saruniitr: I actually need `case` also in that function hence using `advice` list alone won't help.

PR #287, opened 2021-11-01
Add my advice to view advice page

eadpearce:
Screenshot 2021-11-01 at 18 45 34


caseworker/advice/services.py
```diff
@@ -23,6 +23,12 @@ def filter_current_user_advice(all_advice, user_id):
     ]
 
 
+def get_advice_subjects(case):
+    return [(destination["name"], destination["id"]) for destination in case.destinations] + [
```
saruniitr, line 27: this doesn't look correct to me, could you please explain why we are sending destination details here?
IIRC `case.destinations` gives only `end_user` details

eadpearce: Yeah the api only gives us end_user, weirdly. But the frontend goes through the rest of the case data and adds them to the case.destinations attribute so I'm just making use of that https://github.com/uktrade/lite-frontend/blob/master/caseworker/cases/objects.py#L42

caseworker/advice/services.py
```diff
@@ -23,6 +23,12 @@ def filter_current_user_advice(all_advice, user_id):
     ]
 
 
+def get_advice_subjects(case):
```
alixedi, line 26: This can probably use a docstring to describe what `subjects` mean in this case.

alixedi (approved):
A minor comment that I will leave with you.

PR #288, opened 2021-11-03
LTD-1382 Require comments from applicant when no document provided

Pithikos:
Screenshot from 2021-11-03 16-57-57


saruniitr (changes requested):
looks good.
The caseworker site template changes to show the `no_document_comments` are missing, specifically updating below template,
https://github.com/uktrade/lite-frontend/blob/b02cc81613d229fe23ad9b19c5d9e45062434987/caseworker/templates/case/product-on-case.html#L376

saruniitr (approved):

PR #289, opened 2021-11-04
[LTD-1386] Country Select for FCDO users

alixedi:
This one does a bunch of interesting things. E.g. -

1. Checks for `FCO` users and renders a different form for them by overriding the `get_form` method.
2. It adds a method to `CaseMixin` that returns the list of countries for which advice has not been given (yet).

### DIFF
--- a/caseworker/advice/services.py
+++ b/caseworker/advice/services.py
@@ -48,7 +48,7 @@ def get_advice_subjects(case, countries=None):
             if dest["country"]["id"] not in countries:
                 continue
         destinations.append((dest["type"], dest["id"]))
-    goods = [("good", good["id"]) for good in case.goods if good["is_good_controlled"]]
+    goods = [("good", good["id"]) for good in case.goods if good["is_good_controlled"]["key"] == "True"]
     return destinations + goods
 
 
--- a/caseworker/advice/services.py
+++ b/caseworker/advice/services.py
@@ -2,10 +2,6 @@ from core import client
 from caseworker.cases.services import put_case_queues
 
 
-def filter_licenceable_products(products):
-    return [product for product in products if product["is_good_controlled"]["key"] == "True"]
-
-
 def filter_nlr_products(products):
     return [
         product
@@ -65,12 +61,11 @@ def post_approval_advice(request, case, data):
             "note": data["instructions_to_exporter"],
             "footnote_required": True if data["footnote_details"] else False,
             "footnote": data["footnote_details"],
-            "good": product["id"],
+            subject_name: subject_id,
             "denial_reasons": [],
         }
-        for product in filter_licenceable_products(case["data"]["goods"])
+        for subject_name, subject_id in get_advice_subjects(case, data.get("countries"))
     ]
-
     response = client.post(request, f"/cases/{case['id']}/user-advice/", json)
     response.raise_for_status()
     return response.json(), response.status_code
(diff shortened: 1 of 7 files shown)


caseworker/advice/services.py
```diff
@@ -23,6 +19,21 @@ def filter_current_user_advice(all_advice, user_id):
...
+    key fields on the Advice model.
+    """
+    destinations = []
+    for dest in case.destinations:
+        if countries is not None:
+            if dest["country"]["id"] not in countries:
+                continue
+        destinations.append((dest["type"], dest["id"]))
+    goods = [("good", good["id"]) for good in case.goods if good["is_good_controlled"]]
+    return destinations + goods
```
alixedi, line 34: I would like to draw attention to this. Does this make sense?

wkeeling: makes sense :+1:

saruniitr: I think we should check for `good['is_good_controlled']['key'] == True` as it can be `False` also

alixedi: I will do the following -

```
good['is_good_controlled']['key']
```

caseworker/advice/views.py
```diff
@@ -25,6 +26,20 @@ def case_id(self):
...
 
+    @property
+    def caseworker(self):
+        user_id = str(self.request.session["lite_api_user_id"])
+        data, _ = get_gov_user(self.request, user_id)
+        return data["user"]
+
+    def unadvised_countries(self):
+        """Returns a dict of countries for which advice has not been given"""
+        dest_types = ("end_user", "ultimate_end_user", "consignee")
```
saruniitr, line 37: there are third parties also in some cases

alixedi: I will include them here. ty 👍

caseworker/advice/services.py
saruniitr: :+1:

caseworker/advice/views.py
```diff
@@ -77,9 +86,14 @@ class GiveApprovalAdviceView(LoginRequiredMixin, CaseContextMixin, FormView):
     Form to recommend approval advice for all products on the application
     """
 
-    form_class = forms.GiveApprovalAdviceForm
     template_name = "advice/give-approval-advice.html"
 
+    def get_form(self):
+        if self.caseworker["team"]["name"] == "FCO":
```
saruniitr, line 92: probably one for next revision - relying on name looks fragile eg if it is typed in lower case or name changed to FCDO etc. Should we consider specifying configuration parameters for Team model eg need countersigning, need advice on all destinations, need consolidation etc which can be used to differentiate here?

saruniitr (approved):
one clarification requested

caseworker/advice/views.py
```diff
@@ -30,7 +30,16 @@ def case(self):
...
     def caseworker(self):
         user_id = str(self.request.session["lite_api_user_id"])
-        return get_gov_user(self.request, user_id)[0]
+        data, _ = get_gov_user(self.request, user_id)
+        return data["user"]
+
+    def unadvised_countries(self):
+        """Returns a dict of countries for which advice has not been given"""
+        dest_types = ("end_user", "ultimate_end_user", "consignee", "third_parties")
+        dests = {advice.get(dest_type) for dest_type in dest_types for advice in self.case.advice} - {None}
```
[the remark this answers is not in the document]
alixedi: Yes, probably. I will do this before merging. ty!

caseworker/advice/views.py
alixedi: Very good point. 

I will create a new follow-up ticket for this.

caseworker/advice/views.py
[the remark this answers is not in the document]
alixedi: On second thought, this is being called before the advice has been consolidated into the team advice and therefore it is ok to not filter for specific advice levels.

This might blow up in my fact at some point but I will deal with it then.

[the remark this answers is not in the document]
saruniitr: ok

PR #295, opened 2021-11-10
LTD-1238: Add option to return to advisor when countersign is refused

saruniitr:
Update form to select the queue to be returned to in case countersign
is refused. Case gets moved to the selected queue if countersigning is
refused.

### DIFF
--- a/caseworker/advice/forms.py
+++ b/caseworker/advice/forms.py
@@ -126,15 +126,23 @@ class CountersignAdviceForm(forms.Form):
         label="Explain why this recommendation needs to be sent back to the advisor",
     )
 
-    def __init__(self, *args, **kwargs):
+    def __init__(self, queues, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.helper = FormHelper()
         self.helper.form_tag = False
+
+        choices = [("", "")] + [(q[0], q[1]) for q in queues]
+        self.fields["queue_to_return"] = forms.ChoiceField(
+            required=False, label="Choose where to return this recommendation", choices=choices
+        )
 
     def clean(self):
         cleaned_data = super().clean()
         option_selected = cleaned_data.get("agree_with_recommendation")
         if option_selected == "yes" and cleaned_data.get("approval_reasons") == "":
             self.add_error("approval_reasons", "Enter your reasons for agreeing with the recommendation")
-        if option_selected == "no" and cleaned_data.get("refusal_reasons") == "":
-            self.add_error("refusal_reasons", "Enter why you do not agree with the recommendation")
+        if option_selected == "no":
+            if cleaned_data.get("refusal_reasons") == "":
+                self.add_error("refusal_reasons", "Enter why you do not agree with the recommendation")
+            if cleaned_data.get("queue_to_return") == "":
+                self.add_error("queue_to_return", "Select where you want to return this recommendation")
--- a/caseworker/advice/views.py
+++ b/caseworker/advice/views.py
@@ -1,13 +1,12 @@
 from django.forms import formset_factory
+from django.forms.formsets import BaseFormSet
 from django.http import HttpResponseRedirect
 from django.views.generic import FormView, TemplateView
 from django.urls import reverse
 
 from caseworker.advice import forms, services
 from caseworker.advice.constants import DECISION_TYPE_VERB_MAPPING
-
 from core import client
-
 from caseworker.cases.services import get_case
 from caseworker.core.services import get_denial_reasons
 from caseworker.users.services import get_gov_user
@@ -262,6 +261,15 @@ class AdviceView(CaseContextMixin, TemplateView):
         }
 
 
+class CountersignAdviceFormSet(BaseFormSet):
+    def get_form_kwargs(self, index):
+        kwargs = super().get_form_kwargs(index)
+        request = kwargs.pop("request")
+        user_pks = kwargs.pop("user_pks")
+        queues, _ = services.get_users_team_queues(request, user_pks[index])
+        return {**kwargs, "queues": queues["queues"]}
+
+
 class ReviewCountersignView(LoginRequiredMixin, CaseContextMixin, TemplateView):
     template_name = "advice/review_countersign.html"
     form_class = forms.CountersignAdviceForm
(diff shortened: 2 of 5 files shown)


caseworker/advice/views.py
```diff
@@ -262,27 +262,55 @@ def get_context(self, **kwargs):
         }
 
 
+class CountersignAdviceFormSet(BaseFormSet):
+    def get_form_kwargs(self, index):
+        kwargs = super().get_form_kwargs(index)
+        request = kwargs.pop("request")
```
alixedi, line 268: You can use `self.request` instead of popping it off here.

caseworker/advice/views.py
```diff
@@ -262,27 +262,55 @@ def get_context(self, **kwargs):
...
+        user_pks = kwargs.pop("user_pks")
```
alixedi, line 269: What is `user_pks` here? 

Also no need to pop. You can just use it.

caseworker/advice/views.py
```diff
@@ -262,27 +262,55 @@ def get_context(self, **kwargs):
...
+        CountersignAdviceFormsetFactory = formset_factory(
+            self.form_class, formset=CountersignAdviceFormSet, extra=num_forms, min_num=num_forms, max_num=num_forms
+        )
+        if not form_kwargs:
+            form_kwargs = {}
+
+        if data:
+            return CountersignAdviceFormsetFactory(data=data, form_kwargs=form_kwargs)
+
+        return CountersignAdviceFormsetFactory(form_kwargs=form_kwargs)
```
alixedi, line 288: Do we have to do this?

caseworker/teams/services.py
```diff
@@ -18,10 +18,11 @@ def get_teams(request, converted_to_options=False):
 
 
 def get_users_team_queues(request, user, convert_to_options=True):
-    data = client.get(request, f"/users/{user}/team-queues/").json()
+    data = client.get(request, f"/users/{user}/team-queues/")
     if convert_to_options:
+        data = data.json()
         return [Option(key=queue[0], value=queue[1], description=None) for queue in data["queues"]]
-    return data, data.status_code
+    return data.json(), data.status_code
```
alixedi, line 25: Sure this is not being used elsewhere?

caseworker/advice/forms.py
```diff
@@ -126,15 +126,23 @@ class CountersignAdviceForm(forms.Form):
...
-    def __init__(self, *args, **kwargs):
+    def __init__(self, queues, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.helper = FormHelper()
         self.helper.form_tag = False
 
+        choices = [("", "")] + [(q[0], q[1]) for q in queues]
+        self.fields["queue_to_return"] = forms.ChoiceField(
+            required=False, label="Choose where to return this recommendation", choices=choices
+        )
```
alixedi, line 137: 1. Why do we have the empty choice?
2. Surely, we can we replace `(q[0], q[1])` with `q`?
3. Why is this not required?

alixedi (commented):
Some initial comments.

caseworker/advice/views.py
saruniitr: Actually no, `request` is not available in `self` as this class is not part of the view, it is only used for creating the formset.
I had to subclass it from `BaseFormSet` to override `get_form_kwargs` so that I can send custom parameters for each formset. In this case the list of queues which can be different in each formset depending on which user provided that advice.

caseworker/advice/views.py
saruniitr: If `request` and `user_pks` are not popped here then the form `__init__` complains that it received unkown parameter `request` and `user_pks`.
I tried adding them to the `kwargs` itself so that it will go as part of form kwargs bu django seems to separate them at a later stage.

caseworker/advice/views.py
saruniitr: Not really since I had to this in two places I have kept this in a function. Two places,
1. in `get_context` to initialise the empty formset and
2. in `post` to initialise with form data

caseworker/teams/services.py
saruniitr: This is used in `get_form_kwargs()` to get the list of user's queues.
If we set `convert_to_options=True` then it converts them into `Option` fields required by `lite_forms` but here I need to be `False` hence has to update this function. In their case they were always calling with True so they didn't come across this issue.

caseworker/advice/forms.py
saruniitr: Designs show empty value when nothing is selected initially. I think it makes sense because we have validation to prompt user to select if nothing is selected.
Also if we don't use empty and show the first item in the list of queues, it would appear that user has selected that even though that is not what he selected. May be instead of completely empty should we have a placeholder text like "Select queue to be returned to"?

`required=False` because of the presentation issue we have in this form. We are supposed to conditionally include approval/refusal reasons based on user selecting Yes/No but we have a problem presenting validation messages so we are displaying everything in this case, making them as optional and handling the validation in `clean()` method below.

alixedi: Ah! Makes sense. TY for explaining @saruniitr

caseworker/advice/views.py
alixedi: Ah! ok.

One thing that might simplify this is to get the `queues` in the view and pass them here directly instead of passing `request` and `user_pks`?

caseworker/advice/views.py
alixedi: I think in this case, separate is better because you have 2 different conditional paths.

caseworker/teams/services.py
alixedi: I meant if you are sure this function is not being used by other code? 

Might be a good idea to leave this alone and create a separate method in `advice.services`?

alixedi (approved):
Had a chat with @saruniitr over the ☎️ 

A couple of things are outstanding but I will leave them with you 🙏

caseworker/advice/views.py
saruniitr: `queues` are different for each formset so we have to make all the requests in the view and pass list of queues for all advisors here or other option is to individually request here.
As discussed leaving as is.

caseworker/advice/views.py
saruniitr: ok splitting as it was before.

caseworker/teams/services.py
saruniitr: :+1:

PR #297, opened 2021-11-10
LTD-1382

Pithikos:
Changes to some text and buttons

### DIFF
--- a/exporter/goods/forms.py
+++ b/exporter/goods/forms.py
@@ -605,7 +605,7 @@ def firearms_number_of_items(firearm_type):
             HiddenField("number_of_items_step", True),
             TextInput(name="number_of_items"),
         ],
-        default_button_name="Save and continue",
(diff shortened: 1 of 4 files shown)


exporter/goods/forms.py
```diff
@@ -444,7 +444,7 @@ def check_document_available_form(back_url):
             ),
         ],
         back_link=BackLink("Back", back_url),
-        default_button_name="Save and continue",
+        default_button_name=DocumentAvailabilityForm.SUBMIT_BUTTON,
```
saruniitr, line 447: actually inline is just fine but since you created it is ok

saruniitr (approved):

exporter/goods/forms.py
Pithikos: Yeah, just did it like this with the aim in the future to make changes a bit more centralized

PR #301, opened 2021-11-17
Add views for countersign and consolidate advice.

alixedi:
Started this as `LTD-1232` but it seemed that it would be uncontroversial to address `LTD-1230` here as well.

As usual, I have taken some liberties in deviating from designs where I felt that I could keep the functionality and address the semantics in a follow-up ticket.

### DIFF
--- a/caseworker/advice/views.py
+++ b/caseworker/advice/views.py
@@ -350,3 +350,13 @@ class CountersignEditAdviceView(EditAdviceView):
 
     def get_context(self, **kwargs):
         return {**super().get_context(), "subtitle": self.subtitle, "edit": True}
+
+
+class CountersignAdviceView(AdviceView):
+    def get_context(self, **kwargs):
+        return {**super().get_context(**kwargs), "countersign": True}
+
+
+class ConsolidateAdviceView(AdviceView):
+    def get_context(self, **kwargs):
+        return {**super().get_context(**kwargs), "consolidate": True}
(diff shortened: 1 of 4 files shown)


caseworker/advice/views.py
```diff
@@ -350,3 +350,13 @@ class CountersignEditAdviceView(EditAdviceView):
 
     def get_context(self, **kwargs):
         return {**super().get_context(), "subtitle": self.subtitle, "edit": True}
+
+
+class CountersignAdviceView(AdviceView):
+    def get_context(self, **kwargs):
+        return {**super().get_context(**kwargs), **{"countersign": True}}
```
wkeeling, line 357: Maybe specify the `countersign` key directly here, rather than wrapping it in it's own dictionary:

```
return {**super().get_context(**kwargs), "countersign": True}
```

caseworker/advice/views.py
```diff
@@ -350,3 +350,13 @@ class CountersignEditAdviceView(EditAdviceView):
...
+
+
+class ConsolidateAdviceView(AdviceView):
+    def get_context(self, **kwargs):
+        return {**super().get_context(**kwargs), **{"consolidate": True}}
```
wkeeling, line 362: same as previous comment :point_up:

wkeeling (approved):
Looks good to me. One very minor comment but feel free to ignore.

saruniitr (approved):
looks good to me as well

eadpearce (approved):

PR #302, opened 2021-11-18
[LTD-1342] Move case forward

alixedi:
We were using system flags to ensure that the case is only moved to countersigning queue when we have advice for all destinations.

While working on it, I realized - the routing is run when the user presses the `I am Done` button. What if we could hide/disable the button until we have advice for all the destinations?

If we could do this (this PR), we don’t need any routing changes or create new system flags, etc.

Just one little thing that is bothering me - when this button is disabled, we should probably display a message that tells users why it is disabled.

### DIFF
--- a/caseworker/advice/templates/advice/view_my_advice.html
+++ b/caseworker/advice/templates/advice/view_my_advice.html
@@ -5,6 +5,20 @@
 <div class="govuk-width-container">
     <a href="#" class="govuk-back-link">Back</a>
     <main class="govuk-main-wrapper">
+        {% for error in form.non_field_errors %}
+        <div class="govuk-error-summary" aria-labelledby="error-summary-title" role="alert" tabindex="-1" data-module="govuk-error-summary">
+          <h2 class="govuk-error-summary__title" id="error-summary-title">
+            There is a problem
+          </h2>
+          <div class="govuk-error-summary__body">
+            <ul class="govuk-list govuk-error-summary__list">
+              <li>
+                {{ error }}
+              </li>
+            </ul>
+          </div>
+        </div>
+        {% endfor %}
         <div class="govuk-grid-row">
             <div class="govuk-grid-column-three-quarters">
               <h1 class="govuk-heading-xl">View Advice</h1>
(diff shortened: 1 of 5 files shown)


caseworker/advice/templates/advice/view_my_advice.html
```diff
@@ -5,6 +5,20 @@
 <div class="govuk-width-container">
     <a href="#" class="govuk-back-link">Back</a>
     <main class="govuk-main-wrapper">
+        {% for error in form.non_field_errors %}
```
saruniitr, line 8: is it done like this because it is just a template and not a form?

alixedi: It is a form but I wanted to show the errors at a different place.

unit_tests/caseworker/advice/views/test_view_my_advice_view.py
```diff
@@ -19,6 +19,35 @@ def url(data_queue, data_standard_case):
...
+            "level": "user",
+            "note": "additional notes",
+            "proviso": "no conditions",
+            "text": "meets the criteria",
+            "third_party": "95c2d6b7-5cfd-47e8-b3c8-dc76e1ac9747",
+            "type": {"key": "proviso", "value": "Proviso"},
+            "ultimate_end_user": None,
+            "user": current_user,
+        }
+        for good_id in ("0bedd1c3-cf97-4aad-b711-d5c9a9f4586e", "6daad1c3-cf97-4aad-b711-d5c9a9f4586e")
```
wkeeling, line 47: Doesn't look as though `good_id` is being used in the fixture data, but that probably doesn't matter for the test?

saruniitr, line 47: `good_id` is not used

alixedi: Good catch. It doesn't matter but it's probably best to use it.

alixedi: Good catch. Using it now.

saruniitr (commented):
Could you add a test to check "Move case forward" is not rendered on the template if advice is not completed? similarly that it is visible when advice is completed.
We can extend it later when advice is sent back etc

saruniitr (approved):

PR #306, opened 2021-11-25
[LTD-1486] Not display advice for LU on consolidate.

alixedi:


caseworker/advice/views.py
```diff
@@ -374,4 +374,6 @@ def get_context(self, **kwargs):
 
 class ConsolidateAdviceView(AdviceView):
     def get_context(self, **kwargs):
-        return {**super().get_context(**kwargs), "consolidate": True}
+        # For LU, we do not want to show the advice summary
+        hide_advice = self.caseworker["team"]["name"] == "Licensing Unit"
```
saruniitr, line 378: `hide_advice_summary` may be as we are hiding the summary only?

saruniitr (approved):

PR #308, opened 2021-11-26
Update test helper function with location changes

saruniitr:
The way we provide location data is changed recently hence updating the
helper function otherwise tests fail.

### DIFF
--- a/tests_common/helpers/applications.py
+++ b/tests_common/helpers/applications.py
@@ -77,7 +77,11 @@ def create_standard_application(api_test_client, context, app_data, submit=True)
             "is_suspected_wmd": False,
             "is_eu_military": False,
         },
-        route_of_goods={"is_shipped_waybill_or_lading": True},
+        route_of_goods={
+            "is_shipped_waybill_or_lading": True,
+            "goods_starting_point": "GB",
+            "goods_recipients": "via_consignee",
+        },
     )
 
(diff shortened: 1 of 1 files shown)


tests_common/helpers/applications.py
```diff
@@ -77,7 +77,11 @@ def create_standard_application(api_test_client, context, app_data, submit=True)
             "is_suspected_wmd": False,
             "is_eu_military": False,
         },
-        route_of_goods={"is_shipped_waybill_or_lading": True},
+        route_of_goods={
+            "is_shipped_waybill_or_lading": True,
+            "goods_starting_point": "GB",
```
eadpearce, line 82: does this work? I think this is the wrong section for these new fields

saruniitr: yes it works, initially I updated in the `draft` section here https://github.com/uktrade/lite-frontend/pull/308/files#diff-26c045785fbccb7f7a5efadacb07051ded8bf6e15f9f754621587e16906dcb46R18 but it was getting overwritten from this section.
Once it is sent to API it is updating in the correct location.

eadpearce: that is weird. it's difficult to see how this whole function works at a glance. but if it works that's fine

eadpearce (approved):

PR #315, opened 2021-12-01
Fix countersign review.

alixedi:
A case with e.g. 2 destinations and 2 goods, has 4 distinct **advice-subjects**. 

As a result, when we save `approval-all` or `refuse-all` advice, we create 4 advice records - one for each subject and send it to the API. 

P.S. This is important and will come in handy when we tackle variable advice.

The flip side to this is that e.g. the countersigner is having to process advice for each subject separately whereas according to the designs, he/she should be able to process the advice from a given advisor as a whole.

What we need to do is to somehow group advice from the same user but different destinations and products into the same block for the purpose of countersigning.

This PR fixes this & hopefully improves on a few other minor things E.g. -

* I have put in a new template `advice_details` that does both approvals and refusals. I am hoping that in another PR, I will be able to get rid of `approval-advice-details` & `refusal-advice-details` entirely.
* I have put in a different template for `countersign_view`. Before this PR, `countersign_review` was being used for this but looking at the code, it seemed like not much was common.

### DIFF
--- a/caseworker/advice/templates/advice/advice_details.html
+++ b/caseworker/advice/templates/advice/advice_details.html
@@ -1,0 +1,140 @@
+{% with user=advice.0.user decision=advice.0.type.key %}
+<table id="table-licenceable-products-approve-all" class="govuk-table">
+    {% if decision == 'approve' or decision == 'proviso' %}
+    <h2 class="app-bg-colour--green govuk-heading-m govuk-!-padding-2">
+        Approved by {{ user|full_name }}
+    </h2>
+    {% elif decision == 'refuse' %}
+    <h2 class="govuk-heading-m app-bg-colour--red govuk-!-padding-2">
+        Refused by {{ user|full_name }}
+    </h2>
+    {% endif %}
+    <thead class="govuk-table__head">
+        <tr class="govuk-table__row">
+            <th scope="col" class="govuk-table__header">Country</th>
+            <th scope="col" class="govuk-table__header">Type</th>
+            <th scope="col" class="govuk-table__header">Name</th>
+            <th scope="col" class="govuk-table__header">Approved products</th>
+        </tr>
+    </thead>
+    <tbody class="govuk-table__body">
+        {% for item in advice %}
+            {% if item.consignee %}
+            {% with consignee=case.data.consignee %}
+            <tr class="govuk-table__row">
+                <td class="govuk-table__cell">{{ consignee.country.name }}</td>
+                <td class="govuk-table__cell">{{ consignee.type|sentence_case }}</td>
+                <td class="govuk-table__cell">{{ consignee.name }}</td>
+                <td class="govuk-table__cell">All</td>
+            </tr>
+            {% endwith %}
+            {% endif %}
+            {% if item.end_user %}
+            {% with end_user=case.data.end_user %}
+            <tr class="govuk-table__row">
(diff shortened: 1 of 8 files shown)


caseworker/advice/templates/advice/advice_details.html
```diff
@@ -0,0 +1,140 @@
...
+    <h2 class="govuk-heading-m app-bg-colour--red govuk-!-padding-2">
+        Refused by {{ user|full_name }}
+    </h2>
+    {% endif %}
+    <thead class="govuk-table__head">
+        <tr class="govuk-table__row">
+            <th scope="col" class="govuk-table__header">Country</th>
+            <th scope="col" class="govuk-table__header">Type</th>
+            <th scope="col" class="govuk-table__header">Name</th>
+            <th scope="col" class="govuk-table__header">Approved products</th>
```
saruniitr, line 17: If the decision is `refuse`, this needs to be different, we also show refusal criteria.

caseworker/advice/templates/advice/review_countersign.html
```diff
@@ -15,66 +14,23 @@
...
-                            {% endfor %}
-                        {% endif %}
-                        <div class="form-actions"> <input type="submit" name="submit" value="Submit" class="govuk-button" id="submit-id-submit"> </div>
-                    </form>
-                {% else %}
-                    <a role="button" draggable="false" class="govuk-button govuk-button--secondary" href="#">
-                        Edit recommendation
-                    </a>
-                    {% if advice_to_countersign|length %}
                         {% for advice in advice_to_countersign %}
```
saruniitr, line 20: `advice_to_countersign` is dictionary of lists with user id as keys correct?

caseworker/advice/urls.py
```diff
@@ -14,6 +14,6 @@
     path("countersign/", views.CountersignAdviceView.as_view(), name="countersign_advice_view"),
     path("countersign/review-advice/", views.ReviewCountersignView.as_view(), name="countersign_review"),
     path("countersign/view-advice/", views.ViewCountersignedAdvice.as_view(), name="countersign_view"),
-    path("countersign/edit-advice", views.CountersignEditAdviceView.as_view(), name="countersign_edit"),
+    path("countersign/edit-advice/", views.CountersignEditAdviceView.as_view(), name="countersign_edit"),
```
saruniitr, line 17: @wkeeling this is fixed

saruniitr (commented):
few comments

caseworker/advice/templates/advice/review_countersign.html
alixedi: Yup.

saruniitr: so in that `advice` here will be the `user id` correct?

alixedi: Sorry, my bad. Yes, `advice_to_countersign` is originally a `dict` but we do a `.values()` when we pass it into the context.

alixedi: [This](https://github.com/uktrade/lite-frontend/pull/315/files#diff-a84e0d397fc7d7cb3b6f58747a8354a7df56357dcd706b4f425eebf5f27a744aR349) is what you are looking for.

saruniitr: ty, makes sense now.

saruniitr (approved):

PR #317, opened 2021-12-02
[LTD-1231] Refactor Formset View

alixedi:
We have a few views that render a list of user-advice with formsets. Each form in the formset renders a `ChoiceField` of queues that are accessble to the user.

This PR introduces `QueueFormset` so that we can re-use some of the code for this in other places.


saruniitr (approved):
looks good to me, mostly refactoring.

PR #319, opened 2021-12-02
LTD-1345: Select advice approval, refusal, footnotes text from picklists

saruniitr:
## Change description
Add option to select advice approval, refusal, footnote text from picklists (list of user defined text blocks).

Approve, Refuse advice forms are updated to use picklists


caseworker/advice/templates/advice/picklist.html
```diff
@@ -0,0 +1,140 @@
...
+        <a id="link-next-page" href="#" class="lite-pagination__navigation-link">
+            <span>Next page</span>
+            {% svg 'next' %}
+        </a>
+    </div>
+</div>
+
+{% block javascript %}
+<script nonce="{{ request.csp_nonce }}">
+    $("a[target={{ target }}]").click(function () {
```
saruniitr, line 32: When the link on the form is clicked, this gets called and brings up the Modal to select a picklist item

caseworker/advice/templates/advice/picklist.html
```diff
@@ -0,0 +1,140 @@
...
+        var target = item.data("target")
+        $("#" + target + "-text").removeClass("app-hidden");;
+        $("#" + target + "-text output")[0].value = item.data("text")
+        $("#" + target + "-input").val(item.data("text"));
+        $("#link-" + target + "-picklist-picker").hide();
+        $("#link-" + target + "-picklist-picker-change").removeClass("app-hidden");
+        LITECommon.Modal.closeAllModals();
+    }
+
+    function setTextareaValue(textareaId) {
```
saruniitr, line 131: this is updated the target text field with the text from the selected picklist item

caseworker/advice/templates/advice/picklist.html
alixedi: @saruniitr Might be better to have a single `js` function defined separately and call it with the target as an argument here. Something like -

```
$("a[target={{ target }}]").click(picklist({{ target }}))
```

Whereas, separately -

```
def picklist(target) {
    LITECommon.Modal.....
}
```

caseworker/advice/forms.py
```diff
@@ -41,22 +42,35 @@ def __init__(self, *args, **kwargs):
 
 class GiveApprovalAdviceForm(forms.Form):
 
-    proviso = forms.CharField(
-        label="Add a licence condition (optional)", widget=forms.Textarea(attrs={"rows": "10"}), required=False
-    )
     approval_reasons = forms.CharField(
-        widget=forms.Textarea(attrs={"rows": "10"}),
+        widget=forms.Textarea(attrs={"rows": "10", "style": "margin-top:20px"}),
```
wkeeling, line 46: I'm wondering if using a fixed pixel `margin-top` here is going to cause problems on other types of devices - e.g. mobile?

saruniitr: good point, I will have a quick search and replace accordingly.

saruniitr: @wkeeling there is a govuk class, I have updated to use that class

caseworker/advice/templates/advice/picklist.html
saruniitr: Anonymous function is moved into a separate function

wkeeling (approved):

caseworker/advice/forms.py
```diff
@@ -41,22 +41,31 @@ def __init__(self, *args, **kwargs):
...
 class GiveApprovalAdviceForm(forms.Form):
 
-    proviso = forms.CharField(
-        label="Add a licence condition (optional)", widget=forms.Textarea(attrs={"rows": "10"}), required=False
-    )
     approval_reasons = forms.CharField(
-        widget=forms.Textarea(attrs={"rows": "10"}),
+        widget=forms.Textarea(attrs={"rows": "10", "class": "govuk-!-margin-top-4"}),
         label="What are your reasons for approving?",
+        help_text='<a class="govuk-link govuk-link--no-visited-state" href="#" target="approval_reasons">Choose an approval reason from the template list</a>',
```
alixedi, line 47: Can we not subclass `CharField` so that we don't have to put the `help_text` incantation every time for `picklist` inputs?

Something like -

```
class PicklistCharField(forms.CharField):
    def __init__(self, target, *args, **kwargs):
        help_text = f'<a class="govuk-link govuk-link--no-visited-state" href="#" target="{target}">Choose an approval reason from the template list</a>'
        super.__init__(*args, **kwargs, help_text=)
```

caseworker/advice/templates/advice/picklist.html
```diff
@@ -0,0 +1,142 @@
...
+            {% svg 'next' %}
+        </a>
+    </div>
+</div>
+
+{% block javascript %}
+<script nonce="{{ request.csp_nonce }}">
+    $("a[target={{ target }}]").on("click", {"picklist_name": `{{ picklist_name }}`, "target": `{{ target }}`}, showPickListItems);
+
+    function showPickListItems(event) {
```
alixedi, line 34: The reason I asked you to put this in a separate function was so that this bit is not repeated 3 times on the rendered page. 

ATM, whenever you do -

```
{% include "advice/picklist.html" with target="approval_reasons" picklist_type="standard_advice" picklist_name="Standard Advice" %}
```

It copies over everything in this template (including this function) into the rendered template. This should be fixable by moving the function to another template and including it once.

alixedi (changes requested):

caseworker/advice/templates/advice/picklist.html
```diff
@@ -0,0 +1,142 @@
...
+        LITECommon.Modal.showModal(`Select a ${event.data.name} picklist item`, $(`#${event.data.target}-picklist-picker`).html(), false, true, { maxWidth: '600px' });
+        filter{{ target }}PicklistItems(true);
+        $(".lite-modal .app-picklist__search-input").on('input propertychange paste', function () {
+            $(`.lite-modal #${event.data.target}-picklist-picker__container`).addClass("app-picklist-picker__container--loading");
+            filter{{ target }}PicklistItems(true, $(this).val());
```
alixedi, line 39: Can we not replace this with -

```
def FilterPicklistItems(true, $(this).val(), target)
```

The fact that we are choosing to template a function where can just pass arguments doesn't make sense.

caseworker/advice/forms.py
saruniitr: :+1:

caseworker/advice/templates/advice/picklist.html
saruniitr: moved this into a template and only included once now so no repetition and not templating the function.

caseworker/advice/templates/advice/picklist.html
saruniitr: Moving this function to a template is not bringing up the modal when the link is clicked.

saruniitr: repetition is avoided now

caseworker/advice/forms.py
```diff
@@ -27,6 +27,17 @@ def get_refusal_advice_form_factory(advice, denial_reasons_choices):
     return RefusalAdviceForm(data=data, denial_reasons=denial_reasons_choices)
 
 
+class PicklistCharField(forms.CharField):
+    def get_help_link(self, picklist_attrs, help_text):
+        picklist_tags = f'picklist_type="{picklist_attrs.get("type")}" picklist_name="{picklist_attrs.get("name")}" target="{picklist_attrs.get("target")}"'
+        return f'<a class="govuk-link govuk-link--no-visited-state" href="#" {picklist_tags}>{help_text}</a>'
+
+    def __init__(self, picklist_attrs, label, help_text, required=True, min_rows=10, **kwargs):
```
alixedi, line 35: I think we can remove `required` and `min_rows` from here, no? Because you are passing through `kwargs` to super and these would be included without having to put them explicitly in the parameters.

alixedi (approved):
One comment that I will leave with you. 

A reminder that we will need to add browser tests for this later. Because of all the `js`.

caseworker/advice/forms.py
saruniitr: I was expecting the same but when removed django-forms complaining that it received `unexpected keyword argument 'min_rows'`

saruniitr: yeah, I will remove them.

PR #335, opened 2021-12-08
[LTD-1487] Radio button conditional form

alixedi:
This works now but please checkout this branch on your local and try it before approving.


wkeeling (approved):

saruniitr (approved):
When neither yes/no selected the error message displayed is not as per design, I think it should be `Select yes if you agree with the recommendation`
with that fixed it looks good to me

PR #337, opened 2021-12-09
LTD-1529: Retain case queues when countersigning is refused

saruniitr:
## Change description

When the countersigner doesn't agree with the recommendation the case is
sent back to advisor along with the queue it needs to go back but this
was unassigning the case from it's current set of queues. We only want
it to be unassigned from the countersigning queue and it fixes this.

### DIFF
--- a/caseworker/advice/services.py
+++ b/caseworker/advice/services.py
@@ -145,7 +145,7 @@ def get_users_team_queues(request, user):
 
 def countersign_advice(request, case, caseworker, formset_data):
     data = []
-    queues = []
+    queues_to_return = []
     case_pk = case["id"]
     advice_to_countersign = get_advice_to_countersign(case.advice, caseworker)
 
@@ -156,16 +156,16 @@ def countersign_advice(request, case, caseworker, formset_data):
             comments = form_data["approval_reasons"]
         elif form_data["agree_with_recommendation"] == "no":
             comments = form_data["refusal_reasons"]
-            if form_data["queue_to_return"] not in queues:
-                queues.append(form_data["queue_to_return"])
+            if form_data["queue_to_return"] not in queues_to_return:
+                queues_to_return.append(form_data["queue_to_return"])
         for advice in user_advice:
             data.append({"id": advice["id"], "countersigned_by": caseworker["id"], "countersign_comments": comments})
 
     response = client.put(request, f"/cases/{case_pk}/countersign-advice/", data)
     response.raise_for_status()
 
-    if queues:
-        put_case_queues(request, case_pk, json={"queues": queues})
+    current_queues = [queue_id for queue_id in case.queues if queue_id != "FCO Counter-signing"]
+    put_case_queues(request, case_pk, json={"queues": current_queues + queues_to_return})
 
     return response.json(), response.status_code
 
--- a/caseworker/advice/services.py
+++ b/caseworker/advice/services.py
@@ -5,6 +5,8 @@ from requests.exceptions import HTTPError
 from core import client
 from caseworker.cases.services import put_case_queues
 
+FCDO_COUNTERSIGNING_QUEUE = "5e772575-9ae4-4a16-b55b-7e1476d810c4"
+
 
 def filter_nlr_products(products):
     return [
@@ -145,7 +147,7 @@ def get_users_team_queues(request, user):
 
 def countersign_advice(request, case, caseworker, formset_data):
     data = []
-    queues_to_return = []
+    queues_to_return = set()
     case_pk = case["id"]
     advice_to_countersign = get_advice_to_countersign(case.advice, caseworker)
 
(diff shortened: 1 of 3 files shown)


caseworker/advice/services.py
```diff
@@ -156,16 +156,19 @@ def countersign_advice(request, case, caseworker, formset_data):
...
             data.append({"id": advice["id"], "countersigned_by": caseworker["id"], "countersign_comments": comments})
 
     response = client.put(request, f"/cases/{case_pk}/countersign-advice/", data)
     response.raise_for_status()
 
-    if queues:
-        put_case_queues(request, case_pk, json={"queues": queues})
+    current_queues = [id for id, name in zip(case.queues, case.queue_names) if name != "FCO Counter-signing"]
+    if queues_to_return:
+        put_case_queues(request, case_pk, json={"queues": current_queues + queues_to_return})
```
saruniitr, line 169: intentionally duplicated in `if-else` to make it more clear and readable

caseworker/advice/services.py
```diff
@@ -156,16 +156,19 @@ def countersign_advice(request, case, caseworker, formset_data):
...
+                queues_to_return.append(form_data["queue_to_return"])
         for advice in user_advice:
             data.append({"id": advice["id"], "countersigned_by": caseworker["id"], "countersign_comments": comments})
 
     response = client.put(request, f"/cases/{case_pk}/countersign-advice/", data)
     response.raise_for_status()
 
-    if queues:
-        put_case_queues(request, case_pk, json={"queues": queues})
+    current_queues = [id for id, name in zip(case.queues, case.queue_names) if name != "FCO Counter-signing"]
```
saruniitr, line 167: currently we don't have a better way other than relying on the `name` or `id`

wkeeling: Just wondering if using `id` would be slightly safer than `name` (thinking that names more likely to change than ids - e.g. FCO -> FCDO)?

saruniitr: ok, I will go with the `id` and add a comment about which queue that `id` is for.

caseworker/advice/services.py
wkeeling: You could eliminate the `if-else` completely though I think, since if `queues_to_return` is empty, then nothing additional gets appended to the `current_queues`. So this should cover both cases:
```
put_case_queues(request, case_pk, json={"queues": current_queues + queues_to_return})
```

saruniitr: ah yes, that's correct, will change it

caseworker/advice/services.py
```diff
@@ -156,16 +156,17 @@ def countersign_advice(request, case, caseworker, formset_data):
...
         for advice in user_advice:
             data.append({"id": advice["id"], "countersigned_by": caseworker["id"], "countersign_comments": comments})
 
     response = client.put(request, f"/cases/{case_pk}/countersign-advice/", data)
     response.raise_for_status()
 
-    if queues:
-        put_case_queues(request, case_pk, json={"queues": queues})
+    # Remove the case from the FCO countersigning queue
+    current_queues = [queue_id for queue_id in case.queues if queue_id != "5e772575-9ae4-4a16-b55b-7e1476d810c4"]
```
wkeeling, line 168: Maybe use a named constant for clarity?, e.g.
```

current_queues = [queue_id for queue_id in case.queues if queue_id != FCDO_COUNTER_SIGNING_QUEUE
```

saruniitr: :+1:

caseworker/advice/services.py
```diff
@@ -145,7 +147,7 @@ def get_users_team_queues(request, user):
 
 def countersign_advice(request, case, caseworker, formset_data):
     data = []
-    queues = []
+    queues_to_return = []
```
alixedi, line 150: 1) This could be a set.

alixedi (approved):

caseworker/advice/services.py
saruniitr: :+1:

caseworker/advice/services.py
```diff
@@ -156,16 +158,17 @@ def countersign_advice(request, case, caseworker, formset_data):
...
         for advice in user_advice:
             data.append({"id": advice["id"], "countersigned_by": caseworker["id"], "countersign_comments": comments})
 
     response = client.put(request, f"/cases/{case_pk}/countersign-advice/", data)
     response.raise_for_status()
 
-    if queues:
-        put_case_queues(request, case_pk, json={"queues": queues})
+    # Remove the case from the FCO countersigning queue
+    current_queues = [queue_id for queue_id in case.queues if queue_id != FCDO_COUNTERSIGNING_QUEUE]
```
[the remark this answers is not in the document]
saruniitr: :+1:

PR #338, opened 2021-12-13
Consider team when filtering destinations

wkeeling:
Only exclude a destination when it has already been advised on by the current user's team.

### DIFF
--- a/caseworker/advice/views.py
+++ b/caseworker/advice/views.py
@@ -37,11 +37,22 @@ class CaseContextMixin:
         return data["user"]
 
     def unadvised_countries(self):
-        """Returns a dict of countries for which advice has not been given"""
+        """Returns a dict of countries for which advice has not been given by the current user's team."""
         dest_types = ("end_user", "ultimate_end_user", "consignee", "third_party")
-        dests = {advice.get(dest_type) for dest_type in dest_types for advice in self.case.advice} - {None}
+
+        advised_on = {
+            # Map of destinations advised on -> team that gave the advice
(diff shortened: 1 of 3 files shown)


caseworker/advice/views.py
```diff
@@ -37,11 +37,22 @@ def caseworker(self):
...
         dest_types = ("end_user", "ultimate_end_user", "consignee", "third_party")
-        dests = {advice.get(dest_type) for dest_type in dest_types for advice in self.case.advice} - {None}
+
+        advised_on = {
+            # Map of destinations advised on -> team that gave the advice
+            advice.get(dest_type): advice["user"]["team"]["id"]
+            for dest_type in dest_types
+            for advice in self.case.advice
+        }
+        advised_on.pop(None)
```
saruniitr, line 49: does this read better if we filter it using `if` condition when generating the map?

saruniitr (approved):

caseworker/advice/views.py
wkeeling: I was trying to eliminate a duplicate `advice.get(dest_type)` call, because if we used `if` then we'd need to do:

```
advised_on = {
    # Map of destinations advised on -> team that gave the advice
    advice.get(dest_type): advice["user"]["team"]["id"]
    for dest_type in dest_types
    for advice in self.case.advice
    if advice.get(dest_type) is not None  # Duplicated call
}
```

wkeeling: Actually `advised_on.pop(None)` falls over anyway when `None` is not present. Have updated.

PR #339, opened 2021-12-13
LTD-1571: Fix display of security grading value in case details summary

saruniitr:
## Change description

Currently when displaying security grading we are looking for pv_grading_details
which is actually a dict, we should be using is_pv_graded which is a string and
the values of all products are joined and displayed as a comma separated value.

If the product is security graded we should also display the actual grading
instead of yes/no so a follow up is required to display more details.

### DIFF
--- a/caseworker/advice/templates/advice/case_detail.html
+++ b/caseworker/advice/templates/advice/case_detail.html
@@ -99,7 +99,7 @@
                 <p class="govuk-body"><strong>Security grading</strong></p>
             </div>
             <div class="govuk-grid-column-one-half">
-                <p class="govuk-body">{{ case.goods | get_security_grading | join:', ' }}<br></p>
+                <p class="govuk-body">{{ case.goods | get_security_grading | join:', ' | title }}<br></p>
             </div>
             <div class="govuk-grid-column-one-half">
                 <p class="govuk-body"><strong>Total value</strong></p>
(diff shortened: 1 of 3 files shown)


agalamatis (approved):
+1 from a superficial look

caseworker/advice/templates/advice/case_detail.html
```diff
@@ -99,7 +99,7 @@ <h2 class="govuk-heading-m">Case Details</h2>
                 <p class="govuk-body"><strong>Security grading</strong></p>
             </div>
             <div class="govuk-grid-column-one-half">
-                <p class="govuk-body">{{ case.goods | get_security_grading | join:', ' }}<br></p>
+                <p class="govuk-body">{{ case.goods | get_security_grading | join:', ' | title }}<br></p>
```
wkeeling, line 102: This looks as though it will now potentially output a comma separated list of `yes` or `no` strings, so you could have e.g. `No, No, Yes, Yes, Yes`. Just double-checking that is definitely intended as it seems a bit strange?

wkeeling: Nevermind - just noticed you're using a `set` (so no repetition) and I've also just read the Teams chat which explains!

wkeeling (approved):

PR #340, opened 2021-12-14
[LTD-1564] Consolidate edit

alixedi:
This is the consolidate edit view. 

It has a few clever bits in how (1) it inherits from the `consolidate-review` to re-use some bits (2) gets into trouble and (3) uses `setup` method to get out of trouble.


wkeeling (approved):
Looks fine to me.

PR #342, opened 2021-12-15
LTD-1549: View consolidated outcome

saruniitr:
## Change description

Add template to show the consolidated outcome from LU.


caseworker/advice/services.py
```diff
@@ -92,6 +97,24 @@ def get_advice_to_consolidate(advice):
     return grouped_user_advice
 
 
+def arrange_by_party_type(all_advice):
+    arranged_advice = []
```
saruniitr, line 101: reorders the list of advice objects in the logical order of parties ie., consignee first, end user etc

caseworker/advice/templates/advice/view_consolidate.html
```diff
@@ -0,0 +1,76 @@
...
+                        <a role="button" draggable="false" class="govuk-button govuk-button--secondary" href="{% url 'cases:consolidate_edit' queue_pk case.id %}">
+                                Edit outcome
+                        </a>
+                    {% endif %}
+                    {% with advice=consolidated_advice %}
+                        <br><br>
+                        {% include "advice/advice_details.html" %}
+                        <br>
+                    {% endwith %}
+                    <h2 class="govuk-heading-m">Other recommendations for this case</h2>
```
saruniitr, line 26: I will move this into a template

caseworker/advice/views.py
```diff
@@ -449,4 +449,17 @@ def get_form_kwargs(self):
...
+
+class ViewConsolidatedAdvice(AdviceView):
+    template_name = "advice/view_consolidate.html"
+
+    def get_context(self, **kwargs):
+        context = super().get_context()
+        consolidated_advice = services.get_consolidated_advice(self.case.advice, services.LICENSING_UNIT_TEAM_ID)
+        context["consolidated_advice"] = consolidated_advice
+        context["nlr_products"] = services.filter_nlr_products(self.case["data"]["goods"])
+        context["user_can_finalise"] = self.caseworker["team"]["id"] == services.LICENSING_UNIT_TEAM_ID
```
saruniitr, line 463: to prevent users from other teams from finalising the case.

caseworker/advice/services.py
alixedi: Try -

```
sorted(x, key=lambda x: party_types.index(advice["party_type"]))
```

alixedi: Also, consider renaming this function to `order_by_party_type`

caseworker/advice/services.py
```diff
@@ -92,6 +97,24 @@ def get_advice_to_consolidate(advice):
...
+        for index, advice in enumerate(all_advice):
+            if advice.get(party_type):
+                arranged_advice.append(all_advice.pop(index))
+
+    return arranged_advice
+
+
+def get_consolidated_advice(advice, team_id):
+    team_advice = filter_advice_by_level(advice, ["team"])
+    consolidated_advice = filter_advice_by_team(team_advice, team_id)
```
alixedi, line 113: Can we remove `get_my_team_advice` and refactor the code that uses it to instead use `filter_advice_by_team` please?

caseworker/advice/services.py
```diff
@@ -92,6 +97,24 @@ def get_advice_to_consolidate(advice):
...
+    consolidated_advice = arrange_by_party_type(consolidated_advice)
```
alixedi, line 114: Please consider `return arrange_by_party_type(consolidated_advice)` or if you really want to keep this line, come up with a new variable name?

caseworker/advice/services.py
```diff
@@ -6,6 +6,7 @@
 from caseworker.cases.services import put_case_queues
 
 FCDO_COUNTERSIGNING_QUEUE = "5e772575-9ae4-4a16-b55b-7e1476d810c4"
+LICENSING_UNIT_TEAM_ID = "58e77e47-42c8-499f-a58d-94f94541f8c6"
```
alixedi, line 9: Presume LU has a single team?

caseworker/advice/views.py
alixedi: Might want to double-check in the API if all LU users can finalize.

alixedi (commented):
I have a general thought about this PR -

Do you think that screen you are building here is different enough from `advice_view` to deserve its own view and template?

As opposed to something along these lines https://github.com/uktrade/lite-frontend/blob/0a04c184c3776b25b270fdcfac4aa59ffd3cf432/caseworker/advice/templates/advice/view-advice.html#L11-L17

caseworker/advice/templates/advice/view_consolidate.html
```diff
@@ -0,0 +1,76 @@
...
+
+                    {% for team_advice in grouped_advice %}
+                        <details class="govuk-details" data-module="govuk-details">
+                            <summary class="govuk-details__summary">
+                                <span class="govuk-details__summary-text govuk-!-font-weight-bold">
+                                    {{ team_advice.team.name }}
+                                </span>
+                            </summary>
+
+                            {% for user_decision_advice in team_advice.advice %}
```
alixedi, line 36: Is this different to `advice_details.html`?

caseworker/advice/views.py
saruniitr: API checks if the user has permission to manage licence final advice

caseworker/advice/services.py
saruniitr: :+1:

caseworker/advice/services.py
saruniitr: :+1: was actually going to remove it

saruniitr:
> I have a general thought about this PR -
> 
> Do you think that screen you are building here is different enough from `advice_view` to deserve its own view and template?
> 
> As opposed to something along these lines
> 
> https://github.com/uktrade/lite-frontend/blob/0a04c184c3776b25b270fdcfac4aa59ffd3cf432/caseworker/advice/templates/advice/view-advice.html#L11-L17

As per the design there is no title for this page, the above `if-else` can be modified to make it work for this page with more conditions specific to this view.

caseworker/advice/services.py
saruniitr: yes, only one team can finalise

wkeeling: Maybe adopt the same naming convention as the constant above - i.e. drop the `_ID` or append an `_ID` to `FCDO_COUNTERSIGNING_QUEUE`, but keep them the same either way.

saruniitr: :+1:

caseworker/advice/services.py
```diff
@@ -92,6 +97,24 @@ def get_advice_to_consolidate(advice):
     return grouped_user_advice
 
 
+def arrange_by_party_type(all_advice):
+    arranged_advice = []
+    party_types = ("consignee", "end_user", "ultimate_end_user", "third_party")
+    for party_type in party_types:
+        for index, advice in enumerate(all_advice):
+            if advice.get(party_type):
+                arranged_advice.append(all_advice.pop(index))
```
wkeeling, line 106: This looks like it's modifying the `all_advice` list by `pop`-ing whilst iterating over it? I guess by using `pop` you're trying to prevent the same advice from being added to `arranged_advice` more than once (if that advice were to have more than one `party_type`)? Rather than `pop`, maybe check whether the advice has already been added to `arranged_advice`?
```
if advice.get(party_type) and advice not in arranged_advice:
    arranged_advice.append(advice)
```

saruniitr: yes popping to avoid it being added twice. It is possible to have more than one party type eg for ultimate end users or third parties but since we are popping by index it is ok.
I will add the additional check as well to be safe.

wkeeling: Won't popping mean that the list indices will change, which could result in not all elements being iterated? e.g. https://stackoverflow.com/questions/13939341/why-does-a-for-loop-with-pop-method-or-del-statement-not-iterate-over-all-list

saruniitr: good point. Since we are checking if that advice item is already added or not we can leave it in the original list so no need to `pop`.

caseworker/advice/views.py
```diff
@@ -295,9 +295,15 @@ def group_advice(self):
...
+        consolidated_advice = services.get_consolidated_advice(self.case.advice, services.LICENSING_UNIT_TEAM)
+        nlr_products = services.filter_nlr_products(self.case["data"]["goods"])
+
         return {
             "queue": self.queue,
             "grouped_advice": self.grouped_advice,
+            "consolidated_advice": consolidated_advice,
+            "nlr_products": nlr_products,
+            "user_can_finalise": self.caseworker["team"]["id"] == services.LICENSING_UNIT_TEAM,
         }
```
wkeeling, line 307: I imagined that reusing `AdviceView` would make sense, but it would still make sense to have a concrete subclass for the consolidate specific attributes (in the same way that `CountersignAdviceView` and `ConsolidateAdviceView` work) - e.g.
```python
class ViewConsolidatedAdviceView(AdviceView):
        consolidated_advice = services.get_consolidated_advice(self.case.advice, services.LICENSING_UNIT_TEAM)

        return {
            **super().get_context(**kwargs), 
            "consolidated_advice": consolidated_advice,
            "nlr_products": nlr_products,
            "user_can_finalise": self.caseworker["team"]["id"] == services.LICENSING_UNIT_TEAM,
        }
```
That would mean we wouldn't need to pollute `AdviceView` with consolidate specific code?

wkeeling: Sorry if that wasn't clear from my original comment :( -- what are your thoughts?

saruniitr: :+1:

wkeeling (approved):

caseworker/advice/views.py
```diff
@@ -368,7 +368,7 @@ def get_context(self, **kwargs):
 class ConsolidateAdviceView(AdviceView):
     def get_context(self, **kwargs):
         # For LU, we do not want to show the advice summary
-        hide_advice = self.caseworker["team"]["name"] == "Licensing Unit"
+        hide_advice = self.caseworker["team"]["id"] == services.LICENSING_UNIT_TEAM
```
alixedi, line 371: ty

caseworker/advice/services.py
```diff
@@ -38,19 +39,17 @@ def filter_advice_by_users_team(all_advice, caseworker):
     return [advice for advice in all_advice if advice["user"]["team"]["id"] == caseworker["team"]["id"]]
 
 
+def filter_advice_by_team(all_advice, team_id):
+    return [advice for advice in all_advice if advice["user"]["team"]["id"] == team_id]
```
alixedi, line 43: I think this can also call `filter_advice_by_level` before filtering by the given `team_id`.

In places where we are calling this, the call is always preceded by `filter_advice_by_level`.

alixedi (approved):
Looks good. A couple of comments that I will leave with you.

caseworker/advice/services.py
saruniitr: intentionally that is not done because we could use this for `user` level advice as well.
If we filter by `level` then we need to reflect that in the name also.

PR #343, opened 2021-12-16
Use good on licence, fallback to good for NLR

wkeeling:


caseworker/advice/templates/advice/case_detail.html
```diff
@@ -21,7 +21,15 @@ <h2 class="govuk-heading-m">Case Details</h2>
...
                     <div class="govuk-grid-column-one-half">
-                        <p class="govuk-body">{{ good.good.is_good_controlled.value }}</p>
+                        <p class="govuk-body">
+                        {% if good.is_good_controlled is not None %}
+                            {{ good.is_good_controlled.value }}
+                        {% elif good.good.is_good_controlled is not None %}
+                            {{ good.good.is_good_controlled.value }}
+                        {% else %}
+                            Not added
+                        {% endif %}
```
wkeeling, line 31: This logic is now in alignment with the logic that displays the NLR value against the good on the main case detail page. i.e. prefer "good on licence" when present, otherwise fallback to the underlying "good".

saruniitr: yes, if it is reviewed then `good.is_good_controlled` is set and we should be using that, otherwise fallback to `good.good.is_good_controlled`

saruniitr (approved):

PR #344, opened 2021-12-16
[LTD-1574] [LTD-1575] Redirect fixes

alixedi:


caseworker/advice/views.py
```diff
@@ -406,7 +408,7 @@ def get_success_url(self):
             recommendation = self.request.POST.get("recommendation")
```
saruniitr, line 408: Do we usually get into this condition?
`if self.kwargs.get("advice_type") is None:`

Users will either approve or refuse so we always get `advice_type` correct?

alixedi: Yes, we get into this when we display the `ConsolidateSelectAdviceForm`.

wkeeling (approved):
LGTM

PR #347, opened 2021-12-17
LTD-1563: Design fixes and add move case forward

saruniitr:
## Change description

Fixes below issues,
- the green banner with ‘Review successful.’ shouldn’t be there
- Display ‘Edit recommendation’ button
- Display ‘Move case forward’ button when the case cannot be finalised
- Display approver team
- Remove excess white space above ' Other recommendations for this case'


caseworker/advice/views.py
```diff
@@ -455,17 +453,40 @@ def get_success_url(self):
...
+        queue_id = str(self.kwargs["queue_pk"])
+        try:
+            services.move_case_forward(self.request, self.case.id, queue_id)
+        except HTTPError as e:
+            errors = e.response.json()["errors"]["queues"]
+            for error in errors:
+                form.add_error(None, error)
+            return super().form_invalid(form)
+        return super().form_valid(form)
+
```
alixedi, line 490: Might be useful to think about refactoring this code because this seems like the same as `AdviceDetailView.form_valid`.

alixedi: Like don't have to do this in this PR but maybe make a separate ticket?

caseworker/advice/views.py
```diff
@@ -455,17 +453,40 @@ def get_success_url(self):
...
+    form_class = forms.MoveCaseForwardForm
+
     def get_context(self, **kwargs):
         user_team_id = self.caseworker["team"]["id"]
         consolidated_advice = []
         if user_team_id in [services.LICENSING_UNIT_TEAM, services.MOD_ECJU_TEAM]:
             consolidated_advice = services.get_consolidated_advice(self.case.advice, user_team_id)
         nlr_products = services.filter_nlr_products(self.case["data"]["goods"])
 
+        lu_countersign_flags = set([services.LU_COUNTERSIGN_REQUIRED, services.LU_SR_MGR_CHECK_REQUIRED])
```
alixedi, line 466: Might be worth considering using a `dict` to define these flags? Or even a set like -

```
LU_COUNTERSIGN_FLAGS = {<id-1>, <id-2>}
```

caseworker/advice/views.py
```diff
@@ -455,17 +453,40 @@ def get_success_url(self):
...
+        case_flag_ids = set([flag["id"] for flag in self.case.flags])
```
alixedi, line 467: Matter of taste but I prefer the `{}` set notation -

```
{flag["id"] for flag in self.case.flags}
```

caseworker/advice/views.py
```diff
@@ -455,17 +453,40 @@ def get_success_url(self):
...
+        lu_countersign_required = bool(lu_countersign_flags.intersection(case_flag_ids))
```
alixedi, line 468: So this would return `true` if either (and not both) of those flags are present, right? Is that what we want?

caseworker/advice/views.py
```diff
@@ -455,17 +453,40 @@ def get_success_url(self):
...
             **super().get_context(**kwargs),
             "consolidated_advice": consolidated_advice,
             "nlr_products": nlr_products,
-            "user_can_finalise": user_team_id == services.LICENSING_UNIT_TEAM,
+            "finalise_case": finalise_case,
+            "lu_countersign_required": lu_countersign_required,
         }
+
+    def form_valid(self, form):
+        queue_id = str(self.kwargs["queue_pk"])
```
alixedi, line 481: Since we inherit from `AdviceView`, I think we get this for free - `self.queue_id`?

caseworker/advice/views.py
saruniitr: correct, `list` is not required, `set` can be used directly

caseworker/advice/views.py
saruniitr: :+1: initially was planning to do with lists but set is better

caseworker/advice/views.py
saruniitr: yes, if either of them is present then we need to countersign

caseworker/advice/views.py
saruniitr: ah ok, nice.

caseworker/advice/views.py
saruniitr: will do separately

caseworker/advice/views.py
saruniitr: :+1:

wkeeling (approved):

PR #348, opened 2021-12-20
LTD-1573: Update display of PV security grading details in case details section

saruniitr:
## Change description
In Case details side panel currently we only display yes/no to indicate
whether a product is pv graded or not in both individual product section
and in summary section below which is not very useful. This is updated to
show the classification level as well if a product is pv graded.

If none of the products are pv graded then that field is completely removed
from individual section and it is mentioned as 'No' in the summary.

If any of the product is pv graded then we should the classification and
mention it as 'Yes' in the summary section.


caseworker/advice/templates/advice/case_detail.html
```diff
@@ -104,10 +115,10 @@ <h2 class="govuk-heading-m">Case Details</h2>
                 <p class="govuk-body">{{ case.goods | get_clc | join:', ' }}<br></p>
             </div>
             <div class="govuk-grid-column-one-half">
-                <p class="govuk-body"><strong>Security grading</strong></p>
+                <p class="govuk-body"><strong>Security graded</strong></p>
             </div>
             <div class="govuk-grid-column-one-half">
-                <p class="govuk-body">{{ case.goods | get_security_grading | join:', ' | title }}<br></p>
+                <p class="govuk-body">{% if display_pv_grading %} Yes {% else %} No {% endif %}<br></p>
```
eadpearce, line 121: as a shortcut here you can do `{{ display_pv_grading|yesno|capfirst }}` https://docs.djangoproject.com/en/2.2/ref/templates/builtins/#yesno

caseworker/advice/templatetags/advice_tags.py
```diff
@@ -15,10 +15,11 @@ def get_case_value(goods):
...
 
-@register.filter()
-def get_security_grading(goods):
-    gradings = {good.get("good", {}).get("is_pv_graded") for good in goods}
-    return sorted(gradings - {None})
+@register.filter
+def is_case_pv_graded(products):
+    """Returns True if pv_grading is True for atleast one of the products on the application"""
+    gradings = {product.get("good", {}).get("is_pv_graded") for product in products}
+    return bool({"yes"}.intersection(gradings))
```
eadpearce, line 22: I think `"yes" in gradings` would be easier to read here

eadpearce (approved):

caseworker/advice/templatetags/advice_tags.py
saruniitr: :+1:

PR #349, opened 2021-12-20
LTD-1544: Add option to finalise the case

saruniitr:
## Change description

Consider the LU consolidate as the final advice as that is required before
the case can be finalised


wkeeling (approved):
LGTM

alixedi (approved):

PR #351, opened 2021-12-20
LTD 1548: Tidy other recommendations, align with designs

wkeeling:
image


caseworker/advice/templates/advice/other-recommendations.html
```diff
@@ -0,0 +1,117 @@
...
+		<summary class="govuk-details__summary">
+			<span class="govuk-details__summary-text govuk-!-font-weight-bold">
+				{{ team_advice.team.name }} {{ user_advice.decision }}
+			</span>
+		</summary>
+        {% for decision_advice in user_advice.advice %}
+            <div class="govuk-details__text">
+                <h2 class="govuk-!-font-weight-bold govuk-body govuk-!-padding-2 {% if decision_advice.decision == "Approve" or decision_advice.decision == "Proviso" %}app-bg-colour--green{% elif decision_advice.decision == "Refuse" %}app-bg-colour--red{% endif %}">
+                    {{ decision_advice.decision_verb|capfirst }} by {% if user_advice.user.first_name %}{{ user_advice.user.first_name|add:' '|add:user_advice.user.last_name }}{% else %}{{ user_advice.user.email }}{% endif %}
+                    <span class="govuk-!-font-weight-regular" style="float:right">{{ decision_advice.advice.0.advice.created_at|parse_date|date:"d F Y" }}</span>
```
saruniitr, line 16: The date is to be right aligned but in the screenshot attached to the PR it is not as per the design.
We have a separate ticket for displaying the ticket, perhaps we can remove the date here and fix it other ticket as the styling is not correct?

wkeeling: Thanks. Yes floating the date to the right is proving troublesome. Will remove it and we can fix in the other ticket.

caseworker/advice/views.py
```diff
@@ -242,59 +242,77 @@ def grouped_advice(self):
...
+            "name": destination["name"],
             "address": destination["address"],
             "licence_condition": advice_item["proviso"],
             "country": destination["country"]["name"],
+            "denial_reasons": advice_item["denial_reasons"],
             "advice": advice_item,
         }
 
-    def group_user_decision_advice(self, user_advice, team_user, decision):
+    def group_user_advice_by_decision(self, user_advice, decision):
```
saruniitr, line 256: This function and the above, are they only expected to work with user level advice or required for team level advice?
If it needs to be generic may be send level as argument and filter by level within the function?

caseworker/advice/constants.py
```diff
@@ -1,7 +1,13 @@
...
-    "Proviso": "approved with proviso",
+    "Proviso": "approved",
     "Refuse": "refused",
     "Conflicting": "given conflicting advice",
     "No Licence Required": "no licence required",
 }
+
+
+TEAM_DECISION_APPROVED = "has approved"
+TEAM_DECISION_APPROVED_REFUSED = "has approved and refused"
```
saruniitr, line 11: does this mean there are approve and refuse advice objects?
`has approved and refused` seems to indicate the same advice instance is approved and refused.

caseworker/advice/templates/advice/other-recommendations.html
```diff
@@ -0,0 +1,117 @@
...
+                    </thead>
+                    <tbody class="govuk-table__body">
+                        {% for advice in decision_advice.advice %}
+                        <tr class="govuk-table__row">
+                            <td class="govuk-table__cell">{{ advice.country }}</td>
+                            <td class="govuk-table__cell">
+                            {% if advice.advice.consignee %}
+                                Consignee
+                            {% elif advice.advice.end_user %}
+                                End user
```
saruniitr, line 41: one of the content change is to show End user as `End-user`

caseworker/advice/templates/advice/other-recommendations.html
```diff
@@ -0,0 +1,117 @@
...
+                        <div class="govuk-grid-row">
+                            <div class="govuk-grid-column-full">
+                                <h3 class="govuk-heading-s">Licence condition</h3>
+                            </div>
+                            <div class="govuk-grid-column-one-third">
+                                <p class="govuk-body govuk-!-margin-0">
+                                {% if advice.consignee %}
+                                    Consignee
+                                {% elif advice.end_user %}
+                                    End user
```
saruniitr, line 94: same as above

saruniitr (commented):
few minor comments

caseworker/advice/constants.py
wkeeling: This allows for this:

image

caseworker/advice/views.py
wkeeling: Thanks - have added an optional `level` argument.

saruniitr (approved):

alixedi (approved):

PR #356, opened 2021-12-22
[LTD-1586] Countersign move

alixedi:
Small one. It does 2 things on the `/countersign` screen -

1) It only shows the `Edit recommendation` button if you wrote the recommendation. This is a bit gnarly because of multiple advice objects etc. I have tried to get around this by using sets.

2) It puts the `move-case-forward` machinery in here. I have no idea how or why it was missing up until now.

### DIFF
--- a/caseworker/advice/views.py
+++ b/caseworker/advice/views.py
@@ -279,12 +279,22 @@ class ReviewCountersignView(LoginRequiredMixin, CaseContextMixin, TemplateView):
 class ViewCountersignedAdvice(LoginRequiredMixin, CaseContextMixin, TemplateView):
     template_name = "advice/view_countersign.html"
 
+    def can_edit(self, advice_to_countersign):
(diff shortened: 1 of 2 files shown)


wkeeling (approved):
Looks fine to me

caseworker/advice/views.py
```diff
@@ -276,15 +276,25 @@ def get_success_url(self):
...
 
 
-class ViewCountersignedAdvice(LoginRequiredMixin, CaseContextMixin, TemplateView):
+class ViewCountersignedAdvice(AdviceDetailView):
     template_name = "advice/view_countersign.html"
 
-    def get_context(self, **kwargs):
-        context = super().get_context()
+    def can_edit(self, advice_to_countersign):
+        """Determine of the current user can edit the countersign comments.
```
saruniitr, line 283: small correction `Determine if`

saruniitr (approved):

PR #357, opened 2021-12-23
LTD-1595: Post advice for NLR items also

saruniitr:
## Change description

If the licence is being issued for an application and it contains
NLR items then we have to generate NLR letter besides the main licence
document. The NLR letter is only generated if there is a final advice
for NLR items also.

To keep it consistent with Advice1.0, post NLR advice for other
advice levels as well.

### DIFF
--- a/caseworker/advice/services.py
+++ b/caseworker/advice/services.py
@@ -159,7 +159,20 @@ def post_approval_advice(request, case, data, level="user-advice"):
         }
         for subject_name, subject_id in get_advice_subjects(case, data.get("countries"))
     ]
-    response = client.post(request, f"/cases/{case['id']}/{level}/", json)
+    json_nlr_products = [
+        {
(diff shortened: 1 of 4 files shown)


caseworker/advice/services.py
```diff
@@ -159,7 +159,20 @@ def post_approval_advice(request, case, data, level="user-advice"):
...
+            "proviso": "",
+            "note": "",
+            "footnote_required": False,
+            "footnote": "",
+            "good": good["id"],
+            "denial_reasons": [],
+        }
+        for good in filter_nlr_products(case["data"]["goods"])
+    ]
+    response = client.post(request, f"/cases/{case['id']}/{level}/", json + json_nlr_products)
```
alixedi, line 175: Do we need to do this for refusal @saruniitr ?

alixedi (approved):

PR #359, opened 2021-12-23
Add a method to populate countries for FCO.

alixedi:
Small but gnarly. I have left comments for my 6m-later-self.

### DIFF
--- a/caseworker/advice/views.py
+++ b/caseworker/advice/views.py
@@ -4,7 +4,7 @@ from django.urls import reverse
 from django.utils.functional import cached_property
 from requests.exceptions import HTTPError
 
-from caseworker.advice import forms, services
+from caseworker.advice import forms, services, constants
 from core import client
 from caseworker.cases.services import get_case
 from caseworker.core.services import get_denial_reasons
@@ -37,7 +37,7 @@ class CaseContextMixin:
 
     def unadvised_countries(self):
         """Returns a dict of countries for which advice has not been given by the current user's team."""
-        dest_types = ("end_user", "ultimate_end_user", "consignee", "third_party")
+        dest_types = constants.DESTINATION_TYPES
 
         advised_on = {
             # Map of destinations advised on -> team that gave the advice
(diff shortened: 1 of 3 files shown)


caseworker/advice/views.py
```diff
@@ -187,8 +187,32 @@ def get_form(self):
         else:
             raise ValueError("Invalid advice type encountered")
 
+    def advised_countries(self):
+        """Returns a dict of countries for which advice has given by the current user."""
+        dest_types = ("end_user", "ultimate_end_user", "consignee", "third_party")
```
saruniitr, line 192: please rearrange in logical order `consignee, end_user, ultimate_end_user, third_party`.
Not sure about the last two which comes first.

caseworker/advice/views.py
```diff
@@ -187,8 +187,32 @@ def get_form(self):
...
+            if dest["id"] in advised_on
+        ]
+
     def form_valid(self, form):
         data = form.cleaned_data
+        # When an FCO officer edits the advice, we don't allow for changing the countries
+        # & therefore, we render the normal forms and not the FCO ones.
+        # This means that here data here doesn't include the list of countries for which
+        # the advice should be applied and so we pop that in using a method.
+        if self.caseworker["team"]["name"] == "FCO":
```
saruniitr, line 214: I think team id is better here

caseworker/advice/views.py
```diff
@@ -187,8 +187,32 @@ def get_form(self):
         else:
             raise ValueError("Invalid advice type encountered")
 
+    def advised_countries(self):
+        """Returns a dict of countries for which advice has given by the current user."""
```
wkeeling, line 191: Looks like the method returns a `list` rather than a `dict`?

caseworker/advice/views.py
wkeeling: Maybe it's worth moving this tuple of destination types to `constants.py` - I think `CaseContextMixin.unadvised_countries()` also uses it.

caseworker/advice/views.py
alixedi: Is `team-id` stable across envs?

saruniitr: not entirely but better than name. We need to update the name soon for this.

saruniitr (approved):

PR #361, opened 2021-12-23
LTD-1587: During consolidation/combining see advice from appropriate teams

saruniitr:
## Change description

When MoD consolidates they should see advice given by its sub-teams and
similarly when LU combines, they should only see from FCDO and MOD teams.


caseworker/advice/services.py
```diff
@@ -98,14 +116,20 @@ def get_advice_to_countersign(advice, caseworker):
...
     """
     user_advice = filter_advice_by_level(advice, ["user"])
-    grouped_user_advice = group_advice_by_user(user_advice)
-    return grouped_user_advice
+    if caseworker["team"]["id"] == LICENSING_UNIT_TEAM:
+        advice_from_teams = filter_advice_by_teams(user_advice, LU_CONSOLIDATE_TEAMS)
+    elif caseworker["team"]["id"] == MOD_ECJU_TEAM:
+        advice_from_teams = filter_advice_by_teams(user_advice, MOD_CONSOLIDATE_TEAMS)
+    else:
+        advice_from_teams = user_advice
```
saruniitr, line 130: should we raise an exception here as only the above teams perform this action?

wkeeling: I think raising an exception would make sense, since displaying all the unfiltered advice could lead to some other problem?

caseworker/advice/services.py
```diff
@@ -6,9 +6,18 @@
...
+FCDO_TEAM = "67b9a4a3-6f3d-4511-8a19-23ccff221a74"
 LICENSING_UNIT_TEAM = "58e77e47-42c8-499f-a58d-94f94541f8c6"
 MOD_ECJU_TEAM = "b7640925-2577-4c24-8081-b85bd635b62a"
-
+MOD_CONSOLIDATE_TEAMS = [
+    "2e5fab3c-4599-432e-9540-74ccfafb18ee",
+    "4c62ce4a-18f8-4ada-8d18-4b53a565250f",
+    "809eba0f-f197-4f0f-949b-9af309a844fb",
+    "a06aec31-47d7-443b-860d-66ab0c6d7cfd",
+]
```
wkeeling, line 19: Is it worth maybe putting a comment next to these IDs to indicate which teams they are? Alternatively make them constants themselves, and then reference those constants in the list (like what `LU_CONSOLIDATE_TEAMS` does).

caseworker/advice/services.py
```diff
@@ -52,6 +61,14 @@ def filter_advice_by_team(all_advice, team_id):
     return [advice for advice in all_advice if advice["user"]["team"]["id"] == team_id]
 
 
+def filter_advice_by_teams(all_advice, teams_list):
```
wkeeling, line 64: We should probably introduce some consistency into `services.py`. We've currently got:
*  singular named functions that filter on singular arguments (e.g. `team_id` in `filter_advice_by_team`)
*  singular named functions that filter on list based arguments (e.g. `advice_types` in `filter_advice_by_type`)
*  plural named functions that filter on list based arguments (e.g. `teams_list` in `filter_advice_by_teams`)

Probably we should fix (2) as that's most likely to lead to programmer error. Maybe one for a different PR though.

caseworker/advice/services.py
```diff
@@ -98,14 +115,20 @@ def get_advice_to_countersign(advice, caseworker):
     return grouped_user_advice
 
 
-def get_advice_to_consolidate(advice):
+def get_advice_to_consolidate(advice, caseworker):
```
alixedi, line 118: Might want to pass in `team_id` instead of `caseworker` because we only need the former?

alixedi, line 118: Might want to pass in `team_id` instead of `caseworker` because we only need the former?

caseworker/advice/services.py
saruniitr: will do in a different PR

caseworker/advice/services.py
saruniitr: :+1:

caseworker/advice/services.py
saruniitr: ok

alixedi (approved):

PR #371, opened 2022-01-04
[LTD-1613] Stop FCO advice when all destinations have been advised upon

alixedi:


unit_tests/caseworker/advice/views/test_view_ogd_advice.py
```diff
@@ -1,10 +1,17 @@
...
 import pytest
 from bs4 import BeautifulSoup
 from django.urls import reverse
 
 from core import client
 
 
+@pytest.fixture(autouse=True)
+def setup(mock_queue, mock_case):
+    yield
```
alixedi, line 12: We should do this in all other tests. Gets rid of the horrible `requests_mock` code everywhere.

caseworker/advice/views.py
```diff
@@ -253,9 +253,16 @@ def teams(self):
             key=lambda a: a["name"],
         )
 
+    def can_advice(self):
```
wkeeling, line 256: Maybe use the verb, as in `can_advise`?

wkeeling (approved):

PR #372, opened 2022-01-05
LTD-1630: Show countersigning warning message only for LU

saruniitr:
## Change description

Because this message is only applicable if the case contains flags
relevant for LU team.


alixedi (commented):
Looks straight forward but we should add a unit test for this.

saruniitr:
> Looks straight forward but we should add a unit test for this.

ok, will add one

unit_tests/caseworker/advice/views/test_consolidate.py
```diff
@@ -472,3 +473,69 @@ def test_consolidate_raises_exception_for_other_team(
...
+            "priority": 0,
+        },
+    ]
+
+    case_id = data_standard_case["case"]["id"]
+    requests_mock.get(client._build_absolute_uri(f"/cases/{case_id}"), json=data_standard_case)
+    requests_mock.get(
+        client._build_absolute_uri("/gov-users/2a43805b-c082-47e7-9188-c8b3e1a83cb0"),
+        json={"user": {"id": "2a43805b-c082-47e7-9188-c8b3e1a83cb0", "team": {"id": team_id, "name": team_name},}},
+    )
```
alixedi, line 531: One more little thing that I will leave with you. This is as good an opportunity as any to use the [setup trick](https://github.com/uktrade/lite-frontend/blob/master/unit_tests/caseworker/advice/views/test_view_ogd_advice.py#L10-L12) and get rid of the `requests_mock` bits - not only in this test but all the tests in this module.

alixedi (approved):
Leaving one small comment with you 🙏

PR #380, opened 2022-01-06
LTD-1633: Advice 2.0 tab - navigation and button visibility

wkeeling:
Also addresses [LTD-1637](https://uktrade.atlassian.net/browse/LTD-1637) and [LTD-1658](https://uktrade.atlassian.net/browse/LTD-1658)


caseworker/advice/services.py
```diff
@@ -4,7 +4,17 @@
 
 from core import client
 
+# Queues
+FCDO_CASES_TO_REVIEW_QUEUE = "f458094c-1fed-4222-ac70-ff5fa20ff649"
 FCDO_COUNTERSIGNING_QUEUE = "5e772575-9ae4-4a16-b55b-7e1476d810c4"
+MOD_CASES_TO_REVIEW_QUEUE = "a4c5bd9d-0d06-4856-abe1-c71d73abe636"
```
saruniitr, line 10: I think we also need LU countersigning queue

wkeeling: I'm not sure that's mentioned in the spreadsheet: https://docs.google.com/spreadsheets/d/1GDxPPHblzqXaPh1zUOav8qIlA6qDIeCe1zvA6PEddHE/edit#gid=0
I can follow up with Jess on that though.

caseworker/core/objects.py
```diff
@@ -1,9 +1,10 @@
 class Tab:
-    def __init__(self, id, name, url, count=0):
+    def __init__(self, id, name, url, count=0, has_content=True):
         self.id = "tab-" + id
         self.name = name
         self.url = url
         self.count = count
+        self.has_content = has_content
```
wkeeling, line 7: This might be better named `has_template`

caseworker/advice/services.py
```diff
@@ -252,3 +275,67 @@ def move_case_forward(request, case_id, queue_id):
     except HTTPError as e:
         raise HTTPError(response=response) from e
     return response.json()
+
+
+def get_advice_tab_context(case, caseworker, queue_id):
```
saruniitr, line 280: except for unit tests I think this looks good

caseworker/advice/services.py
alixedi: P.S. this is why I was talking about queues being immutable. There is also an interesting problem to be solved with team IDs that we are referring to in code. They are not stable across envs, which means everything we are building will only ever work on UAT (& hopefully PROD).

wkeeling: Agreed.

unit_tests/caseworker/advice/test_services.py
```diff
@@ -0,0 +1,94 @@
...
+        advice[0]["level"] = advice_level
+        if countersigned:
+            advice[0]["countersigned_by"] = current_user
+        data_standard_case["case"]["advice"] = advice
+    current_user["team"]["id"] = team
+
+    context = get_advice_tab_context(Case(data_standard_case["case"]), current_user, queue)
+
+    assert context["url"] == url
+    for k, v in context["buttons"].items():
```
saruniitr, line 93: instead of `k, v` could you rename them as `button_name`, `active` or similar?

saruniitr (approved):

PR #393, opened 2022-01-11
Upgrade to Django 3.2.x LTS

kevincarrogan:


unit_tests/core/test_middleware.py
```diff
@@ -134,4 +134,4 @@ def test_x_robots_tag_middleware(rf):
     # We should get a 200 and the token should be cached
     response = instance(request)
     assert response.status_code == status.HTTP_200_OK
-    assert response._headers.get("x-robots-tag", "") == ("X-Robots-Tag", "noindex,nofollow")
+    assert response.headers["x-robots-tag"] == "noindex,nofollow"
```
saruniitr, line 137: `x-robots-tag` always expected to be present in the headers?

saruniitr (commented):
requesting clarification

unit_tests/core/test_middleware.py
kevincarrogan: Yes. The middleware is adding this parameter so if it's not there then it should be a failure. I felt the `.get` was less explicit as to what exactly we want to be testing. We want to know whether the parameter is there (may result in an IndexError) and if it's the right thing (which is what the assertion is doing). The .get will hide the possible IndexError case.

saruniitr: :+1:

saruniitr: :+1:

saruniitr (approved):

PR #398, opened 2022-01-14
 LTD-1675-WMB-MED-contect-Change 

depsiatwal:
Display /advice/view-my-advice/ on
Screenshot 2022-01-14 at 11 42 14


unit_tests/caseworker/advice/views/test_consolidate.py
```diff
@@ -550,6 +550,7 @@ def test_view_consolidate_approve_outcome(
 def test_view_consolidate_refuse_outcome(
     requests_mock, authorized_client, data_standard_case, view_consolidate_outcome_url, consolidated_refusal_outcome
 ):
+    data_standard_case["denial_reasons_display"] = {}
```
saruniitr, line 553: not sure this is required

saruniitr (commented):
few comments

depsiatwal:
image
group template in action

unit_tests/caseworker/advice/views/test_consolidate.py
depsiatwal: removed

caseworker/advice/views.py
```diff
@@ -65,6 +69,7 @@ def get_context_data(self, **kwargs):
         # doesn't have anything to do with e.g. lite-forms
         # P.S. the case here is needed for rendering the base
         # template (layouts/case.html) from which we are inheriting.
+
         return {
             **context,
             **self.get_context(case=self.case),
```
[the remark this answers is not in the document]
depsiatwal: good point this convert_to_display_dict was done before I shifted to property 
Done

caseworker/advice/templatetags/advice_tags.py
```diff
@@ -67,6 +67,17 @@ def group_advice(case):
     return {"grouped_advice": grouped_advice}
 
 
+@register.filter
+def get_denial_reason_display_values(denial_reasons, denial_reasons_display):
+    denial_reasons = (
+        denial_reasons[0] if len(denial_reasons) == 1 and isinstance(denial_reasons[0], list) else (denial_reasons)
```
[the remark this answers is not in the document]
depsiatwal: as discussed decide not to shift to inclusion_tag

regarding denial reason being a list this is because in the tests denial_reasons  are contained in a list i.e 
[[reason_a, reason_b]] I wasn't sure is this reflected backend but when testing on UI it's a flattened list. Hence the code. We can fix the tests of the structure is incorrect

caseworker/advice/templates/advice/advice_details.html
```diff
@@ -71,7 +72,7 @@ <h2 class="govuk-heading-m app-bg-colour--red govuk-!-padding-2">
                 <td class="govuk-table__cell">{{ ultimate_end_user.name }}</td>
                 <td class="govuk-table__cell">All</td>
                 {% if decision == 'refuse' %}
-                <td class="govuk-table__cell">{{ item.denial_reasons|join:", "}}</td>
+                <td class="govuk-table__cell">{{ item.denial_reasons|get_denial_reason_display_values:denial_reasons_display}}</td>
```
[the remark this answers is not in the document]
depsiatwal: Done and sifted to context so it's available in the template

caseworker/advice/templatetags/advice_tags.py
```diff
@@ -70,8 +70,22 @@ def group_advice(case):
...
-    return {"grouped_advice": grouped_advice}
+@register.filter
+def get_denial_reason_display_values(denial_reasons, denial_reasons_display):
+    if denial_reasons and denial_reasons_display:
+        denial_reasons = (
+            denial_reasons[0] if len(denial_reasons) == 1 and isinstance(denial_reasons[0], list) else (denial_reasons)
+        )
+        display_values = []
+        for item in denial_reasons:
+            display_values.append(denial_reasons_display.get(item, item))
```
wkeeling, line 85: Minor, but could reduce to a single line with a list comp?
```python
display_values = [denial_reasons_display.get(item, item) for item in denial_reasons]
```

depsiatwal: ahh nice one done

wkeeling (approved):
Looks good

caseworker/advice/templatetags/advice_tags.py
```diff
@@ -70,8 +70,20 @@ def group_advice(case):
...
+@register.filter
+def get_denial_reason_display_values(denial_reasons, denial_reasons_display):
+    if denial_reasons and denial_reasons_display:
+        denial_reasons = (
+            denial_reasons[0] if len(denial_reasons) == 1 and isinstance(denial_reasons[0], list) else (denial_reasons)
+        )
+        display_values = [denial_reasons_display.get(item, item) for item in denial_reasons]
+        return ", ".join(display_values)
+    else:
+        return ""
```
saruniitr, line 86: I think `denial_reasons` is always a list with the criteria or an empty list
If for any view that is not the case then we need to fix that.
so the above can just be replaced with,
```
return ", ".join([denial_reasons_display[item] for item in denial_reasons])
```

caseworker/advice/templatetags/advice_tags.py
```diff
@@ -70,8 +70,16 @@ def group_advice(case):
...
+    return context
+
 
-    return {"grouped_advice": grouped_advice}
+@register.filter
+def get_denial_reason_display_values(denial_reasons, denial_reasons_display):
+    if denial_reasons and denial_reasons_display:
+        return ", ".join([denial_reasons_display.get(item, item) for item in denial_reasons])
+    else:
+        return ""
```
saruniitr, line 82: sorry the `else` is not required. Below takes care of empty list also,
```
return ", ".join([denial_reasons_display[item] for item in denial_reasons])
```

Also instead of using `.get` it is intentional to access it by key because `denial_reasons` are seeded and we come across something that is not there then it should fail and we can investigate further instead of taking the default value.

saruniitr (approved):

PR #402, opened 2022-01-17
Ltd 1690 consolidation missing form title

kevincarrogan:


caseworker/advice/views.py
```diff
@@ -344,15 +344,27 @@ def is_advice_approve_only(self):
         approve_advice_types = ("approve", "proviso", "no_licence_required")
         return all([a["type"]["key"] in approve_advice_types for a in self.case.advice])
 
+    def get_team_name(self):
+        user_team_id = self.caseworker["team"]["id"]
```
saruniitr, line 348: There is team `name` in `self.caseworker['team']` so it can used instead of the map below

kevincarrogan: The mapping was due to the fact that the original requirement was that it _wasn't_ the team name that's passed through in regards to `MOD ECJU` so this mapping was required.

Now that the requirement has changed I can change this piece of code.

saruniitr (approved):

PR #409, opened 2022-01-21
Add implicit waits on flaky steps and remove sleeps

PippoRaimondi:
The intent of this PR is make steps which have been flaky in CI a bit more robust by adding implicit waits prior to interacting with them.

By adding implicit waits we also removed any hardcoded sleep within the steps.


saruniitr (approved):
looks good to me

PR #412, opened 2022-01-24
LTD-1607-fix-backlink

depsiatwal:
fix back button to approve/refuse advice 

image

### DIFF
--- a/caseworker/advice/views.py
+++ b/caseworker/advice/views.py
@@ -228,6 +228,11 @@ class EditAdviceView(LoginRequiredMixin, CaseContextMixin, FormView):
     def get_success_url(self):
         return reverse("cases:view_my_advice", kwargs=self.kwargs)
 
+    def get_context_data(self, **kwargs):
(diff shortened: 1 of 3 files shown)


sekharpanja (approved):

caseworker/advice/views.py
```diff
@@ -119,7 +119,7 @@ def form_valid(self, form):
         return super().form_valid(form)
 
     def get_success_url(self):
-        return reverse("cases:view_my_advice", kwargs=self.kwargs)
+        return reverse("cases:select_advice", kwargs=self.kwargs)
```
saruniitr, line 122: the back link to be changed is in the template, previous value of `cases:view_my_advice` is correct here and below as well.

saruniitr (changes requested):
Back link to be corrected is in the template

depsiatwal:
This flag fixes the templates to also show when we in edit mode 

image

since this flag was already there I assume it's some how disappeared from the context

saruniitr (approved):

PR #415, opened 2022-01-26
LTD-1733: Skip inactive refusal criteria items

saruniitr:
## Change description

In the updated refusal criteria some are deprecated so skip them so that
user cannot select these items.
Previous refusal advice may be referring to these deprecated items so only
skip in the form but display them as is when viewing the advice.

### DIFF
--- a/caseworker/advice/forms.py
+++ b/caseworker/advice/forms.py
@@ -134,6 +134,9 @@ class RefusalAdviceForm(forms.Form):
     def _group_denial_reasons(self, denial_reasons):
         grouped = defaultdict(list)
         for item in denial_reasons:
+            # skip the ones that are not active anymore
+            if item["deprecated"]:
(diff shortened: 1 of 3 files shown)


caseworker/advice/forms.py
```diff
@@ -134,6 +134,9 @@ class RefusalAdviceForm(forms.Form):
     def _group_denial_reasons(self, denial_reasons):
         grouped = defaultdict(list)
         for item in denial_reasons:
+            # skip the ones that are not active anymore
```
alixedi, line 137: worth double-checking if the API returns deprecated `denial_reasons`.

saruniitr: yeah, checked that API is returning all fields in the serializer

alixedi (approved):

PR #418, opened 2022-02-01
[LTD-663] I want to choose which cookie types are turned-on

alixedi:
This PR does a bunch of things so I would recommend a commit-wise review -

- [x] Adds the `cookies` app from [here](https://github.com/uktrade/export-support/tree/master/export_support/cookies) (ty @kevincarrogan). I have made minor modifications to fit this in `lite-frontend` e.g. using bare `<script>` instead of `webpack`.
- [x] Replaces the current GA/GTM code with the template tags that come out of the box with `cookies`.
- [x] The `cookies` app has a nice naming scheme so I chose to rename the current `GOOGLE_ANALYTICS_KEY`.
- [x] Add the `js` and `scss` assets that are required to make this work. 
- [x] Modify the content in the banner and the cookies page to match our requirements.


wkeeling:
Looks fine as far as I can tell.

PR #420, opened 2022-02-01
 LTD-1664-alias-teams-queues 

depsiatwal:
This removed UUID and depends on alias flags set in the DB.

### DIFF
--- a/caseworker/advice/views.py
+++ b/caseworker/advice/views.py
@@ -109,7 +109,7 @@ class GiveApprovalAdviceView(LoginRequiredMixin, CaseContextMixin, FormView):
     template_name = "advice/give-approval-advice.html"
 
     def get_form(self):
-        if self.caseworker["team"]["id"] == services.FCDO_TEAM:
+        if self.caseworker["team"]["alias"] == services.FCDO_TEAM:
             return forms.FCDOApprovalAdviceForm(self.unadvised_countries(), **self.get_form_kwargs())
         else:
             return forms.GiveApprovalAdviceForm(**self.get_form_kwargs())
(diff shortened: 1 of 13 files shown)


caseworker/advice/views.py
```diff
@@ -374,14 +373,17 @@ def get_form(self):
 
     def get_context(self, **kwargs):
         context = super().get_context()
-        advice_to_consolidate = services.get_advice_to_consolidate(self.case.advice, self.caseworker["team"]["id"])
+        team_alias = (
+            self.caseworker["team"]["alias"] if self.caseworker["team"]["alias"] else self.caseworker["team"]["id"]
```
alixedi, line 377: Might be easier to do `self.caseworker["team"].get("alias", "")` if the aim if to fail the equality test for alias down the stack.

caseworker/advice/views.py
```diff
@@ -455,19 +458,19 @@ class ViewConsolidatedAdviceView(AdviceView, FormView):
...
-            consolidated_advice = services.get_consolidated_advice(self.case.advice, user_team_id)
+        if user_team_alias in [services.LICENSING_UNIT_TEAM, services.MOD_ECJU_TEAM]:
+            consolidated_advice = services.get_consolidated_advice(self.case.advice, user_team_alias)
         nlr_products = services.filter_nlr_products(self.case["data"]["goods"])
 
         lu_countersign_flags = {services.LU_COUNTERSIGN_REQUIRED, services.LU_SR_MGR_CHECK_REQUIRED}
-        case_flag_ids = {flag["id"] for flag in self.case.all_flags}
-        lu_countersign_required = user_team_id == services.LICENSING_UNIT_TEAM and bool(
-            lu_countersign_flags.intersection(case_flag_ids)
+        case_flag_aliases = {flag["alias"] for flag in self.case.all_flags}
```
alixedi, line 468: I don't think the `id` here is the one we are replacing with `alias`.

alixedi (commented):
Had a quick look and left some comments.

caseworker/advice/views.py
depsiatwal: my thinking was fail on key because something unexpected happened. I'm conscious we might fall into an else statement where something unexpected happens.

caseworker/advice/views.py
depsiatwal: since in BE we using alias we check intersection on alias names.

alixedi: Oh! I get it now. We have moved flags to use alias as well haven't we?

alixedi (approved):
Awesome stuff, ty @depsiatwal

wkeeling (approved):

PR #421, opened 2022-02-02
LTD-1745: Port AddGood LITE forms to Django forms SessionWizardView

wkeeling:
- [ ] Port 17 forms and check validation behaviour
- [ ] Implement final data submission
- [ ] Remove redundant API validation
- [ ] Port unit tests


exporter/applications/urls.py
```diff
@@ -61,6 +61,7 @@
     # Goods
     path("<uuid:pk>/goods/", goods.ApplicationGoodsList.as_view(), name="goods"),
     path("<uuid:pk>/goods/add-new/", goods.AddGood.as_view(), name="new_good"),
+    path("<uuid:pk>/goods/add-new2/", goods.AddGood2.as_view(), name="new_good2"),
```
wkeeling, line 64: Temporary URL mapping

exporter/goods/forms.py
```diff
@@ -895,3 +901,728 @@ def has_valid_section_five_certificate(application):
...
+        except ValueError:
+            year_of_manufacture = 0
+
+        if year_of_manufacture <= 0:
+            self.add_error("year_of_manufacture", "Year of manufacture must be valid")
+
+        if year_of_manufacture > timezone.now().date().year:
+            self.add_error("year_of_manufacture", "Year of manufacture must be in the past")
+
+        return cleaned_data
```
kevincarrogan, line 1327: If the cleaning only pertains to a single field i.e. the validation isn't dependent on two or more fields, you can use the `clean_year_of_manufacture` method instead.

exporter/goods/forms.py
```diff
@@ -895,3 +901,728 @@ def has_valid_section_five_certificate(application):
...
+        label="Expiry date",
+        help_text="For example 12 3 2022",
+    )
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        # The date field is styled correctly when the input type is 'number'.
+        # This workaround switches the underlying widgets to 'number' type.
+        convert_to_number_input(self.fields['expiry_date'])
```
kevincarrogan, line 1451: GDS has [specific guidance on entering dates](https://design-system.service.gov.uk/patterns/dates/) and more general guidance around [_not_ using `number`](https://design-system.service.gov.uk/components/text-input/#avoid-using-inputs-with-a-type-of-number) as an input type.

core/file_handler.py
```diff
@@ -23,3 +25,13 @@ def receive_data_chunk(self, raw_data, start):
...
+
+    def file_complete(self, *args, **kwargs):
+        super().file_complete(*args, **kwargs)
+
+        self.file = S3Boto3StorageFile(self.s3_key, 'r', self.storage)
+        self.file.original_name = self.file_name
+        self.file.content_type = self.content_type
+        self.file.charset = None
+
+        return self.file
```
kevincarrogan, line 37: This probably needs some comments to explain why I did this in the first place.

I'm also not completely sure that `charset = None` is the right thing to do, it needs to be set to something though but not sure if `None` is the right thing.

wkeeling: Thanks, yes comment added. I've also propagated the charset rather than explicitly setting it to `None` which seems to work fine, e.g.
```python
self.file.charset = self.charset
```

exporter/goods/forms.py
wkeeling: I've created a subtask to address this: https://uktrade.atlassian.net/browse/LTD-1875.

saruniitr (approved):
looks good to me, thank you @kevincarrogan and @wkeeling

PR #429, opened 2022-02-14
Ltd 1745 port to django forms ltd 1791

kevincarrogan:


exporter/applications/views/goods.py
```diff
@@ -971,6 +982,270 @@ def on_submission(self, request, **kwargs):
...
+            ).get("number_of_items", 0)
+
+        if step in (
+            AddGoodToApplicationFormSteps.FIREARM_UNIT_QUANTITY_VALUE,
+            AddGoodToApplicationFormSteps.COMPONENT_OF_A_FIREARM_UNIT_QUANTITY_VALUE,
+            AddGoodToApplicationFormSteps.COMPONENT_OF_A_FIREARM_AMMUNITION_UNIT_QUANTITY_VALUE,
+            AddGoodToApplicationFormSteps.UNIT_QUANTITY_VALUE,
+        ):
+            good, _ = get_good(self.request, self.kwargs["good_pk"])
+            kwargs["good"] = good
```
wkeeling, line 1133: Would it make sense to pass the (potentially) cached `self.good` property here?

kevincarrogan: Absolutely.

PR #431, opened 2022-02-15
 ltd-1814-pre-commit-pii-hook-2

depsiatwal:


.pre-commit-config.yaml
```diff
@@ -0,0 +1,45 @@
+exclude: ^exporter/assets/built/
```
wkeeling, line 1: Is there a reason why we don't exclude `caseworker/assets/built/` too?

depsiatwal: ahh this was the one throwing the error in pii checker hence it was added i've added both now anyway

depsiatwal:
@wkeeling  and @kevincarrogan  I agree with both your comments regarding excluding
example.caseworker.env
example.exporter.env
But the file is breaking as pii checker doesn't seem to allow this file type. Even after adding /PS_IGNORE. So we would have to exclude .env files from the repo or upgrade p2 checker. 

Another note is that this may be fixed in the latest version. But the p2 checker is breaking when I upgrade to latter version  i'll flag this with Ross Miller (owner of pii)

wkeeling (approved):

PR #436, opened 2022-02-22
Move linting checks into pre-commit

kevincarrogan:


makefile
```diff
@@ -10,15 +10,6 @@ clean:
 	-find . -type f -name "*.pyc" -delete
 	-find . -type d -name "__pycache__" -delete
 
-lint:
-	-pipenv run bandit -r . --skip=B101 --exclude=/ui_tests,/unit_tests,/tests_common
-	-pipenv run prospector exporter
-	-pipenv run prospector caseworker
-	-pipenv run black .
```
PippoRaimondi, line 17: are we not running black anymore as part of linting? not sure I saw it in the pre commit hook...

also, sometimes it's nice to have the ability to run linting before attempting to commit... so keeping the makefile could help with that... but that could be me only so nvm if that's the case...

kevincarrogan: Black was already in the `pre-commit` hook before this change.

I couldn't decide whether it was as useful having it in the makefile due to it happening automatically on commit but I can add it back.

PippoRaimondi (approved):

PR #443, opened 2022-02-24
LTD-1873: Update Advice approval form label when consolidating for MoD and LU

saruniitr:


caseworker/advice/views.py
```diff
@@ -370,7 +370,12 @@ def get_form(self):
             return forms.RefusalAdviceForm(denial_reasons=denial_reasons, **form_kwargs)
 
         if self.kwargs.get("advice_type") == "approve" or self.is_advice_approve_only():
-            return forms.ConsolidateApprovalForm(**form_kwargs)
+            team_alias = (
+                self.caseworker["team"]["alias"]
+                if self.caseworker["team"]["alias"]
+                else self.caseworker["team"]["name"]
+            )
```
alixedi, line 377: How about - 

```
team_alias = self.caseworker["team"].get("alias", None)
```

caseworker/advice/forms.py
```diff
@@ -123,8 +124,13 @@ def __init__(self, *args, **kwargs):
...
     """Approval form minus some fields."""
 
-    def __init__(self, *args, **kwargs):
+    def __init__(self, team_alias, *args, **kwargs):
         super().__init__(*args, **kwargs)
+        if team_alias == services.MOD_ECJU_TEAM:
+            self.fields["approval_reasons"].label = "Enter MOD’s overall reason for approval"
+        elif team_alias == services.LICENSING_UNIT_TEAM:
+            self.fields["approval_reasons"].label = "Enter Licensing Unit’s reason for approval"
+
```
alixedi, line 133: How about we define the following for the `class`:

```
ALIAS_LABELS = {services.MOD_ECJU_TEAM: "Enter MOD’s overall reason for approval", services.LICENSING_UNIT_TEAM: "Enter Licensing Unit’s reason for approval"}
```

And inside `__init__`:

```
self.fields["approval_reasons"].label = ALIAS_LABELS[team_alias]
```

caseworker/advice/views.py
saruniitr: :+1:

caseworker/advice/forms.py
saruniitr: With above we have to provide the default label (`What are your reasons for approving?`) if the team is not in `ALIAS_LABELS` which causes duplication as it is already set in `GiveApprovalAdviceForm` so ok to go with `if-else` ?

saruniitr: @alixedi could you take another look?

saruniitr: updated

alixedi (approved):

PR #445, opened 2022-02-25
Ltd 1358 serial number question updates

kevincarrogan:
This updates the question posed as to whether a product has serial numbers or not. Originally this was a boolean choice, whereas now the exporter can choose to indicate that their product has serial numbers but would like to provide them at a later date.

There are some additional changes to some of the affected HTML files to tidy up their indentation which was inconsistent and sometimes confusing, the [ first commit](https://github.com/uktrade/lite-frontend/pull/445/commits/ffff340ac2ae476c7360bc0923367a7d8bbf588a) has the bulk of the core changes without the HTML indentation changes.

### DIFF
--- a/lite_content/lite_exporter_frontend/goods.py
+++ b/lite_content/lite_exporter_frontend/goods.py
@@ -256,11 +256,10 @@ class CreateGoodForm:
             DONT_KNOW = "I don't know"
 
         class IdentificationMarkings:
-            TITLE = "Has the product been marked with a serial number or other type of identification marking?"
-            MARKINGS_DETAILS = "Serial number or other type of marking"
-            MARKINGS_HELP_TEXT = "Enter one or more"
+            TITLE = "Will each product have a serial number or other identification marking?"
             NO_MARKINGS_DETAILS = "Explain why the product has not been marked"
-            YES = "Yes"
+            YES_NOW = "Yes, I can add serial numbers now"
+            YES_LATER = "Yes, I can add serial numbers later"
             NO = "No"
 
             BUTTON_TEXT = "Continue"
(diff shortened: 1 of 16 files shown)


lite_content/lite_exporter_frontend/goods.py
```diff
@@ -256,11 +256,10 @@ class FirearmsActCertificate:
...
 
         class IdentificationMarkings:
-            TITLE = "Has the product been marked with a serial number or other type of identification marking?"
-            MARKINGS_DETAILS = "Serial number or other type of marking"
-            MARKINGS_HELP_TEXT = "Enter one or more"
+            TITLE = "Will each product have a serial number or other identification marking?"
             NO_MARKINGS_DETAILS = "Explain why the product has not been marked"
-            YES = "Yes"
+            YES_NOW = "Yes, I can add serial numbers now"
+            YES_LATER = "Yes, I can add serial numbers later"
```
wkeeling, line 262: I think we're considering the `lite_content` repo as deprecated these days - so we're not making any new changes in it. For new additions, we just make them directly in the relevant place in `lite-frontend`.

saruniitr: yes correct but the `lite_content` in frontend repo is not a submodule, it is just a directory (unlike in lite-api) so I think its fine in frontend repo.

unit_tests/caseworker/cases/test_views.py
```diff
@@ -389,8 +389,8 @@ def test_good_on_application_detail_clc_entries(
     assert response.status_code == 200
     soup = BeautifulSoup(response.content.decode("utf-8"), "html.parser")
     clc_entries = soup.find(id="control-list-entries-value").text
-    clc_entries = clc_entries.replace("\n", "").replace("\t", "")
-    assert clc_entries == "ML1, ML2"
+    clc_entries = "".join(line.strip() for line in clc_entries.split("\n"))
+    assert clc_entries == "ML1,ML2"
```
saruniitr, line 393: when there are multiple entries, design require us to include space also

kevincarrogan: This test isn't particularly trying to work out what the whitespace is but just that the values are correct as the whitespace from the HTML is stripped here. I don't think we need to worry about how exactly its represented in the HTML but that the values are correct.

saruniitr (approved):

PR #446, opened 2022-02-25
LTD-1704-gov-sigin-login 

depsiatwal:
New Gov.uk sign-in follow
completely replaces GREAT gov.uk signin. Please don't merge as we will need to be sure we are in the state to shift to GOV.UK


exporter/templates/core/register-name.html
```diff
@@ -0,0 +1,19 @@
...
+            <div class="govuk-grid-column-one-half">
+                <form method="post" enctype="multipart/form-data" novalidate="novalidate">
+                    {% csrf_token %}
+                    {{ form|crispy }}
+                    <button type="submit" class="govuk-button">Save</button>
+                </form>
+            </div>
+        </div>
+    </div>
+{% endblock %}
```
wkeeling, line 19: We've been reusing the generic template `exporter/templates/core/form.html` for new Django crispy forms. I wonder whether it's possible to use that here, rather than creating a new template?

depsiatwal: @wkeeling thanks for your help on this, It's now done :-)

exporter/core/views.py
```diff
@@ -198,6 +203,38 @@ def get(self, request, *args, **kwargs):
...
+    def get_success_url(self):
+        return resolve_url(settings.LOGIN_URL)
+
+    def dispatch(self, request, *args, **kwargs):
+        if self.request.session.get("first_name") and self.request.session.get("last_name"):
+            return HttpResponseRedirect(resolve_url(settings.LOGIN_URL))
+        return super().dispatch(request, *args, **kwargs)
+
+    def form_valid(self, form):
+        data = {**form.cleaned_data}
```
kevincarrogan, line 219: Is there any need to make a copy of the data here?

kevincarrogan, line 219: Is there any need to make a copy of the data here?

unit_tests/exporter/core/test_views.py
```diff
@@ -0,0 +1,42 @@
...
+    assert response.status_code == 302
+    assert response.url == settings.LOGIN_URL
+
+
+def test_register_name_redirects_name_known(authorized_client):
+    url = reverse("core:register_name")
+    response = authorized_client.get(url)
+
+    assert response.status_code == 302
+    assert response.url == settings.LOGIN_URL
```
kevincarrogan, line 42: It would be nice to know in this test exactly why this should pass. Is there anything we could add in to make it clear that it's passing because there is data somewhere (I think in session)?

Maybe just a couple of asserts at the beginning to show the data is there that the rest of this test expects.

exporter/core/forms.py
```diff
@@ -233,3 +237,25 @@ def register_an_individual_group(request, location):
             site_form(request, True, location),
         ]
     )
+
+
+class RegisterNameForm(forms.Form):
+    first_name = forms.CharField(
+        label="First name",
+        widget=forms.TextInput(),
```
kevincarrogan, line 245: I _think_ that the widget here is superfluous as this is what CharField defaults to anyway.

conf/exporter.py
```diff
@@ -34,12 +34,20 @@
...
+    AUTHBROKER_SCOPE = "openid,email"
+    AUTHBROKER_AUTHORIZATION_URL = urljoin(AUTHBROKER_URL, "authorize/")
+    AUTHBROKER_TOKEN_URL = urljoin(AUTHBROKER_URL, "token")
+    AUTHBROKER_PROFILE_URL = urljoin(AUTHBROKER_URL, "userinfo")
+else:
+    LOGOUT_URL = f"{AUTHBROKER_URL}/sso/accounts/logout/?next="
+    AUTHBROKER_SCOPE = "profile"
+    AUTHBROKER_AUTHORIZATION_URL = urljoin(AUTHBROKER_URL, "sso/oauth2/authorize/")
+    AUTHBROKER_TOKEN_URL = urljoin(AUTHBROKER_URL, "sso/oauth2/token/")
+    AUTHBROKER_PROFILE_URL = urljoin(AUTHBROKER_URL, "sso/oauth2/user-profile/v1/")
```
alixedi, line 50: Reckon this shouldn't be in the `else` as long as it is before the `if FEATURE_FLAG_GOVUK_SIGNIN_ENABLED`.

alixedi (commented):
Left some initial feedback. Will take another look later.

exporter/core/views.py
depsiatwal: ahh well spotted probably a handover from previous code removed

exporter/core/forms.py
depsiatwal: removed

conf/exporter.py
depsiatwal: done :-)

conf/base.py
```diff
@@ -204,6 +204,7 @@
 if additional_logger_config:
     LOGGING["loggers"] = additional_logger_config
 
+
```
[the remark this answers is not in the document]
depsiatwal: removed

kevincarrogan (approved):

PR #449, opened 2022-02-28
Update End-user journey

saruniitr:
## Change description

In the journey of adding End-user to the application we are asking for few more details and based on the choices selected the user may need to upload additional documents. For this new forms have been added.
Additional details that we are asking are,
- Whether there is any undertaking document available or not, if not available then the reason for not having it
- Whether the document is in English or not? if not then an English translation document need to be uploaded
- Whether atleast one page is on company letterhead? If yes then document on company letterhead need to be uploaded

The existing journey was using `lite-forms` and to make the implementation easier the existing journey is also replaced with django forms using SessionWizard.

A new summary page is added using which user can edit individual fields.
While editing user can change document options as well - eg initially he may have selected that the document is not in english and uploaded a translation document but while editing he may change to document in english. This is implemented in another wizard and depending on the updated options user is presented forms to upload document or delete previously uploaded documents.

All of these changes are only applicable to End-user. The other parties such as Consignee, ultimate end-user, third-parties continue to use other views hence they are not updated.


exporter/applications/views/parties/end_users.py
```diff
@@ -116,3 +130,37 @@ def __init__(self):
...
+
+    def get_success_url(self):
+        reuse_party = self.request.POST.get("reuse_party")
+        if reuse_party == "yes":
+            return reverse("applications:end_users_copy", kwargs=self.kwargs)
+        else:
+            return reverse("applications:add_end_user2", kwargs=self.kwargs)
+
+
+class AddEndUserForm(LoginRequiredMixin, SessionWizardView):
```
wkeeling, line 147: This is a view not a form, so better named as `AddEndUserView`

core/forms/templates/forms/autocomplete_field.html
```diff
@@ -6,8 +6,8 @@
   </optgroup>{% endif %}{% endfor %}
 </select>
 
-<script type="text/javascript" src="{% static '/javascripts/accessible-autocomplete.min.js' %}" nonce="{{ request.csp_nonce }}"></script>
-<script type="text/javascript" nonce="{{ request.csp_nonce }}">
+<script type="text/javascript" src="{% static '/javascripts/accessible-autocomplete.min.js' %}" nonce="{{ widget.attrs.nonce }}"></script>
```
saruniitr, line 9: one of the form has a input field to select country which is an autocomplete field

exporter/applications/forms/parties.py
```diff
@@ -129,3 +135,305 @@ def new_party_form_group(request, application, strings, back_url, clearance_opti
         forms.append(party_signatory_name_form("Signatory name", "Save and continue"))
 
     return FormGroup(forms)
+
+
+class PartyReuseForm(forms.Form):
```
saruniitr, line 140: Replacing existing lite-forms in this journey with Django forms

exporter/applications/views/parties/end_users.py
```diff
@@ -116,3 +138,435 @@ def __init__(self):
...
+        "name": getattr(document, "original_name", document.name),
+        "s3_key": document.name,
+        "size": int(document.size // 1024) if document.size else 0,  # in kilobytes
+    }
+
+    response, status_code = post_party_document(request, application_id, party_id, data)
+    assert status_code == HTTPStatus.CREATED
+
+
+class SetPartyView(LoginRequiredMixin, SessionWizardView):
```
saruniitr, line 166: New wizard using Django session wizard to add End-user. This ticket is specific to End-user only hence this journey is replaced.

exporter/templates/applications/party-summary.html
```diff
@@ -0,0 +1,164 @@
...
+                            </a>
+                        </dd>
+                    </div>
+                </dl>
+            {% endif %}
+
+        {% endfor %}
+    </div>
+
+    <div class="govuk-summary-list">
```
kevincarrogan, line 156: This shouldn't be a `govuk-summary-list`.

exporter/applications/views/parties/end_users.py
```diff
@@ -1,118 +1,514 @@
...
+        return reverse("applications:end_user_summary", kwargs=self.kwargs)
+
+
+class PartyDocumentDownloadView(LoginRequiredMixin, PartyContextMixin, TemplateView):
+    def get(self, request, **kwargs):
+        document = [doc for doc in self.party["documents"] if doc["id"] == kwargs["document_pk"]]
+        if not document:
+            return error_page(request, "Requested document not associated with this party")
+
+        document = document[0]
```
kevincarrogan, line 513: Should we do something here if there's more than one? Why is it the first one?

exporter/applications/views/parties/end_users.py
```diff
@@ -1,118 +1,514 @@
...
+                self.existing_documents[PartyDocumentType.END_USER_COMPANY_LETTERHEAD_DOCUMENT],
+            )
+
+        return redirect(reverse("applications:end_user_summary", kwargs=self.kwargs))
+
+
+class PartyDocumentEditView(LoginRequiredMixin, PartyContextMixin, FormView):
+    def get_form(self):
+        form_kwargs = self.get_form_kwargs()
+        if self.kwargs.get("document_type") == "english_translation":
```
kevincarrogan, line 474: `english_translation` and `company_letterhead` could be constants.

exporter/applications/views/parties/end_users.py
```diff
@@ -1,118 +1,514 @@
...
+        if step != SetPartyFormSteps.PARTY_DOCUMENT_UPLOAD:
+            return {"end_user_document_available": True}
+
+        document = [
+            doc for doc in self.party["documents"] if doc["type"] == PartyDocumentType.END_USER_UNDERTAKING_DOCUMENT
+        ]
+        if not document:
+            raise ValueError("End-user undertaking document not yet uploaded, editing not allowed")
+
+        document = document[0]
```
kevincarrogan, line 372: What happens if there are more than one? Why is it the first one?

exporter/applications/views/parties/end_users.py
```diff
@@ -1,118 +1,514 @@
...
+class PartyEditMixin(LoginRequiredMixin, PartyContextMixin, FormView):
+    def form_valid(self, form):
+        update_party(self.request, self.application_id, self.party_id, form.cleaned_data)
+        return super().form_valid(form)
+
+    def get_success_url(self):
+        return reverse("applications:end_user_summary", kwargs=self.kwargs)
+
+
+class PartySubTypeEditView(PartyEditMixin):
```
kevincarrogan, line 281: `PartyEditMixin` isn't really a mixin if it's the _only_ class we're deriving from.

exporter/applications/views/parties/end_users.py
```diff
@@ -1,118 +1,514 @@
...
+        return str(self.kwargs["pk"])
+
+    @property
+    def party_id(self):
+        return str(self.kwargs["obj_pk"])
+
+    @property
+    def party(self):
+        party = get_party(self.request, self.kwargs["pk"], self.kwargs["obj_pk"])
+        party_type = list(party.keys())[0]
```
kevincarrogan, line 246: Why is it the first one? Is it definitely always the first one?

exporter/templates/applications/party-summary.html
```diff
@@ -0,0 +1,164 @@
...
+            {% endif %}
+
+            {% if document.type == "end_user_english_translation_document" %}
+                <dl class="govuk-summary-list">
+                    <div class="govuk-summary-list__row">
+                        <dt class="govuk-summary-list__key">
+                            Upload an English translation of the end-user document
+                        </dt>
+                        <dd class="govuk-summary-list__value">
+                            <a {% if document.safe %} href="{% url 'applications:party_document_download' pk obj_pk document.id %}"{% endif %} class="govuk-link--no-visited-state">{{ document.name }}</a>
```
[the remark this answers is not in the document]
saruniitr: No, it is shown as link only if the document is safe so user can download, otherwise we just show document name.

exporter/templates/applications/party-summary.html
saruniitr: correct but I used this to correctly align these buttons with the table

exporter/applications/views/parties/end_users.py
```diff
@@ -1,118 +1,514 @@
...
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        return {**context, "party": self.party}
+
+
+class PartySummaryView(LoginRequiredMixin, PartyContextMixin, TemplateView):
+    template_name = "applications/party-summary.html"
+
+    def get_success_url(self):
+        return "#"
```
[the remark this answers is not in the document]
saruniitr: no, will fix this

exporter/applications/views/parties/end_users.py
```diff
@@ -1,118 +1,514 @@
...
+        if step == SetPartyFormSteps.PARTY_DOCUMENTS:
+            cleaned_data = {"end_user_document_available": True}
+
+        if cleaned_data is None:
+            return {}
+        return cleaned_data
+
+    def done(self, form_list, **kwargs):
+        all_data = {k: v for form in form_list for k, v in form.cleaned_data.items()}
+        print(all_data)
```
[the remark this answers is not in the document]
saruniitr: :+1:

exporter/applications/views/parties/end_users.py
```diff
@@ -1,118 +1,514 @@
...
+
+        data = {
+            "type": party_document_type,
+            "name": getattr(document, "original_name", document.name),
+            "s3_key": document.name,
+            "size": int(document.size // 1024) if document.size else 0,  # in kilobytes
+        }
+
+        _, status_code = post_party_document(self.request, self.application_id, self.party_id, data)
+        assert status_code == HTTPStatus.CREATED
```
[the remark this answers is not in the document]
saruniitr: ok

exporter/templates/applications/party-summary.html
```diff
@@ -0,0 +1,164 @@
+{% extends 'layouts/base.html' %}
+
+{% block back_link %}
+{% endblock %}
+
+{% block body %}
+    <div class="govuk-width-container">
+        <a href="#" class="govuk-back-link">Back</a>
```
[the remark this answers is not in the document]
saruniitr: will fix this

exporter/applications/views/parties/end_users.py
saruniitr: yes because `get_party` only returns one party identified by `obj_pk`. We get a dict back with the key as `party_type`.
May be I can update the serializer in the API to include `type` as separate key without breaking other changes.

exporter/applications/views/parties/end_users.py
saruniitr: sorry I don't follow, `PartyEditMixin` derives three other classes which are all required in all of the edit forms.

exporter/applications/views/parties/end_users.py
```diff
@@ -1,118 +1,514 @@
...
+    }
+
+    def get_form_kwargs(self, step=None):
+        kwargs = super().get_form_kwargs(step)
+
+        # get existing document status for the end-user
+        self.existing_documents = {document["type"]: document["id"] for document in self.party["documents"]}
+        self.english_translation_exists = bool(
+            {PartyDocumentType.END_USER_ENGLISH_TRANSLATION_DOCUMENT}.intersection(self.existing_documents.keys())
         )
```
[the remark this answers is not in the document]
saruniitr: yeah true, will change this

exporter/applications/views/parties/end_users.py
saruniitr: There will only be one document for each type, if another one is uploaded then the previous one gets deleted.

exporter/applications/views/parties/end_users.py
saruniitr: :+1:

exporter/applications/views/parties/end_users.py
saruniitr: we are retrieving document by `document_pk` so there will only be one

exporter/templates/applications/party-summary.html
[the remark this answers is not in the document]
kevincarrogan: In that case if we just want to display the name then it shouldn't be wrapped in an `a` tag at all, it shouldn't just be removing the href.

exporter/applications/views/parties/end_users.py
kevincarrogan: Usually when I see a Mixin of some sort I would expect it to be one of multiple classes that a class derives from, it seems unexpected that something solely derives from a Mixin (a Mixin usually means that it provides a subset of functionality but isn't fully complete on its own).

saruniitr: understand, I called it incorrectly as a mixin, it should have been `PartyEditView`

exporter/applications/views/parties/end_users.py
```diff
@@ -1,118 +1,514 @@
...
+
+
+class SetEndUserView(SetPartyView):
+    party_type = "end_user"
+
+    def get_success_url(self, party_id):
+        return reverse("applications:end_user_summary", kwargs={"pk": self.kwargs["pk"], "obj_pk": party_id})
+
+
+class CopyEndUserView(SetEndUserView):
```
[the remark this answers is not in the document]
saruniitr: `SetPartyView` is created to be independent of party type. Currently we are only changing `end-user` but we have similar journey for other parties (consignee, ultimate end-user and third parties) and eventually we have to replace them. End-user is the important one hence we are starting with that.
For this reason `CopyEndUserView` extends `SetEndUserView`.

kevincarrogan (approved):

PR #454, opened 2022-03-02
Enter serial numbers later

kevincarrogan:
Adds a form view, and link to said form view from the application details page, to allow serial numbers to be added later.

There are two different cases where we want to allow exporters to update serial numbers, either because they've explicitly said they would like to do it later or because they've said they could add them straight away yet only added a few of them.

### DIFF
--- a/core/constants.py
+++ b/core/constants.py
@@ -1,6 +1,64 @@
+import re
+
+
 class GoodsTypeCategory:
     MILITARY = "military"
     CRYPTOGRAPHIC = "cryptographic"
     MEDIA = "media"
     UK_CONTINENTAL_SHELF = "uk_continental_shelf"
     DEALER = "dealer"
+
+
+class CaseStatusEnum:
+    APPEAL_FINAL_REVIEW = "appeal_final_review"
+    APPEAL_REVIEW = "appeal_review"
+    APPLICANT_EDITING = "applicant_editing"
+    CHANGE_INTIAL_REVIEW = "change_initial_review"
+    CHANGE_UNDER_FINAL_REVIEW = "change_under_final_review"
+    CHANGE_UNDER_REVIEW = "change_under_review"
+    CLC = "clc_review"
+    OPEN = "open"
+    UNDER_INTERNAL_REVIEW = "under_internal_review"
+    RETURN_TO_INSPECTOR = "return_to_inspector"
+    AWAITING_EXPORTER_RESPONSE = "awaiting_exporter_response"
+    CLOSED = "closed"
+    DEREGISTERED = "deregistered"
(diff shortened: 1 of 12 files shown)


core/constants.py
```diff
@@ -1,6 +1,64 @@
...
+
 class GoodsTypeCategory:
     MILITARY = "military"
     CRYPTOGRAPHIC = "cryptographic"
     MEDIA = "media"
     UK_CONTINENTAL_SHELF = "uk_continental_shelf"
     DEALER = "dealer"
+
+
+class CaseStatusEnum:
```
depsiatwal, line 12: nice enum :-)

exporter/applications/helpers/check_your_answers.py
```diff
@@ -632,3 +643,19 @@ def get_application_type_string(application):
         return applications.ApplicationPage.Summary.Licence.TRADE_CONTROL
     else:
         return APPLICATION_TYPE_STRINGS[application_type]
+
+
+def requires_serial_numbers(good_on_application):
+    try:
+        firearm_details = good_on_application["firearm_details"]
+    except KeyError:
```
depsiatwal, line 651: not a big fan of using exceptions for normal programming flow 
wonder if it's better to use 

if good_on_application.get("firearm_details") 
else:
    return false

kevincarrogan: https://devblogs.microsoft.com/python/idiomatic-python-eafp-versus-lbyl/#:~:text=One%20idiomatic%20practice%20in%20Python,ask%20for%20forgiveness%20than%20permission%E2%80%9D. - I guess it's kind of a philosophical thing, I tend to write using the EAFP style.

kevincarrogan: Changed.

exporter/goods/views.py
```diff
@@ -1174,3 +1179,90 @@ def post(self, request, **kwargs):
...
+
+    def get_context_data(self, **kwargs):
+        ctx = super().get_context_data(**kwargs)
+
+        ctx["back_link_url"] = self.get_success_url()
+        ctx["title"] = self.get_form().title
+
+        return ctx
+
+    def generate_serial_numbers_data(self, cleaned_data):
```
depsiatwal, line 1239: worth moving to helper class ?

depsiatwal (commented):
few small comments

exporter/goods/views.py
kevincarrogan: Considering this is highly specific to this particular view I don't think moving it out makes any sense. If and when this could be used somewhere else then I think it should be moved.

depsiatwal (approved):
lgtm

PR #455, opened 2022-03-03
LTD 1745: Add missing validation and test

wkeeling:


exporter/goods/forms.py
```diff
@@ -1196,15 +1202,26 @@ def __init__(self, *args, **kwargs):
...
+            ("yes_modified", "modified_details", "Enter the details of the modifications and the hardware"),
+            (
+                "yes_general",
                 "general_details",
                 "Enter the details of the types of applications the component is intended to be used in",
-            )
+            ),
+        )
+
+        max_chars = 2000
```
saruniitr, line 1215: I think in general we don't want to limit the user input, could you check if this is a requirement?

wkeeling: Will do, although this replicates the current behaviour we currently have in prod. The underlying model in the API imposes a `max_length=2000` restriction on the `Good.component_details` field, so sounds like we need to potentially drop that too. Will double-check the requirement.

wkeeling: Colin confirmed that limits should be scrapped, so have raised a separate ticket to address (will need both API and FE changes).

saruniitr: :+1:

saruniitr (approved):

PR #457, opened 2022-03-03
Improve the way we login into sso

PippoRaimondi:
The intent of this PR is to prevent us from having to go through sso via client side, adding credentials in the browser and then navigating to internal domain.

The way we workaround all these steps and manage successfully authenticate to the app is by:

- submit a GET to sso login url and extract the csrftoken
- submit a POST to sso login with correct body/header and csrftoken from the first step
- persist the sessionId cookie generated by the POST

All the above is done within less than a second which means we save quite some time compared to previous way we were logging in.

### DIFF
--- a/cypress/integration/caseworker/application.spec.js
+++ b/cypress/integration/caseworker/application.spec.js
@@ -1,69 +1,64 @@
 import { createApplication } from '../../api/createApplication'
 
-describe('Login', () => {
-  it('login', () => {
+
+describe('Application', () => {
+  const counterSignQueue = Cypress.env('fco_counter_sign_queue')
+  let response
+
+  before(async () => {
     cy.login()
+    cy.visit('/')
+    response = await createApplication()
   })
 
-  describe('Application', () => {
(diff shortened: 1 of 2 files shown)


cypress/integration/caseworker/application.spec.js
```diff
@@ -1,69 +1,64 @@
...
-        .should('contain', 'Ukraine')
-        .should('contain', 'FCDO Approval Reason')
-      cy.findByText('Move case forward').click()
+    cy.visit(`/queues/${response.defaultQueue}/?case_reference=${response.submittedApplication.reference_code}`)
+    cy.get('#form-cases').should('contain', 'There are no new cases')
+  })
 
-      cy.visit(`/queues/${response.defaultQueue}/?case_reference=${response.submittedApplication.reference_code}`)
-      cy.get('#form-cases').should('contain', 'There are no new cases')
+  it('should approve a case in Counter-Sign queue with the correct information', () => {
```
saruniitr, line 37: we are using "countersign" in all places

kevincarrogan (approved):

PR #461, opened 2022-03-08
Add list view for applications requiring serial numbers

kevincarrogan:


depsiatwal (approved):
LGTM would be nice to see some screen shots and may it working but trust you have it working :-)

PR #465, opened 2022-03-09
 LTD-1706-sign-out-gov-uk

depsiatwal:
This upgrades the sign-out view to support GOV UK signout


wkeeling (approved):

unit_tests/exporter/auth/test_views.py
```diff
@@ -0,0 +1,33 @@
...
+
+def test_log_out_logged_in(authorized_client):
+    session = authorized_client.session
+
+    assert session[settings.TOKEN_SESSION_KEY]
+    url = reverse("auth:logout")
+    response = authorized_client.get(url)
+    assert response.status_code == 302
+    assert authorized_client.session.get(settings.TOKEN_SESSION_KEY) is None
+    assert response.url == f"{settings.LOGOUT_URL}http://testserver/"
```
kevincarrogan, line 14: This seems weird.

unit_tests/exporter/auth/test_views.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+def test_log_out_logged_in(authorized_client):
+    session = authorized_client.session
+
+    assert session[settings.TOKEN_SESSION_KEY]
+    url = reverse("auth:logout")
+    response = authorized_client.get(url)
+    assert response.status_code == 302
+    assert authorized_client.session.get(settings.TOKEN_SESSION_KEY) is None
```
kevincarrogan, line 13: This should either be
```
assert settings.TOKEN_SESSION_KEY not in authroized_client.session
```
or
```
assert authorized_client.session[settings.TOKEN_SESSION_KEY] is None
```

to avoid the ambiguity of this current assertion.

unit_tests/exporter/auth/test_views.py
depsiatwal: yeah it's the way LOGOUT URL is current constructed it has parameters and ?next the end

kevincarrogan: Ok. I guess why it's confusing me is in this test there isn't an explicit `?` but in the other test there is.

kevincarrogan (approved):

PR #467, opened 2022-03-10
[LTD-1952] TAU App

alixedi:
This PR adds an app that will host TAU 2.0 functionality that we are looking to build.

The `home` view is just a placeholder and I fully expect it to be replaced/renamed down the line.

### DIFF
--- a/caseworker/tau/urls.py
+++ b/caseworker/tau/urls.py
@@ -1,0 +1,9 @@
+from django.urls import path
+
+from caseworker.tau.views import TAUHome
+
+app_name = "tau"
+
+urlpatterns = [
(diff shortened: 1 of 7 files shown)


depsiatwal (commented):
LGTM just failing tests

caseworker/tau/urls.py
```diff
@@ -0,0 +1,7 @@
+from django.urls import path
+
+from caseworker.tau.views import TAUHome
+
```
wkeeling, line 4: Wondering if we need to specify a namespace, e.g.
```python
app_name = "tau"
```

wkeeling (approved):

PR #470, opened 2022-03-10
Ensure actions columns remain balanced across summary rows

kevincarrogan:


exporter/applications/helpers/check_your_answers.py
```diff
@@ -225,6 +225,11 @@ def _convert_hmrc_query(application, editable=False):
 
 def convert_goods_on_application(goods_on_application, is_exhibition=False, is_summary=False):
     converted = []
+
+    def requires_actions(good_on_application):
+        return not is_summary and requires_serial_numbers(good_on_application)
```
saruniitr, line 230: just want to double check the condition is correct.
On the summary page (i.e., "Check your answers before submitting page"), `is_summary` is going to be `True` so if some products require serial numbers then it will return `False` but we want to show actions in that case correct?

kevincarrogan: These actions should only be displayed if it's not the summary page.

saruniitr: :+1:

saruniitr (approved):

PR #474, opened 2022-03-11
LTD-1368: English translation changes review fixes

saruniitr:
## Change description

Product review fixes
- save undertaking document description
- Error handling when editing documents
- Correct links in 'Check your answers before submitting' page
- refactoring repeating bits into a common function

### DIFF
--- a/exporter/applications/helpers/check_your_answers.py
+++ b/exporter/applications/helpers/check_your_answers.py
@@ -15,6 +15,7 @@ from exporter.core.constants import (
     PERMANENT,
     CaseTypes,
     APPLICATION_TYPE_STRINGS,
+    PartyDocumentType,
 )
 from core.constants import GoodsTypeCategory
 from core.builtins.custom_tags import (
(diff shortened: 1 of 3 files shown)


exporter/applications/helpers/check_your_answers.py
```diff
@@ -489,19 +490,34 @@ def convert_party(party, application, editable):
         data["Role"] = party.get("role_other") if party.get("role_other") else party.get("role").get("value")
 
     if application["case_type"]["sub_type"]["key"] != OPEN:
-        if party.get("document"):
-            document_type = party["type"] if party["type"] != "end_user" else "end-user"
-            document = _convert_document(party, document_type, application["id"], editable)
+        if party["type"] == "end_user":
+            party_type = "end-user"
+            for doc in party.get("documents"):
```
kevincarrogan, line 495: The `.get` call may return `None` which is going to make the `for` fail.

exporter/applications/views/parties/end_users.py
```diff
@@ -83,16 +83,25 @@ def form_valid(self, form):
...
-    assert status_code == HTTPStatus.CREATED
+    response, status_code = post_party_document(request, application_id, party_id, data)
+    if status_code == HTTPStatus.CREATED:
+        log.error(
+            "Error uploading party document of type %s - response was: %s - %s",
+            data["type"],
+            status_code,
+            response,
+            exc_info=True,
+        )
```
kevincarrogan, line 103: I think this should be `if status_code != HTTPStatus.CREATED`, we should probably also direct to an error page if this happens.

exporter/applications/views/parties/end_users.py
```diff
@@ -177,41 +185,38 @@ def done(self, form_list, **kwargs):
...
+            _, status_code = _post_party_document(
+                self.request,
+                self.application.id,
+                party_id,
+                PartyDocumentType.END_USER_UNDERTAKING_DOCUMENT,
+                party_document,
+                all_data["description"],
+            )
+            if status_code != HTTPStatus.CREATED:
+                return error_page(self.request, "Unexpected error uploading party document")
```
kevincarrogan, line 197: Should we stick with the same pattern in other places and log an error here?

exporter/applications/helpers/check_your_answers.py
saruniitr: right, will fix it.

exporter/applications/views/parties/end_users.py
saruniitr: The log is added in `_post_party_document()` which contains the repeating bits eg preparing the data payload and logging

saruniitr: The log is added in `_post_party_document()` which contains the repeating bits eg preparing the data payload and logging

exporter/applications/helpers/check_your_answers.py
```diff
@@ -489,19 +490,34 @@ def convert_party(party, application, editable):
         data["Role"] = party.get("role_other") if party.get("role_other") else party.get("role").get("value")
 
     if application["case_type"]["sub_type"]["key"] != OPEN:
-        if party.get("document"):
-            document_type = party["type"] if party["type"] != "end_user" else "end-user"
-            document = _convert_document(party, document_type, application["id"], editable)
+        if party["type"] == "end_user":
```
saruniitr, line 493: this is specific to end user because these changes are only applicable to end-user and we only updated the forms, urls specific to end-user.
When we replace other party forms with django forms then we can combine this in a better way.

kevincarrogan (approved):

PR #487, opened 2022-03-15
add AWS_S3_ENDPOINT_URL to enable local endpoint

sekharpanja:
add AWS_S3_ENDPOINT_URL to enable local endpoint


Dockerfile
```diff
@@ -6,6 +6,7 @@ RUN wget -q -O - https://dl-ssl.google.com/linux/linux_signing_key.pub | apt-key
 RUN sh -c 'echo "deb [arch=amd64] http://dl.google.com/linux/chrome/deb/ stable main" >> /etc/apt/sources.list.d/google.list'
 RUN apt-get update --fix-missing
 RUN apt-get -y install google-chrome-stable
+RUN apt-get install -y git
```
saruniitr, line 9: why do we need this?

PR #497, opened 2022-03-17
Change because dict isn't hashable.

alixedi:


seijihg (approved):

wkeeling (approved):

caseworker/tau/templatetags/tau_tags.py
```diff
@@ -7,4 +7,4 @@
 @register.filter()
 def is_ultimate_end_user(destinations):
     """If the case does not have an ultimate_end_user, then the product might be exported onwards to another country"""
-    return {d for d in destinations if d.get("type") == "ultimate_end_user"} != {}
+    return {d.get("type") for d in destinations if d.get("type") == "ultimate_end_user"} != {}
```
kevincarrogan, line 10: This is also trying to equate a set with a dictionary `{}` is not an empty set.

kevincarrogan: I think this could just be `return any(d.get("type") == "ultimate_end_user" for d in destinations]`

holdenweb: The easiest test for whether a container object `c` is empty is surely `len(c) == 0`? Whether it's a set, dict, list, tuple or [name container of choice].

holdenweb: Interestingly, since no empty set is ever a dictionary, I'm wondering what defect in the tests this excellent spot reveals. :)

unit_tests/caseworker/tau/test_home.py
```diff
@@ -49,7 +49,7 @@ def test_case_info(authorized_client, url):
         "Is the product for military use?",
         "No",
         "Will the product be onward exported to any additional countries?",
-        "Yes",
+        "No",
```
alixedi, line 52: "I think it would be the other way round, if there is an ultimate_end_user then that means the product is being onward exported." - Jess

PR #504, opened 2022-03-21
LTD-1642: Provide option to upload EC3 document

saruniitr:
## Change description

An exporter may need to provide End-user EC3 document depending on goods starting location, destination and product type.
End-user journey is updated to ask for this document if the above conditions are met.
We only need this if goods are starting from NI, going to EU and firearms type is not software or technology related.
User can either upload a document or give an explanation for not having it. But both of these are optional.

The details required to determine (location, destination, type) are not all captured in one place, since user can complete these details in any order it is not possible to determine this requirement when user entering end-user details because he may not have added product details or starting location etc hence we display a banner notifying the user that he may need to provide these details. Once the details are provided then banner is also not shown.

### DIFF
--- a/caseworker/cases/views/main.py
+++ b/caseworker/cases/views/main.py
@@ -162,8 +162,8 @@ class CaseDetail(CaseView):
         self.slices = [
             Slices.GOODS,
             Slices.DESTINATIONS,
-            Slices.DENIAL_MATCHES,
-            Slices.SANCTION_MATCHES,
+            conditional(self.case.data["denial_matches"], Slices.DENIAL_MATCHES),
+            conditional(self.case.data["sanction_matches"], Slices.SANCTION_MATCHES),
             conditional(self.case.data["end_user"], Slices.END_USER_DOCUMENTS),
             conditional(self.case.data["inactive_parties"], Slices.DELETED_ENTITIES),
             Slices.LOCATIONS,
--- a/caseworker/templates/case/slices/end-user-documents.html
+++ b/caseworker/templates/case/slices/end-user-documents.html
@@ -69,6 +69,32 @@
                 </tr>
             {% endif %}
 
+            {% if end_user.documents|length %}
+                {% for document in end_user.documents %}
+                    {% if document.type == "end_user_ec3_document" %}
+                        <tr class="govuk-table__row">
+                            <th scope="row" class="govuk-table__header">Upload an EC3 form</th>
+                            <td class="govuk-table__cell">
+                                {% if document.safe %}
+                                    <a target="_blank" href="{% url 'cases:document' queue_pk=queue.id pk=case.id file_pk=document.id %}" class="govuk-link--no-visited-state">
+                                        EC3 form ({{ document.name|document_extension|upper}} opens in new tab)
+                                    </a>
+                                {% else %}
+                                    {{ document.name }}
+                                {% endif %}
+                            </td>
+                        </tr>
+                    {% endif %}
+                {% endfor %}
+            {% endif %}
+
+            {% if end_user.ec3_missing_reason %}
+                <tr class="govuk-table__row">
+                    <th scope="row" class="govuk-table__header">If you do not have an EC3 form, explain why (optional)</th>
+                    <td class="govuk-table__cell">{{ end_user.ec3_missing_reason }}</td>
+                </tr>
+            {% endif %}
+
             {% if consignee and consignee.documents|length %}
                 {% with document=consignee.documents.0 %}
                     <tr class="govuk-table__row">
--- a/exporter/applications/helpers/parties.py
+++ b/exporter/applications/helpers/parties.py
@@ -1,3 +1,6 @@
+from exporter.core.constants import FirearmsProductType
+
+
 def party_requires_ec3_document(application):
     """
     Helper function to determine EC3 document requirement status for End-user
@@ -25,13 +28,17 @@ def party_requires_ec3_document(application):
         eu_destination = True
 
     ec3_product_types = {
-        "firearms",
-        "components_for_firearms",
-        "ammunition",
-        "components_for_ammunition",
-        "firearms_accessory",
+        FirearmsProductType.FIREARMS,
+        FirearmsProductType.AMMUNITION,
+        FirearmsProductType.COMPONENTS_FOR_FIREARMS,
+        FirearmsProductType.COMPONENTS_FOR_AMMUNITION,
+        FirearmsProductType.FIREARMS_ACCESSORY,
     }
-    product_types = {product["firearm_details"]["type"]["key"] for product in application.get("goods", [])}
-    firearms_products = bool(ec3_product_types.intersection(product_types))
+    product_types = {
+        product["firearm_details"]["type"]["key"]
+        for product in application.get("goods", [])
+        if product.get("firearm_details")
+    }
+    firearms_products = ec3_product_types.intersection(product_types)
 
     return goods_from_NI and eu_destination and firearms_products
--- a/exporter/applications/helpers/parties.py
+++ b/exporter/applications/helpers/parties.py
@@ -1,0 +1,37 @@
+def party_requires_ec3_document(application):
+    """
+    Helper function to determine EC3 document requirement status for End-user
+
+    We only need to ask if the goods originate from NI, going to EU and the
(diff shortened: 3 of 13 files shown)


exporter/applications/helpers/parties.py
```diff
@@ -0,0 +1,37 @@
...
+    destination = application.get("destinations", {})
+    if (
+        destination
+        and destination["type"] == "end_user"
+        and destination["data"]
+        and destination["data"]["country"]["is_eu"]
+    ):
+        eu_destination = True
+
+    ec3_product_types = {
```
depsiatwal, line 27: Worth putting as constant some where ?
easier to maintain

exporter/applications/helpers/parties.py
```diff
@@ -0,0 +1,37 @@
...
+        "firearms",
+        "components_for_firearms",
+        "ammunition",
+        "components_for_ammunition",
+        "firearms_accessory",
+    }
+    product_types = {product["firearm_details"]["type"]["key"] for product in application.get("goods", [])}
+    firearms_products = bool(ec3_product_types.intersection(product_types))
```
wkeeling, line 35: Minor but the `bool()` isn't needed here since `firearms_products` is evaluated in a boolean context in the `return` statement.

exporter/applications/views/parties/end_users.py
```diff
@@ -223,7 +229,7 @@ class SetEndUserView(SetPartyView):
     party_type = "end_user"
 
     def get_success_url(self, party_id):
-        return reverse("applications:end_user_summary", kwargs={"pk": self.kwargs["pk"], "obj_pk": party_id})
+        return super().get_success_url(party_id)
```
wkeeling, line 232: Since this method now just delegates to the superclass, it can be removed?

unit_tests/exporter/applications/views/parties/test_end_users.py
```diff
@@ -80,6 +82,8 @@ def test_set_end_user_view(url, authorized_client, requests_mock, data_standard_
         "applications:end_user_summary", kwargs={"pk": data_standard_case["case"]["id"], "obj_pk": party_id}
     )
 
+    _ = requests_mock.request_history.pop()
+
     letterhead_data = requests_mock.request_history.pop().json()
```
wkeeling, line 87:
```python
requests_mock.request_history.pop(-2).json()
```
then you wouldn't need to pop the request on the line before?

wkeeling, line 87: Maybe change to:
```python
requests_mock.request_history.pop(-2).json()
```
then you wouldn't need to pop the request on the line before?

unit_tests/exporter/applications/views/parties/test_end_users.py
```diff
@@ -321,3 +348,84 @@ def test_edit_end_user_document(
...
+    data = end_user_documents[DOCUMENT_TYPE_PARAM_EC3_DOCUMENT]
+    ec3_edit_url = reverse(
+        f"applications:end_user_ec3_document",
+        kwargs={"pk": application_id, "obj_pk": end_user["id"]},
+    )
+    response = authorized_client.post(ec3_edit_url, data=data)
+    assert response.status_code == 302
+
+    _ = requests_mock.request_history.pop()
+    actual = requests_mock.request_history.pop().json()
```
wkeeling, line 425: Maybe use `pop(-2)` then could do away with the earlier `.pop()`?

exporter/applications/helpers/parties.py
saruniitr: :+1:

exporter/applications/helpers/task_lists.py
```diff
@@ -68,6 +70,21 @@ def get_application_task_list(request, application, errors=None):
...
+    end_user = application.get("end_user", {})
+    ec3_details_available = False
+    if end_user:
+        ec3_document_available = bool(
+            [
+                document
+                for document in end_user["documents"]
+                if document["type"] == PartyDocumentType.END_USER_EC3_DOCUMENT
+            ]
+        )
```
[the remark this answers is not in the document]
saruniitr: :+1: your suggestion is more readable as well

exporter/applications/helpers/parties.py
saruniitr: agree it is not required, wanted to be explicit about the type. I will remove.

exporter/applications/views/parties/end_users.py
saruniitr: yes

unit_tests/exporter/applications/views/parties/test_end_users.py
saruniitr: yes but we need to update `pop` similarly for the documents after this one, translation document and undertaking data so instead if we just pop the first one then we don't have to change the remaining.

unit_tests/exporter/applications/views/parties/test_end_users.py
saruniitr: here it makes sense as this is the only assertion, updated it

caseworker/cases/views/main.py
```diff
@@ -162,8 +162,8 @@ def get_standard_application(self):
         self.slices = [
             Slices.GOODS,
             Slices.DESTINATIONS,
-            Slices.DENIAL_MATCHES,
-            Slices.SANCTION_MATCHES,
+            conditional(self.case.data["denial_matches"], Slices.DENIAL_MATCHES),
+            conditional(self.case.data["sanction_matches"], Slices.SANCTION_MATCHES),
```
saruniitr, line 166: without this change we see extra space between the sections in the case detail page. It was an existing issue but highlighted recently when I added a new section to show End-user documents in one place.

caseworker/templates/case/slices/end-user-documents.html
```diff
@@ -69,6 +69,32 @@ <h2 class="govuk-heading-m">End-user documents</h2>
                 </tr>
             {% endif %}
 
+            {% if end_user.documents|length %}
```
saruniitr, line 72: additional change - this shows EC3 document details on the internal side

exporter/templates/applications/party-summary.html
```diff
@@ -192,6 +192,73 @@ <h1 class="govuk-heading-l">
                 </dd>
             </div>
         {% endif %}
+
+        {% if not "end_user_ec3_document"|has_party_document:party.documents and not party.ec3_missing_reason %}
```
saruniitr, line 196: this `if` block is the new change suggested after initial testing. When neither document nor reason is provided we now show questions in the summary page with blank values so user can still click change links and add them.

the `else` block was already reviewed, if either of them are available then we only show that field

wkeeling: Wonder whether this logic would be better placed inside the `PartySummaryView` and the outcome stored as a context variable, rather than doing this rather convoluted logic with a filter? e.g.

```python
class PartySummaryView(LoginRequiredMixin, PartyContextMixin, TemplateView):
    template_name = "applications/party-summary.html"

    def get_context_data(self, **kwargs):
        context = super().get_context_data(**kwargs)
        has_ec3 = any(
            document["type"] == "end_user_ec3_document" for document in self.party["documents"]
        )
        context["has_ec3"] = has_ec3
        return context
```
and then the template:
```
{% if not has_ec3 and not party.ec3_missing_reason %}
```
It may be a good idea to make `party` a `@cached_property`.

saruniitr: @wkeeling Extended this further to cover other document types so it is easier to check their availability in the template

Intentionally the `party` is not cached because user can edit various sections here and the documents we want to fetch updated data rather than the cached value.

exporter/applications/views/parties/end_users.py
```diff
@@ -268,6 +265,11 @@ def get_context_data(self, **kwargs):
 class PartySummaryView(LoginRequiredMixin, PartyContextMixin, TemplateView):
     template_name = "applications/party-summary.html"
 
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        available_documents = {document["type"]: document for document in self.party["documents"]}
+        return {**context, **available_documents}
```
wkeeling, line 271: Nice :+1:

wkeeling (approved):

PR #506, opened 2022-03-22
Create FE e2e docker and docker-compose capability

PippoRaimondi:
This PR consists on:

- Dockerfile with the necessary dependencies for the tests to run locally
- Docker-compose with the right parameters to bring the FE up pointing to devdata
- Makefile to ease running tests locally vs CI
- Enabling headless runs and video recording on failures for the playwright test
- Adding a circleci job that run the ui test on every branch/PR
- Ensuring we store video recording to circleci artifacts (https://app.circleci.com/pipelines/github/uktrade/lite-frontend/3549/workflows/4f16cbd2-7c9d-4571-9008-74df6c226547/jobs/13656/artifacts)


.circleci/config.yml
```diff
@@ -238,14 +260,29 @@ jobs:
       - ui_tests
       - vpn_disconnect
 
+  caseworker_e2e_test:
+    machine:
+      docker_layer_caching: true
```
kevincarrogan, line 265: From what I recall this doesn't do anything because it's a machine instance and/or because we don't have the right circleci subscription for this to work.

ci.caseworker.env
```diff
@@ -0,0 +1,43 @@
...
+
+# Below values can be found in vault
+export PW_SSO_USER=
+export PW_EXPORT_SSO_USER=
+export PW_SSO_PASSWORD=
+export PW_SSO_URL
+export PW_DIRECTORY_SSO_API_CLIENT_BASE_URL=
+export PW_DIRECTORY_SSO_API_CLIENT_API_KEY=
+export PW_AUTHBROKER_CLIENT_ID=
+export PW_AUTHBROKER_CLIENT_SECRET=
```
kevincarrogan, line 43: I've never seen this syntax in the env variables before. What does it do?

.circleci/config.yml
PippoRaimondi: You are right, I think this can even cause a problem (set as true) given we don't have access to its functioanlity... I don't remember what triggers it... I might as well set it to false (not sure what's the default value)

ci.caseworker.env
PippoRaimondi: Actually, it doesn't do anything other than remind you that you need these values exported in your bash instance or manually add these values to the docker-compose... given they don't break anything I left it there as a reminder... if you have a better idea or if this is too funky let me know...

PippoRaimondi: also, you can source the env file with the values and it exports the env variables with the `exports` command I.E `source caseworker.env`

.circleci/config.yml
PippoRaimondi: default is false. I removed it

ci.caseworker.env
PippoRaimondi: your options to successfully simulate circleci setup:

1:. Add the values for the environment variables directly in the docker-compose.e2e.yml file
2:. Copy ci.caseworker.env to caseworker.env and add the values into the export commands and run `source caseworker.env`

kevincarrogan (approved):

PR #511, opened 2022-03-23
[LTD-1953] TAU Form

alixedi:
This PR brings in a draft version of the TAU assessment form so that product peeps can review it in UAT -

Screenshot 2022-03-24 at 15 26 55

The formatting of this page requires more work (& cleanup of WIP things in the template especially) which I am planning to do in `LTD-1949`.

I would recommend reviewing this commit-wise.


caseworker/assets/javascripts/review-good.js
```diff
@@ -72,9 +72,11 @@ export default function initReviewGood() {
 
   if (!(controlListEntriesField || controlRationgField)) return;
 
-  controlRationgField.style.display = 'none';
+  if (controlRationgField) {
+    controlRationgField.style.display = 'none';
 
-  progressivelyEnhanceMultipleSelectField(controlRationgField)
+    progressivelyEnhanceMultipleSelectField(controlRationgField)
+  }
```
kevincarrogan, line 79: Isn't the `if` on line 73 already terminating this function before it gets here? At least that's what it reads like it's supposed to be doing and if it's not then I think just change that line to work instead of adding this other `if` statement.

alixedi: The if on line 73 returns if both `controlListEntriesField` & `controlRationgField` are false-y, no?

We are dealing with a case where we have a `controlListEntriesField` but not `controlRationgField`, in which case the function does not return.

caseworker/tau/forms.py
```diff
@@ -0,0 +1,72 @@
...
+
+
+class TAUAssessmentForm(forms.Form):
+    """
+    This is replacing caseworker.cases.forms.review_goods.ExportControlCharacteristicsForm.
+
+    TODO: Delete ExportControlCharacteristicsForm after this goes live.
+    """
+
+    MESSAGE_NO_CLC_MUTEX = "This is mutually exclusive with Control list entries"
```
wkeeling, line 15: Maybe use a lowercase `C` for `Control` here here just so it's consistent with the other messages.

caseworker/tau/forms.py
```diff
@@ -0,0 +1,72 @@
...
+        )
+
+    def clean(self):
+        cleaned_data = super().clean()
+        has_none = cleaned_data["does_not_have_control_list_entries"]
+        has_some = bool(cleaned_data["control_list_entries"])
+        if has_none and has_some:
+            raise forms.ValidationError({"does_not_have_control_list_entries": self.MESSAGE_NO_CLC_MUTEX})
+        elif not has_none and not has_some:
+            raise forms.ValidationError({"does_not_have_control_list_entries": self.MESSAGE_NO_CLC_REQUIRED})
```
wkeeling, line 71: These `ValidationError`s will trigger an error message at the top of the form. Just wondering if they needed to be shown against the individual field as well (i.e. with `self.add_error()`)?

alixedi: P.S. I copied all this from the form that this is trying to replace.

But, the problem that I see with this is that there is (1) a text field and (2) a checkbox and the form is invalid if (1) is empty and (2) is not checked. In that, the error doesn't really belong to a field per se. 

That said, it's not the most efficient UI design either. One could just add a help text that reads like - "Leave this field empty if the product does not have any control list entries".

wkeeling: > the error doesn't really belong to a field per se
Yes, makes sense

wkeeling (approved):

PR #518, opened 2022-03-24
Upgrade to supported version of ubuntu when running UI tests

kevincarrogan:


.circleci/config.yml
```diff
@@ -220,8 +220,7 @@ jobs:
 
   exporter_ui_test:
     machine:
-      image: ubuntu-1604:201903-01
-      docker_layer_caching: true
+      image: ubuntu-2004:202201-02
```
saruniitr, line 223: In the latest `master` we have one more job `caseworker_e2e_test` that needs updating

PR #531, opened 2022-03-29
LU-2037: UI test for checking denials

wkeeling:


ui_tests/caseworker/pages/add_denial_records_page.py
```diff
@@ -0,0 +1,55 @@
...
+class AddDenialRecordsPage(BasePage):
+    CSV_FILE_LOCATION = "/tmp/example-denials.csv"
+
+    def download_example_csv_file(self):
+        self.driver.find_element(by=By.LINK_TEXT, value="Download an example .csv file").click()
+
+        for _ in range(20):
+            if os.path.exists(self.CSV_FILE_LOCATION):
+                return
+            time.sleep(0.2)
```
saruniitr, line 20: you can use `driver.implicitly_wait(delay)` for this, not sure if that is any better than this.

ui_tests/caseworker/features/denials.feature
```diff
@@ -0,0 +1,41 @@
...
+    When I select end user "Joe Bloggs"
+    And I click "View related denials"
+    Then I should see "Joe Bloggs" listed
+    When I select "Joe Bloggs"
+    And I click "Add as exact match"
+    Then I should see "Joe Bloggs" as an exact match
+    When I select "Joe Bloggs" under denial matches
+    And I click "Remove denial match"
+    Then I should not see "Joe Bloggs" as an exact match
+    And I cleanup the temporary files created
```
saruniitr, line 37: I think it is better we cleanup the files at the beginning (before downloading the template) because if there is an error in between then we don't get a chance to cleanup and the test might read from previous files.

ui_tests/caseworker/features/denials.feature
```diff
@@ -0,0 +1,41 @@
...
+
+    Examples:
+    | name        | product  | part_number | clc_rating | end_user_name | end_user_address | country | consignee_name | consignee_address | address          | notifying_govmt  | final_dest | item_list_codes  | item_desc | end_use   |
+    | Joe Bloggs  | Rifle1   | PN-ABC-123  | PL9002     | Joe Bloggs    | 123 Main Street  | GE      | Joe Bloggs     | 123 Main Street   | 123 Main Street  | France           | Germany    | 1234             | gun       | Kill      |
```
saruniitr, line 41: can we change the `end_use` to less scary example

ui_tests/caseworker/features/denials.feature
wkeeling: Good idea.

ui_tests/caseworker/pages/application_page.py
```diff
@@ -293,3 +293,35 @@ def go_to_cases_activity_tab_for_eua(self, internal_url, context):
...
+
+    def get_exact_matches(self):
+        """Return a list of names that have exact denial matches."""
+        return [
+            td.text
+            for td in self.driver.find_elements(
+                by=By.XPATH,
+                value="//table[@id='table-denial-matches']/tbody/tr/td[3][contains(preceding-sibling::td/strong/text(), 'EXACT MATCH')]",
+            )
+        ]
```
saruniitr, line 321: Is it worth combining into a single function with `match_type` as an argument as these are exactly the same?

ui_tests/caseworker/features/denials.feature
saruniitr: Also both end-user and consignee names are same which can be confusing

saruniitr (approved):

PR #543, opened 2022-03-31
[LTD-1949] TAU list view

alixedi:
Not perfect in terms of layout but it is one step closer to how we want it to be. There will be follow-up tickets to fix some of the layout issues, add the edit version of this form for assessed goods, some changes to good details etc.

Screenshot 2022-03-31 at 10 51 55

### DIFF
--- a/caseworker/tau/templatetags/tau_tags.py
+++ b/caseworker/tau/templatetags/tau_tags.py
@@ -8,3 +8,10 @@ register = template.Library()
 def is_ultimate_end_user(destinations):
     """If the case does not have an ultimate_end_user, then the product might be exported onwards to another country"""
     return any(d.get("type") == "ultimate_end_user" for d in destinations)
+
+
+@register.filter()
+def get_destinations(case):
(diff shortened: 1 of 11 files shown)


caseworker/tau/templatetags/tau_tags.py
```diff
@@ -8,3 +8,10 @@
 def is_ultimate_end_user(destinations):
     """If the case does not have an ultimate_end_user, then the product might be exported onwards to another country"""
     return any(d.get("type") == "ultimate_end_user" for d in destinations)
+
+
+@register.filter()
+def get_destinations(case):
+    """Get unique destinations for the case in a nice comma-separated string"""
+    unique_destinations = {(dest.get("country") or {}).get("name") or "" for dest in case.destinations}
+    return sorted(list(unique_destinations - {""}))
```
wkeeling, line 17: You could also filter out the empty string with:
```python
return sorted(filter(None, unique_destinations))
```

wkeeling (approved):

PR #544, opened 2022-03-31
Ltd 1433 certificates section

kevincarrogan:
This refactors some common pieces of functionality:

  - Getting file data to post to the API
  - Handling errors when posting to the API from the firearms wizard
  - Setting up crispy form layouts in firearm forms
  - How we build the payload to create the product

### DIFF
--- a/exporter/goods/forms/firearms.py
+++ b/exporter/goods/forms/firearms.py
@@ -35,11 +35,33 @@ class TextChoice(Choice):
         super().__init__(choice.value, choice.label, **kwargs)
 
 
-class FirearmCategoryForm(forms.Form):
+class BaseFirearmForm(forms.Form):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        self.helper = FormHelper()
+        for field in self.fields.values():
+            if isinstance(field, forms.FileField):
+                self.helper.attrs = {"enctype": "multipart/form-data"}
+                break
+
+        self.helper.layout = Layout(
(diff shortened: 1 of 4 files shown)


exporter/goods/forms/firearms.py
```diff
@@ -35,11 +35,29 @@ def __init__(self, choice, **kwargs):
         super().__init__(choice.value, choice.label, **kwargs)
 
 
-class FirearmCategoryForm(forms.Form):
+class BaseFirearmForm(forms.Form):
```
saruniitr, line 38: this is nice, avoids repetition in each form

exporter/goods/forms/firearms.py
```diff
@@ -35,11 +35,29 @@ def __init__(self, choice, **kwargs):
...
+        super().__init__(*args, **kwargs)
+
+        self.helper = FormHelper()
+        for field in self.fields.values():
+            if isinstance(field, forms.FileField):
+                self.helper.attrs = {"enctype": "multipart/form-data"}
+                break
+
+        self.helper.layout = Layout(
+            HTML.h1(self.Layout.TITLE),
```
saruniitr, line 49: this breaks if the derived class doesn't define `Layout` so wondering if we include that in the base class itself?

saruniitr: ok sounds good.
Advantage being derived classes just need to define `Layout` class otherwise they need to change these in `__init__`

saruniitr (approved):

PR #548, opened 2022-04-01
2115-update-sanctions-list

depsiatwal:


sekharpanja (approved):

r4vi:
changing from .html to .xml are you sure that's what people want? there's also a .html version there: https://ofsistorage.blob.core.windows.net/publishlive/2022format/ConList.html

PR #549, opened 2022-04-01
Add firearms act 1968 form

kevincarrogan:


exporter/goods/forms/firearms.py
```diff
@@ -581,3 +582,217 @@ def get_layout_fields(self):
             "product_document",
             "description",
         )
+
+
+class FirearmFirearmAct1968Form(BaseFirearmForm):
```
wkeeling, line 587: Is this meant to have two `Firearm` in the name?

kevincarrogan: Yes. It's a bit weird but I was trying to be consistent with the rest of the forms. All of these forms start with `Firearm` and this one happens to be about the `FirearmAct` hence why it's turned into `FirearmFirearmAct`.

Can chance it if it's too confusing though.

exporter/goods/forms/firearms.py
```diff
@@ -581,3 +582,217 @@ def get_layout_fields(self):
...
+                        "section_certificate_missing_reason",
+                        f"Enter a reason why you do not have a {file_type}",
+                    )
+
+            return cleaned_data
+
+    return BaseAttachFirearmActCertificateForm
+
+
+class FirearmAttachFirearmCertificateForm(get_base_firearm_act_attach_file_form("firearm certificate")):
```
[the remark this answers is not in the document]
kevincarrogan: The difference doing it that way is that there's a lot more code to write to go in and start changing all of the necessary labels and what not. This way it's a little simpler because you can do it all at the class definition but obviously it makes this bit of code slightly more confusing.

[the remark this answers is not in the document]
kevincarrogan: Less confusing is better so I've changed it.

exporter/goods/forms/firearms.py
wkeeling: OK makes sense.

wkeeling (approved):

PR #550, opened 2022-04-01
Add exporter to docker setup

PippoRaimondi:
The intent of this PR is to:

- Add exporter docker setup to compose
- Run smoke test against CI
- Ensure both exporter and caseworker are pointing to local started instance


makefile
```diff
@@ -66,11 +67,21 @@ secrets:
...
+start-exporter:
+	$(docker-e2e) up --build -d exporter redis
+
+stop-exporter:
+	$(docker-e2e) down --remove-orphans
+
 caseworker-e2e-test:
 	@echo "*** Requires starting the caseworker stack, which can be started running: 'make start-caseowkrer' ***"
-	$(docker-e2e) exec caseworker bash -c '$(wait-for-caseworker) && pipenv run pytest playwright_tests/specs/test_smoke.py --video=retain-on-failure --output=/app/playwright_videos'
+	$(docker-e2e) exec caseworker bash -c '$(wait-for-caseworker) && pipenv run pytest playwright_tests/specs/caseworker/test_smoke.py --video=retain-on-failure --output=/app/playwright_videos --base-url=https://internal.lite.service.devdata.uktrade.digital/'
```
wkeeling, line 83: Does the `base_url` need changing to point at `http://localhost:8200/` ?

PippoRaimondi: The way the VPN is setup we can't start caseworker locally (within the CI network) and have it reach devdata API in a different network. for that to happen we would need to build a proxy so that requests coming from the FE would send the auth header when making requests to the BE...

So for now, we need to switch this smoke test to hit devdata stack (which defeats the purpose of starting caseworker in the CI)... but it won't be like that for long, as soon as we have the DB setup properly, we can start the entire stack in CI and VPN will no longer be needed. and then we can switch back to hitting `http://localhost:8200`

wkeeling (approved):

PR #552, opened 2022-04-01
LTD-2142: Edit firearm product type changes from summary screen

saruniitr:


exporter/applications/views/goods/add_good_firearm.py
```diff
@@ -317,3 +321,69 @@ def done(self, form_list, form_dict, **kwargs):
...
+
+    def get_success_url(self):
+        return reverse("applications:product_summary", kwargs=self.kwargs)
+
+
+class FirearmEditProductType(FirearmEditView):
+    form_class = GroupTwoProductTypeForm
+
+    def get_initial(self):
+        return {"type": self.good["firearm_details"]["type"]["key"]}
```
kevincarrogan, line 389: I can't believe it's this simple to be able to change the product type and not require a user to go through other steps.

Surely changing the product type has a huge knock on effect for all the other details that have been entered?

kevincarrogan, line 389: I can't believe it's this simple to be able to change the product type and not require a user to go through other steps.

Surely changing the product type has a huge knock on effect for all the other details that have been entered?

saruniitr: that is correct, when we change product type it affects other questions such as "serial numbers" and "fire arms certificate" etc but they are not yet added to the wizard.

kevincarrogan: But right now we have a whole completely different journey for the first half of the creation of a product depending on the category chosen so it's not just the second half.

saruniitr: ok understand what you mean, will update this.

PR #555, opened 2022-04-05
LTD-2144-SIEL-Change-product-name

depsiatwal:
image
image

### DIFF
--- a/exporter/goods/forms/firearms.py
+++ b/exporter/goods/forms/firearms.py
@@ -113,6 +113,7 @@ class FirearmNameForm(BaseFirearmForm):
     class Layout:
         TITLE = "Give the product a descriptive name"
 
+    title = Layout.TITLE
     name = forms.CharField(
         label="",
         error_messages={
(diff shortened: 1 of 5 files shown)


exporter/goods/forms/firearms.py
```diff
@@ -113,6 +113,7 @@ class FirearmNameForm(BaseFirearmForm):
     class Layout:
         TITLE = "Give the product a descriptive name"
 
+    title = Layout.TITLE
```
saruniitr, line 116: this shouldn't be required

exporter/goods/views.py
```diff
@@ -542,6 +543,31 @@ def get_success_url(self):
             return reverse_lazy("goods:good", kwargs={"pk": self.object_pk})
 
 
+class EditNameView(LoginRequiredMixin, GoodCommonMixin, FormView):
```
saruniitr, line 546: This file `exporter/goods/views.py` is the current journey but the new wizard that we are adding now is behind a feature flag in https://github.com/uktrade/lite-frontend/blob/master/exporter/applications/views/goods/add_good_firearm.py.
I think all of the edit changes also should come from there without affecting this journey.

PR #557, opened 2022-04-05
LTD-2143: Add view to edit firearm category

saruniitr:


exporter/applications/urls.py
```diff
@@ -75,6 +75,11 @@
         goods.AttachFirearmActSectionDocument.as_view(),
         name="attach-firearms-certificate-existing-good",
     ),
+    path(
+        "<uuid:pk>/goods/<uuid:good_pk>/firearm/edit/category",
```
kevincarrogan, line 79: Can we be consistent and put a slash at the end of all URLs?

kevincarrogan (commented):
If we're splitting out the tests into a separate `test_edit_firearm.py` file maybe this is also a good indicator that we should split out the views in the same manner.

unit_tests/exporter/applications/views/test_edit_firearm.py
```diff
@@ -0,0 +1,43 @@
...
+    response = authorized_client.post(
+        url,
+        data={
+            "category": ["NON_AUTOMATIC_SHOTGUN", "NON_AUTOMATIC_RIM_FIRED_HANDGUN"],
+        },
+    )
+
+    assert response.status_code == 302
+    assert requests_mock.last_request.json() == {
+        "item_category": "group2_firearms",
```
kevincarrogan, line 41: Why do we also need to post the `item_category` here?

exporter/applications/views/goods/add_good_firearm.py
```diff
@@ -555,7 +561,7 @@ def application_id(self):
     def good_id(self):
         return str(self.kwargs["good_pk"])
 
-    @property
+    @cached_property
     def good(self):
         return get_good(self.request, self.good_id, full_detail=True)[0]
```
kevincarrogan, line 566: We should really do something sensible here if this fails. For example, if this doesn't bring back a result because someone has changed the URL so that it's an invalid id then we should make the page 404.

unit_tests/exporter/applications/views/test_edit_firearm.py
saruniitr: just a safeguard in the edit function that it is the firearms category (group2), can be removed in edit.

exporter/applications/urls.py
saruniitr: :+1:

exporter/applications/views/goods/add_good_firearm.py
saruniitr: :+1:

PR #560, opened 2022-04-05
LTD-2159: Switch to using django-chunk-s3-av-upload-handlers

wkeeling:


conf/base.py
```diff
@@ -152,15 +153,10 @@
 STATICFILES_STORAGE = env.str("STATICFILES_STORAGE", "whitenoise.storage.CompressedManifestStaticFilesStorage")
 
 # File Upload
-# https://github.com/uktrade/s3chunkuploader
-S3_DOCUMENT_ROOT_DIRECTORY = ""
-S3_APPEND_DATETIME_ON_UPLOAD = True
```
saruniitr, line 157: The setting `S3_APPEND_DATETIME_ON_UPLOAD` appends timestamp when we upload the file to S3 which is useful for us to keep it unique if user uploads another file with the same name.
Do we know if we have this behaviour in the new package?

wkeeling: The new package uses a UUID instead. Unfortunately that is not configurable - so it has to be a UUID. However, I did test file download and the filename used is the original name. So it doesn't seem that the UUID is publicly exposed.

saruniitr: :+1:

saruniitr:
lgtm unless there are more changes to be done

conf/base.py
wkeeling: It turns out that the file is stored on S3 with a datetime in the filename. The UUID is temporarily used to name the file whilst the chunks are being uploaded. The downloaded file uses the original filename.

saruniitr: ok so the timestamp is stripped off when we download the file?

core/file_handler.py
```diff
@@ -23,7 +64,7 @@ def receive_data_chunk(self, raw_data, start):
         if start == 0:
             mime = magic.from_buffer(raw_data, mime=True)
             if mime not in self.ACCEPTED_FILE_UPLOAD_MIME_TYPES:
-                self.abort(PermissionDenied("Unsupported file type"))
+                raise UploadFailed(PermissionDenied("Unsupported file type"))
```
saruniitr, line 67: Could you add the mime type determined to the error message, I think it helps with the debug otherwise we will have ask the user the type of document they are trying to upload etc

conf/base.py
wkeeling: Effectively, yes

saruniitr (approved):

PR #561, opened 2022-04-05
LTD-2144-SIEL-Change-product-name-2

depsiatwal:
I have had to add a mixin since the name is effectively editing the 'good' so edit fire arms is technically incorrect and confusing. Hence the break up from the EditfirearmsForm and mixin

image
image

### DIFF
--- a/exporter/applications/views/goods/add_good_firearm.py
+++ b/exporter/applications/views/goods/add_good_firearm.py
@@ -557,7 +557,7 @@ class FirearmProductSummary(LoginRequiredMixin, TemplateView):
         }
 
 
-class EditView(LoginRequiredMixin, FormView):
+class BaseEditView(LoginRequiredMixin, FormView):
     template_name = "core/form.html"
 
     @cached_property
@@ -577,12 +577,24 @@ class EditView(LoginRequiredMixin, FormView):
 
         return good
 
+    def dispatch(self, request, *args, **kwargs):
+        if not settings.FEATURE_FLAG_PRODUCT_2_0:
+            raise Http404
+
+        return super().dispatch(request, *args, **kwargs)
+
     def get_success_url(self):
         return reverse("applications:product_summary", kwargs=self.kwargs)
 
+    def form_valid(self, form):
+        self.process_valid_form(form)
+        return super().form_valid(form)
 
-class EditNameView(EditView):
-    template_name = "core/form.html"
+    def process_valid_form(self, form):
+        raise NotImplementedError(f"`process_valid_form` should be implemented on {self.__class__.__name__}")
+
+
+class EditNameView(BaseEditView):
     form_class = FirearmNameForm
 
     def get_initial(self):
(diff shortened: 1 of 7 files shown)


unit_tests/exporter/applications/views/conftest.py
```diff
@@ -0,0 +1,21 @@
...
+@pytest.fixture
+def mock_good_put(requests_mock, data_standard_case):
+    good = data_standard_case["case"]["data"]["goods"][0]["good"]
+    url = client._build_absolute_uri(f'/goods/{good["id"]}/')
+    yield requests_mock.put(url=url, json={})
+
+
+@pytest.fixture(autouse=True)
+def setup(mock_good_get, mock_good_put):
+    pass
```
kevincarrogan, line 21: There are a lot of tests in this folder. Do we really want them all to automatically be provided with this?

exporter/applications/views/goods/add_good_firearm.py
```diff
@@ -570,6 +575,27 @@ def good(self):
 
         return good
 
+
+class EditNameView(GoodsMixin, FormView):
```
kevincarrogan, line 579: Can you add the `LoginRequiredMixin` here.

exporter/applications/views/goods/add_good_firearm.py
```diff
@@ -570,6 +575,27 @@ def good(self):
...
+
+    def form_valid(self, form):
+        edit_good(self.request, self.good_id, form.cleaned_data)
+        return super().form_valid(form)
+
+    def get_success_url(self):
+        return reverse("applications:product_summary", kwargs=self.kwargs)
+
+
+class FirearmEditView(LoginRequiredMixin, GoodsMixin, FormView):
```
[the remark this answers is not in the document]
depsiatwal: Problem here is form_valid as base class 
so I can override and not call it which is kind of odd or just add an abstract signature in the super class  and Derived class does the work

unit_tests/exporter/applications/views/conftest.py
depsiatwal: Fair point 
removed setup to the test files where it's required and make it more explicit

exporter/applications/views/goods/add_good_firearm.py
```diff
@@ -550,7 +557,7 @@ def get_context_data(self, **kwargs):
         }
 
 
-class FirearmEditView(LoginRequiredMixin, FormView):
+class EditView(LoginRequiredMixin, FormView):
```
saruniitr, line 560: I am doing something similar and I think it can be improved further. We need a common class(eg GoodCommonMixin) that allows editing `good` details and another class `FirearmEditView` using which we can edit `firearm_details` in the `good` and individual field class derive from whichever is relevant.

exporter/applications/views/goods/add_good_firearm.py
```diff
@@ -570,18 +577,35 @@ def good(self):
...
+
+    def get_initial(self):
+        return {
+            "name": self.good["name"],
+        }
+
     def form_valid(self, form):
-        edit_firearm(self.request, self.good_id, {"firearm_details": form.cleaned_data})
+        edit_good(self.request, self.good_id, form.cleaned_data)
         return super().form_valid(form)
```
saruniitr, line 595: this can be moved to `GoodCommonMixin`

exporter/applications/views/goods/add_good_firearm.py
saruniitr: @kevincarrogan does this refactoring sound reasonable?

depsiatwal: I think this is good allows called it decide 
thought I imagine EDITGoods might be repeated but as Arun mentions when need be we can shift to mixin when over repeated

kevincarrogan (approved):

PR #562, opened 2022-04-06
LTD-2145: Add view to edit control list options from summary page

saruniitr:


exporter/applications/views/goods/add_good_firearm.py
```diff
@@ -587,32 +587,45 @@ def get_success_url(self):
         return reverse("applications:product_summary", kwargs=self.kwargs)
 
     def form_valid(self, form):
-        self.process_valid_form(form)
+        edit_good(self.request, self.good_id, form.cleaned_data)
```
saruniitr, line 590: moved it here otherwise `process_valid_form` gets repeated in each good attribute edit view

exporter/applications/views/goods/add_good_firearm.py
```diff
@@ -587,32 +587,45 @@ def get_success_url(self):
...
         return super().form_valid(form)
 
+
+class FirearmBaseEditView(BaseEditView):
```
saruniitr, line 594: similarly created this to avoid repetition in each firearm attribute edit view

exporter/applications/views/goods/add_good_firearm.py
kevincarrogan: Funnily enough I was just doing the same thing.

kevincarrogan (approved):

PR #564, opened 2022-04-06
Remove the rfd certificate if user marks it as invalid

kevincarrogan:
This updates the flow for when the exporter is asked to add and/or validate their RFD certificate.

In particular this is to handle the following:
  - When an exporter marks their firearm certificate as invalid we remove it from the organisation
  - When an exporter marks their firearm certificate as valid we also attach it as a supporting document
  - When an exporter adds a new firearm certificate we add it to the organisation and attach it as a supporting document
  - The system attached firearm certificate is not able to be deleted by the user

### DIFF
--- a/exporter/applications/templatetags/additional_documents.py
+++ b/exporter/applications/templatetags/additional_documents.py
@@ -1,0 +1,11 @@
+from django import template
+
+from exporter.core.constants import DocumentType
+
+
+register = template.Library()
+
+
+@register.filter(name="is_system_document")
+def is_system_document(document):
+    return document.get("document_type") in [DocumentType.RFD_CERTIFICATE]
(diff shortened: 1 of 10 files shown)


exporter/applications/templatetags/additional_documents.py
```diff
@@ -0,0 +1,11 @@
+from django import template
+
+from exporter.core.constants import DocumentType
+
+
+register = template.Library()
+
+
+@register.filter(name="is_system_document")
+def is_system_document(document):
```
saruniitr, line 10: is_organisation_document?

kevincarrogan: Kind of but not necessarily. I'm trying to signal that this is something that the system did and therefore isn't something that should be allowed to be deleted. It just _happens_ that this one is an organisation document.

saruniitr (approved):

PR #565, opened 2022-04-06
Refactor base views for editing firearm details

kevincarrogan:


depsiatwal (approved):
LGTM

exporter/applications/urls.py
```diff
@@ -38,12 +38,11 @@
     UpdateSerialNumbersView,
 )
 from exporter.applications.views.goods.add_good_firearm import (
-    FirearmProductSummary,
     FirearmEditCalibre,
-    EditNameView,
     FirearmEditCategory,
-    EditNameView,
-    EditControlListEntry,
```
saruniitr, line 46: as this is not only specific to firearms I didn't add the Firearm prefix but may be ok for consistency

PR #567, opened 2022-04-06
LTD-2149-change-replica-firearm

depsiatwal:


exporter/applications/views/goods/add_good_firearm.py
```diff
@@ -644,3 +644,14 @@ def get_initial(self):
...
+
+
+class FirearmEditReplica(BaseFirearmEditView):
+    form_class = FirearmReplicaForm
+
+    def get_initial(self):
+        firearm_details = self.good["firearm_details"]
+        return {
+            "is_replica": firearm_details["is_replica"],
+            "replica_description": firearm_details["replica_description"],
```
saruniitr, line 656: should we reset the description if the user answers no to this question?

kevincarrogan: There are plenty of other cases where we do exactly this (the API does this in some cases when you make a change to something and wipes out fields that no longer make sense).

It depends whether having the `description` and `is_replica` as `True` ends up putting something in a weird inconsistent state elsewhere.

It will also end up being kind of confusing to look at the data and try to understand why there appears to be an inconsistency between those two fields i.e. why has a user said "no" to `is_replica` but we've somehow captured the description like they said "yes".

I do kind of feel that the user changing the answer and us hiding the description from them is kind of an explicit action saying we no longer need the description.

depsiatwal: Done in clean form :-)

saruniitr: ty

saruniitr (approved):

PR #569, opened 2022-04-06
[LTD-2169] TAU Edit

alixedi:
This adds an edit form - linked from the assessed products on the home screen.

### DIFF
--- a/caseworker/tau/forms.py
+++ b/caseworker/tau/forms.py
@@ -5,11 +5,9 @@ from crispy_forms_gds.layout import Layout, Submit
 from caseworker.tau.widgets import GoodsMultipleSelect
 
 
-class TAUAssessmentForm(forms.Form):
+class TAUEditForm(forms.Form):
     """
-    This is replacing caseworker.cases.forms.review_goods.ExportControlCharacteristicsForm.
-
-    TODO: Delete ExportControlCharacteristicsForm after this goes live.
+    This is for editing product assessment.
     """
 
     MESSAGE_NO_CLC_MUTEX = "This is mutually exclusive with control list entries"
@@ -42,19 +40,11 @@ class TAUAssessmentForm(forms.Form):
         widget=forms.Textarea,
     )
 
-    def __init__(self, goods, control_list_entries_choices, *args, **kwargs):
+    def __init__(self, control_list_entries_choices, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.fields["control_list_entries"].choices = control_list_entries_choices
-        self.fields["goods"] = forms.MultipleChoiceField(
-            choices=goods.items(),
-            widget=GoodsMultipleSelect(),
-            label="Select a product to begin. Or you can select multiple products to give them the same assessment.",
-            error_messages={"required": "Select the products that you want to assess"},
-        )
         self.helper = FormHelper()
-        self.helper.form_tag = False
         self.helper.layout = Layout(
-            "goods",
             "control_list_entries",
             "does_not_have_control_list_entries",
             "report_summary",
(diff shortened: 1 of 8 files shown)


caseworker/tau/forms.py
```diff
@@ -65,9 +55,39 @@ def __init__(self, goods, control_list_entries_choices, *args, **kwargs):
...
+        )
+        self.helper = FormHelper()
+        self.helper.form_tag = False
+        self.helper.layout = Layout(
+            "goods",
+            "control_list_entries",
+            "does_not_have_control_list_entries",
+            "report_summary",
+            "comment",
+            Submit("submit", "Submit"),
```
wkeeling, line 92: Wondering if it's possible to reuse the fields from the super `Layout` - save having to specify them again?

```python
self.helper.layout = Layout(
    "goods",
    **super().helper.layout.fields,
)
```

alixedi: Good shout! Let me try that.

alixedi: Done.

wkeeling (approved):

caseworker/tau/views.py
```diff
@@ -79,3 +93,49 @@ def form_valid(self, form):
...
+        return TAUEditForm(
+            self.control_list_entries,
+            data=self.request.POST
+            or {
+                "control_list_entries": [cle["rating"] for cle in good["control_list_entries"]],
+                "does_not_have_control_list_entries": good["control_list_entries"] == [],
+                "report_summary": good["good"]["report_summary"],
+                "comment": good["comment"],
+            },
+        )
```
saruniitr, line 117: Instead of this is it not better to override `get_form_kwargs()` and handle it in Form? it will also be consistent with `TAUHome`

caseworker/tau/views.py
```diff
@@ -79,3 +93,49 @@ def form_valid(self, form):
...
+    def form_valid(self, form):
+        data = form.cleaned_data
+        # API does not accept `does_not_have_control_list_entries` but it does require `is_good_controlled`.
+        # `is_good_controlled`.has an explicit checkbox called "Is a licence required?" in
+        # ExportControlCharacteristicsForm. Going forwards, we want to deduce this like so -
+        is_good_controlled = not data.pop("does_not_have_control_list_entries")
+        payload = {
+            **form.cleaned_data,
+            "current_object": self.good_id,
+            "objects": [self.get_good()["good"]["id"]],
```
saruniitr, line 137: `get_good()` can return `None` then this breaks

unit_tests/caseworker/tau/test_edit.py
```diff
@@ -0,0 +1,84 @@
...
+    """
+    Tests the submission of a valid form only. More tests on the form itself are in test_forms.py
+    """
+    # Remove assessment from a good
+    good = data_standard_case["case"]["data"]["goods"][0]
+    good["is_good_controlled"] = None
+    good["control_list_entries"] = []
+    edit_good = data_standard_case["case"]["data"]["goods"][1]
+    edit_good["control_list_entries"] = [{"rating": "ML1"}, {"rating": "ML1a"}]
+    requests_mock.get(client._build_absolute_uri(f"/cases/{data_standard_case['case']['id']}"), json=data_standard_case)
```
saruniitr, line 51: probably you don't need this as `mock_case` is already included in `setup()`?

caseworker/tau/views.py
alixedi: Good spot! Fixed now.

caseworker/tau/views.py
```diff
@@ -79,3 +94,50 @@ def form_valid(self, form):
...
+            "report_summary": good["good"]["report_summary"],
+            "comment": good["comment"],
+        }
+        return form_kwargs
+
+    def get_good(self):
+        for good in self.case.goods:
+            if good["id"] == self.good_id:
+                return good
+        raise Http404
```
saruniitr, line 124: :+1:

alixedi: This was a particularly good spot b/c users bookmark things.

PR #571, opened 2022-04-07
Add view to support editing of product document

saruniitr:
## Change description

Add view to edit product document options i.e., 1. editing product document, 2. editing document sensitivity option, 3. editing document availability option.
The questions 2 and 3 trigger subsequent questions so they are implemented as separate mini-wizards.
The common bits are refactored into a Document specific class and also generic wizard common bits are moved to another class with the idea being this serves as common base for the remaining wizards.

Add unit tests to cover various combinations,
 - uploading new document deletes existing one
 - marking document as sensitive deletes any existing document
 - marking document not available deletes any existing document
 - marking document available but sensitive deletes any existing document

### DIFF
--- a/exporter/applications/views/goods/add_good_firearm.py
+++ b/exporter/applications/views/goods/add_good_firearm.py
@@ -56,6 +56,8 @@ from exporter.goods.forms.firearms import (
 from exporter.goods.services import (
     post_firearm,
     post_good_documents,
+    delete_good_document,
+    update_good_document_data,
     get_good,
     get_good_documents,
     edit_firearm,
@@ -878,3 +880,73 @@ class FirearmEditPvGrading(LoginRequiredMixin, BaseSessionWizardView):
     def done(self, form_list, form_dict, **kwargs):
         self.edit_firearm(form_dict)
         return redirect(reverse("applications:product_summary", kwargs=self.kwargs))
+
+
+class FirearmEditProductDocumentView(BaseGoodEditView):
+    form_class = FirearmDocumentUploadForm
+
+    @cached_property
+    def product_document(self):
+        is_document_available = self.good["is_document_available"]
+        is_document_sensitive = self.good["is_document_sensitive"]
+        if not is_document_available or is_document_sensitive:
+            return None
+
+        if not self.good["documents"]:
+            return None
+
+        # when creating new product we can only add one document but we save it as
+        # a list because from the product detail page user can add multiple documents
+        return self.good["documents"][0]
+
+    def get_form_kwargs(self):
+        kwargs = super().get_form_kwargs()
+        return {**kwargs, "good_id": self.good_id, "document": self.product_document}
+
+    def get_initial(self):
+        return {"description": self.product_document["description"] if self.product_document else ""}
+
+    def form_valid(self, form):
+        product_document = form.cleaned_data.pop("product_document", None)
+        description = form.cleaned_data.pop("description", "")
+        payload = {"description": description}
+        if product_document:
+            payload = {
+                **get_document_data(product_document),
+                **payload,
+            }
+            api_resp_data, status_code = post_good_documents(
+                request=self.request,
+                pk=self.good_id,
+                json=payload,
+            )
+            if status_code != HTTPStatus.CREATED:
+                raise ServiceError(
+                    status_code,
+                    api_resp_data,
+                    "Error product document when creating firearm - response was: %s - %s",
+                    "Unexpected error adding document to firearm",
+                )
+
+            # Delete existing document
+            api_resp_data, status_code = delete_good_document(self.request, self.good_id, self.product_document["id"])
+            if status_code != HTTPStatus.OK:
+                raise ServiceError(
+                    status_code,
+                    api_resp_data,
+                    "Error deleting the product document - response was: %s - %s",
+                    "Unexpected error deleting product document",
+                )
+        elif self.product_document["description"] != description:
+            api_resp_data, status_code = update_good_document_data(
+                self.request, self.good_id, self.product_document["id"], payload
+            )
+            if status_code != HTTPStatus.OK:
+                raise ServiceError(
+                    status_code,
+                    api_resp_data,
+                    "Error deleting the product document - response was: %s - %s",
+                    "Unexpected error deleting product document",
+                )
+
+        return redirect(self.get_success_url())
--- a/exporter/applications/views/goods/add_good_firearm.py
+++ b/exporter/applications/views/goods/add_good_firearm.py
@@ -889,7 +889,7 @@ class FirearmEditProductDocumentView(BaseGoodEditView):
     def product_document(self):
         is_document_available = self.good["is_document_available"]
         is_document_sensitive = self.good["is_document_sensitive"]
-        if not is_document_available or is_document_sensitive:
+        if not is_document_available or (is_document_available and is_document_sensitive):
             return None
 
         if not self.good["documents"]:
@@ -907,6 +907,7 @@ class FirearmEditProductDocumentView(BaseGoodEditView):
         return {"description": self.product_document["description"] if self.product_document else ""}
 
     def form_valid(self, form):
+        existing_product_document = self.product_document
         product_document = form.cleaned_data.pop("product_document", None)
         description = form.cleaned_data.pop("description", "")
         payload = {"description": description}
@@ -929,14 +930,17 @@ class FirearmEditProductDocumentView(BaseGoodEditView):
                 )
 
             # Delete existing document
-            api_resp_data, status_code = delete_good_document(self.request, self.good_id, self.product_document["id"])
-            if status_code != HTTPStatus.OK:
-                raise ServiceError(
-                    status_code,
-                    api_resp_data,
-                    "Error deleting the product document - response was: %s - %s",
-                    "Unexpected error deleting product document",
+            if existing_product_document:
+                api_resp_data, status_code = delete_good_document(
+                    self.request, self.good_id, self.product_document["id"]
                 )
+                if status_code != HTTPStatus.OK:
+                    raise ServiceError(
+                        status_code,
+                        api_resp_data,
+                        "Error deleting the product document - response was: %s - %s",
+                        "Unexpected error deleting product document",
+                    )
         elif self.product_document["description"] != description:
             api_resp_data, status_code = update_good_document_data(
                 self.request, self.good_id, self.product_document["id"], payload
--- a/exporter/applications/views/goods/add_good_firearm.py
+++ b/exporter/applications/views/goods/add_good_firearm.py
@@ -54,6 +54,7 @@ from exporter.goods.forms.firearms import (
     FirearmSection5Form,
 )
 from exporter.goods.services import (
+    edit_good,
     post_firearm,
     post_good_documents,
     delete_good_document,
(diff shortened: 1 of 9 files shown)


exporter/goods/forms/firearms.py
```diff
@@ -579,11 +579,38 @@ class Layout:
...
+        self.document = kwargs.pop("document", None)
+        if self.document:
+            self.product_document_download_url = reverse(
+                "goods:document",
+                kwargs={
+                    "pk": good_id,
+                    "file_pk": self.document["id"],
+                },
+            )
+            self.document_name = self.document["name"]
```
saruniitr, line 593: When user edits document we display empty file upload field, to avoid confusion to the user this will display the document currently upload by the user.
User can choose to continue without uploading or select a new document

exporter/goods/forms/firearms.py
```diff
@@ -579,11 +579,38 @@ class Layout:
...
+        super().__init__(*args, **kwargs)
+
     def get_layout_fields(self):
-        return (
-            "product_document",
-            "description",
-        )
+        layout_fields = ("product_document", "description")
+        if self.document:
+            self.fields["product_document"].required = False
```
saruniitr, line 600: If document already available user may choose not to upload new document hence this needs to be optional in this case

exporter/applications/views/goods/add_good_firearm.py
```diff
@@ -663,3 +665,313 @@ def get_initial(self):
...
+                    status_code,
+                    api_resp_data,
+                    "Error updating the product document description - response was: %s - %s",
+                    "Unexpected error updating product document description",
+                )
+
+        return redirect(self.get_success_url())
+
+
+class AddGoodWizardCommon:
```
saruniitr, line 740: Idea is to move all wizard related common bit here including the ones in `AddGoodFirearmSteps` but since that journey is still in progress it is not updated yet in this PR.

exporter/applications/views/goods/add_good_firearm.py
```diff
@@ -878,3 +880,317 @@ def edit_firearm(self, form_dict):
...
+    def get_form_kwargs(self):
+        kwargs = super().get_form_kwargs()
+        return {**kwargs, "good_id": self.good_id, "document": self.product_document}
+
+    def get_initial(self):
+        return {"description": self.product_document["description"] if self.product_document else ""}
+
+    def form_valid(self, form):
+        existing_product_document = self.product_document
+        product_document = form.cleaned_data.pop("product_document", None)
```
kevincarrogan, line 911: Do you really need to `pop` these? Is removing them required?

exporter/applications/views/goods/add_good_firearm.py
```diff
@@ -878,3 +880,317 @@ def edit_firearm(self, form_dict):
...
+        is_document_sensitive = self.good["is_document_sensitive"]
+        if not is_document_available or (is_document_available and is_document_sensitive):
+            return None
+
+        if not self.good["documents"]:
+            return None
+
+        # when creating new product we can only add one document but we save it as
+        # a list because from the product detail page user can add multiple documents
+        return self.good["documents"][0]
```
kevincarrogan, line 1032: This seems to be identical logic to `product_document` in `FirearmEditProductDocumentView`.

exporter/applications/views/goods/add_good_firearm.py
```diff
@@ -878,3 +880,317 @@ def edit_firearm(self, form_dict):
...
+                    "Unexpected error updating product document description",
+                )
+
+        return redirect(self.get_success_url())
+
+
+class AddGoodWizardCommon:
+    @cached_property
+    def application_id(self):
+        return str(self.kwargs["pk"])
```
kevincarrogan, line 962: We don't ever seem to check that this is a valid application id. We should probably do that and do something sensible if it's not.

unit_tests/exporter/applications/views/test_edit_firearm.py
```diff
@@ -1,37 +1,143 @@
...
+        def save(self, name, content, max_length=None):
+            return name
+
+        def open(self, name, mode="rb"):
+            return None
+
+        def delete(self, name):
+            pass
+
+    with override_settings(FEATURE_FLAG_ONLY_ALLOW_FIREARMS_PRODUCTS=False), patch(
```
kevincarrogan, line 36: Do we actually need to override this setting?

exporter/core/templates/goods/forms/firearms/product_document_download_link.html
```diff
@@ -0,0 +1 @@
+<a class="govuk-link" href="{{ url }}">{{ name }}</a>
```
kevincarrogan, line 1: I think we usually use the pattern where if this document isn't marked as `safe` then we don't allow this to be downloadable. Not sure if this is relevant for this document type.

exporter/applications/views/goods/add_good_firearm.py
```diff
@@ -878,3 +880,317 @@ def edit_firearm(self, form_dict):
...
+    def get_product_document_payload(self):
+        data = self.get_cleaned_data_for_step(AddGoodFirearmSteps.PRODUCT_DOCUMENT_UPLOAD)
+        document = data["product_document"]
+        payload = {
+            **get_document_data(document),
+            "description": data["description"],
+        }
+        return payload
+
+    def get_payload(self, form_dict):
```
depsiatwal, line 1064: looks familiar do we not already have this in helper / super class ?

exporter/applications/views/goods/add_good_firearm.py
```diff
@@ -878,3 +880,317 @@ def edit_firearm(self, form_dict):
...
+    def done(self, form_list, form_dict, **kwargs):
+        all_data = {k: v for form in form_list for k, v in form.cleaned_data.items()}
+        is_document_available = all_data.pop("is_document_available", None)
+        is_document_sensitive = all_data.pop("is_document_sensitive", None)
+
+        try:
+            self.edit_firearm(self.good_id, form_dict)
+
+            existing_product_document = self.product_document
+            if not is_document_available or (is_document_available and is_document_sensitive):
```
depsiatwal, line 1180: isn't this repeated above ? 
may be move to helper classes ?

depsiatwal (commented):
few small comments overall LGTM

exporter/applications/views/goods/add_good_firearm.py
saruniitr: Not really but also we are not submitting the cleaned_data in this case, I will just `get` required fields

exporter/applications/views/goods/add_good_firearm.py
saruniitr: ok you mean by fetching the application details from the api? we use application id in many places, it will result in many api requests.

exporter/applications/views/goods/add_good_firearm.py
saruniitr: yes but that one is a form and this is a mini-wizard, was not sure how best to combine these. Is it ok to just create a class and move there?

exporter/applications/views/goods/add_good_firearm.py
saruniitr: you are right, I was intending to move this to `AddGoodWizardCommon`

exporter/core/templates/goods/forms/firearms/product_document_download_link.html
saruniitr: correct but here we are editing an already uploaded document so I have assumed that is already marked as safe. But you are right, it may not be the case. I will change it to only show the name if it is not safe.

unit_tests/exporter/applications/views/test_edit_firearm.py
saruniitr: I will check again, this is copied from another file

exporter/applications/views/goods/add_good_firearm.py
```diff
@@ -878,3 +880,317 @@ def edit_firearm(self, form_dict):
...
+            api_resp_data, status_code = post_good_documents(
+                request=self.request,
+                pk=self.good_id,
+                json=payload,
+            )
+            if status_code != HTTPStatus.CREATED:
+                raise ServiceError(
+                    status_code,
+                    api_resp_data,
+                    "Error product document when creating firearm - response was: %s - %s",
```
wkeeling, line 928: Are these substitution params `%s` meant to be substituted with something?

wkeeling: Ah just seen `handle_service_error()` -- ignore me

exporter/applications/views/goods/add_good_firearm.py
```diff
@@ -878,3 +880,317 @@ def edit_firearm(self, form_dict):
...
+        api_resp_data, status_code = post_good_documents(
+            request=self.request,
+            pk=good_pk,
+            json=document_payload,
+        )
+        if status_code != HTTPStatus.CREATED:
+            raise ServiceError(
+                status_code,
+                api_resp_data,
+                "Error product document when creating firearm - response was: %s - %s",
```
wkeeling, line 1084: Should this message be `Error adding product document...`

exporter/applications/views/goods/add_good_firearm.py
depsiatwal: ok fair enough

exporter/applications/views/goods/add_good_firearm.py
kevincarrogan: We don't need to check every time we access the application id but it would be good that there is some kind of verification that the application id actually relates to a real application.

`AddGoodFirearm` already does something similar in the `dispatch` method.

exporter/applications/views/goods/add_good_firearm.py
kevincarrogan: This doesn't necessarily have to be a part of a class to be reusable this could just be a function that's used in the relevant classes that require it.

unit_tests/exporter/applications/views/test_edit_firearm.py
```diff
@@ -1,37 +1,143 @@
...
+            ),
+            AddGoodFirearmSteps.PRODUCT_DOCUMENT_SENSITIVITY: (
+                "firearm_edit_product_document_availability",
+                edit_product_availability_url,
+            ),
+            AddGoodFirearmSteps.PRODUCT_DOCUMENT_UPLOAD: (
+                "firearm_edit_product_document_availability",
+                edit_product_availability_url,
+            ),
+        },
```
kevincarrogan, line 130: At this point all we're really using this for is to get the URL to post to so I don't think it's that useful (the new `wizard_name` parameter is practically identical to the view name we're trying to get from this big config dictionary).

exporter/applications/views/goods/add_good_firearm.py
saruniitr: ok the same can be done here, didn't include initially because `dispatch` is also called but it is only for each request it is probably fine.

exporter/applications/views/goods/add_good_firearm.py
saruniitr: :+1:

exporter/goods/forms/firearms.py
```diff
@@ -579,11 +582,38 @@ class Layout:
         required=False,
     )
 
+    def __init__(self, *args, **kwargs):
+        good_id = kwargs.pop("good_id", None)
+        self.document = kwargs.pop("document", None)
```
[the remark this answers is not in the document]
saruniitr: :+1:

unit_tests/exporter/applications/views/test_edit_firearm.py
saruniitr: we are also getting the step_name when posting to a particular wizard. In this case most of the steps are common but they can be different.

kevincarrogan: Is that last comment suggesting that this is a good thing or a bad thing?

saruniitr: I think it is a good thing because this allows us to use `goto_step` in these tests where there are more mini-wizards possible when editing product attributes. Perhaps this can be improved to avoid repetition. I had a similar mechanism initially in my PR but when the pv_grading changes merged, I have updated mine to fit in this so this mechanism is not introduced in this PR.

saruniitr: > we might as well just be sending through firearm_edit_pv_grading in the first place
yes but we also need the step name. This setup is mainly required for document wizards rather than pv_grading.
Eg the step `AddGoodFirearmSteps.PRODUCT_DOCUMENT_SENSITIVITY` is common in two wizards which has different urls.

saruniitr: @kevincarrogan please ignore previous comment.
I knew the issue was the wizard prefix but It didn't occur to me to mock that. I have looked it again and able to get rid of that unnecessary setup by mocking the wizard prefix. I guess this is the solution you have in mind?

kevincarrogan (approved):

PR #573, opened 2022-04-08
Ltd 2193 firearm certificate summary

kevincarrogan:


exporter/applications/views/goods/add_good_firearm.py
```diff
@@ -374,13 +379,51 @@ def get_form_kwargs(self, step=None):
 
         return kwargs
 
+    def has_skipped_firearms_attach_step(self, form_dict, firearm_details, section_value, attach_step_name):
+        firearms_act_section = firearm_details["firearms_act_section"]
+        return firearms_act_section == section_value and attach_step_name not in form_dict
+
     def get_payload(self, form_dict):
```
depsiatwal, line 386: getting quite complex now 
wonder if at some point we might have to move to helper class may be think of more generic function.

exporter/applications/views/goods/add_good_firearm.py
```diff
@@ -374,13 +379,51 @@ def get_form_kwargs(self, step=None):
...
+                            )
+                            .date()
+                            .isoformat(),
+                        }
+                    )
+                    break
+
         return payload
 
     def has_organisation_rfd_certificate_data(self):
```
depsiatwal, line 429: has ? 
to me implies a boolean but this return a dict of items.

depsiatwal (commented):
just one small comment really other then that LGTM

exporter/applications/views/goods/add_good_firearm.py
kevincarrogan: It doesn't return a dictionary of items. It returns a boolean.

exporter/applications/views/goods/add_good_firearm.py
kevincarrogan: Are you meaning a different bit of code? This comment seems to be pointing to a very specific small function that we wouldn't be able to generalise.

depsiatwal (approved):
my bad the had_ function looked like it was returning a dict I must have read it wrong sure I made the same mistake before. 

payload function is the one i'm referring to more of an observation that it's getting complex now and be break out to helper class.

PR #577, opened 2022-04-11
TD-2147-change-grading 

depsiatwal:
i've moved out post_to_step/goto_step in attempt to reuse. 
if happy we could move to conftest and use across the other tests

### DIFF
--- a/exporter/applications/views/goods/add_good_firearm.py
+++ b/exporter/applications/views/goods/add_good_firearm.py
@@ -1,6 +1,7 @@
 import datetime
 import logging
 import requests
+from datetime import date
 
 from deepmerge import always_merger
 from http import HTTPStatus
@@ -27,7 +28,9 @@ from exporter.core.helpers import (
     get_rfd_certificate,
     has_firearm_act_document as _has_firearm_act_document,
     has_valid_rfd_certificate as has_valid_organisation_rfd_certificate,
+    str_to_bool,
 )
+
 from exporter.core.wizard.conditionals import C
 from exporter.core.wizard.views import BaseSessionWizardView
 from exporter.goods.forms.firearms import (
@@ -198,7 +201,7 @@ def get_pv_grading_payload(form):
 def get_pv_grading_good_payload(form):
     payload = form.cleaned_data.copy()
     payload["date_of_issue"] = payload["date_of_issue"].isoformat()
-    return payload
+    return {"pv_grading_details": payload}
 
 
 @firearm_details_payload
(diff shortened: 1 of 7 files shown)


unit_tests/exporter/applications/views/test_edit_firearm.py
```diff
@@ -1,12 +1,57 @@
...
+    return {
+        AddGoodFirearmSteps.PV_GRADING: (
+            "firearm_edit_pv_grading",
+            reverse("applications:firearm_edit_pv_grading", kwargs={"pk": application_id, "good_pk": good["id"]}),
+        ),
+        AddGoodFirearmSteps.PV_GRADING_DETAILS: (
+            "firearm_edit_pv_grading",
+            reverse("applications:firearm_edit_pv_grading", kwargs={"pk": application_id, "good_pk": good["id"]}),
+        ),
+    }
```
kevincarrogan, line 25: These are just the same values. Is there any real need to make it a look up?

kevincarrogan, line 25: These are just the same values. Is there any real need to make it a look up?

exporter/applications/views/goods/add_good_firearm.py
```diff
@@ -792,3 +795,90 @@ def get_initial(self):
...
+        }
+
+    def get_form_initial(self, step):
+        initial = {}
+        if step == AddGoodFirearmSteps.PV_GRADING:
+            initial = {
+                "is_pv_graded": str_to_bool(self.good["is_pv_graded"]["key"])
+                if self.good["is_pv_graded"].get("key")
+                else None
+            }
```
kevincarrogan, line 841: If you can get here can this value ever actually be none? Doesn't the fact that you're on this edit screen in the first place imply that there is a value for this?

unit_tests/exporter/applications/views/test_edit_firearm.py
kevincarrogan: Having read the comment for the pull request I'm presuming this is for you to potentially refactor later.

My thoughts are, either refactor or don't refactor, having this kind of weird halfway house is a bit confusing.

I don't think we can pull this out into `conftest` either but it may still be useful in just this particular file.

depsiatwal: ok if we don't pull into conftest then for now this is complete no ?
Other edit wizards will definitely require these functions.

kevincarrogan: I don’t know. It might make more sense to put other edit tests in a different file. There may be a different way of refactoring. 

We can keep it I just never like seeing code that’s for a hypothetical future thing. It’s premature optimisation.

kevincarrogan (approved):

PR #578, opened 2022-04-11
LTD-2189: Reduce sleep() calls in UI tests

wkeeling:


ui_tests/caseworker/step_defs/test_enforcement.py
```diff
@@ -37,7 +37,7 @@ def xml_file_downloaded(driver):
     i = 5
     EU_XML_PATH = "/tmp/enforcement_check.xml"
     while not os.path.exists(EU_XML_PATH) and i > 0:
-        time.sleep(1)
+        time.sleep(0.1)
```
saruniitr, line 40: File operations are usually slow may be 500ms may not be sufficient sometimes? although this is only checking if file exists or not so may be increase `i` to `10` atleast?

PR #579, opened 2022-04-12
Ltd 2150 change firearm dealer

kevincarrogan:
This has the start of the editing for RFD status but I ended up doing a lot of refactoring whilst still implementing this so I think it makes sense to pull this in before we end up in merge hell.

Highlights:
  - Extracts out common functionality into mixins and base classes e.g. getting application and good objects
  - Pulls out common logic for posting to services into helper classes (only section certificates at the moment)
  - Add a factory fixture for creating fixtures to post to wizard steps

### DIFF
--- a/unit_tests/exporter/conftest.py
+++ b/unit_tests/exporter/conftest.py
@@ -3,9 +3,14 @@ import pytest
 
 from dotenv import load_dotenv
 
+from urllib.parse import urljoin, urlparse
+
 from django.conf import settings
 from django.core.files.storage import Storage
 from django.test import Client
+from django.urls import resolve
+
+from formtools.wizard.views import normalize_name
 
 from conf import exporter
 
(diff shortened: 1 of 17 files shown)


unit_tests/exporter/conftest.py
```diff
@@ -209,3 +214,58 @@ def delete(self, name):
...
+@pytest.fixture
+def goto_step_factory(authorized_client):
+    def goto_step(url):
+        step_gotoer = WizardStepGoto(authorized_client, url)
+        return step_gotoer
+
+    return goto_step
+
+
+class WizardStepPost(WizardStepPoster):
```
depsiatwal, line 249: like all the new wizard stuff :-)

depsiatwal (approved):
LGTM wondering if you forwards we should append 
Form/View the the relevant classes ?

PR #580, opened 2022-04-12
LTD-2190: Support running tests in parallel

wkeeling:


tests_common/fixtures/core.py
```diff
@@ -20,14 +20,25 @@ def exporter_info(request, environment):
 
 
 @fixture(scope="session")
-def internal_info(request, environment):
+def internal_info(request, environment, worker_id):
+    # For parallel execution, ensure that a TEST_SSO_EMAIL_n and TEST_SSO_PASSWORD_n
+    # pair of environment variables exist for each worker. n is zero-based and is
+    # supplied by the -n option on the command line.
+    try:
+        worker_num = int(worker_id[2:])
```
saruniitr, line 28: Given that we only have max two accounts, is it worth checking `worker_num` is within that max number and limiting to it if that is not the case?

ui_tests/caseworker/step_defs/test_enforcement.py
```diff
@@ -50,7 +50,7 @@ def enforcement_file_content_check(party_type, tag, value):
 
     # get values of all party_types as the file can contain multiple entries
     nodes = root.findall(f".//STAKEHOLDER[SH_TYPE='{party_type}']")
-    party_values = set([child.text for node in nodes for child in node.getchildren() if child.tag == tag])
+    party_values = set([child.text for node in nodes for child in node if child.tag == tag])
```
saruniitr, line 53: does this still result in the expected behaviour or is it correcting existing code?

ui_tests/caseworker/step_defs/test_organisation.py
```diff
@@ -221,3 +224,13 @@ def organisation_warning(driver):
 def step_impl(driver, context):
     pass
     # assert CasePage(driver).is_flag_applied(context.flag_name), "Flag " + context.flag_name + " is not applied"
+
+
+@fixture(scope="function")
+def get_eori_number():
+    return "GB" + "".join(["{}".format(randint(0, 9)) for _ in range(12)])
```
saruniitr, line 231: something to consider in some cases as it can be `XI` also

tests_common/fixtures/core.py
wkeeling: Yeah good call, it would be safer to be explicit. On a related note, SRE have created us a third test SSO user - so now we can run across 3 workers.

ui_tests/caseworker/step_defs/test_enforcement.py
wkeeling: The behaviour is the same, but the `getchildren()` method is now deprecated in preference of iterating the `node` directly (which by default will iterate the children).

saruniitr (approved):

PR #581, opened 2022-04-12
Ltd 2150 initial refactoring

kevincarrogan:


exporter/applications/views/goods/add_good_firearm/views/edit.py
```diff
@@ -0,0 +1,545 @@
...
+
+
+class BaseEditView(LoginRequiredMixin, FormView):
+    template_name = "core/form.html"
+
+    @cached_property
+    def application_id(self):
+        return str(self.kwargs["pk"])
+
+    @cached_property
```
depsiatwal, line 72: is it worth having this Cached properties as a mixin as I notice it's used 4 times in this file ?.

exporter/applications/views/goods/add_good_firearm/views/edit.py
```diff
@@ -0,0 +1,545 @@
...
+                    status_code,
+                    api_resp_data,
+                    "Error updating the product document description - response was: %s - %s",
+                    "Unexpected error updating product document description",
+                )
+
+        return redirect(self.get_success_url())
+
+
+class AddGoodWizardCommon:
```
depsiatwal, line 322: is this better named as mixin to make it Django like  ?

exporter/applications/views/goods/add_good_firearm/views/edit.py
```diff
@@ -0,0 +1,545 @@
...
+        if status_code != HTTPStatus.OK:
+            raise ServiceError(
+                status_code,
+                api_resp_data,
+                "Error updating firearm - response was: %s - %s",
+                "Unexpected error updating firearm",
+            )
+
+
+class AddGoodDocumentWizardCommon(AddGoodWizardCommon):
```
depsiatwal, line 390: two commons ? 
personally I find it confusing but may be it's just me

depsiatwal (commented):
few commons also does it make more sense to move all the extra classes that don't contain views in the /core/ subfolder ?

exporter/applications/views/goods/add_good_firearm/views/edit.py
kevincarrogan: Yes. We will eventually do that, Arun has already done some of that but I think we are waiting until these views have stabilised.

kevincarrogan:
No. I don’t think it makes sense to move them to `core`. This code is only relevant (right now at least) to these views. Moving it to `core` moves the code further away from its place of use and infers it’s more common than it really is.

depsiatwal (approved):
LGTM

exporter/applications/views/goods/add_good_firearm/views/edit.py
saruniitr: these can be renamed with better names, currently these views are not fully stable.

PR #582, opened 2022-04-13
[LTD-2170] Regimes

alixedi:
Adds the `is_wassenaar` field that is chucked to the API which sets the corresponding flag. 

It is a boolean field for now but as we introduce more regimes, it may end up becoming a `MultipleChoiceField`. Didn't want to jump the gun though so keeping it simple for now.

P.S. Also fixed a small bug in the first commit.

### DIFF
--- a/caseworker/tau/templates/tau/home.html
+++ b/caseworker/tau/templates/tau/home.html
@@ -70,7 +70,7 @@
                         {% for good in assessed_goods %}
                         <tr class="govuk-table__row">
                             <td class="govuk-table__cell">{{ forloop.counter }}</td>
-                            <td class="govuk-table__cell">{{ good.name }}</td>
+                            <td class="govuk-table__cell">{{ good.good.name }}</td>
                             <td class="govuk-table__cell">{{ good.quantity|floatformat }}</td>
                             <td class="govuk-table__cell">{{ good | get_clc | join:',' }}</td>
                             <td class="govuk-table__cell">{{ good.is_good_controlled.value }}</td>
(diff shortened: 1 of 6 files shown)


caseworker/tau/templates/tau/home.html
```diff
@@ -70,7 +70,7 @@ <h1 class="govuk-heading-l">Product Assessment</h1>
                         {% for good in assessed_goods %}
                         <tr class="govuk-table__row">
                             <td class="govuk-table__cell">{{ forloop.counter }}</td>
-                            <td class="govuk-table__cell">{{ good.name }}</td>
+                            <td class="govuk-table__cell">{{ good.good.name }}</td>
```
saruniitr, line 73: It is confusing with respect to good on application vs good. As there are new views perhaps we should've used `good_on_application` and `good`

unit_tests/caseworker/tau/test_forms.py
```diff
@@ -52,6 +52,28 @@
...
+            {
+                "report_summary": "test",
+                "does_not_have_control_list_entries": False,
+                "control_list_entries": ["test-rating"],
+                "is_wassenaar": False,
+            },
+            True,
+            [],
+        ),
+        # Set is_wassenaar to False
```
saruniitr, line 66: comment differs from the value being set :)

saruniitr (approved):

caseworker/tau/templates/tau/home.html
alixedi: Yes, please. If you can do that that would be wonderful!

PR #585, opened 2022-04-13
Upgrade version of parcel to fix `async` library security advisory

kevincarrogan:
https://github.com/uktrade/lite-frontend/security/dependabot/19


depsiatwal (approved):
just tests seem to be failing

PR #587, opened 2022-04-13
LTD-1436: Add wizard to capture remaining product details

saruniitr:
Add forms to ask year of manufacture related questions


exporter/applications/views/goods/add_good_firearm/views/add.py
```diff
@@ -447,3 +453,108 @@ def done(self, form_list, form_dict, **kwargs):
...
+        all_data = {k: v for form in form_list for k, v in form.cleaned_data.items()}
+
+        good_on_application_payload["firearm_details"] = self.get_firearm_details_payload(all_data)
+
+        return good_on_application_payload
+
+    def post_firearm_to_application(self, form_list):
+        all_data = self.get_good_on_application_data(form_list)
+
+        api_resp_data, status_code = post_good_on_application(self.request, self.application["id"], all_data)
```
saruniitr, line 536: in its current state it fails because we need to provide other mandatory fields such as `number_of_items, serial_numbers etc`

exporter/applications/views/goods/add_good_firearm/views/summary.py
```diff
@@ -52,3 +52,7 @@ def get_context_data(self, **kwargs):
             "documents": documents,
             "organisation_documents": organisation_documents,
         }
+
+
+class FirearmProductSummary2(FirearmProductSummary):
```
saruniitr, line 57: We can probably use the existing `product-summary.html` template itself to cover the second summary screen but kept it separate till we complete this, then we can refactor it easily.
Also the edit links makes it messy. Clicking on editing links related to permanent product attributes takes to the corresponding forms but after that they current go to summary screen1 so we need a way to route them back to summary screen2 in this case.

exporter/applications/urls.py
```diff
@@ -49,7 +49,10 @@
     FirearmEditProductDocumentSensitivity,
     FirearmEditProductDocumentAvailability,
 )
-from exporter.applications.views.goods.add_good_firearm.views.summary import FirearmProductSummary
+from exporter.applications.views.goods.add_good_firearm.views.summary import (
+    FirearmProductSummary,
+    FirearmProductSummary2,
```
depsiatwal, line 54: Summary2 possible for better name ? 
FirearmProductSummary (extra/additional) ?

exporter/applications/views/goods/add_good_firearm/views/add.py
```diff
@@ -447,3 +453,108 @@ def done(self, form_list, form_dict, **kwargs):
...
+class AddGoodFirearmToApplication(LoginRequiredMixin, BaseSessionWizardView):
+    form_list = [
+        (AddGoodFirearmSteps.MADE_BEFORE_1938, FirearmMadeBefore1938Form),
+        (AddGoodFirearmSteps.YEAR_OF_MANUFACTURE, FirearmYearOfManufactureForm),
+    ]
+
+    condition_dict = {AddGoodFirearmSteps.YEAR_OF_MANUFACTURE: C(is_product_made_before_1938)}
+
+    @cached_property
+    def application_id(self):
```
depsiatwal, line 467: see this a few times now possible to shift to a common mixin/view ?

exporter/applications/views/goods/add_good_firearm/views/add.py
kevincarrogan: We could just put some hardcoded numbers in to get it so it doesn't fail/error so we can see the whole journey go through.

exporter/applications/urls.py
kevincarrogan: Yes. Lets come up with a better name for this (I think even better than extra/additional).

exporter/applications/views/goods/add_good_firearm/views/add.py
kevincarrogan: I'm going to do this as another piece of work so lets leave it like this for now.

exporter/applications/views/goods/add_good_firearm/views/add.py
```diff
@@ -447,3 +453,108 @@ def done(self, form_list, form_dict, **kwargs):
...
+
+    def get_success_url(self, pk, good_pk):
+        return reverse(
+            "applications:product_summary_2",
+            kwargs={"pk": pk, "good_pk": good_pk},
+        )
+
+    def get_firearm_details_payload(self, all_data):
+        payload = {}
+        firearm_details = self.good.get("firearm_details", {})
```
kevincarrogan, line 516: This shouldn't be a `.get`. Because of this journey we are guaranteed to have a firearm_details object so we should rely on it. Alternatively we should fail if we don't have this (in fact the next line of code would just `KeyError` if it did return `{}`).

depsiatwal (commented):
few small comments

exporter/applications/views/goods/add_good_firearm/views/constants.py
```diff
@@ -17,3 +17,5 @@ class AddGoodFirearmSteps:
     ATTACH_SHOTGUN_CERTIFICATE = "ATTACH_SHOTGUN_CERTIFICATE"
     ATTACH_SECTION_5_LETTER_OF_AUTHORITY = "ATTACH_SECTION_5_LETTER_OF_AUTHORITY"
     IS_COVERED_BY_SECTION_5 = "IS_COVERED_BY_SECTION_5"
+    MADE_BEFORE_1938 = "MADE_BEFORE_1938"
+    YEAR_OF_MANUFACTURE = "YEAR_OF_MANUFACTURE"
```
kevincarrogan, line 21: These steps are for a separate journey so we shouldn't just add to the previous set of steps.

exporter/goods/forms/firearms.py
```diff
@@ -831,3 +836,42 @@ def clean(self):
...
+    class Layout:
+        TITLE = "What year was it made?"
+
+    year_of_manufacture = forms.IntegerField(
+        label="For example, 2007",
+        error_messages={
+            "required": "Enter the year it was made",
+        },
+        validators=[
+            RelativeDeltaYearValidator("The year must be before 1938", year=1938),
```
kevincarrogan, line 872: This doesn't need to be this complicated. We're just checking if a number is smaller than another number, we don't need to do anything with relative dates here.

exporter/applications/views/goods/add_good_firearm/views/add.py
saruniitr: Yes I tried that but the functions within `post_good_on_application()` manipulate them again especially in this https://github.com/uktrade/lite-frontend/blob/master/exporter/applications/services.py#L138.
Changing that function may break existing behaviour.

exporter/applications/views/goods/add_good_firearm/views/add.py
```diff
@@ -447,3 +453,108 @@ def done(self, form_list, form_dict, **kwargs):
...
+        payload["type"] = firearm_details["type"]["key"]
+        payload["category"] = [category["key"] for category in firearm_details["category"]]
+
+        payload["is_made_before_1938"] = all_data.get("is_made_before_1938", None)
+        payload["year_of_manufacture"] = all_data.get("year_of_manufacture", None)
```
kevincarrogan, line 521: The default return of `.get` is `None`.

exporter/applications/urls.py
saruniitr: I agree it is not a better name but couldn't think of anything at that time, looking for suggestions.

unit_tests/exporter/applications/views/conftest.py
```diff
@@ -6,7 +6,7 @@
 def mock_application_get(requests_mock, data_standard_case):
     application = data_standard_case["case"]["data"]
     url = client._build_absolute_uri(f'/applications/{application["id"]}/')
-    yield requests_mock.get(url=url, json={})
+    yield requests_mock.get(url=url, json={**application})
```
kevincarrogan, line 9: Do we need to make a copy of the `application` data? Can the JSON just be `application`?

exporter/applications/views/goods/add_good_firearm/views/add.py
saruniitr: right, this was before I found `post_good_on_application` failing because of `number_of_items` and other fields so I tried to extract relevant bits but you are right, we should directly use the `firearm_details` from `good`.
I will update this.

exporter/applications/views/goods/add_good_firearm/views/add.py
kevincarrogan: Actually. That's an interesting point. This is the reason on the other wizard I wrote a whole new service function because of all the manipulation that function does (I don't really like that it does all that).

Could we possibly do the same again here?

That function does so much because it's trying to be generic but can we make another one that's just specific for firearms and simplify it?

exporter/applications/views/goods/add_good_firearm/views/constants.py
saruniitr: ok I thought we could somehow use this to route back to second summary screen if user tries to edit some of the initial attributes, I will update this.

exporter/goods/forms/firearms.py
saruniitr: yes but it is still a validation. Without this we have mark the field as optional and do it in `clean`

unit_tests/exporter/applications/views/conftest.py
saruniitr: ah it can be. The test only requires `id` but then I thought of adding all fields and it didn't occur to just assign it directly.

exporter/applications/views/goods/add_good_firearm/views/add.py
saruniitr: Are you referring to `get_payload()` in the other wizard?
If yes, then yes we need something similar here. As it takes sometime I thought of getting this out first (as it contains the setup for second summary screen). I will try to do that here.

exporter/goods/forms/firearms.py
kevincarrogan: That's not correct. Why would we need to mark it as optional to run additional validation?

But even so, it's more that this isn't really anything to do with being a relative date. We're just checking that the number supplied is less than the number 1938, we don't need to do anything using relative date. In fact, the relativedelta thing is just getting the year number back out again anyway, so you're passing in the number 1938, converting it to a relativedelta thing to just get the same number out again.

saruniitr: I see what you mean. I only used relative because we are comparing it specifically with 1938. This can be similar to `PastDateValidator` but with 1938. Anyway I am easy on this, I will do this in `clean` without validator.

kevincarrogan (approved):

PR #610, opened 2022-04-22
Ltd 1452 quantity and value

kevincarrogan:


exporter/applications/services.py
```diff
@@ -130,7 +130,12 @@ def get_application_goods_types(request, pk):
     return data.json().get("goods") if data.status_code == HTTPStatus.OK else None
 
 
-def post_firearm_good_on_application(request, pk, json):
+def post_firearm_good_on_application(request, pk, good_id, json):
```
depsiatwal, line 133: ahh nice I think this might break my wizard since I was doing some of this in the view.

depsiatwal (approved):
LGTM

exporter/applications/services.py
depsiatwal: ohh seen you took that out

PR #615, opened 2022-04-26
Ltd 1722 product section

kevincarrogan:


depsiatwal (approved):
LGTM

exporter/applications/views/goods/add_good_firearm/views/add.py
```diff
@@ -446,8 +449,10 @@ def handle_service_error(self, service_error):
 
     def done(self, form_list, form_dict, **kwargs):
         try:
-            self.post_firearm_to_application(form_dict)
+            good_on_application, _ = self.post_firearm_to_application(form_dict)
         except ServiceError as e:
             return self.handle_service_error(e)
 
-        return redirect(self.get_success_url(self.application["id"], self.good["id"]))
+        self.good_on_application = good_on_application["good"]
```
saruniitr, line 456: May be we need to clarify the naming, `good_on_application["good"]` is the underlying `Good` instance so calling that `self.good_on_application` confuses things I think.

saruniitr: Actually I see why it is like this, the serializer is sending the data with the key `good`, above makes sense.

kevincarrogan: Yeah. It's not great is it? Maybe I'll add a comment for this.

saruniitr: :+1:

PR #617, opened 2022-04-26
LTD-2137: Product details view

saruniitr:
## Change description

Update Product detail page to include new RFI questions and edit links.
This is the same as product summary page but part of a different journey.

The edit views, unit tests are all duplicated as we have to have distinct urls. The only change is that these Edit views use `organisation` instead of `application` to deal with certificate upload changes as we don't have access to `application` in product list. Other than that the code is the same from editing product fields from the product summary page.


exporter/goods/urls.py
```diff
@@ -48,10 +67,94 @@
...
+    path(
+        "<uuid:pk>/firearm/edit/firearms-act-1968/",
+        FirearmEditFirearmsAct1968.as_view(),
+        name="firearm_edit_firearms_act_1968",
+    ),
 ]
 
+if settings.FEATURE_FLAG_PRODUCT_2_0:
+    urlpatterns += [
+        path("<uuid:pk>/", FirearmProductDetails.as_view(), name="good"),
```
saruniitr, line 150: It is conditionally included because we want this to have the same `name` as the template that triggers the product detail page uses this.

depsiatwal (approved):
lgtm

PR #618, opened 2022-04-26
LTD-1453-serial-number 

depsiatwal:


exporter/applications/views/goods/add_good_firearm/views/conditionals.py
```diff
@@ -109,3 +109,10 @@ def is_product_made_before_1938(wizard):
 def is_onward_exported(wizard):
     is_onward_exported_data = wizard.get_cleaned_data_for_step(AddGoodFirearmToApplicationSteps.ONWARD_EXPORTED)
     return is_onward_exported_data.get("is_onward_exported", False)
+
+
+def is_serial_available(wizard):
```
kevincarrogan, line 114: I think this could be named slightly better. Maybe `is_serial_numbers_available`.

exporter/applications/views/goods/add_good_firearm/views/conditionals.py
```diff
@@ -109,3 +109,10 @@ def is_product_made_before_1938(wizard):
...
+    serial_numbers_available_data = wizard.get_cleaned_data_for_step(
+        AddGoodFirearmToApplicationSteps.SERIAL_IDENTIFICATION_MARKING
+    )
+    return serial_numbers_available_data.get("serial_numbers_available", False) == "AVAILABLE"
```
kevincarrogan, line 118: `.get` will return `None` by default so we don't really need to return `False`. Also you can use the constant from the form `SerialChoices` instead of the string `AVAILABLE`.

exporter/goods/forms/firearms.py
```diff
@@ -1037,3 +1038,99 @@ def get_layout_fields(self):
...
+            cleaned_data["no_identification_markings_details"] = ""
+
+        return cleaned_data
+
+
+class FirearmSerialNumbersForm(BaseFirearmForm):
+    class Layout:
+        TITLE = "Enter serial numbers or other identification markings"
+
+    serial_numbers = None
```
kevincarrogan, line 1107: This seems like an odd thing to do and a bit confusing, I don't think this is making the intention very clear of what's happening.

kevincarrogan (commented):
Could we get some tests added for the form validation?

The other forms are tested in https://github.com/uktrade/lite-frontend/blob/master/unit_tests/exporter/goods/forms/test_firearms.py

kevincarrogan:
It would also be good to see a test showing that the second question gets skipped depending on the value selected.

exporter/applications/views/goods/add_good_firearm/views/conditionals.py
depsiatwal: lol don't know what I was thinking here

kevincarrogan (approved):

PR #628, opened 2022-04-29
Remove section 1 and section 2 steps from add good firearm wizard

kevincarrogan:
This changes when we ask for the section 1 and section 2 documents to be uploaded (but not section 5 nor RFD).

The difference in the documents is that some pertain to the organisation or the person (section 5 and RFD) and some pertain to the actual tangible firearm itself (a thing with a unique serial number).

It's further complicated by the fact that a particular firearm will always fall under section 1 and section 2 but it will require a different certificate each time for that specific tangible firearm.

This means we now slightly change where we ask someone to upload their section 1 and section 2 documents.


exporter/applications/views/goods/add_good_firearm/views/edit.py
```diff
@@ -723,46 +699,24 @@ def get_edit_payload(self, form):
     def form_valid(self, form):
         response = super().form_valid(form)
 
-        FirearmActCertificateAction(
+        OrganisationFirearmActCertificateAction(
+            self.request,
             self.document_type,
-            self,
+            f"Letter of authority for '{self.good['name']}'",
```
depsiatwal, line 705: seems to be repeated a few times  was just thinking of maintainability 
would it be better to define in the action ?

depsiatwal (commented):
small comment other then that LGTM

depsiatwal (commented):
small comment other then that LGTM

exporter/applications/views/goods/add_good_firearm/views/edit.py
kevincarrogan: Changed.

depsiatwal (approved):
LGTM

PR #629, opened 2022-05-03
[LTD-2242] Styling fixes

alixedi:


caseworker/tau/templates/tau/good_detail.html
```diff
@@ -1,5 +1,5 @@
 {% load advice_tags tau_tags %}
-<details class="govuk-details" data-module="govuk-details">
+<details class="govuk-details" data-module="govuk-details" style="text-align: left; margin-bottom: 0px;">
```
saruniitr, line 2: probably better to use govuk margin classes, eg `govuk-!-margin-bottom-0`

caseworker/tau/templates/tau/goods_checkbox_select.html
```diff
@@ -1,4 +1,4 @@
-<div>
+<div style="margin-top: 20px">
```
wkeeling, line 1: I'm not very well up on govuk styling, but wonder if there's a way of doing this with a govuk class rather than a specific pixel value. Just thinking about rendering on devices with different screen sizes.

saruniitr: There are govuk classes eg `govuk-!-margin-*` so better to go with something that gives the required 20px margin

caseworker/tau/templates/tau/good_detail.html
alixedi: Good one. I also tried to use `govuk-!-text-align-left` but that didn't work.

wkeeling (approved):

PR #632, opened 2022-05-03
[LTD-2241] Line numbers

alixedi:


caseworker/tau/views.py
```diff
@@ -26,7 +26,10 @@ def queue_id(self):
 
     @cached_property
     def case(self):
-        return get_case(self.request, self.case_id)
+        case = get_case(self.request, self.case_id)
+        for (i, good) in enumerate(case.goods):
```
saruniitr, line 30: may be `start=1`?

caseworker/tau/templates/tau/good_detail.html
```diff
@@ -1,7 +1,7 @@
 {% load advice_tags tau_tags %}
 <details class="govuk-details govuk-!-margin-bottom-0" data-module="govuk-details" style="text-align: left;">
     <summary class="govuk-details__summary">
-        <span class="govuk-details__summary-text">{{ good.good.name }}</span>
+        <span class="govuk-details__summary-text">{{ good.line_number }}.  {{ good.good.name }}</span>
```
saruniitr, line 4: looks like there is more than one space after the dot unless that is the requirement

saruniitr (approved):
one small comment otherwise looks good to me

caseworker/tau/templates/tau/good_detail.html
alixedi: It looked better with 2 spaces lol

PR #635, opened 2022-05-04
[LTD-2238] Show good details along edit form

alixedi:
This does some (1) refactoring before (2) putting it to good use. I would recommend a commit-wise review.


wkeeling (approved):
LGTM

PR #638, opened 2022-05-04
Expand validation messages for date field

kevincarrogan:
Handles additional potential error messages that pertain to errors coming from the Python `datetime` module.

We will also full back to a user friendly message if there are any that we can't find an appropriate message for.


depsiatwal (approved):
LGTM

PR #646, opened 2022-05-05
WIP: Enable selenium tests to run against docker stack

PippoRaimondi:


ui_tests/caseworker/features/case_notes.feature
```diff
@@ -11,9 +11,7 @@ Feature: I want to add an internal note to a case and view notes
     When I go to application previously created
     And I click on the notes and timeline tab
     And I enter "case note" as the case note
-    And I click make visible to exporter
```
saruniitr, line 14: Any particular reason these steps are removed, I believe they are required for the exporter to see the note

ui_tests/caseworker/features/case_notes.feature
```diff
@@ -11,9 +11,7 @@ Feature: I want to add an internal note to a case and view notes
...
     And I click post note
-    And I click confirm on confirmation box
```
saruniitr, line 16: this is also required to dismiss the confirmation that shows up if the above is checked.

docker-compose.caseworker.yml
```diff
@@ -14,6 +14,8 @@ services:
       DIRECTORY_SSO_API_CLIENT_API_KEY: ${PW_DIRECTORY_SSO_API_CLIENT_API_KEY}
       AUTHBROKER_CLIENT_ID: ${PW_AUTHBROKER_CLIENT_ID}
       AUTHBROKER_CLIENT_SECRET: ${PW_AUTHBROKER_CLIENT_SECRET}
+      TEST_SSO_EMAIL: ${TEST_SSO_EMAIL}
```
sekharpanja, line 17: Should we be adding additional sso accounts here to run tests in parallel?

ui_tests/caseworker/features/case_notes.feature
sekharpanja: Worth double checking this

ui_tests/caseworker/features/organisation.feature
```diff
@@ -81,4 +81,4 @@ Feature: I want to add a company to LITE
 
     Examples:
     |eori_number    | uk_vat_number | primary_site              | phone_number   |
-    |GB205672212000 | 123456789     | HQ London United Kingdom  | +448922296634  |
+    |GB205672212000 | 123456789     | HQ London United Kingdom  | +441234567890  |
```
sekharpanja, line 84: Do we know why this phone number works and not the other?

sekharpanja (commented):
Looks good to me except few tiny comments inine.

docker-compose.caseworker.yml
PippoRaimondiDIT: No, we don't need these users for parallel runs anymore... I will explain in more details on today's meeting why this is not needed anymore...

ui_tests/caseworker/features/organisation.feature
PippoRaimondiDIT: apparently numbers starting with 8 is not a valid format in the UK?

ui_tests/caseworker/features/case_notes.feature
saruniitr: It is ok even if we don't spin up exporter stack but from caseworker point of view this should still work.
Checking the box "make it visible to exporter" again asks for confirmation when we click create note so we need step to dismiss the browser alert dialog as well.

saruniitr (approved):

PR #652, opened 2022-05-11
adding ui test for checking sanctions

sekharpanja:
LTD-2110 UI test to check sanctions

### DIFF
--- a/ui_tests/caseworker/features/sanctions.feature
+++ b/ui_tests/caseworker/features/sanctions.feature
@@ -1,5 +1,5 @@
-@all @internal @sactions
-Feature: I want to check sactions matches
+@all @internal @sanctions
+Feature: I want to check sanctions matches
   As a logged in government user
   I want to check that the sanction match is highlighted
   For a case is created with a name on it that has a sanction
--- a/ui_tests/caseworker/features/sanctions.feature
+++ b/ui_tests/caseworker/features/sanctions.feature
@@ -1,0 +1,16 @@
+@all @internal @sactions
+Feature: I want to check sactions matches
+  As a logged in government user
+  I want to check that the sanction match is highlighted
+  For a case is created with a name on it that has a sanction
+
+  @check_sanctions
+  Scenario: view a case note added by caseworker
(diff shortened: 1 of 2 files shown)


ui_tests/caseworker/features/sanctions.feature
```diff
@@ -0,0 +1,16 @@
+@all @internal @sactions
+Feature: I want to check sactions matches
+  As a logged in government user
+  I want to check that the sanction match is highlighted
+  For a case is created with a name on it that has a sanction
+
+  @check_sanctions
```
saruniitr, line 7: this need not be present

ui_tests/caseworker/features/sanctions.feature
```diff
@@ -0,0 +1,16 @@
...
+  Scenario: view a case note added by caseworker
```
saruniitr, line 8: please update this to describe the scenario

ui_tests/caseworker/step_defs/test_sanctions.py
```diff
@@ -0,0 +1,22 @@
...
+
+@given("For a case is created with a name on it that has a sanction")  # noqa
+def sactioned_case(driver, sso_sign_in):  # noqa
+    pass
+
+
+@then(parsers.parse('I should see that the sanction match is highlighted as "{end_user_name}"'))
+def check_saction_highlighted(driver, end_user_name):  # noqa
+    sanctions_table = driver.find_element(by=By.ID, value="table-sanction-matches")
+    name_element = sanctions_table.find_element(by=By.XPATH, value="//tbody/tr/td[4]")
```
saruniitr, line 21: don't remember exactly, are we only matching on name or name and address?
If address is also included then please add assertions for address as well.

sekharpanja: There is no specific requirement to check address.

saruniitr: but do we use address also to highlight an entry as sanctions match or just name?

saruniitr (approved):

PR #662, opened 2022-05-16
LTD-2313: Add missing details in Firearm accessory product details page

saruniitr:


caseworker/tau/templatetags/tau_tags.py
```diff
@@ -15,3 +18,11 @@ def get_destinations(case):
     """Get unique destinations for the case in a nice comma-separated string"""
     unique_destinations = {(dest.get("country") or {}).get("name") or "" for dest in case.destinations}
     return sorted(list(unique_destinations - {""}))
+
+
+@register.filter
+def format_date(val):
```
saruniitr, line 24: We have existing filter https://github.com/uktrade/lite-frontend/blob/master/core/builtins/custom_tags.py#L938 that we could've used but because `goods` is rendered `as_crispy_field()` that is causing issue with that filter.
The above filter returns a `datetime` which is causing issues hence created this filter which returns a `str` formatted date in the required format.

depsiatwal (approved):
LGTM just need to fix lint

PR #665, opened 2022-05-17
Ltd 2327 attach firearm bug fix

kevincarrogan:


exporter/goods/services.py
```diff
@@ -69,7 +69,18 @@ def post_firearm(request, json):
...
+
+
+def edit_firearm_for_attaching(request, pk, json):
+    """Edit the permanent attributes of a good for attaching.
+
+    This is distinct from a normal edit as the normal edit has additional checks
+    to ensure we can't edit a good that is already on a submitted application,
+    however, this is an explicit case where there may be some fields we do want
+    to edit regardless of its status.
+    """
```
saruniitr, line 82: just for my understanding, could you give an example of when this is required?

kevincarrogan: Product 1.0 goods are saved without a `category` when we upgrade we want to save the new value of the `category` to the product.

kevincarrogan: The related API changes https://github.com/uktrade/lite-api/pull/1002 should hopefully show you what I'm having to do.

saruniitr (approved):

PR #670, opened 2022-05-18
Skip tests that are affected by firearms 2.0 changes

kevincarrogan:
This just skips the tests that will fail for now because trying to write a test that works with the feature switch is going to be too complex

What we will do for now is temporarily skip the test but have a fixed test waiting in the wings, when we first deploy the tests will be skipped but we can then follow up with deploying the fixed test

### DIFF
--- a/ui_tests/conftest.py
+++ b/ui_tests/conftest.py
@@ -1,7 +1,17 @@
 import os
+import pytest
 
 import tests_common.tools.helpers as utils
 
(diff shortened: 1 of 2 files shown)


kevincarrogan:
Updating of the tests happening https://github.com/uktrade/lite-frontend/pull/678.

depsiatwal (approved):
lgtm

PR #672, opened 2022-05-18
Allow unsafe application documents to be retrieved

kevincarrogan:
There is a point in time where an uploaded document is being virus scanned asynchronously.

It's highly likely that the user has been presented with a summary screen whilst this asynchronous task is occurring and as such the document won't have been marked safe yet. In this case we still want to display the name of the file even though we can't provide the download link.

This allows us to explicitly retrieve unsafe documents as well as safe ones but makes it an explicit flag to maintain backwards compatibility with anything that was expecting the old behaviour of unsafe documents being already filtered out.

### DIFF
--- a/exporter/applications/services.py
+++ b/exporter/applications/services.py
@@ -305,8 +305,12 @@ def delete_additional_document(request, pk, doc_pk):
     return data.status_code
 
 
-def get_application_documents(request, pk, good_pk):
-    response = client.get(request, f"/applications/{pk}/goods/{good_pk}/documents/")
+def get_application_documents(request, pk, good_pk, include_unsafe=False):
+    params = ""
+    if include_unsafe:
+        params = "?include_unsafe=1"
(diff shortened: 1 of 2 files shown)


exporter/applications/services.py
```diff
@@ -305,8 +305,12 @@ def delete_additional_document(request, pk, doc_pk):
...
 
 
-def get_application_documents(request, pk, good_pk):
-    response = client.get(request, f"/applications/{pk}/goods/{good_pk}/documents/")
+def get_application_documents(request, pk, good_pk, include_unsafe=False):
+    params = ""
+    if include_unsafe:
+        params = "?include_unsafe=1"
+    url = f"/applications/{pk}/goods/{good_pk}/documents/{params}"
+    response = client.get(request, url)
```
saruniitr, line 313: makes senses. In templates we check `safe=True` to provide download link anyway so in this case `safe=False` so only the name is rendered instead of link.

saruniitr (approved):

PR #674, opened 2022-05-18
[LTD-2369] Add product assessment tab and remove TAU 1.0 links

alixedi:
1. Put TAU 2.0 in a tab
2. Remove "Review goods" button (which links to TAU 1.0) on the case details tab
3. Remove checkboxes because they are useless without the "Review goods" button

Update - scratch (3) because we have "Edit goods flags" button as well, I have 2xC with Jess and it seems we need to keep the checkboxes for now.


depsiatwal (approved):
wow not too bad to get it to display then :-) 
LGTM

PR #688, opened 2022-05-20
LTD-2237-disabled-evidence-file-upload 

depsiatwal:


alixedi:
If we are not sure about this any more, I am of the opinion to delete the code. It will stay in Git. We can revive it later if needed.

PR #693, opened 2022-05-23
Ltd 2235 add s3 check to healthcheck

kevincarrogan:


conf/base.py
```diff
@@ -179,6 +180,8 @@
 CLAM_AV_PASSWORD = env.str("CLAM_AV_PASSWORD", "")
 CLAM_AV_DOMAIN = env.str("CLAM_AV_DOMAIN", "")
 
+DEFAULT_FILE_STORAGE = "storages.backends.s3boto3.S3Boto3Storage"
```
saruniitr, line 183: just wondering is it possible to have a test for this?
If something goes wrong we get exceptions in Sentry?

saruniitr (approved):

PR #694, opened 2022-05-23
LTD-2353-Validation fix

seijihg:
https://user-images.githubusercontent.com/24960282/169872208-5964d7ac-7045-447a-baab-b1656afccc57.mov

### Workflow

- [x] Ticket exists in Jira https://uktrade.atlassian.net/browse/LTD-2353

### Reviewing help

### Housekeeping

- [x] Frontend assets have been re-compiled.

### Merging


caseworker/assets/javascripts/tau-prod-assesment-headline.js
```diff
@@ -97,6 +107,15 @@ const initTauAssesmentHeadline = () => {
     tauSecondColumn
   );
 
+  // Check for validation error report summary.
+  if (errorMessage) {
+    tauSecondColumn.classList.remove("tau__second-column--hide");
+    productsNumberChecks.number = arrayProducts.map(
```
kevincarrogan, line 113: Doing the map here is always just going to return the same full array. Did you mean `filter`?

caseworker/assets/javascripts/tau-exporter-suggestions.js
```diff
@@ -81,38 +81,41 @@ export const hideUnhideExporterCle = (product, cleList) => {
 };
 
 const initTauControlListEntry = () => {
-  const cleList = document.querySelectorAll(".control-list__list");
+  const cleList = document.querySelectorAll(".tau .control-list__list");
```
kevincarrogan, line 84: I think we could just check for the top level `tau` element and if it doesn’t exist just return from this function early.

caseworker/assets/javascripts/tau-prod-assesment-headline.js
```diff
@@ -74,6 +75,14 @@ const addSelectAllExpandAll = (
   });
 };
 
+const headlineString = (arrayProducts, productsNumberChecks) => {
+  return productsNumberChecks.number > 1
+    ? `Assessing ${productsNumberChecks.number} products`
+    : `Assessing ${
+        arrayProducts.find((product) => product.checked)?.dataset["productName"]
```
kevincarrogan, line 82: The `?` implies this could return null somehow and if it does this is going to return something unexpected.

caseworker/assets/javascripts/tau-prod-assesment-headline.js
seijihg: ah yea wow good catch :D

caseworker/assets/javascripts/tau-exporter-suggestions.js
seijihg: ah yes this is actually a great idea.

caseworker/assets/javascripts/tau-prod-assesment-headline.js
kevincarrogan: But this is just accepting that there's a bug.

Either admit that it won't happen ever and therefore don't code around it or at least try to do something sensible if it were to happen.

caseworker/assets/javascripts/tau-exporter-suggestions.js
```diff
@@ -90,29 +93,32 @@ const initTauControlListEntry = () => {
...
-  );
+  cleList &&
+    cleList.forEach((cle) =>
+      cle.addEventListener("click", (event) => {
+        event.preventDefault();
+        if (!doesNotHaveCleSentence.checked) {
+          createTokenFieldSetItem(event.target.innerText);
+        }
+      })
+    );
```
kevincarrogan, line 104: Now that you're escaping early I don't think you need to check for the existence of these elements.

If you still want to check then change it to an `if` statement instead of `&&`.

seijihg: thats true no need to check them :D

kevincarrogan (approved):

PR #698, opened 2022-05-24
LTD-2276-Authentication-logging

depsiatwal:


depsiatwal:
login - get_redirect_url (initial call to call out to authentication service)

logout- get_redirect_url ( I haven’t captured the user as I didn’t want the risk of key errors in case user isn’t session). Will implement when we remove the Feature flags

user_profile - Used every time we extract to get email from authentication service. 

authenticate_user: called to get lite user when we have a valid [gov.uk](http://gov.uk/) user. Here we log the user

handle_success: Once we have a valid lite user also log user profile. 

fetch_token: called every time we require a token to talk to service.

Introspect: has it’s current logging in place so nothing was added

kevincarrogan (approved):

PR #706, opened 2022-05-26
 LTD-2214-gov-uk-ux-changes 

depsiatwal:


exporter/core/forms.py
```diff
@@ -44,21 +45,27 @@ def select_your_organisation_form(organisations):
 def register_triage():
     from exporter.core.views import RegisterAnOrganisationTriage
 
+    landing_form = (
+        Form(
```
saruniitr, line 49: Are we doing this with `lite-forms` only?

saruniitr, line 49: I think we should rewrite this in `django-forms`

lite_content/lite_exporter_frontend/core.py
```diff
@@ -35,6 +36,31 @@ class StartPage:
     ]
 
 
+class StartPageGovUK:
+    TITLE = "Apply for an export license using LITE"
```
saruniitr, line 40: typo in `license`, should be `licence`
And also in other lines below (42, 44)

PR #709, opened 2022-05-30
Consolidate product summaries

kevincarrogan:


saruniitr (approved):
this is definitely an improvement on what we have and will help to have consistent summaries in different places

PR #713, opened 2022-06-01
LTD-2397-registration-individual-df-draft

depsiatwal:
This is Feature switched as we have decided to do this in phases. 

This is draft and won't yet be switched on. 
There's lots more to do 
Also I have copied some of the functions from exporter TODO we will need to copy some of the common stuff into core.


exporter/core/registration/forms.py
```diff
@@ -0,0 +1,237 @@
...
+        if value:
+            stripped_vat = re.sub(r"[^A-Z0-9]", "", value)
+            if len(stripped_vat) < Validation.UK_VAT_MIN_LENGTH:
+                self.add_error("vat_number", "Standard UK VAT numbers are 9 digits long")
+            elif len(stripped_vat) > Validation.UK_VAT_MAX_LENGTH:
+                self.add_error("vat_number", "Standard UK VAT numbers are 9 digits long")
+            elif not re.match(Validation.UK_VAT_VALIDATION_REGEX, stripped_vat):
+                self.add_error("vat_number", "Invalid UK VAT number")
+            return stripped_vat
+        return value
```
kevincarrogan, line 112: I think it might be a good idea to extract these as separate [validator functions](https://docs.djangoproject.com/en/4.0/ref/validators/#writing-validators) as you've already moved some of it out as constants so might as well go one step further and just make it a validator function

exporter/core/registration/forms.py
```diff
@@ -0,0 +1,237 @@
...
+        return [
+            Submit("submit", getattr(self.Layout, "SUBMIT_BUTTON", "Continue")),
+        ]
+
+
+class RegistrationTypeForm(BaseForm):
+    class Layout:
+        TITLE = "Commercial organisation or private individual"
+
+    type = forms.TypedChoiceField(
```
kevincarrogan, line 35: I think this should just be a `ChoiceField` instead of `TypedChoiceField` as there's no `coerce` function necessary here.

I've been using `TypedChoiceField` a lot for those returning boolean values.

exporter/core/registration/forms.py
```diff
@@ -0,0 +1,237 @@
...
+
+    def get_layout_fields(self):
+        return ("type",)
+
+
+class RegistrationUKBasedForm(BaseForm):
+    class Layout:
+        TITLE = "Where is your organisation based?"
+
+    location = forms.TypedChoiceField(
```
kevincarrogan, line 55: Should be `ChoiceField` (see above comment).

exporter/core/registration/forms.py
```diff
@@ -0,0 +1,237 @@
...
+    def clean(self):
+        if self.is_uk_based:
+            errors_to_remove = {"address", "country"}.intersection(set(self.errors.keys()))
+        else:
+            errors_to_remove = {"address_line_1", "address_line_2", "city", "region", "postcode"}.intersection(
+                set(self.errors.keys())
+            )
+
+        for field in errors_to_remove:
+            del self.errors[field]
```
kevincarrogan, line 201: I feel that maybe the logic to skip error fields that aren't in errors should just happen in the `for` loop instead of doing this set manipulation (it's not instantly obvious what this is doing).

exporter/core/registration/forms.py
```diff
@@ -0,0 +1,237 @@
...
+        return
+
+    def clean_phone_number(self):
+        value = self.cleaned_data["phone_number"]
+        try:
+            phone_number = phonenumbers.parse(value)
+            if not phonenumbers.is_valid_number(phone_number):
+                self.add_error("phone_number", "Invalid phone number")
```
kevincarrogan, line 209: Instead of doing `self.add_error` you can raise a `ValidationError` instead. I tend to just use `add_error` where it's not known what the field is (I'm fairly certain Django knows that because it's in a `clean_*` method that the validation error is specific to this field).

exporter/core/registration/forms.py
```diff
@@ -0,0 +1,237 @@
...
+
+    def clean_website(self):
+        value = self.cleaned_data["website"]
+        if value:
+            try:
+                validator = URLValidator()
+                validator(value)
+            except ValidationError:
+                self.add_error("website", "Enter a valid URL")
+        return value
```
kevincarrogan, line 222: This could just be a `URLField` instead of doing your own validation checking.

exporter/core/registration/services.py
```diff
@@ -0,0 +1,10 @@
+from core import client
+
+
+def register_organisation(request, json):
+    data = {
+        "user": {"email": request.session["email"]},
+    }
+    print(json)
```
kevincarrogan, line 8: There's debug here.

exporter/core/registration/views.py
```diff
@@ -0,0 +1,98 @@
...
+    BaseSessionWizardView,
+    LoginRequiredMixin,
+):
+    form_list = [
+        (RegistrationSteps.REGISTRATION_TYPE, RegistrationTypeForm),
+        (RegistrationSteps.UK_BASED, RegistrationUKBasedForm),
+        (RegistrationSteps.INDIVIDUAL_DETAILS, RegisterIndividualDetailsForm),
+        (RegistrationSteps.ADDRESS_DETAILS, RegisterAddressDetailsForm),
+    ]
+    condition_dict = {}
```
kevincarrogan, line 39: If there's no conditions then I don't think you need to add this attribute.

exporter/core/registration/views.py
```diff
@@ -0,0 +1,98 @@
...
+            self.request,
+            payload,
+        )
+
+    def done(self, form_list, form_dict, **kwargs):
+        try:
+            self.post_registration(form_dict)
+        except ServiceError as e:
+            return self.handle_service_error(e)
+        self.update_authenticate_exporter_user()
```
kevincarrogan, line 64: I think this should be in the `try` and that the method should also potentially raise a `ServiceError` if it gets the wrong response status (at the moment it doesn't do any `expect_status` check).

exporter/core/registration/views.py
```diff
@@ -0,0 +1,98 @@
...
+            service_error.response,
+            exc_info=True,
+        )
+        if settings.DEBUG:
+            raise service_error
+        return error_page(self.request, service_error.user_message)
+
+    @property
+    def is_uk_based(self):
+        return bool(self.get_cleaned_data_for_step(RegistrationSteps.UK_BASED)["location"] == "united_kingdom")
```
kevincarrogan, line 98: You shouldn't need to cast this via `bool` as it will return a boolean anyway.

unit_tests/exporter/core/Registration/test_forms.py
```diff
@@ -0,0 +1,138 @@
...
+def test_register_uk_address_details_form(data, valid, error):
+    form = forms.RegisterAddressDetailsForm(is_uk_based=True, data=data)
+    assert form.is_valid() == valid
+    if not valid:
+        assert form.errors == error
+
+
+def test_register_non_uk_address_details_form():
+    form = forms.RegisterAddressDetailsForm(is_uk_based=False, data={})
+    assert form.is_valid() is False
```
kevincarrogan, line 132: This can just be `assert not form.is_valid()`.

PR #715, opened 2022-06-06
 LTD-2395-store-user-external-id

depsiatwal:


exporter/core/views.py
```diff
@@ -172,16 +172,17 @@ def prettify_data(self, data):
...
                     "last_name": self.request.session["last_name"],
-                },
+                }
             },
         )
+        # attempt to update user
+        response, _ = authenticate_exporter_user(self.request, profile)
         self.request.session["user_token"] = response["token"]
         self.request.session["lite_api_user_id"] = response["lite_api_user_id"]
         return reverse("core:register_an_organisation_confirm") + "?animate=True"
```
kevincarrogan, line 188: I feel that `get_success_url` should be side-effect free and should really only be returning the URL and not much more. I wonder if some of this code would be better elsewhere.

It also kind of looks like this is duplicated in `form_valid` as well so why does this updating of session and user happen in both places?

depsiatwal: I agree and was thinking the same 
this updating of profile is a mind field, can we leave as is ? and hopefully the non-lite version is progress is cleaner

kevincarrogan (approved):

PR #721, opened 2022-06-08
LTD-2375 Join issues

alixedi:


caseworker/tau/templatetags/tau_tags.py
```diff
@@ -24,3 +24,11 @@ def format_date(val):
...
     date_obj = datetime.strptime(val, "%Y-%m-%d")
     return date_obj.strftime("%d %B %Y")
+
+
+@register.filter
+def humanise_list(_list):
+    if len(_list) < 2:
+        return "".join(_list)
+    last = _list.pop()
+    return f"{', '.join(_list)} & {last}"
```
saruniitr, line 34: As per the ticket I think we want `and` right instead of `&`?

alixedi: I thought they were the same but good spot!

PR #722, opened 2022-06-08
Removes headlines when all products are assessed

seijihg:
Without style adjustment:
Screenshot 2022-06-09 at 10 49 13

With style:
Screenshot 2022-06-08 at 17 07 17

### Workflow

- [x] Ticket exists in Jira https://uktrade.atlassian.net/browse/LTD-2349

### Reviewing help

### Housekeeping

- [x] Frontend assets have been re-compiled.

### Merging


unit_tests/caseworker/tau/test_home.py
```diff
@@ -118,7 +118,7 @@ def test_home_content(
 
     # Test elements of case info panel
     soup = BeautifulSoup(response.content, "html.parser")
-    assert soup.find(id="subtitle").text == "Assess 1 product going from Great Britain to Abu Dhabi and United Kingdom"
+    assert soup.find(id="subtitle").text == "Assess 1 product going from Great Britain to Abu Dhabi, United Kingdom"
```
saruniitr, line 121: this change is not related to this, I think if you pull `master` this will go away.

caseworker/assets/javascripts/tau-exporter-suggestions.js
```diff
@@ -114,7 +114,7 @@ export const addDeleteExporterCleSuggestions = (
 
   // See if product is checked if YES then create buttons.
   if (checked) {
-    globalCleMatchToItems[id].forEach((cle) => {
+    globalCleMatchToItems[id]?.forEach((cle) => {
```
saruniitr, line 117: This is also not related. From what I know this `(using ?)` is not good as this is used to supress null checks?
Any particular reason why this needs to be added as it was not required before?

seijihg: It is related to another thing maybe I should stop fixing not related issue to the ticket.

unit_tests/caseworker/tau/test_home.py
seijihg: Because I rebased the master thats why test started failing.

caseworker/tau/templates/tau/home.html
```diff
@@ -10,26 +10,21 @@
 
 <div class="govuk-width-container">
     <main class="govuk-main-wrapper tau">
-
-        <div class="govuk-grid-row">
+        <div class="govuk-grid-row tau__display--flex">
```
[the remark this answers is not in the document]
seijihg: It is still there? just added tau__display--flex

caseworker/assets/javascripts/tau-exporter-suggestions.js
saruniitr: yeah if it is not related to this ticket then separate PR is preferred.

I think a separate null check before this is better and return early if that is the case is better rather than this.

caseworker/tau/templates/tau/home.html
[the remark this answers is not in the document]
saruniitr: yes but is that really required?

unit_tests/caseworker/tau/test_home.py
saruniitr: No, that was added to fix another ticket. This shouldn't fail if you have latest master.

If there are multiple countries we add `add` between the last two countries.

saruniitr (approved):

PR #728, opened 2022-06-13
Ltd 2314 separate fields temporary properties

kevincarrogan:


depsiatwal (approved):
In-terms of code base can't find any obvious faults :-)
for newbee  quite complex (not in a bad way) as we sorting a complex issue which many projects don't handle very well. 

Would be nice to see how it works , perhaps a session or doc would help ?
i'll let other dev chip in as well

PR #749, opened 2022-06-20
Don't allow blank whitespace only values in serial number inputs

kevincarrogan:


unit_tests/exporter/goods/forms/test_firearms.py
```diff
@@ -904,3 +905,57 @@ def test_firearm_identificateion_markings_form(data, is_valid, errors):
...
+            {
+                "serial_numbers_0": "12345",
+            },
+            True,
+            {},
+        ),
+        (
+            {
+                "serial_numbers_0": "11111",
+                "serial_numbers_1": "22222",
```
saruniitr, line 943: could you add a value with leading/trailing spaces?

PR #751, opened 2022-06-20
LTD-2426-add-agent-to-org-2

depsiatwal:


exporter/organisation/members/users/conditional.py
```diff
@@ -0,0 +1,7 @@
+from .constants import AddUserSteps
+from exporter.core.constants import Roles
+
+
+def is_agent_role(wizard):
+    role_select = wizard.get_cleaned_data_for_step(AddUserSteps.SELECT_ROLE)
+    return role_select.get("role") == Roles.AGENT_USER_ROLE[0]
```
kevincarrogan, line 7: I'm wondering if a `namedtuple` for these roles instead of a plain tuple would make this more readable.

exporter/organisation/members/users/payloads.py
```diff
@@ -0,0 +1,13 @@
+from exporter.core.wizard.payloads import MergingPayloadBuilder
+from .constants import AddUserSteps
+
+
+def get_cleaned_data(form):
+    return form.cleaned_data
+
```
kevincarrogan, line 7: We've used this before. I think it could be extracted to a common area.

exporter/organisation/members/users/views.py
```diff
@@ -0,0 +1,64 @@
...
+        HTTPStatus.CREATED,
+        "Error adding user to organisation",
+        "Unexpected error adding user to organisation",
+    )
+    def post_add_user(self, form_dict):
+        payload = AddMemberPayloadBuilder().build(form_dict)
+        return post_users(self.request, payload)
+
+    def done(self, form_list, form_dict, **kwargs):
+        self.post_add_user(form_dict)
```
kevincarrogan, line 56: The called method`post_add_user` might return a `ServiceError` which you might want to catch and do somethign with there's a `handle_service_error` method that I've used in various other wizards.

In the long term we might even want to think about a middleware that can catch `ServiceError` exceptions and display a common errorpage.

unit_tests/exporter/conftest.py
```diff
@@ -147,6 +147,58 @@ def mock_authenticate_user_save(requests_mock, mock_exporter_user):
...
+                        "name": "United Kingdom",
+                        "type": "gov.uk Country",
+                        "is_eu": True,
+                    },
+                },
+            },
+        ]
+    }
+    requests_mock.get(url=url, json=data)
+    yield data
```
kevincarrogan, line 199: This doesn't really need to be `yield` it can just be `return`. `yield` is more for when you want to do cleanup after the test has run.

unit_tests/exporter/organisation/members/test_forms.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+def test_select_role_form(mock_sites):
+    data = {"email": "joe@bloggs.com", "sites": [mock_sites["sites"][0]["id"]]}
+    form = forms.AddUserForm(
+        data=data, request=None, role_id=Roles.ADMINISTRATOR_USER_ROLE[0], sites=mock_sites["sites"]
+    )
+    assert form.is_valid()
+    assert form.fields["sites"].choices[0][0] == mock_sites["sites"][0]["id"]
+    assert form.fields["sites"].choices[0].hint.split("\n")[5] == "    Islington<br />"
```
kevincarrogan, line 56: This assertion is kind of confusing. Why are we only check the 6th element?

exporter/organisation/members/users/forms.py
```diff
@@ -0,0 +1,108 @@
...
+        error_messages={
+            "required": "Please select at least one site",
+        },
+        label="Sites",
+        widget=forms.CheckboxSelectMultiple(),
+    )
+
+    def __init__(self, request, role_id, sites, *args, **kwargs):
+        self.request = request
+        role_name = [item[1] for item in Roles.IMMUTABLE_ROLES if item[0] == role_id]
```
kevincarrogan, line 75: As per my other comments we could just make the tuples in `Roles` `nametuple`s that would make all of this a bit more readable or, for this particular case, it could be `role_name = [name for id, name in Roles.IMMUTABLE_ROLES if id == role_id]`

exporter/organisation/members/users/forms.py
```diff
@@ -0,0 +1,108 @@
...
+        choices=(),
+        error_messages={
+            "required": "Please select at least one site",
+        },
+        label="Sites",
+        widget=forms.CheckboxSelectMultiple(),
+    )
+
+    def __init__(self, request, role_id, sites, *args, **kwargs):
+        self.request = request
```
kevincarrogan, line 74: I don't see `request` being used in this form so I don't think we need to pass it in.

kevincarrogan (approved):

PR #753, opened 2022-06-22
 LTD-2465-validate-email-new-user

depsiatwal:


exporter/organisation/members/users/forms.py
```diff
@@ -83,6 +85,14 @@ def format_address(self, address):
...
         return render_to_string("organisation/members/includes/site-address.html", {"site_address": site_address})
 
+    def clean_email(self):
+        email = self.cleaned_data["email"]
+        users = get_organisation_users(self.request, str(self.request.session["organisation"]), params={})
+        matched_users = list(filter(lambda d: d["email"] == self.cleaned_data["email"], users["results"]))
+        if matched_users:
+            raise forms.ValidationError("Enter an email address that is not registered to this organisation")
+        return email
+
```
alixedi, line 95: Usually, passing `request` into a form is a bad sign but I see what you are trying to do.

Would you consider writing a function in one of them `services.py` that gets all the emails for the `organisation`, passing them into the form instead of `request` and then validating against them?

Also, presume you are trying to check if an element in in a list? Why don't use `set` for this?

alixedi (changes requested):

exporter/organisation/members/users/forms.py
kevincarrogan: Don’t use set either. Use any.

exporter/core/services.py
```diff
@@ -177,6 +177,12 @@ def get_organisation_users(request, pk, params, convert_to_options=False):
     return response.json()
 
 
+def get_organisation_users_list(request, pk):
+    response = client.get(request, f"/organisations/{pk}/users/")
+    response.raise_for_status()
+    return response.json()["results"]
```
alixedi, line 183: A slight improvement that we can do here is to make this function return the set of emails?

Also perhaps rename it to `get_org_user_emails` to reflect the extra bits?

This would remove this particular logic from the view and might make testing even easier :)

alixedi (approved):
One minor comment that would elevate this to 🤌

alixedi:
Also, please clean up your commits before you hit merge @depsiatwal  🙏

PR #755, opened 2022-06-22
Ltd 2470 notes and timelines styling

kevincarrogan:


caseworker/activities/views.py
```diff
@@ -0,0 +1,33 @@
...
+        if not settings.FEATURE_FLAG_NOTES_TIMELINE_2_0:
+            raise Http404("Feature flag disabled")
+        return super().dispatch(*args, **kwargs)
+
+    @cached_property
+    def case_id(self):
+        return str(self.kwargs["pk"])
+
+    @cached_property
+    def case(self):
```
depsiatwal, line 24: don't we have mixin for some of this stuff 
probably not generic enough

caseworker/advice/templates/advice/view-advice.html
```diff
@@ -4,7 +4,7 @@
 {% load crispy_forms_tags %}
 
 {% block header_tabs %}
-	{% include "advice/case-tabs.html" with case=case queue=queue %}
+	{% include "includes/case-tabs.html" with case=case queue=queue selected_tab="advice" %}
```
depsiatwal, line 7: is this existing template do we need to feature flag here ?

caseworker/advice/templates/advice/view_countersign.html
```diff
@@ -4,7 +4,7 @@
 {% load crispy_forms_tags %}
 
 {% block header_tabs %}
-	{% include "advice/case-tabs.html" with case=case queue=queue %}
+	{% include "includes/case-tabs.html" with case=case queue=queue selected_tab="advice" %}
```
depsiatwal, line 7: same as above FF ?

depsiatwal (commented):
LGTM some minor comments which may not be applicable

caseworker/advice/templates/advice/view-advice.html
kevincarrogan: This is me just re-using and moving an include that already exists in multiple places. It doesn't require any feature flags.

caseworker/activities/views.py
kevincarrogan: I contemplated trying to extract this out as a mixin. There are others mixins that have this but they also include other functionality and I didn't want to have to rewrite a whole bunch of mixins and class hierarchies just for this.

depsiatwal (approved):
LGTM

PR #759, opened 2022-06-23
Display team names ("applicant" for exporters).

alixedi:


caseworker/templates/includes/audit-trail.html
```diff
@@ -4,10 +4,10 @@
 			<div id="case-activity-{{ audit.id }}" class="app-activity__item">
 				<p class="govuk-body">
 					{# user not guaranteed to have id e.g, in case of system user #}
-					{% if audit.user.type == 'exporter' or not audit.user.id %}
-						<span class="app-activity__item__user">{{ audit.user.first_name }} {{ audit.user.last_name }}</span>
+					{% if audit.user.type == 'exporter' %}
+						<span class="app-activity__item__user">Applicant: {{ audit.user.first_name }} {{ audit.user.last_name }}</span>
```
saruniitr, line 8: the text `Applicant:` and the user team name are required in bold as per designs.

alixedi: They are. I think because `activity__item__user` makes them.

saruniitr: :+1:

saruniitr: sorry slight issue, we only want "Applicant" and "user team" in bold but not their first and last names

alixedi: I've moved the `</span>` to before the user name now.

PR #766, opened 2022-06-27
Fix Jenkins UI tests job to run against a specified LITE API

sekharpanja:
Current automated tests that we run in parallel of manual regression tests are running against LITE API latest master image. What we want is to run a specified FE release against a specified API release, using Release Tags that we create during deployment.

This PR fixes that issue. Process is that when we create a Github Release Tag, GCR event is triggered that builds an image that is tagged with the release string. This image will be used during the setup of UI e2e tests.

image

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -9,12 +9,19 @@ parameters:
   api_git_tag:
     type: string
     default: latest
+  environment:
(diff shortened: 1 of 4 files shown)


ejayesh (approved):
great, lgtm.

PR #768, opened 2022-06-28
LTD-2482-feature-flag-2FA

depsiatwal:


kevincarrogan:
I'm wondering if, instead of having a flag for this, we just allow the parameters to be passed into the config.

So instead of:
`AUTHBROKER_LOW_SECURITY=True`

We have:
`AUTHBROKER_PARAMS=vtr=['Cl']`

depsiatwal:
> I'm wondering if, instead of having a flag for this, we just allow the parameters to be passed into the config.
> 
> So instead of: `AUTHBROKER_LOW_SECURITY=True`
> 
> We have: `AUTHBROKER_PARAMS=vtr=['Cl']`

I did think about this but I thought hardcoding is more secure and less prone to error. Thoughts ?

kevincarrogan:
> > I'm wondering if, instead of having a flag for this, we just allow the parameters to be passed into the config.
> > So instead of: `AUTHBROKER_LOW_SECURITY=True`
> > We have: `AUTHBROKER_PARAMS=vtr=['Cl']`
> 
> I did think about this but I thought hardcoding is more secure and less prone to error. Thoughts ?

You've convinced me.

kevincarrogan (approved):

PR #771, opened 2022-06-29
LTD-2494-tweak-agents-copy 

depsiatwal:


kevincarrogan:
> Shouldn't `get_layout_fields` return a list or a tuple? Does it definitely work if you just return a single layout item?

```TypeError: type object argument after * must be an iterable, not HTML``` - turns out the failing test answers that question.

depsiatwal:
> > Shouldn't `get_layout_fields` return a list or a tuple? Does it definitely work if you just return a single layout item?
> 
> `TypeError: type object argument after * must be an iterable, not HTML` - turns out the failing test answers that question.

opps by accident fixed now

kevincarrogan (approved):

PR #772, opened 2022-06-29
LTD-2493: Update Notes and Timeline with new design and options

saruniitr:
## Change description

This tab is redesigned to show notifications grouped by date.
Quick filtering options are available to filter the events by team.


tests_common/tools/helpers.py
```diff
@@ -162,6 +162,13 @@ def search_for_correct_date_regex_in_element(element):
     )
 
 
+def search_for_correct_date_heading_regex_in_element(element):
+    return search(
```
saruniitr, line 166: In new Notes and timeline we only show date (time is not included)
eg 27 June 2022

caseworker/cases/views/main.py
```diff
@@ -157,9 +157,8 @@ def get_open_application(self):
     def get_standard_application(self):
         self.tabs = self.get_tabs()
         self.tabs.insert(1, Tabs.LICENCES)
+        self.tabs.append(self.get_notes_and_timelines_new_tab())
```
kevincarrogan, line 160: Let’s rename this so it doesn’t say `new`.

kevincarrogan (approved):

caseworker/cases/views/main.py
saruniitr: :+1:

PR #774, opened 2022-06-29
[LTD-2447] Previous assessments

alixedi:


caseworker/tau/views.py
```diff
@@ -55,12 +54,11 @@ def organisation_documents(self):
...
     def goods(self):
         goods = []
-        precedents = get_recent_precedent(self.request, self.case)
+        precedents = get_first_precedents(self.request, self.case)
         for item in self.case.goods:
+            item_id = item["id"]
             # Populate precedents
-            if precedents[item["id"]]:
-                precedents[item["id"]]["queue"] = precedents[item["id"]]["queue"] or ALL_CASES_QUEUE_ID
-            item["precedent"] = precedents[item["id"]]
```
alixedi, line 63: I have moved this down the stack into the function.

caseworker/tau/services.py
```diff
@@ -16,21 +18,28 @@ def group_gonas_by_good(gonas):
...
+    for gona in case.goods:
+        results[gona["id"]] = get_first_cles_precedents(gona, good_precedents)
+    return results
+
+
+def get_first_cles_precedents(gona, good_precedents):
+    precedents = good_precedents[gona["good"]["id"]]
+    cle_precedents = {}
+    for precedent in precedents:
+        cles = ",".join(precedent["control_list_entries"])
```
currycoder, line 39: Should we do the formatting in the view or in the template?

caseworker/tau/templates/tau/good_detail.html
```diff
@@ -5,8 +5,10 @@
     </summary>
     <div class="govuk-details__text tau__details">
       {% include "tau/good_table.html" with good=good %}
-      {% if good.precedent %}
-      {% include "tau/precedent_table.html" with precedent=good.precedent %}
+      {% if good.precedents %}
```
currycoder, line 8: The if is redundant now, right?

caseworker/tau/views.py
```diff
@@ -55,12 +54,11 @@ def organisation_documents(self):
...
+            item["precedents"] = precedents[item_id]
```
currycoder, line 61: Won't this barf with a KeyError if no precedents exist?

unit_tests/caseworker/tau/test_services.py
```diff
@@ -0,0 +1,138 @@
+import pytest
```
currycoder, line 1: I don't see any tests for the product assessment page.  Do these exist already?

caseworker/tau/templates/tau/good_detail.html
alixedi: Yes, good catch 👏

unit_tests/caseworker/tau/test_services.py
alixedi: Yes.

caseworker/tau/services.py
alixedi: The intention in this particular line was not formatting. In fact, if you notice, we return `.values()` from this function so the keys are only being used for group-by CLEs.

caseworker/tau/views.py
alixedi: It would but it doesn't because of how the precedents are constructed atm. That said, maybe some defensive code here is called for 👍

caseworker/tau/services.py
currycoder: Ah of course - thanks

currycoder (approved):

PR #775, opened 2022-06-29
Ltd 2322 separate product details fields in caseworker product summary

currycoder:
This change achieves the following:
- Standardises the summary logic used for the caseworker "product detail" page for V2 products (firearms only).  This reuses summary logic from the core package.  The net effect is that fields called out in the jira ticket are separated as desired.
- Re-instates RFD fields for exporter product summaries as it appears that certificate upload/expiry/reference fields went missing at some point.

Where possible, common summary logic has been pushed up in to the core package such that caseworker/exporter summaries only override formatting of document fields (which require caseworker/exporter-specific URLs respectively).

Caseworker product details before:
image

Caseworker product details after:
image

Exporter product details before:
image

Exporter product details after:
image


caseworker/templates/goods/includes/serial_numbers.html
```diff
@@ -9,7 +9,7 @@
         </summary>
         <div class="govuk-details__text">
             {% for item in firearm_details.serial_numbers %}
-                {{ forloop.counter}}. {{ item }} {%if not forloop.last %} <br> {% endif %}
```
currycoder, line 12: Bizarrely, this needed trimming down since the additional whitespace was being trimmed for the caseworker app and not for the exporter app.  Not entirely sure why this was happening, but it was easier to squash this at the source than to fully dig in.

unit_tests/caseworker/cases/helpers/test_summaries.py
```diff
@@ -0,0 +1,39 @@
...
+            (
+                "product-document",
+                '<a class="govuk-link govuk-link--no-visited-state" '
+                f'href="/queues/{queue_pk}/cases/8fb76bed-fd45-4293-95b8-eda9468aa254/documents/6c48a2cc-1ed9-49a5-8ca7-df8af5fc2335/" '
+                'target="_blank">data_sheet.pdf</a>',
+                "Upload a document that shows what your product is designed to do",
+            ),
+        ),
+    )
+    assert set(product_summary) == set(expected_summary)
```
kevincarrogan, line 29: We lose testing the ordering of the fields by doing this so it might be worth thinking about another way where we can check the order as well.

unit_tests/caseworker/conftest.py
```diff
@@ -971,6 +971,22 @@ def mock_good_on_appplication_documents(requests_mock, mock_case, data_good_on_a
     yield requests_mock.get(url=re.compile(f"{url}.*"), json={"documents": []})
 
 
+@pytest.fixture
+def mock_firearm_good_on_application(requests_mock, mock_case, data_standard_case):
+    good_data = data_standard_case["case"]["data"]["goods"][0]
+    url = client._build_absolute_uri("/applications/good-on-application")
+    yield requests_mock.get(url=re.compile(f"{url}.*"), json=good_data)
```
kevincarrogan, line 978: I've started switching these to `return` if they're not actually doing any clean up after the `yield`.

unit_tests/core/summaries/test_reducers.py
```diff
@@ -53,13 +53,27 @@ def test_firearm_reducer(is_user_rfd, mocker):
...
+            "document": {},
+            "reference_code": "12345",
+            "expiry_date": "31 May 2025",
+        }
+        extra_result_values = (
+            ("rfd-certificate-document", organisation_documents["rfd-certificate"]),
+            ("rfd-certificate-reference-number", organisation_documents["rfd-certificate"]["reference_code"]),
+            ("rfd-certificate-date-of-expiry", organisation_documents["rfd-certificate"]["expiry_date"]),
+        )
+    result = firearm_reducer(good, is_user_rfd, organisation_documents)
```
kevincarrogan, line 69: I've been mocking the called reducers instead of testing their logic directly here so that it's slightly easier to understand what exactly is happening in each reducer.

At the very least I'd probably like to see the new `rfd_reducer` tested in isolation.

caseworker/cases/views/goods.py
```diff
@@ -231,6 +232,28 @@ def get_initial(self):
...
+            good_on_application,
+            good_on_application_documents,
+            self.kwargs["queue_pk"],
+        )
+        return product_summary + product_on_application_summary
+
+    def is_product_type(self, good_on_application, product_type):
+        try:
+            return good_on_application["firearm_details"]["type"]["key"] == product_type
+        except (KeyError, TypeError):
```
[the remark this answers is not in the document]
currycoder: I have seen an application *somewhere* while trying this out which had `good_on_application["firearm_details"]["type"] == None`.  It's probable that this was an old application where we have since tightened things up in the data layer but I would prefer to be defensive.

unit_tests/caseworker/cases/helpers/test_summaries.py
currycoder: In the end this cast to set is unnecessary since in python3.7 dict order is guaranteed to be in insertion order - so `merge_summaries` will work for cases where we are replacing a value like this.  I didn't end up using this helper for adding fields to a summary so I believe we are good.

unit_tests/core/summaries/test_reducers.py
currycoder: I've added in the test for `rfd_reducer` - good spot, thanks.

core/summaries/formatters.py
```diff
@@ -66,6 +69,8 @@ def identity(x):
 
 def comma_separated_list(item_formatter=identity):
     def _comma_separated_list(iterable):
+        if not iterable:
+            return ""
```
kevincarrogan, line 73: It seems that codecov is complaining about this not being tested. I think it would be fairly quick to add a test for this (the formatters are pretty well covered).

kevincarrogan (approved):

PR #780, opened 2022-07-01
[LTD-1990] Add aria-label tags to explain disabled content for screen readers

currycoder:
This change adds aria-label tags to call out disabled content as "unavailable" for screen readers.  As part of this work, some tests using beautiful soup needed updating to still work if the `class` attribute was missing from a tag.

### DIFF
--- a/caseworker/templates/case/tabs/user-advice.html
+++ b/caseworker/templates/case/tabs/user-advice.html
@@ -25,7 +25,7 @@
 					</button>
 				</div>
 		</div>
-		<div class="{% if not case.advice or not case.advice|filter_advice_by_team_id:current_user.team.id %}app-advice__disabled-buttons{% endif %}">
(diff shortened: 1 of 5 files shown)


seijihg (approved):

caseworker/templates/case/tabs/user-advice.html
```diff
@@ -25,7 +25,7 @@
 					</button>
 				</div>
 		</div>
-		<div class="{% if not case.advice or not case.advice|filter_advice_by_team_id:current_user.team.id %}app-advice__disabled-buttons{% endif %}">
+		<div {% if not case.advice or not case.advice|filter_advice_by_team_id:current_user.team.id %}class="app-advice__disabled-buttons" aria-label="content currently unavailable"{% endif %}>
```
saruniitr, line 28: The `C` in `content` is not capitalized here, it is in other places.

currycoder: Fixed!

PR #782, opened 2022-07-05
[LTD-2447] Previous assessments

alixedi:


kevincarrogan (commented):
Could you add some tests that show what happens when there are multiple control list entries for a good? It would especially be good to see one that checks the logic where you check that ["ML1", "ML1a"] and ["ML1a", "ML1"] are actually the same thing so you only use one as a precedent.

kevincarrogan (approved):

PR #783, opened 2022-07-05
LTD-2483: Remove I'm done button for LU users when case is on post circ queue

saruniitr:
## Change description

When adding post circulation notes LU users are clicking I'm done
inadvertently as it is available on that page. The problem with
doing this is that they are not finalising the case and also because
this is the last step and the case becomes queueless.

A long term solution is in the works but till then we are removing
this button for LU  when the case is on post-circulation queue.

If the case is on any other queue besides this one then there is no
issue in showing this button because it won't become queueless.


caseworker/cases/helpers/case.py
```diff
@@ -86,9 +87,20 @@ class CaseView(TemplateView):
...
+
+    def is_tau_user(self):
+        return self.caseworker["team"]["alias"] == TAU_ALIAS
+
+    def is_lu_user(self):
+        return self.caseworker["team"]["alias"] == LU_ALIAS
+
+    def is_only_on_post_circ_queue(self):
+        queue_alias = tuple(queue["alias"] for queue in self.case.queue_details if queue.get("alias"))
+        return self.is_lu_user() and queue_alias == ("LU_POST_CIRC_FINALISE",)
```
currycoder, line 103: Could these methods be useful beyond this particular view?  I wonder if they could be pushed up to be methods on the model instead?

unit_tests/caseworker/cases/views/test_case_details.py
```diff
@@ -0,0 +1,32 @@
...
+    mock_standard_case_on_post_circulation_queue,
+):
+    pass
+
+
+def test_case_details_im_done_lu_user(authorized_client, data_queue, data_standard_case):
+    url = reverse("cases:case", kwargs={"queue_pk": data_queue["id"], "pk": data_standard_case["case"]["id"]})
+    response = authorized_client.get(url)
+    assertTemplateUsed(response, "layouts/case.html")
+    assert response.context["hide_im_done"] == True
```
currycoder, line 32: Would it help to test that other context attributes are as expected?

saruniitr: I guess I can check few other attributes, doesn't harm.

caseworker/cases/helpers/case.py
saruniitr: No they are only useful in this view. Similar checks are done in `advice` app but I think we already included them in mixins specific to that app.

alixedi: Can we perhaps factor out the `LU_POST_CIRC_FINALISE` to a constant defined at the top of the file.

alixedi (approved):
Left a comment but otherwise, it looks good.

currycoder (approved):

PR #789, opened 2022-07-08
Handle the cases for product summaries when checking product type rai…

kevincarrogan:
…ses TypeError


currycoder (approved):
:raised_hands:

PR #800, opened 2022-07-19
Ltd 2547 product search results

kevincarrogan:
This takes the same search data as used in the proof of concept (which comes from the API) and presents it in the new view with the new design. The API side of the proof of concept has been considered production worthy so it's available for reusing here.

This is a small change to integrate the search results that uses the new styling and uses real data. There will be additional tickets and pull requests that will fill in the other required functionality but this lays the groundwork for people to be able to build that functionality on top.


caseworker/search/views.py
```diff
@@ -2,12 +2,28 @@
...
+class ProductSearchView(LoginRequiredMixin, TemplateView):
     template_name = "search/products.html"
 
-    def dispatch(self, *args, **kwargs):
+    def dispatch(self, request, *args, **kwargs):
         if not settings.FEATURE_FLAG_PRODUCT_SEARCH:
             raise Http404("No feature flag")
 
-        return super().dispatch(*args, **kwargs)
+        return super().dispatch(request, *args, **kwargs)
```
currycoder, line 17: Do you need to pick `request` out anymore?

caseworker/templates/search/products.html
```diff
@@ -12,4 +12,52 @@
...
+                            <th scope="col" class="govuk-table__header">
+                                Report summary
+                            </th>
+                            <th scope="col" class="govuk-table__header">
+                                Assessment notes
+                            </th>
+                        </tr>
+                    </thead>
+                    <tbody class="govuk-table__body">
+                        {% for good_on_application in good.inner_hits.hits %}
```
currycoder, line 49: `good.inner_hits.hits` - finding it a little difficult to understand the datastructure here.  What is returned by the search API?

currycoder (approved):
Just minor comments for my understanding

caseworker/search/views.py
kevincarrogan: Nope. That was a hangover from a previous change. I'll fix that up.

caseworker/templates/search/products.html
kevincarrogan: Yep. At the moment this is mostly me just copying over what is being returned from the API as it stands without making any changes but I think there will be an opportunity to tidy this up when we make all of the other changes.

To be completely honest, I couldn't actually fully work out exactly how `inner_hits` was working as it seems to be an [elasticsearch thing](https://www.elastic.co/guide/en/elasticsearch/reference/current/inner-hits.html).

The thing I _do_ know is that it's the `good_on_application` documents that are related to the `good`. So the search results are the `good` and the `inner_hits` are the related `good_on_application` documents.

PR #808, opened 2022-07-26
Remove raising potential exception coming from API when posting firearms on application

kevincarrogan:
There is an issue where we are receiving a 400 back from the API, however the details of the error message are getting swallowed up due to the 400 only raising a generic exception

In the actual usage of this function in the wizard there is already a decorator in place to handle checking of the status code and raising a more appropriate error and log providing us with more information as to what actually happened

### DIFF
--- a/exporter/applications/views/goods/add_good_firearm/views/add.py
+++ b/exporter/applications/views/goods/add_good_firearm/views/add.py
@@ -456,7 +456,7 @@ class AddGoodFirearmToApplication(
             service_error.response,
             exc_info=True,
         )
-        if settings.DEBUG:
+        if settings.DEBUG:  # pragma: no cover
             raise service_error
         return error_page(self.request, service_error.user_message)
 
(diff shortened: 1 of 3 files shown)


exporter/applications/views/goods/add_good_firearm/views/add.py
```diff
@@ -449,6 +449,17 @@ def get_context_data(self, form, **kwargs):
 
         return ctx
 
+    def handle_service_error(self, service_error):
```
currycoder, line 452: Am I missing something..?  Where is this called?

kevincarrogan: It was always being potentially called but it was missing from this class so I'm rectifying that.

https://github.com/uktrade/lite-frontend/blob/35e24860de210a9d7a55989aca623a66bd336dab/exporter/applications/views/goods/add_good_firearm/views/add.py#L486

currycoder: Understood - thanks!

currycoder (approved):
I'm OK with this, but I think codecov has some reservations.  I can't disagree that these lines are probably good to cover but I will leave it with you

PR #815, opened 2022-08-01
Ltd 2601 error when attaching an existing product

kevincarrogan:
This handles the case where a good was made prior to the low RFI changes and didn't have the relevant section 5 details attached to it.

This copies those details to the original good if required. Essentially it's upgrading these goods to contain information that wasn't necessary before.

### DIFF
--- a/exporter/applications/views/goods/add_good_firearm/views/payloads.py
+++ b/exporter/applications/views/goods/add_good_firearm/views/payloads.py
@@ -228,31 +228,26 @@ class FirearmsActPayloadBuilder:
         return firearms_act_section == section_value and attach_step_name not in form_dict
 
     def get_payload(self, form_dict):
-        if not self.firearm_details.get("is_covered_by_firearm_act_section_one_two_or_five") == "Yes":
+        if self.firearm_details.get("is_covered_by_firearm_act_section_one_two_or_five") != "Yes":
             return {}
 
-        for section_value, attach_step_name, document_type in (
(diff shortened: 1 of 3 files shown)


exporter/applications/views/goods/add_good_firearm/views/payloads.py
```diff
@@ -228,36 +230,67 @@ def has_skipped_firearms_attach_step(self, form_dict, firearm_details, section_v
...
+        self.application = application
+        self.good = good
+        self.firearm_details = good["firearm_details"]
+
+    def build(self, form_dict):
+        if self.firearm_details.get("is_covered_by_firearm_act_section_one_two_or_five") != "Yes":
+            return {}
+
+        if self.firearm_details.get("section_certificate_missing") is not None:
+            return {}
```
currycoder, line 269: Shouldn't these both be booleans?  I suppose firearms_details is just an established data structure which is hard to change now

currycoder (approved):

exporter/applications/views/goods/add_good_firearm/views/payloads.py
kevincarrogan: Yes, these probably should be booleans.

The API has been setup in a way to present formatted data back to the frontend, this is something we're attempting to move away from but there's a lot of legacy to unwind so that's not going to be possible in the short term (probably not even in the long term for everything).

PR #819, opened 2022-08-02
LTD-2458: Only allow specified users to change config rules

saruniitr:
## Change description

Config rules include Routing rules and flagging rules, we don't want
everyone (even if they are Admin) to change these rules so only
allow specified list of users.


depsiatwal (approved):
LGTM

unit_tests/caseworker/flags/test_views.py
```diff
@@ -12,10 +12,20 @@ def setup(
...
 
 
+def test_view_flagging_rules(authorized_client):
+
+    url = reverse("flags:flagging_rules")
+    response = authorized_client.get(url)
+    assert response.status_code == 200
+    assert response.context["can_change_config"] == False
+
+
```
kevincarrogan, line 28: Can you add the test that asserts the `can_change_config == True` case and then this is all good.

depsiatwal:
did we have any e2e tests around this

kevincarrogan (approved):

unit_tests/caseworker/flags/test_views.py
saruniitr: :+1:

PR #820, opened 2022-08-02
Remove MANAGE_PICKLIST permission check from picklist json view

kevincarrogan:
There was originally a permission check for all picklist related views that checked that a user had `MANAGE_PICKLIST` however this is just a readonly view of the picklist data so isn't just necessary for people who can add/edit those picklists.

### DIFF
--- a/caseworker/picklists/urls.py
+++ b/caseworker/picklists/urls.py
@@ -8,11 +8,13 @@ app_name = "picklists"
 
 urlpatterns = [
     path("", views.Picklists.as_view(), name="picklists"),
-    path(".json", views.PicklistsJson.as_view(), name="picklists_json"),
     path("<uuid:pk>/", views.ViewPicklistItem.as_view(), name="picklist_item"),
     path("add/", views.AddPicklistItem.as_view(), name="add"),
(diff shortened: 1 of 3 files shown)


caseworker/picklists/urls.py
```diff
@@ -8,11 +8,13 @@
...
     path("<uuid:pk>/", views.ViewPicklistItem.as_view(), name="picklist_item"),
     path("add/", views.AddPicklistItem.as_view(), name="add"),
     path("<uuid:pk>/edit/", views.EditPicklistItem.as_view(), name="edit"),
     path("<uuid:pk>/<str:status>/", views.ChangeStatusView.as_view(), name="change_status"),
 ]
+decorate_patterns_with_permission(urlpatterns, Permission.MANAGE_PICKLISTS)
 
-url_patterns = decorate_patterns_with_permission(urlpatterns, Permission.MANAGE_PICKLISTS)
+urlpatterns += [
+    path(".json", views.PicklistsJson.as_view(), name="picklists_json"),
```
currycoder, line 19: Seems like a bit of a bizarre URL, but that is out of scope for this change.

currycoder (approved):

PR #822, opened 2022-08-03
LTD-1696 Accessibility

seijihg:


caseworker/templates/layouts/case.html
```diff
@@ -120,10 +120,15 @@ <h1 id="heading-reference-code" class="app-case-header__heading">
...
 					<div class="lite-tabs {% for tab in tabs %}{% for tab in tab.children %}{% if tab.url == current_tab %}lite-tabs__container--fade{% endif %}{% endfor %}{% endfor %}">
+
 						{% for tab in tabs %}
 							{% if tab.url %}
 								<a id="{{ tab.id }}" href="{% if tab.has_template %}{% url 'cases:case' queue.id case.id tab.url %}{% else %}{% url tab.url queue.id case.id %}{% endif %}{{ CURRENT_PATH_ONLY_PARAMS }}" class="lite-tabs__tab {% if tab.url == current_tab %}lite-tabs__tab--selected{% endif %}">
-									{{ tab.name }} {% if tab.count %}<span class="lite-tabs__tab-notification">{{ tab.count }}</span>{% endif %}
+									{{ tab.name }} 
+									{% if tab.count %}
+										<span class="lite-tabs__tab-notification" aria-hidden="true">{{ tab.count }}</span>
+										<strong><span class="govuk-visually-hidden lite-tabs__tab-notification">(Requires attention {{ tab.count }} items)</span></strong>
```
kevincarrogan, line 130: You probably don't need the `<span>` in the `<strong>` you could just have the `<strong>` and apply `govuk-visually-hidden` to it. You also don't need to have the `lite-tabs__tab-notification` in the `govuk-visually-hidden` element.

seijihg: it's needed for JS to hide it when being clicked on. since its being selected via `lite-tabs__tab-notification` to hide so screen reader wont read it when its hidden with js.

kevincarrogan (approved):

PR #823, opened 2022-08-03
LTD-2592-add-name-non-firearms 

depsiatwal:
lot more work to be done here for example 
make all reusable components generic and refactor

This is first draft kind of framework


core/summaries/formatters.py
```diff
@@ -51,6 +51,17 @@
     "section-5-certificate-missing-reason": "Explain why you do not have a section 5 letter of authority",
 }
 
+COMPLETE_PRODUCT_LABELS = {
```
kevincarrogan, line 54: What does "COMPLETE_PRODUCT" actually mean, it seems like a confusing name for this?

exporter/applications/summaries.py
```diff
@@ -4,6 +4,8 @@
     document_formatter,
 )
 from core.summaries.summaries import firearm_product_summary as core_firearm_product_summary
+from core.summaries.summaries import complete_product_summary as core_complete_product_summary
```
kevincarrogan, line 7: You don't need the two separate `from` lines, you can do:

```
from core.summaries.summaries import (
    firearm_product_summary as core_firearm_product_summary,
    complete_product_summary as core_complete_product_summary,
)
```

exporter/applications/views/goods/add_good_complete_product/views/add.py
```diff
@@ -0,0 +1,111 @@
...
+            },
+        )
+        ctx["title"] = form.Layout.TITLE
+
+        return ctx
+
+    def get_payload(self, form_dict):
+        good_payload = AddGoodCompleteProductPayloadBuilder().build(form_dict)
+        payload = good_payload
+        return payload
```
kevincarrogan, line 72: You don't really need to reassign to a variable in between here.

exporter/goods/services.py
```diff
@@ -299,3 +299,13 @@ def get_good_document_sensitivity(request, pk):
...
+
+
+def post_complete_product(request, json):
+    json["item_category"] = PRODUCT_CATEGORY_PLATFORM
+    json["is_good_controlled"] = None
+    data = client.post(request, "/goods/", json)
+
+    if data.status_code == HTTPStatus.OK:
+        data.json().get("good"), data.status_code
+    return data.json(), data.status_code
```
kevincarrogan, line 311: In the view you're doing:

```
            good, _ = self.post_complete_product(form_dict)
            self.good = good["good"]
```
so why are you doing this `.get("good")`.

Is `HTTPStatus.OK` what you're expecting to get back from a `post`?

core/summaries/formatters.py
depsiatwal: renamed to PLATFORM which is what i'm calling this 'product' for now

exporter/applications/summaries.py
depsiatwal: done

exporter/applications/views/goods/add_good_complete_product/views/add.py
depsiatwal: simplified for now

exporter/goods/services.py
depsiatwal: legacy copy fixed up

kevincarrogan (approved):

PR #824, opened 2022-08-04
Ltd 2620 prevent multiple assessment submissions

kevincarrogan:
We are experiencing an intermittent error when caseworkers are submitting products for assessment. 

Our current theory is that, due to the time it takes to process on the backend, that this error is occurring due to the caseworker clicking the submit button again whilst we are still processing the previous request. 

This change explicitly disables the button once it has been clicked.


depsiatwal (approved):
lgtm

PR #827, opened 2022-08-08
Fix typo in include for verified control list template

kevincarrogan:


exporter/templates/applications/goods/preexisting.html
```diff
@@ -42,7 +42,7 @@ <h1 class="govuk-heading-l">{% block title %}{% lcs 'goods.AddPreexistingGoodToA
 								Firearms
 							</td>
 							<td class="govuk-table__cell" id="good-control_code">
-								{% include 'includes/verfied-control-list-entries.html' with control_list_entries=good.control_list_entries status=good.status%}
+								{% include 'includes/verified-control-list-entries.html' with control_list_entries=good.control_list_entries status=good.status%}
```
currycoder, line 45: Shouldn't something have failed loudly in a testcase for this?  Could we add a small one in that asserts that this page 200's?

currycoder (approved):

PR #831, opened 2022-08-10
Add edit name for platform

kevincarrogan:
This also includes some large refactoring to extract common functionality between firearms and platform


depsiatwal (approved):
approving for now

PR #837, opened 2022-08-11
Ltd 2664 summary one edit links product document

kevincarrogan:


depsiatwal (commented):
all looks sensible to me 
any reason why some lines aren't covered by unit tests ?

kevincarrogan:
> all looks sensible to me any reason why some lines aren't covered by unit tests ?

Because I'm refactoring and moving stuff around they are probably things that weren't covered from the previous wizards.

depsiatwal (approved):
code cov can be addressed separately as long as the gap isn't increasing

PR #839, opened 2022-08-12
LTD-2592-part-number 

depsiatwal:
image

slightly different design but will confirm with Collin when he returns


exporter/applications/views/goods/common/payloads.py
```diff
@@ -20,6 +20,19 @@ def get_pv_grading_details_payload(form):
     }
 
 
+def get_part_number_payload(form):
+    if form.cleaned_data["is_part_number"]:
```
saruniitr, line 24: if we provide spaces as part number then we still take the `if` path so unless API complains we will try to save it correct?

exporter/goods/forms/common.py
```diff
@@ -216,3 +216,55 @@ def get_layout_fields(self):
                 render_to_string("goods/forms/common/help_with_security_gradings.html"),
             ),
         )
+
+
+class ProductPartNumberForm(BaseForm):
+    class Layout:
+        TITLE = "Do you have the part number ?"
```
saruniitr, line 223: there is an extra space before `?`

exporter/goods/forms/common.py
```diff
@@ -216,3 +216,55 @@ def get_layout_fields(self):
...
+
+    is_part_number = forms.TypedChoiceField(
+        choices=(
+            (True, "Yes"),
+            (False, "No"),
+        ),
+        coerce=coerce_str_to_bool,
+        label="",
+        widget=forms.RadioSelect,
```
saruniitr, line 232: In the prototype links provided in ticket LTD-2592 there are no radio buttons on the form so just wondering are they not updated or not the correct links?

depsiatwal: correct please see my comment above we will confirm with collin when he returns.

saruniitr: :+1:

kevincarrogan (approved):

PR #866, opened 2022-08-30
 LTD-2794-material-substance-options 

depsiatwal:


exporter/applications/views/goods/goods.py
```diff
@@ -269,11 +271,37 @@ class NonFirearmCategory(LoginRequiredMixin, NonFirearmsFlagMixin, FormView):
 
     def form_valid(self, form):
         category = form.cleaned_data["no_firearm_category"]
+        redirect_url = reverse("applications:new_good_platform", kwargs={"pk": self.kwargs["pk"]})
+        if category.lower() == "software":
+            redirect_url = reverse("applications:new_good_software", kwargs={"pk": self.kwargs["pk"]})
+        elif category.lower() == "material_category":
+            redirect_url = reverse("applications:is_material_substance", kwargs={"pk": self.kwargs["pk"]})
```
kevincarrogan, line 278: We are in control of the form so why do we need to do this `category.lower()`? Is the category not in an enum or a constant anyway?

kevincarrogan: I think the enum you could use here is `NonFirearmCategoryChoices`.

exporter/goods/forms/goods.py
```diff
@@ -766,6 +769,33 @@ def get_layout_fields(self):
         return ("no_firearm_category",)
 
 
+class IsMaterialSubstanceCategoryForm(BaseForm):
+    class Layout:
+        TITLE = "Is it material or substance ?"
```
kevincarrogan, line 774: This should be "Is it a material or substance?".

unit_tests/exporter/goods/forms/test_non_firearms.py
```diff
@@ -37,19 +54,46 @@ def test_non_firearm_category_form(data, is_valid, errors):
...
+    ),
+)
+def test_non_firearm_category_form_ff_enabled(settings, disable_non_firearms, feature, choice):
+    setattr(settings, feature, True)
     form = NonFirearmCategoryForm()
     choice_values = [c.value for c in form.fields["no_firearm_category"].choices]
-    assert form.NonFirearmCategoryChoices.COMPONENTS.value in choice_values
-    assert form.NonFirearmCategoryChoices.SOFTWARE.value in choice_values
+    assert len(choice_values) == 1
+    assert choice_values[0] == choice
```
kevincarrogan, line 71: I think this could just be `assert choices_values == [choice]`

kevincarrogan (approved):

PR #868, opened 2022-08-30
split text into paragraphs

sekharpanja:
add missing line break in Onward export form


exporter/goods/forms/common.py
```diff
@@ -420,9 +420,9 @@ class Layout:
 
     def get_layout_fields(self):
         return (
-            HTML.p(
-                "Tell us if the item will be exported again, beyond its first destination."
-                "This includes when the product has been incorporated into another item."
+            HTML(
+                "<p>ell us if the item will be exported again, beyond its first destination.</p>"
+                "<p>This includes when the product has been incorporated into another item.</p>"
```
kevincarrogan, line 425: You're missing the `T` in `Tell`.

I would also probably do this as two `HTML.p` fields as the `p` method adds some css classes.

kevincarrogan (approved):

PR #879, opened 2022-09-05
Ltd 2799 product document edit

kevincarrogan:
This adds the product description question to the firearm "add product" journey as it wasn't there previously as well as adding it to the edit flow for product document availability on material and platform.


depsiatwal (approved):
lgtm

PR #884, opened 2022-09-06
Ltd 2805 platform tau screen

kevincarrogan:
This uses the Python based summaries for the TAU screens for non-firearm products.


depsiatwal (approved):
LGTM

PR #886, opened 2022-09-07
Raise suspicious operation when state doesn't exist or differs on aut…

kevincarrogan:
An exporter has informed us that they have had issues when logging in where the callback URL responds with a 500. 

We aren’t receiving anything in our Sentry logs showing this error so I’ve added some additional logging and additional checks around the state parameter that we receive back. 

Hopefully all of the above will give us a better understanding as to what is happening when this 500 error occurs.


depsiatwal (approved):
lgtm

PR #902, opened 2022-09-14
Update unit values to reflect updated values from API

kevincarrogan:
This includes updating tests to have mock values that are more inline with what the API would produce

This also includes the removal of ITG/intangible as a selection and the additional validation and form manipulation related to that value


depsiatwal (approved):
lgtm

depsiatwal (approved):
lgtm

PR #919, opened 2022-09-20
Update logic for displaying pluralise quantity

kevincarrogan:
The original function was catering for some eventualitites that we've now removed:

  - the quantity and unit are always on the good_on_application model
  - these values are never null
  - all units can have their ending `s` removed to make the singular


depsiatwal (commented):
lgtm through looks like we need more test coverage

PR #921, opened 2022-09-20
Regimes on TAU form

kevincarrogan:
Updates the regimes question in the TAU form to display MTCR.

This removes the existing Wassennar question as this will be replaced after this in the same way that MTCR has been done. No one appears to have used or be using Wassennar at the moment and the way it was implemented is different to now so removing it is actually a better interim solution than trying to feature switch it.


caseworker/assets/javascripts/tau-edit.js
```diff
@@ -16,7 +17,16 @@ const initAssessmentForm = () => {
     "#control_list_entries"
   );
 
-  const suggestionsEl = document.querySelector(".tau__cle-suggestions");
+  const suggestionsEl = document.createElement("div");
```
currycoder, line 20: El..?

caseworker/regimes/services.py
```diff
@@ -0,0 +1,7 @@
+from core import client
+
+
+def get_mtcr_entries(request):
+    response = client.get(request, "/static/regimes/mtcr/entries/")
```
currycoder, line 5: More of a long term thought, but I just realised that we have two separate API URLs for these entries. IMO they should have probably been a single URL with a filter parameter. It made me think; A) We don't really have established rest design guidelines and, B) The whole idea of our app needing to be an API + frontend is just flawed and adds an unnecessary level of indirection

caseworker/regimes/services.py
```diff
@@ -0,0 +1,7 @@
...
+
+    return response.json(), response.status_code
```
currycoder, line 7: I'm concerned that if this was a 4XX we would fall over later on where this is used as the usage ignores the status `entries, _ = get_mtcr_entries(self.request)`.  Is that the case or do we have some configuration where non 2XX responses result in an exception?

caseworker/tau/views.py
```diff
@@ -179,12 +193,18 @@ def form_valid(self, form):
...
                 "current_object": good["id"],
                 "objects": [good["good"]["id"]],
                 "is_good_controlled": is_good_controlled,
+                "regime_entries": get_regime_entries(data),
             }
 
+            # These are used to determine the `regime_entries` and aren't needed
+            # when sending to the backend
+            del payload["mtcr_entries"]
+            del payload["regimes"]
```
currycoder, line 206: I feel like it might be better to just pick out what's needed from `data` than using everything and unticking things later with `del`

caseworker/tau/views.py
```diff
@@ -253,20 +287,26 @@ def get_context_data(self, **kwargs):
...
             "objects": [good["good"]["id"]],
             "is_good_controlled": is_good_controlled,
+            "regime_entries": get_regime_entries(data),
         }
 
+        # These are used to determine the `regime_entries` and aren't needed
+        # when sending to the backend
+        del payload["mtcr_entries"]
+        del payload["regimes"]
+
```
currycoder, line 309: Duplication...?  Can it be factored out at all?

caseworker/assets/javascripts/tau-edit.js
```diff
@@ -25,6 +35,11 @@ const initAssessmentForm = () => {
     }
   );
 
+  const mtcrEntriesEl = document.querySelector("#mtcr_entries");
+  progressivelyEnhanceMultipleSelectField(mtcrEntriesEl, (option) => {
```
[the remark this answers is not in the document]
kevincarrogan: Yes. That is correct.

caseworker/assets/javascripts/tau-edit.js
kevincarrogan: Element.

caseworker/regimes/services.py
kevincarrogan: A. Yes. I have plans to change it to an eventually but just wanted to get it done quickly now. 

B. Yes. But there are other reasons it’s like this, but I agree that fundamentally they’re so tightly coupled that it’s really just one application that happens to be split in two.

caseworker/tau/views.py
kevincarrogan: There is a second pull request coming soon afterwards where I think I'm already doing this.

caseworker/regimes/services.py
kevincarrogan: I'll add this in. I do have something where I handle this.

depsiatwal (approved):
lgtm

caseworker/tau/views.py
kevincarrogan: I think I did it like this to make it more obvious what was happening here. It's so it's a lot clearer that I'm removing things.

PR #928, opened 2022-09-26
Ltd 2859 regimes UI wassenaar

kevincarrogan:


caseworker/tau/forms.py
```diff
@@ -51,11 +51,24 @@ class TAUEditForm(forms.Form):
...
+        ),
+        error_messages={
+            "required": "Add a regime, or select none",
+        },
         widget=forms.CheckboxSelectMultiple,
     )
 
+    wassenaar_entries = forms.ChoiceField(
+        label="Choose the highest applicable sensitivity level",
+        choices=[],  # set in __init__
```
currycoder, line 67: Doubt this will ever really become an issue but I usually prefer to use immutable iterables at the file level; tuple in this case

caseworker/tau/forms.py
```diff
@@ -115,10 +134,21 @@ def clean(self):
...
             regimes = cleaned_data.get("regimes", [])
-            is_mtcr_regime = "MTCR" in regimes
-            mtcr_entries = cleaned_data.get("mtcr_entries", [])
-            if is_mtcr_regime and not mtcr_entries:
-                self.add_error("mtcr_entries", "Type an entry for the Missile Technology Control Regime")
+
+            has_selected_none = "NONE" in regimes
+            only_selected_none = regimes == ["NONE"]
+            if has_selected_none and not only_selected_none:
+                self.add_error("regimes", "Add a regime, or select none")
```
currycoder, line 141: return early and save the indentation below?

caseworker/tau/views.py
```diff
@@ -143,10 +159,23 @@ def get_context_data(self, *args, **kwargs):
...
+    for key, entry_field in [
+        ("MTCR", "mtcr_entries"),
+        ("WASSENAAR", "wassenaar_entries"),
+    ]:
+        if key not in regimes:
+            continue
 
-    return form_cleaned_data["mtcr_entries"]
+        values = form_cleaned_data[entry_field]
+        if not isinstance(values, list):
```
currycoder, line 173: When might this occur?

currycoder (approved):

caseworker/tau/forms.py
kevincarrogan: I did contemplate that. I didn't want to trip someone up who then wanted to add something after.

This code actually originally raised ValidationErrors which was problematic as you might want to add more than one error so returning early felt like just introducing the same thing again.

caseworker/tau/views.py
kevincarrogan: `mtcr_entries` is a multichoice and `wassenaar_entries` is a single choice. I'm going to rename `wassenaar_entries` in the refactor to more explicitly show that it's a single choice throughout the code.

caseworker/tau/forms.py
kevincarrogan: I agree. I chose consistency with other things that did the same, maybe I should just change them all (but now I fear the dreaded code coverage).

PR #929, opened 2022-09-26
 LTD-2857-approval-security-view-internal 

depsiatwal:


caseworker/advice/views.py
```diff
@@ -33,6 +35,13 @@ def denial_reasons_display(self):
         denial_reasons_data = get_denial_reasons(self.request)
         return {denial_reason["id"]: denial_reason["display_value"] for denial_reason in denial_reasons_data}
 
+    @cached_property
+    def security_approvals_classified_display(self):
+        security_approvals = self.case["data"].get("security_approvals")
+        if security_approvals:
+            security_approvals_dict = dict(SecurityClassifiedApprovalsType.choices)
+            return ", ".join([security_approvals_dict[approval] for approval in security_approvals])
+
```
kevincarrogan, line 44: Do you really want this to implicitly return `None` if there isn't any `security_approvals`?

caseworker/advice/views.py
```diff
@@ -33,6 +35,13 @@ def denial_reasons_display(self):
         denial_reasons_data = get_denial_reasons(self.request)
         return {denial_reason["id"]: denial_reason["display_value"] for denial_reason in denial_reasons_data}
 
+    @cached_property
```
kevincarrogan, line 38: This really doesn't need to be a `cached_property` you're not needing to cache anything that's expensive to compute.

kevincarrogan (approved):

PR #937, opened 2022-09-28
Save screenshot artifacts on UI test failures

kevincarrogan:
The tests are run within a Docker container so we need to save these out to the host system to get them uploaded


depsiatwal (approved):
lgtm

PR #941, opened 2022-09-30
Update component used for autocompletion of ARS entries.

kevincarrogan:
Update component used for autocompletion of ARS entries.

This moves to the accessible autocomplete https://github.com/alphagov/accessible-autocomplete as the previous component had accessibility issues.

This also tidies up some of the previous code that was initialising this component.


depsiatwal (approved):
lgtm

PR #944, opened 2022-10-04
Refactor out product document upload into an action

kevincarrogan:
We have code duplicated across multiple product wizards that handles logic for uploading the product document.

Through creating the wizards we have moved away from having individual chunks of logic in each wizard and moved over to creating `Action` classes that groups together the specific logic for a certain step or steps, however the product document code was never ported over to using an `Action` object. This pull request now rectifies this.


depsiatwal (approved):
was dying to move this to common function :-)

PR #948, opened 2022-10-06
Remove feature flag to only allow firearm products

kevincarrogan:
This removes the feature flag that pertains to whether we were just showing firearm questions or we would also show non-firearm questions in the product creation wizard.

We are now handling the non-firearm questions in a separate flow and so this old feature flag and code is redundant.


depsiatwal (approved):
lgtm

depsiatwal (approved):
LGTM

PR #949, opened 2022-10-07
Edit view refactoring

kevincarrogan:
This is the first part of probably a few PRs to refactor out how we handle wizards with a focus on consolidating functionality that's commonly reused into a new concept of a `Step`.

This first PR transitions three such forms over across all of the product types.

These three forms were picked as they are simple enough but also require `initial_data` changes, custom payload changes and `form_kwargs` changes which are the key things that we are trying to consolidate for reuse.

### DIFF
--- a/exporter/applications/views/goods/common/edit.py
+++ b/exporter/applications/views/goods/common/edit.py
@@ -18,7 +18,6 @@ from exporter.goods.forms.common import (
     ProductDocumentAvailabilityForm,
     ProductDocumentSensitivityForm,
     ProductDocumentUploadForm,
-    ProductNameForm,
     ProductPartNumberForm,
 )
 from exporter.goods.services import (
@@ -33,10 +32,7 @@ from .conditionals import (
 )
 from . import constants
 from .helpers import get_product_document
-from .initial import (
-    get_control_list_entry_initial_data,
-    get_name_initial_data,
-)
+from .initial import get_control_list_entry_initial_data
 from .mixins import (
     ApplicationMixin,
     GoodMixin,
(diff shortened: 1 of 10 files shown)


exporter/applications/views/goods/add_good_component/views/edit.py
```diff
@@ -90,8 +94,38 @@ def get_edit_payload(self, form):
...
+
+    def run(self, view, form):
+        self.edit_component(
+            view.request,
+            view.good["id"],
+            form,
+        )
+
+
+class ComponentEditName(
```
depsiatwal, line 120: wondering why do we even need a EditComponent
is it not more generic then that 

EditName could be used across all 'Products' 
I imagine the only change here is the success url

kevincarrogan: This is an action to edit the component which is going to be used across all the views for this particular set of views. Pushing an `EditName` makes it work for the `Name` part but isn't that helpful as that's used four times, this will be used way more times for this particular set of views.

There is a point where we'll make this overly generic and I think this gives us flexibility.

kevincarrogan: Having a single view or a single base view to handle all name editing is what I tried to do previously and it's confusing and inelegant.

depsiatwal: ahh ok I was hoping to remove writing views/urls/tests and basically masses of code to add another product. 

But see what your saying if you've been down this path then I trust your mistake hence LGTM

depsiatwal (approved):
nice I like it not sure how to test is it worth deploying to demo to sanity check ?

kevincarrogan:
> nice I like it not sure how to test is it worth deploying to demo to sanity check ?

I've tested this locally and we do actually have very good coverage of these in regards to the unit tests. I guess the thing that's missing here is the UI tests, but I'm not even certain we'd really want to add all of these paths.

I wouldn't necessarily put it onto `demo` to test, we can just test this as part of the normal UAT process but maybe get some manually testing of these parts if we want to be more confident that nothing has broken.

PR #950, opened 2022-10-07
Ltd 2819 ars UI

kevincarrogan:
This fixes two issues related to the TAU UI:

  - If the MTCR entries isn't displayed, because the feature flag is off, this was resulting in an error and no other JS was executed. This now correctly skips instantiating the JS component if the element doesn't exist.
  - The contrast for the hover state of the new ARS autocomplete wasn't high enough. This has now been solved by setting the hover font colour to white.


depsiatwal (approved):
lgtm

PR #951, opened 2022-10-07
Stop regimes being a required field on TAU form if feature flag is off

kevincarrogan:
The new regimes UI is behind a feature switch, however the TAU forms validation wasn't taking this into account and so when the new regimes field was missing it was still expecting a value to be posted and was causing a hidden error to occur.

This now changes the TAU forms to take this feature switch into account and correctly ignore the regimes field if it isn't meant to be present.

### DIFF
--- a/unit_tests/caseworker/tau/test_forms.py
+++ b/unit_tests/caseworker/tau/test_forms.py
@@ -207,6 +207,127 @@ def test_tau_assessment_form(data, valid, errors, rf):
 
 
 @pytest.mark.parametrize(
+    "data, valid, errors",
+    (
+        # Empty form
+        (
+            {},
+            False,
+            {
+                "does_not_have_control_list_entries": [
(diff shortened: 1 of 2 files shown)


unit_tests/caseworker/tau/test_forms.py
```diff
@@ -206,6 +206,112 @@ def test_tau_assessment_form(data, valid, errors, rf):
...
+                "does_not_have_control_list_entries": True,
+                "control_list_entries": ["test-rating"],
+            },
+            False,
+            {"does_not_have_control_list_entries": ["This is mutually exclusive with control list entries"]},
+        ),
+    ),
+)
+def test_tau_assessment_form_without_feature_flag(data, valid, errors, rf, settings):
+    settings.FEATURE_FLAG_REGIMES = False
```
currycoder, line 312: Is this being fleshed out properly later or is there something missing?

currycoder (approved):

PR #952, opened 2022-10-12
LTD-2928: Denial frontend changes

saruniitr:
## Change description

- Includes below UI changes,
     - rename Final destination to Country
     - Display entity type
     - remove consignee name

- Include Party country in Denial search critiera.
- Also fixes an issue where ultimate end users and third party details were not included in the criteria. This is because there can be multiple of these entities and we were using incorrect key.


caseworker/cases/views/main.py
```diff
@@ -613,13 +613,37 @@ def get_context_data(self, **kwargs):
...
         search = []
-        for key in self.request.GET.keys():
-            if key in case.data:
-                search.append(case.data[key]["name"])
-                search.append(case.data[key]["address"])
+        filter = {
+            "country": set(),
+        }
+        parties_to_search = []
+        for party_type in self.request.GET.keys():
```
saruniitr, line 620: request data is, `party_type` as the key and `party_id` as its value.
In case of ultimate end users and third parties, it is a list as there can be multiple of those parties

depsiatwal (approved):
lgtm

kevincarrogan (approved):

PR #953, opened 2022-10-12
Flip defaults for C5 feature flags for launch

kevincarrogan:


depsiatwal (commented):
covers all non-firearms and F680

only thing was some of the non-firearms were mean't to change as we do renaming. 
but that won't be possible in this release and is one for tech debt. perhaps we just remove then when we know it's stable

PR #954, opened 2022-10-13
disable-denials-uploads

depsiatwal:


caseworker/core/context_processors.py
```diff
@@ -90,7 +90,6 @@ def lite_menu(request):
                 or Permission.MANAGE_ALL_ROUTING_RULES.value in permissions,
                 {"title": "Routing rules", "url": reverse_lazy("routing_rules:list"), "icon": "menu/routing-rules"},
             ),
-            {"title": "Denial records", "url": reverse_lazy("external_data:denials-upload"), "icon": "menu/cases"},
```
saruniitr, line 93: we can still get to that page if we have the url saved

saruniitr: we need to disable view https://github.com/uktrade/lite-frontend/blob/master/caseworker/external_data/views.py#L18

saruniitr (approved):

PR #971, opened 2022-10-27
Update URLs for firearms pages

kevincarrogan:
This consolidates the firearms URLs into a single include for all the firearms specific URLs

This also update the root path of these URLs to make them more consistent


depsiatwal (approved):
lgtm

PR #978, opened 2022-11-02
Ltd 2974 end user name validation

MarkThomasLite:


exporter/applications/forms/validators.py
```diff
@@ -0,0 +1,10 @@
+from django.core.exceptions import ValidationError
+
+from exporter.applications.forms.constants import Validation
+
+
+def validate_name(value):
+    if value:
+        max_length = Validation.END_USER_NAME_MAX_LENGTH
+        if len(value) > max_length:
+            raise ValidationError(f'End user name should be {max_length} characters or less')
```
saruniitr, line 10: looks good to me, please add unit tests

MarkThomasLite: test for raised error has been added, negative scenario already covered

exporter/applications/forms/parties.py
```diff
@@ -210,7 +212,11 @@ def clean(self):
 
 class PartyNameForm(forms.Form):
     title = "End user name"
-    name = forms.CharField(label="", error_messages={"required": "Enter a name"})
+    name = forms.CharField(
+        label="",
+        error_messages={"required": "Enter a name"},
+        validators=[validate_name]
```
kevincarrogan, line 218: If we are sticking with using our own custom validator (although I think `MaxLengthValidator` would be better) then I think this could be named to indicate that it's just for this purpose `validate_name` is slightly too generic.

saruniitr: agree, less custom code is better.

kevincarrogan (approved):

exporter/applications/forms/constants.py
```diff
@@ -0,0 +1,2 @@
+class Validation:
```
saruniitr, line 1: this can be removed now

MarkThomasLite: ok will remove, I keep trying to follow patterns used in other parts of the app so kept the constants file

PR #980, opened 2022-11-04
LTD-2967-error-registration 

depsiatwal:


exporter/core/middleware.py
```diff
@@ -1,13 +1,46 @@
...
+        if request.path == reverse("auth:logout"):
+            return self.get_response(request)
+        if not request.path == reverse("core:register_an_organisation_confirm"):
+            # Check if the user is allowed to make API calls and logged in
+            if request.authbroker_client.token and request.session.get("user_token"):
+                if self.is_organisations_in_status(request, OrganisationStatus.REVIEW):
+                    # An application is still in review
+                    return redirect(reverse("core:register_an_organisation_confirm") + "?animate=True")
+        # If the user doesn't have a LITE account don't allow them to homepage since they need to register
+        if request.path == reverse("core:home"):
```
kevincarrogan, line 30: Why do we check just for the home page here? Doesn’t this logic actually hold true for all pages really?

exporter/templates/layouts/base.html
```diff
@@ -61,7 +61,6 @@
 							</span>
 						</a>
 					</div>
-					{% if not user_in_limbo %}
```
kevincarrogan, line 64: What was this `user_in_limbo` for?

exporter/core/middleware.py
depsiatwal: possibly since this can impact the whole site i'm trying to minimise to where the problem exists atm 
other parts of the site are handled by LoginrequiredMixin but home page isn't hence why we checking here. 

Homepage doesn't check and starts making API calls when the user's registration hasn't been confirmed resulting in 403 permission issue which logs 'HAWK ERROR'

exporter/templates/layouts/base.html
depsiatwal: user_in_limbo is when a user has registered but their organisation is in review 

it's hiding the logout button I think by showing the logout button it will allow the user to clean out their session when things go wrong. Since a user is logged into GOV.UK there is now way to start fresh until session timeouts without logout

This can potentially stop endless loops when things do go wrong.

unit_tests/conftest.py
```diff
@@ -2,10 +2,17 @@
...
 
+from django.conf import settings
+
 from core import client
 from core.constants import OrganisationDocumentType
 
 
+@pytest.fixture(autouse=True)
+def setup():
+    settings.HAWK_AUTHENTICATION_ENABLED = False
```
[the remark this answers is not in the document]
depsiatwal: Most of the tests only appear to work when HAWK_AUTHENTICATION_ENABLED=False
I've now set this to true in my config since I think it's good practice to have it on locally. 

Eventually we should default to true in our templates (pending DEV Meeting discussion)

unit_tests/exporter/core/test_middleware.py
```diff
@@ -63,3 +67,71 @@ def test_service_error_handler_service_error_with_debug(rf, mocker, caplog, sett
...
+    request = rf.get("/")
+    request.authbroker_client = mock.Mock()
+    request.authbroker_client.token = {"access_token": "test"}
+    request.authbroker_client.get = mock.Mock()
+    request.session = {"user_token": "xyz"}
+    get_response = mock.Mock(return_value=Response())
+    mock_get_user.return_value = {"organisations": [{"status": {"key": "in_review"}}]}
+    # Instantiate and call the middleware
+    instance = OrganisationRedirectMiddleWare(get_response)
+    # We should get a 200 use to allowed to log out
```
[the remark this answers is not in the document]
depsiatwal: ahh damn c&p removed

kevincarrogan (approved):

PR #982, opened 2022-11-08
 LTD-2894-shift-generic-fields-to-GoodOnApplication 

depsiatwal:


depsiatwal:
The following fields have changed in API.
            "is_onward_exported"
            "is_onward_altered_processed"
            "is_onward_altered_processed_comments"
            "is_onward_incorporated"
            "is_onward_incorporated_comments"

Added to firearm_details originally which is an extension of good when an item is a firearm.  When adding non-firearm products during C5 it became these transparent these fields are require for other products also. In addition these fields can be set each time a good is added to any application. Hence they were added to 'GoodOnApplication' 

These PR shifts the firearms fields from "Goods" and shifts them to GoodOnapplication as the fields are removed. The UI has changed so that they are set/saved using the relevent payload. 

There's also an accompanying  PR that migrates the values and removes the fields.

exporter/applications/views/goods/firearm/views/edit.py
```diff
@@ -706,14 +705,17 @@ class FirearmProductOnApplicationSummaryEditOnwardAltered(BaseGoodOnApplicationE
     form_class = ProductOnwardAlteredProcessedForm
 
     def get_initial(self):
-        return get_onward_altered_processed_initial_data(self.good_on_application["firearm_details"])
+        return get_onward_altered_processed_initial_data(self.good_on_application)
+
+    def get_edit_payload(self, form):
+        return get_cleaned_data(form)
```
currycoder, line 711: What does this do?

currycoder (approved):

exporter/applications/views/goods/firearm/views/edit.py
depsiatwal: it gets key value pairs for the form

Reason why it's here is because there's a base class implementation which it overrides.

PR #986, opened 2022-11-09
LTD-2990-enable-hawk 

depsiatwal:


sekharpanja (approved):

unit_tests/conftest.py
```diff
@@ -9,7 +9,7 @@
 
 
 @pytest.fixture(autouse=True)
-def setup():
+def disable_hawk():
```
kevincarrogan, line 12: You should use the `settings` fixture here instead of changing the value directly on the imported `settings`

`def disable_hawk(settings):`

PR #993, opened 2022-11-17
 LTD-2997-remove-f680-ff 

depsiatwal:


depsiatwal:
This PR removes the F680 feature flag no longer required since it's in production for few weeks. 
also this is required https://github.com/uktrade/lite-api/pull/1183 since payload is impacted.

exporter/applications/views/common.py
```diff
@@ -138,8 +137,7 @@ def get(self, request, **kwargs):
     def post(self, request, **kwargs):
         application_id = str(kwargs["pk"])
         application = get_application(request, str(kwargs["pk"]))
-        payload = {"check_security_approvals": settings.FEATURE_FLAG_F680_SECURITY_CLASSIFIED_ENABLED}
-        data, status_code = submit_application(request, application_id, payload)
+        data, status_code = submit_application(request, application_id)
```
currycoder, line 140: I agree that we should get this covered with a test

currycoder (approved):

PR #1006, opened 2022-12-05
Enable Make recommendation button for BEIS teams

saruniitr:
## Change description

Users from BEIS teams can give recommendation for the cases in their queues.


kevincarrogan:
Are there no tests for this or the other logic for displaying this button? Feels like something we should probably add.

saruniitr:
@kevincarrogan we have tests for this, updated them to include BEIS teams and queues.

kevincarrogan (approved):

PR #1007, opened 2022-12-06
Add NSG entries form field for TAU assessment

kevincarrogan:


conf/caseworker.py
```diff
@@ -114,6 +114,7 @@
 
 FEATURE_PRODUCTPEDIA_ON = env.bool("FEATURE_PRODUCTPEDIA_ON", False)
 FEATURE_FLAG_PRODUCT_SEARCH = env.bool("FEATURE_FLAG_PRODUCT_SEARCH", False)
+FEATURE_NSG_REGIMES = env.bool("FEATURE_NSG_REGIMES", False)
```
saruniitr, line 117: Can this be a feature flag for C6 overall? because we likely need similar for other regimes (eg AG)

caseworker/tau/forms.py
```diff
@@ -144,6 +167,11 @@ def clean(self):
             if is_mtcr_regime and not mtcr_entries:
                 self.add_error("mtcr_entries", "Type an entry for the Missile Technology Control Regime")
 
+            is_nsg_regime = "NSG" in regimes
```
saruniitr, line 170: `Regimes.NSG in regimes`

conf/caseworker.py
kevincarrogan: Yep, you're right. I was originally thinking of having it separate for each new regime but overall I think one for all of them is better.

saruniitr (approved):

PR #1016, opened 2022-12-14
LTD-3128: Add form to assess potentially trigger list products

saruniitr:
## Change description

For BEIS Nuclear before giving advice they need to assess products potentially on the trigger list.
These are determined by the regime entry marked on those products.

BEIS reviews these products and confirm whether they are actually on the trigger list and covered under any Government assurances.

 - Template added to show this form for BEIS Nuclear
 - Update required JS


caseworker/advice/views.py
```diff
@@ -543,3 +544,83 @@ def form_valid(self, form):
...
+            **context,
+            "case": self.case,
+            "queue_id": self.queue_id,
+            "assessed_trigger_list_goods": self.assessed_trigger_list_goods,
+            "unassessed_trigger_list_goods": self.unassessed_trigger_list_goods,
+        }
+
+    def form_valid(self, form):
+        data = {**form.cleaned_data}
+        data["is_nca_applicable"] = data["is_nca_applicable"] == "Yes"
```
kevincarrogan, line 623: This seems odd that this is a string and not a boolean value in the first place.

caseworker/advice/views.py
```diff
@@ -543,3 +544,83 @@ def form_valid(self, form):
...
+        services.post_trigger_list_assessment(self.request, case_id=self.kwargs["pk"], data=data)
```
kevincarrogan, line 624: What happens if this call fails?

caseworker/advice/forms.py
```diff
@@ -242,3 +244,88 @@ def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.helper = FormHelper()
         self.helper.layout = Layout(Submit("submit", "Move case forward"))
+
+
+class BEISTriggerListFormBase(forms.Form):
+    NSG_LIST_TYPE_CHOICES = [("TRIGGER_LIST", "Trigger list"), ("DUAL_USE", "Dual use")]
+    NCA_CHOICES = [("Yes", "Yes"), ("No", "No")]
```
kevincarrogan, line 251: Isn't this really just a boolean field instead of choices? Not sure if that can be presented as a RadioSelect though.

kevincarrogan (commented):
Why are there three different `trigger_list_home.html` files in different places?

caseworker/advice/forms.py
saruniitr: yeah it was coming out as checkbox but designs require radio button hence using like this

saruniitr:
@kevincarrogan your changes looks good to me

kevincarrogan (approved):

PR #1017, opened 2022-12-15
 LTD-3123-filter-nca 

depsiatwal:


depsiatwal:
This adds the filter to the UI to allow filter for cases that have a GoodOnApplication
is_nca_applicable

caseworker/cases/helpers/filters.py
```diff
@@ -141,5 +141,12 @@ def case_filters_bar(request, filters, is_system_queue) -> FiltersBar:
                 title=CasesListPage.Filters.FLAGS,
                 options=[Option(flag["id"], flag["name"]) for flag in get_flags(request, disable_pagination=True)],
             ),
+            Checkboxes(
+                name="is_nca_applicable",
+                options=[
+                    Option(True, "Filter by Nuclear Cooperation Agreement", id="show-nuclear-cooperation-agreement")
+                ],
+                classes=["govuk-checkboxes--small"],
+            ),
```
currycoder, line 150: Is there any way to test this?

Harriethw: don't know if this is right but in this PR we couldn't find any existing tests (except UI ones that are currently skipped) so added a couple to at least check the box is checked/unchecked based on params https://github.com/uktrade/lite-frontend/pull/1018/commits/d6d46f84d3a2f817134b35da3b09f1ef17ad491e

depsiatwal: added some tests 
we have some unit tests that do a get on 'cases' page and checks response. However I have added a couple of extra ones now as a baseline. One tests the context and the other sets a filter and checks the response of the api call.

PR #1019, opened 2022-12-19
Add dynamic headline component to BEIS advice form

kevincarrogan:
This adds the dynamic heading component from the TAU assessment onto the BEIS assessment.

This was removed from a previous pull request due to it not working and wanting to unblock other tickets so this now rectifies that.


depsiatwal (approved):
lgtm

PR #1024, opened 2022-12-19
Ltd 3146 refactor assessment form

kevincarrogan:
This moves out the generic parts of the assessment form away from it being explicitly just for tau and as such this renames and moves various files and css classes to achieve that.


caseworker/tau/templates/tau/home.html
```diff
@@ -42,12 +42,12 @@ <h1 class="govuk-heading-l">Product assessment</h1>
...
                     <form id="tau-form" action="." method="POST" {% if form.helper.attrs.enctype %}enctype="{{ form.helper.attrs.enctype }}"{% endif %}>
                         {% csrf_token %}
-                        <div class="govuk-grid-column-two-thirds tau__first-column">
+                        <div class="govuk-grid-column-two-thirds assessment-form__first-column">
                             {{ form.goods|as_crispy_field }}
                         </div>
 
-                        <div class="govuk-grid-column-one-third tau__second-column tau__second-column--hide">
-                            <h1 class="govuk-heading-m tau__headline"></h1>
+                        <div class="govuk-grid-column-one-third assessment-form__second-column">
```
currycoder, line 49: This was a `--hide` before..  Is this an intentional change?

kevincarrogan: It sure was. As part of making these changes I realised that the JS component adds this itself so doesn't need it explicitly here.

currycoder (approved):

PR #1026, opened 2022-12-20
LTD 3129 assessed products table

Harriethw:
### aim

Show 'assessed products' table on the "trigger list home page" - this will be used by BEIS Nuclear team members to see products that need assessing or have been assessed.

to set this up locally:
1. set `FEATURE_C6_REGIMES` to True
1. look for a case that has multiple products
1. As TAU officer, assess all products and assign trigger list specifc regimes (e.g. T1) to some of those products. 
1. Then manually assign that case to BEIS Nuclear review queue
1. As BEIS Nuclear case officer, go to BEIS Nuclear queue
1. go to recommendations tab then you should see "Assess products" button because of step2
1. Assess one of the products
1. Observe table appears

### feedback

I've added some checks to the `beis.js` just to stop some console errors but not sure if these are fully needed/i've understood it completely.

### screenshot
NOTE: the original design had 'None' on Nuclear Cooperation Agreement column, but it is mean to be 'No' as confirmed by designer.

Screenshot 2022-12-21 at 12 28 04


seijihg:
your test description is great :D

caseworker/advice/templates/advice/trigger_list_home.html
```diff
@@ -35,6 +37,61 @@ <h1 class="govuk-heading-m assessment-form__headline"></h1>
...
+                                <th scope="col" class="govuk-table__header">Control entry</th>
+                                <th scope="col" class="govuk-table__header">Licence</th>
+                                <th scope="col" class="govuk-table__header">Trigger</th>
+                                <th scope="col" class="govuk-table__header">Dual use</th>
+                                <th scope="col" class="govuk-table__header">Nuclear Cooperation Agreement</th>
+                                <th scope="col" class="govuk-table__header">Assessment note</th>
+                                </tr>
+                            </thead>
+                            <tbody class="govuk-table__body">
+                                {% for good in assessed_trigger_list_goods %}
```
saruniitr, line 60: suggest naming the variable as `good_on_application` instead of `good` as it is deserialized instance of `GoodOnApplication`.
Also it avoids confusion with `good` variable within that (eg line63)

saruniitr (approved):

PR #1028, opened 2022-12-22
LTD-3130-Clear-Assessments

seijihg:
https://user-images.githubusercontent.com/24960282/209237678-c3ab5c0c-4414-45b0-9a17-1c10a787fa79.mov

### Workflow

- [x] Ticket exists in Jira https://uktrade.atlassian.net/browse/LTD-3130

### Reviewing help

### Housekeeping

### Merging

### DIFF
--- a/caseworker/advice/templates/advice/clear_trigger_list_assesment.html
+++ b/caseworker/advice/templates/advice/clear_trigger_list_assesment.html
@@ -23,7 +23,7 @@
 				{% csrf_token %}
 				<div class="govuk-button-group">
 					<input class="govuk-button" type="submit" value="Continue">
-					<a href="{% url 'cases:tau:home' queue_pk=queue_id pk=case.id %}" class="govuk-link--no-visited-state">Cancel</a>
+					<a href="{% url 'cases:assess_trigger_list' queue_pk=queue_pk pk=case.id %}" class="govuk-link--no-visited-state">Cancel</a>
 				</div>
 			</form>
 		</div>
--- a/caseworker/advice/templates/advice/clear_trigger_list_assesment.html
+++ b/caseworker/advice/templates/advice/clear_trigger_list_assesment.html
@@ -1,0 +1,32 @@
+{% extends 'layouts/case.html' %}
+
+{% load static custom_tags crispy_forms_tags advice_tags tau_tags %}
+
+{% block header_tabs %}
+	<div id="tab-bar" class="app-case-tab-bar">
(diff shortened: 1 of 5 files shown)


caseworker/advice/views.py
```diff
@@ -573,6 +574,48 @@ def get_success_url(self):
         return reverse("queues:cases", kwargs={"queue_pk": self.kwargs["queue_pk"]})
 
 
+class BEISNuclearMixin:
```
saruniitr, line 577: This is already merged in `master` may be you need to rebase.

caseworker/advice/views.py
```diff
@@ -573,6 +574,48 @@ def get_success_url(self):
...
+            for product in services.filter_trigger_list_products(self.case["data"]["goods"])
+            if self.is_trigger_list_assessed(product)
+        ]
+
+    @expect_status(
+        HTTPStatus.OK,
+        "Error saving trigger list assessment",
+        "Unexpected error saving trigger list assessment",
+    )
+    def post_trigger_list_assessment(self, request, case_id, selected_good_ids, data):
```
saruniitr, line 603: This should not be in the mixin, we should just add these methods in the views wherever it is required.
If it is in the mixin then it is available in the views where that is not required also.

caseworker/advice/templates/advice/clear_trigger_list_assesment.html
```diff
@@ -0,0 +1,32 @@
...
+	</div>
+{% endblock %}
+
+{% block full_width %}
+
+<div class="govuk-width-container">
+    <main class="govuk-main-wrapper tau">
+        <div class="govuk-grid-row">
+			<h1 class="govuk-heading-l">Clear all assessments</h1>
+			<p class="govuk-body">Are you sure you want to permanently delete all {{assessed_goods.count}} product assessments on this case?</p>
```
saruniitr, line 21: As per designs this screen is not there but probably good to ask for confirmation before clearing assessment (similar to TAU) however this question content is not correct as we are not deleting product assessments.

caseworker/advice/views.py
seijihg: yea but the reason I put it in the mixer it is because two views are using it so kinda no point having it twice.

caseworker/advice/views.py
seijihg: It is the merging conflict thats why.

saruniitr: we need to resolve conflicts in that case.

PR #1029, opened 2022-12-23
Add a UI test for caseworker good detail page

currycoder:
This change adds a UI test to ensure that the caseworker good detail page is working end to end.

### DIFF
--- a/ui_tests/caseworker/features/view_goods.feature
+++ b/ui_tests/caseworker/features/view_goods.feature
@@ -5,10 +5,5 @@ Feature: I want to be able to view goods on a case
     Given I sign in to SSO or am signed into SSO
     And I create standard application or standard application has been previously created
     When I go to application previously created
-    And I assign the case to "FCDO Cases to Review" queue
-    And I go to my profile page
-    And I change my team to "FCDO" and default queue to "FCDO Cases to Review"
-    And I go to my case list
-    And I click the application previously created
     And I click the first good on the case
     Then I see the good details
--- a/ui_tests/caseworker/features/view_goods.feature
+++ b/ui_tests/caseworker/features/view_goods.feature
@@ -1,0 +1,14 @@
+@all @internal @view_goods
+Feature: I want to be able to view goods on a case
+
+  Scenario: View goods detail page
+    Given I sign in to SSO or am signed into SSO
+    And I create standard application or standard application has been previously created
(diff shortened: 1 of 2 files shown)


ui_tests/caseworker/features/view_goods.feature
```diff
@@ -0,0 +1,14 @@
+@all @internal @view_goods
+Feature: I want to be able to view goods on a case
+
+  Scenario: View goods detail page
+    Given I sign in to SSO or am signed into SSO
+    And I create standard application or standard application has been previously created
+    When I go to application previously created
+    And I assign the case to "FCDO Cases to Review" queue
+    And I go to my profile page
+    And I change my team to "FCDO" and default queue to "FCDO Cases to Review"
```
saruniitr, line 10: this is not really necessary but no harm in assigning to a queue

ui_tests/caseworker/step_defs/test_view_goods.py
```diff
@@ -0,0 +1,14 @@
...
+
+@when("I click the first good on the case")  # noqa
+def click_good_on_case(driver, context):
+    driver.find_element_by_link_text(context.goods[0]["good"]["name"]).click()
+
+
+@then("I see the good details")  # noqa
+def good_details_visible(driver, context):
+    assert driver.find_element_by_tag_name("h2").text == "Product details"
+    assert driver.find_element_by_css_selector("div.govuk-caption-l").text == context.goods[0]["good"]["name"]
```
saruniitr, line 14: these methods are deprecated and we are replacing them with their updated counter parts where possible.

Please see,
```
Author: Will Keeling <will@zifferent.com>
Date:   Tue Apr 12 15:49:38 2022 +0100

    Use non-deprecated methods


```

ui_tests/caseworker/features/view_goods.feature
kevincarrogan: I kind of think the smaller the surface area for a test the better so maybe we should just remove this.

If we skipped this we would just check on the "All cases" queue right @saruniitr?

saruniitr: yes that's correct as the default team is Admin.
The existing routing rules will route it to Licensing Reception when application is submitted so we could check it from that queue as well.

saruniitr (approved):

PR #1032, opened 2022-12-30
core changes

sekharpanja:
This PR adds assessed products table to View Recommendations page as per LTD-3135

[LTD-3135]: https://uktrade.atlassian.net/browse/LTD-3135?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


caseworker/advice/templates/advice/view_my_advice.html
```diff
@@ -56,6 +56,10 @@ <h1 class="govuk-heading-xl">View recommendation</h1>
               {% endif %}
             </div>
         </div>
+        <!-- Assessed products table -->
+        {% if assessed_trigger_list_goods %}
+          {% include "advice/assessed_products.html" %}
```
saruniitr, line 61: I think it is better to send `assessed_trigger_list_goods` as a variable to this template.
If we use a variable then even if the name changes outside then we don't need to update the template.

kevincarrogan (approved):

PR #1033, opened 2022-12-30
LTD-3131: Add view and form to edit trigger list assessment

saruniitr:
## Change description

Add view and form to edit trigger list assignment.

Some templates are used outside of TAU in Advice app hence they are moved to a common location caseworker/core/templates/goods/includes

If these are moved to any other location in future then ensure they are in a templates directory in an app because these used in good_detail.html which is rendered as form widget and we get `TemplateDoesnotExist` error if they are in a directory other than app's templates directory.


caseworker/advice/templates/advice/trigger_list_edit.html
```diff
@@ -0,0 +1,32 @@
...
+    <h1 class="govuk-heading-l govuk-!-margin-top-6">Product assessment</h1>
+    <p id="subtitle" class="govuk-body">Editing assessment for <strong>{{ good_on_application.good.name }}</strong></p>
+    <main class="govuk-main-wrapper">
+        <div class="govuk-grid-row">
+            <div class="govuk-grid-column-two-thirds">
+                <div>
+                    <h3 class="govuk-heading-m">{{ good_on_application.line_number }}. {{ good_on_application.good.name }}</h3>
+                    <div>
+                      {% include "tau/good_table.html" with good=good_on_application summary=summary %}
+                      {% include "tau/precedent_table.html" with precedents=good_on_application.precedents %}
```
kevincarrogan, line 14: If we're using this outside of just TAU we should put these in a folder that is more generic.

caseworker/advice/templates/advice/trigger_list_edit.html
```diff
@@ -0,0 +1,32 @@
...
+                    </div>
+                </div>
+            </div>
+            <div class="govuk-grid-column-one-third">
+                <form id="tau-form" method="POST">
```
kevincarrogan, line 19: This isn't a `tau-form` so the id should be different here. We actually might not even need the id really.

caseworker/advice/templates/advice/trigger_list_edit.html
```diff
@@ -0,0 +1,32 @@
...
+                    {% csrf_token %}
+                    <h1 class="govuk-heading-m tau__headline">Assessing {{ good_on_application.good.name}}</h1>
```
kevincarrogan, line 21: I think this class is now wrong and shouldn't be `tau__headline`.

caseworker/advice/templates/advice/trigger_list_edit.html
```diff
@@ -0,0 +1,32 @@
...
+                </form>
+            </div>
+        </div>
+    </main>
+</div>
+{% endblock %}
+
+{% block javascript %}
+    {{ block.super }}
+{% endblock %}
```
kevincarrogan, line 32: If this is just including `{{ block.super }}` then we don't need the block at all.

caseworker/advice/views.py
```diff
@@ -574,15 +582,7 @@ def get_success_url(self):
...
-class BEISProductAssessment(AdviceView, BEISNuclearMixin, FormView):
-    """This renders trigger list product assessment for BEIS Nuclear"""
-
-    template_name = "advice/trigger_list_home.html"
-    form_class = BEISTriggerListAssessmentForm
-
-    def get_success_url(self):
-        return reverse("cases:advice_view", kwargs=self.kwargs)
-
+class BEISAssessmentView:
```
kevincarrogan, line 585: This isn't really a `View` so I don't think it should be called this.

caseworker/advice/views.py
```diff
@@ -592,6 +592,38 @@ def organisation_documents(self):
...
+                **good_on_application_map[item_id],
+                **data,
+            }
+            for item_id in selected_good_ids
+        ]
+
+        return services.post_trigger_list_assessment(self.request, case_id=self.kwargs["pk"], data=data)
+
+
+class BEISProductAssessment(AdviceView, BEISNuclearMixin, BEISAssessmentView, FormView):
```
kevincarrogan, line 618: This is actually a view but isn't called `View`.

caseworker/advice/templates/advice/trigger_list_edit.html
kevincarrogan: Also, for the TAU edit there's a different JS script for this. I think we might need to do the same thing here. Is the JS working on this page? Do you see any errors in the console when you're on this page?

saruniitr: will remove this as there is no need for JS in this case

caseworker/advice/templates/advice/trigger_list_edit.html
saruniitr: removed

caseworker/advice/templates/advice/trigger_list_edit.html
saruniitr: :+1:

kevincarrogan (approved):

PR #1037, opened 2023-01-04
goods total value rounding

MarkThomasLite:
rounding of total value in filter in case of floating point imprecision during conversion from decimal

### DIFF
--- a/core/builtins/custom_tags.py
+++ b/core/builtins/custom_tags.py
@@ -538,11 +538,11 @@ def format_heading(text):
 @register.filter()
 def goods_value(goods):
     total_value = 0
-
     for good in goods:
+        # need to return float in html
         total_value += float(good.get("value", 0))
-
-    return total_value
+    # rounding in case of floating point imprecision during conversion
+    return round(total_value, 2)
 
 
 @register.filter()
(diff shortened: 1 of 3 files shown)


kevincarrogan (commented):
Instead of converting this to a float and then getting out of the precision problem by rounding can't we just switch to using decimals for this whole thing and not worry about floats?

kevincarrogan:
It would also be good to get some tests for this filter.

core/builtins/custom_tags.py
```diff
@@ -538,11 +539,10 @@ def format_heading(text):
 @register.filter()
 def goods_value(goods):
     total_value = 0
-
     for good in goods:
-        total_value += float(good.get("value", 0))
-
-    return total_value
+        total_value += Decimal(good["value"])
```
kevincarrogan, line 543: You’ll wasn’t to retain the `.get` here so that a good without the value key stills gets set as zero.

core/builtins/custom_tags.py
```diff
@@ -1,4 +1,5 @@
 from __future__ import division
+from _decimal import Decimal
```
saruniitr, line 2: is it not just `decimal`?
`from decimal import Decimal`

kevincarrogan (approved):

PR #1043, opened 2023-01-05
LTD-3186-Summary-Template

seijihg:
Ideally, if they haven’t filled in the consignee/third party section we shouldn’t show it at all in this final check your answers summary.

SS shows third parties section is not filled so it is not showing up:
Screenshot 2023-01-05 at 23 30 57

### Workflow

- [x] Ticket exists in Jira https://uktrade.atlassian.net/browse/LTD-3186

### Reviewing help

### Housekeeping

### Merging

### DIFF
--- a/exporter/templates/applications/summary.html
+++ b/exporter/templates/applications/summary.html
@@ -1,9 +1,18 @@
 {% if not type %}
 	{% for key, value in answers.items %}
-		<h2 class="govuk-heading-m">{{ key }}</h2>
+		{% if value %}
+					<h2 class="govuk-heading-m">{{ key }}</h2>
+			{% else %}
+					{% if key != "Third parties" and key != "Consignee" %}
+						<h2 class="govuk-heading-m">{{ key }}</h2>
+			{% endif %}
+		{% endif %}
+
 		<div>
 			{% if not value %}
-				<p class="govuk-caption-m">{% lcs "applications.ApplicationPage.NO_INFORMATION_PROVIDED" %}</p>
+				{% if key != "Third parties" and key != "Consignee" %}
+					<p class="govuk-caption-m">{% lcs "applications.ApplicationPage.NO_INFORMATION_PROVIDED" %}</p>
+				{% endif %}
 			{% elif value|classname == 'str' %}
 				<dl class="govuk-summary-list">
 					<div class="govuk-summary-list__row">


exporter/templates/applications/summary.html
```diff
@@ -1,9 +1,22 @@
...
+				<h2 class="govuk-heading-m">{{ key }}</h2>
+		{% endif %}
+
+		{% if not value %}
+			{% if key == "Third parties" or key == "Consignee" %}
+			{% else %}
+				<h2 class="govuk-heading-m">{{ key }}</h2>
+			{% endif %}
+		{% endif %}
+
```
Harriethw, line 13: minor/ suggestion:
hiya! I'm wondering if this could be simplified to make it a bit more readable? I'm not super familiar with the domain or templating so feel free to ignore!
```
		{% if value %}
				<h2 class="govuk-heading-m">{{ key }}</h2>
		{% else %}
				{% if key != "Third parties" or key != "Consignee" %}
					<h2 class="govuk-heading-m">{{ key }}</h2>
				{% endif %}
		{% endif %}
```

Harriethw:
As this is adding a bit of branching logic to the template, do you think there's any test coverage that could be added? I see there's quite a lot of `test_summary` or `test_summaries` tests but don't know if any of them actually test the html rendered?

exporter/templates/applications/summary.html
saruniitr, line 13: In addition to above simplification something to keep in mind is that this whole template needs to be rewritten to handle this type of changes better. I am not suggesting you rewrite as part of this ticket.

seijihg: Hi Harriet!, I am not sure how to make this better since I am green myself in Django. Template does not allow much room to manoeuvre. I could move it to the View but then these whole thing would need solid rewriting. If you have a suggestion I would welcome a lot!

Harriethw: I included this suggestion in the original comment :) 

```
		{% if value %}
				<h2 class="govuk-heading-m">{{ key }}</h2>
		{% else %}
				{% if key != "Third parties" or key != "Consignee" %}
					<h2 class="govuk-heading-m">{{ key }}</h2>
				{% endif %}
		{% endif %}
```

to me this makes the either/or logic a bit more explicit, and rather than returning null if the key equals  "Third Parties"/"Consignee" you are returning something if it doesn't.

seijihg: ah right I thought u just requoted the code hahaha thanks!!

Harriethw: hmm that's strange, as `!=` is in the docs https://docs.djangoproject.com/en/3.2/ref/templates/builtins/#id1 
and we're using the syntax elsewhere (e.g. lite_forms/templates/form.html) - 

I think I got the logic wrong, it should actually be `if key != "Third parties" AND key != "Consignee"` rather than OR ?

No worries if you'd rather just get this story submitted, as it's going to be re-written anyway

seijihg: *double `or not`  or `!= ` with `or` wont work

Harriethw: >ah yea AND is the correct logic. I did with OR at first as well hmmm

cool, glad we got to the bottom of that mystery! So it wasn't that Django wasn't working, it was just how the expression was being evaluated, does that make sense?

Harriethw (approved):

saruniitr (approved):

PR #1045, opened 2023-01-09
LTD-3188: combine product and regular flags column

Harriethw:
On the 'all cases view'
Remove the 'Products' flags column and put the flags into the 'Flags' column
https://uktrade.atlassian.net/jira/software/c/projects/LTD/boards/204?modal=detail&selectedIssue=LTD-3188 

#### before
Screenshot 2023-01-10 at 11 13 20

#### after
Screenshot 2023-01-10 at 10 19 45

### DIFF
--- a/caseworker/templates/includes/case-row-flags.html
+++ b/caseworker/templates/includes/case-row-flags.html
@@ -12,4 +12,9 @@
 				{% if flag.label %}data-tooltip="{{ flag.label }}"{% endif %}>{{ flag.name }}</li>
 		{% endfor %}
 	{%endif%}
+	{% for flag in goods_flags %}
+		<li
+			class="app-flag app-flag--{{ flag.colour }}"
+			{% if flag.label %}data-tooltip="{{ flag.label }}"{% endif %}>{{ flag.name }}</li>
+	{% endfor %}
 </ol>
--- a/caseworker/templates/queues/cases.html
+++ b/caseworker/templates/queues/cases.html
@@ -77,8 +77,7 @@
 					<th class="govuk-table__header app-table__header--skeleton" scope="col">{% lcs 'cases.CasesListPage.Table.CASE' %}</th>
 					<th class="govuk-table__header app-table__header--skeleton" scope="col">{% lcs 'cases.CasesListPage.Table.ASSIGNEES' %}</th>
 					<th class="govuk-table__header app-table__header--skeleton lite-tablet-hide" width="15%" scope="col">{% lcs 'cases.CasesListPage.Table.DESTINATIONS' %}</th>
-					<th class="govuk-table__header app-table__header--skeleton lite-tablet-hide" width="15%" scope="col">Flags</th>
-					<th class="govuk-table__header app-table__header--skeleton lite-tablet-hide" width="15%" scope="col">{% lcs 'cases.CasesListPage.Table.GOODS' %}</th>
+					<th class="govuk-table__header app-table__header--skeleton lite-tablet-hide" width="15%" scope="col">{% lcs 'cases.CasesListPage.Table.FLAGS' %}</th>
 				</tr>
 			</thead>
 
--- a/caseworker/templates/includes/case-row.html
+++ b/caseworker/templates/includes/case-row.html
@@ -88,6 +88,7 @@
 			{% endif %}
 		</ul>
 	</td>
+
 	{# destinations #}
 	<td class="govuk-table__cell lite-mobile-hide">
 		<ol class="govuk-list destinations__list">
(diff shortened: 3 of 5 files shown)


caseworker/templates/includes/case-row-flags.html
```diff
@@ -12,4 +12,11 @@
 				{% if flag.label %}data-tooltip="{{ flag.label }}"{% endif %}>{{ flag.name }}</li>
 		{% endfor %}
 	{%endif%}
+	{%if goods_flags %}
+		{% for flag in goods_flags %}
+			<li
+				class="app-flag app-flag--{{ flag.colour }}"
+				{% if flag.label %}data-tooltip="{{ flag.label }}"{% endif %}>{{ flag.name }}</li>
+		{% endfor %}
+	{%endif%}
```
currycoder, line 21: It sort of feels like there's a fair bit of unnecessary repetition here now.  I wonder if it could be possible to rework this include slightly so that it just takes a single flat list of `flags`?  I guess that the `all_flags` variable we pass in could be constructed in the view...  What do you think?

caseworker/templates/queues/cases.html
```diff
@@ -77,8 +77,7 @@ <h1 class="govuk-heading-l">
 					<th class="govuk-table__header app-table__header--skeleton" scope="col">{% lcs 'cases.CasesListPage.Table.CASE' %}</th>
 					<th class="govuk-table__header app-table__header--skeleton" scope="col">{% lcs 'cases.CasesListPage.Table.ASSIGNEES' %}</th>
 					<th class="govuk-table__header app-table__header--skeleton lite-tablet-hide" width="15%" scope="col">{% lcs 'cases.CasesListPage.Table.DESTINATIONS' %}</th>
-					<th class="govuk-table__header app-table__header--skeleton lite-tablet-hide" width="15%" scope="col">Flags</th>
-					<th class="govuk-table__header app-table__header--skeleton lite-tablet-hide" width="15%" scope="col">{% lcs 'cases.CasesListPage.Table.GOODS' %}</th>
+					<th class="govuk-table__header app-table__header--skeleton lite-tablet-hide" width="15%" scope="col">{% lcs 'cases.CasesListPage.Table.FLAGS' %}</th>
```
currycoder, line 80: Small question, any idea what `lcs` stands for? :joy:

caseworker/templates/includes/case-row-flags.html
```diff
@@ -12,4 +12,11 @@
 				{% if flag.label %}data-tooltip="{{ flag.label }}"{% endif %}>{{ flag.name }}</li>
 		{% endfor %}
 	{%endif%}
+	{%if goods_flags %}
```
saruniitr, line 15: is the `if` really necessary?

currycoder (commented):
Just a couple of minor comments.

I think the testcases are good and prove that things are popping up where we expect them to.

caseworker/templates/includes/case-row-flags.html
Harriethw: I was also not super happy about it, but looking at the view (caseworker/queues/views.py) I saw that `goods_flags` and `destinations_flags` are added on the API, so I wasn't sure if the logic to combine them belonged in the View or the API...  also I saw on the ticket that there's some work coming up to reduce some of the flags that are shown, could the flat list be built in that ticket, when we know what flags we definitely want to render? (not to kick the can down the road ☺️ )

caseworker/templates/includes/case-row-flags.html
Harriethw: oh probably not, it won't error if `goods_flags` is null? I just copied and pasted from the code above without thinking

caseworker/templates/includes/case-row-flags.html
currycoder: Good points.  Yeah IMO it's ok for the API to give a higher resolution of data than the frontend presents things in; so in this case it would be OK to get the frontend view to do a little bit of reformatting to supply a single variable (other bits of the frontend consuming this API might want these flags separate).
I see your point about deferring the thinking until when we pick up the piece for filtering flags later.  I would guess that we will still do that filtering/exclusion in the frontend rather than adding extra complexity to the API.  
But yes it makes sense if you would prefer to defer this for that ticket - if so, I would suggest adding a comment to that ticket to point out that this bit could be cleaned up so that we don't forget.

Harriethw: cool, thanks, that makes sense about the frontend. I could only see this story for the flag audit so popped a comment on there https://uktrade.atlassian.net/browse/LTD-3218 and will make a note in the commit for this PR

currycoder (approved):

PR #1049, opened 2023-01-13
View cases with open queries tab

Harriethw:
### aim

Add tabs beneath the filters on the cases views (both all cases and for specific queues), one tab shows all cases (with whatever filters have been applied) the other tab shows cases that have an ECJU Query that has NOT been responded to by the user.

***NOTE: this PR is just for the tabs, and a PR to add the `count` to the deselected tab will be upcoming. (as well as potentially no longer 'hiding' cases with open queries on the queue search)***

corresponding API PR to apply the query param= https://github.com/uktrade/lite-api/pull/1247 

all cases view:

Screenshot 2023-01-13 at 10 10 53

team queue view:

Screenshot 2023-01-13 at 11 55 48

### DIFF
--- a/caseworker/queues/views.py
+++ b/caseworker/queues/views.py
@@ -60,8 +60,6 @@ class Cases(LoginRequiredMixin, TemplateView):
 
     @cached_property
     def data(self):
-        hidden = self.request.GET.get("hidden")
-
         params = {"page": int(self.request.GET.get("page", 1))}
         for key, value in self.request.GET.items():
             if key != "flags[]":
@@ -69,8 +67,11 @@ class Cases(LoginRequiredMixin, TemplateView):
 
         params["flags"] = self.request.GET.getlist("flags[]", [])
 
-        if hidden:
-            params["hidden"] = hidden
+        # if the hidden param is not true
+        # then cases with open queries are filtered out on team queue views
+        only_open_queries = self.request.GET.get("only_open_queries")
+        if only_open_queries == "True":
+            params["hidden"] = "True"
 
         data = get_cases_search_data(self.request, self.queue_pk, params)
         return data
@@ -78,6 +79,25 @@ class Cases(LoginRequiredMixin, TemplateView):
     @property
     def filters(self):
         return self.data["results"]["filters"]
+
+    def open_queries_tabs(self):
+        only_open_queries = self.request.GET.get("only_open_queries") or "False"
+
+        params_with_queries = self.request.GET.copy()
+        params_with_queries["only_open_queries"] = "True"
+        params_all_cases = self.request.GET.copy()
+        params_all_cases["only_open_queries"] = "False"
+
+        open_queries_url = "{}?{}".format(self.request.path, params_with_queries.urlencode())
+        all_cases_url = "{}?{}".format(self.request.path, params_all_cases.urlencode())
+
+        open_queries_tabs = {
+            "only_open_queries": only_open_queries,
+            "open_queries_url": open_queries_url,
+            "all_cases_url": all_cases_url,
+        }
+
+        return open_queries_tabs
 
     def get_context_data(self, *args, **kwargs):
 
@@ -112,6 +132,7 @@ class Cases(LoginRequiredMixin, TemplateView):
             "enforcement_check": Permission.ENFORCEMENT_CHECK.value in get_user_permissions(self.request),
             "updated_cases_banner_queue_id": UPDATED_CASES_QUEUE_ID,
             "show_updated_cases_banner": show_updated_cases_banner,
+            "open_queries_tabs": self.open_queries_tabs,
         }
 
         return super().get_context_data(*args, **context, **kwargs)
--- a/caseworker/templates/queues/cases.html
+++ b/caseworker/templates/queues/cases.html
@@ -61,6 +61,24 @@
 
 	{% include 'filters.html' %}
 
+	<div class="lite-tabs__container">
+		<div class="lite-tabs">
+			<a href="{{ open_queries_tabs.all_cases_url }}" class="lite-tabs__tab  {% if open_queries_tabs.only_open_queries == "False" %}lite-tabs__tab--selected{% endif %}" id="view-all-queries-tab">
+				{% if queue.is_system_queue %}
+					All cases
+				{% else %}
+					Cases to review
+				{% endif %}
+				{% if open_queries_tabs.only_open_queries == "False" %}
+					({{ data.count }})
+				{% endif %}
+			</a>
+			<a href="{{ open_queries_tabs.open_queries_url }}" class="lite-tabs__tab {% if open_queries_tabs.only_open_queries == "True" %}lite-tabs__tab--selected{% endif %}" id="view-open-queries-tab">
+				Open queries {% if open_queries_tabs.only_open_queries == "True" %}({{ data.count }}){% endif %}
+			</a>
+		</div>
+	</div>
+
 	<form id="form-cases" method="get" action="{% url 'queues:case_assignments' queue.id %}">
 		{% if not data.results.cases %}
 			{% include "includes/notice.html" with text='cases.CasesListPage.NO_CASES' %}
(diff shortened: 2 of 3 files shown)


caseworker/queues/views.py
```diff
@@ -79,6 +79,25 @@ def data(self):
...
+    def open_queries_tabs(self):
+        has_open_queries = self.request.GET.get("has_open_queries") or "False"
+
+        params_with_queries = self.request.GET.copy()
+        params_with_queries.__setitem__("has_open_queries", "True")
+        params_all_cases = self.request.GET.copy()
+        params_all_cases.__setitem__("has_open_queries", "False")
+
+        open_queries_url = "{}?{}".format(self.request.path, params_with_queries.urlencode())
+        all_cases_url = "{}?{}".format(self.request.path, params_all_cases.urlencode())
```
Harriethw, line 91: this needs to work on both the index url and the queues view, so I'm just getting the current path from the request.

caseworker/queues/views.py
```diff
@@ -79,6 +79,25 @@ def data(self):
     def filters(self):
         return self.data["results"]["filters"]
 
+    def open_queries_tabs(self):
+        has_open_queries = self.request.GET.get("has_open_queries") or "False"
+
+        params_with_queries = self.request.GET.copy()
+        params_with_queries.__setitem__("has_open_queries", "True")
+        params_all_cases = self.request.GET.copy()
+        params_all_cases.__setitem__("has_open_queries", "False")
```
saruniitr, line 88: My understanding is "Open queries" tab is just a quick way of looking at cases that has open queries but they are still visible under "All cases" but this seems to filter those out.
Looking at the comments in the ticket also suggests that they need to be included but not clear what is the conclusion.
So 'All cases' will exclude cases with open queries?

Harriethw: hmm I see the confusion - the param should probably be called `has_open_queries_filter`? if it's true we only show open queries, if it's not then we return the regular search (on the all cases queue that will include cases with open queries, on the team queue view that will hide cases with open queries)

Harriethw: or `is_filtering_open_queries`?

saruniitr: thanks, that clarifies.
`has_open_queries_filter` looks good. is `filter_open_queries` any better?

Harriethw: as it's a boolean value (which makes sense for the backend) i was hoping to keep it beginning with `is` or `has`

caseworker/queues/views.py
```diff
@@ -79,6 +79,25 @@ def data(self):
     def filters(self):
         return self.data["results"]["filters"]
 
+    def open_queries_tabs(self):
+        has_open_queries_filter = self.request.GET.get("has_open_queries_filter") or "False"
+
+        params_with_queries = self.request.GET.copy()
+        params_with_queries.__setitem__("has_open_queries_filter", "True")
```
currycoder, line 86: `_` and `__` usually denote that an interface is designed to be used internally in a given object/module rather than publicly.  So in this case could you just do standard dictionary set item syntax; `params_with_queries["has_open_queries_filter"] = True`?

caseworker/templates/queues/cases.html
```diff
@@ -61,6 +61,24 @@ <h1 class="govuk-heading-l">
...
+					All cases
+				{% else %}
+					Cases to review
+				{% endif %}
+				{% if open_queries_tabs.has_open_queries_filter == "False" %}
+					({{ data.count }})
+				{% endif %}
+			</a>
+			<a href="{{ open_queries_tabs.open_queries_url }}" class="lite-tabs__tab {% if open_queries_tabs.has_open_queries_filter == "True" %}lite-tabs__tab--selected{% endif %}" id="view-open-queries-tab">
+				Open queries {% if open_queries_tabs.has_open_queries_filter == "True" %}({{ data.count }}){% endif %}
```
currycoder, line 77: I'm guessing that when the count API is present these will be reworked to use that for the inactive tab?

caseworker/queues/views.py
```diff
@@ -112,6 +131,7 @@ def get_context_data(self, *args, **kwargs):
             "enforcement_check": Permission.ENFORCEMENT_CHECK.value in get_user_permissions(self.request),
             "updated_cases_banner_queue_id": UPDATED_CASES_QUEUE_ID,
             "show_updated_cases_banner": show_updated_cases_banner,
+            "open_queries_tabs": self.open_queries_tabs,
```
currycoder, line 134: I'm sort of surprised that there was no change needed to send the query param up to the API; are all query params just sent to the API automatically?

currycoder (commented):
A few observations and questions

caseworker/queues/views.py
currycoder: The thing is that this also needs to be consistent on the API.  And TBH I feel like `has_open_queries` was the best option all along as it stays consistent with other params on the API (which do not use the word "filter").
Instead of changing a variable name to try remedy this confusion, could some inline comments do the trick?

caseworker/queues/views.py
Harriethw: It's already all setup to format the params in the request and then to handle them on the API, in the `data` method on the view, and then  in `get_cases_search_data` in `services.py`

caseworker/queues/views.py
Harriethw: I wouldn't say we need to worry too much about being consistent with the existing params, I feel that each param should represent itself and say what it's doing is more important? And I would rather have the code reflect its purpose than rely on comments if possible. I agree that `has_open_queries` is confusing because if it's set to false it would imply it's asking for cases that do not have open queries. some other suggestions- 
`open_queries_tab_is_selected`
`open_queries_tab_is_active`
`is_filtering_open_queries`
`is_showing_open_queries`

Looking at discussions about future work this might need to change anyway in the future, as it looks like there's plans to add more tabs, in which case it won't be a simple either/or situation and we might want to have one param to pass through a key for the active tab.

caseworker/queues/views.py
Harriethw: TIL about magic methods! have changed to use standard syntax

caseworker/queues/views.py
Harriethw: I feel like `include_open_queries` has the same problem as `has_open_queries`, with the added downside that it doesn't follow boolean naming conventions - do we definitely not like `is_filtering_open_queries` ?

considering that this might change soon if we can't come to a compromise I'm happy to use `has_open_queries` and put a comment or two in.

Harriethw: we'v agreed to change it to `only_open_queues`

caseworker/queues/views.py
```diff
@@ -60,17 +60,18 @@ def queue_pk(self):
...
-        hidden = self.request.GET.get("hidden")
-
         params = {"page": int(self.request.GET.get("page", 1))}
         for key, value in self.request.GET.items():
             if key != "flags[]":
                 params[key] = value
 
         params["flags"] = self.request.GET.getlist("flags[]", [])
 
-        if hidden:
```
Harriethw, line 72: as far as I can tell, we no longer needed to check explicitly for the `hidden` param, because we iterate through GET items above.

currycoder: Yeah agreed

caseworker/queues/views.py
```diff
@@ -60,17 +60,18 @@ def queue_pk(self):
...
-            params["hidden"] = hidden
+        # if the hidden param is not true
+        # then cases with open queries are filtered out on team queue views
+        only_open_queries = self.request.GET.get("only_open_queries")
+        if only_open_queries == "True":
+            params["hidden"] = True
```
currycoder, line 74: Perhaps this should be `"True"` so that the datatype is consistent with the other values in `params`?

currycoder (approved):

caseworker/queues/views.py
Harriethw: yep that slipped through, have changed to "True"

PR #1050, opened 2023-01-13
[LTD-3237] Add optional support for mock SSO endpoints

currycoder:
This change adds configurable support for mocked oauth endpoints.  This also switches these endpoints on as the authbroker for CI tests.


caseworker/urls.py
```diff
@@ -33,6 +33,12 @@
     urlpatterns.append(path("spire/", include("caseworker.spire.urls")))
 
 
+if settings.MOCK_SSO_ACTIVATE_ENDPOINTS:
+    urlpatterns = [
+        path("", include("caseworker.mock_sso.urls")),
+    ] + urlpatterns
```
currycoder, line 39: Would be good to get opinions on whether this is sufficient protection to avoid this being accidentally switched on in environments where it shouldn't?
I thought about doing also requiring the env to be CI only, but there's not much that we can easily hook in to.  Perhaps we could identify this if `VCAP_SERVICES` exists, but that seems brittle as we are soon to move away from cf..

saruniitr: can be joined with `DEBUG` so doesn't get switched on in `prod`

kevincarrogan: The problem with `DEBUG` is that I think we eventually want to set `DEBUG=False` during the e2e tests.

pii-secret-exclude.txt
```diff
@@ -1,6 +1,7 @@
 lite_content
 example.caseworker.env
 example.exporter.env
+ci.caseworker.env
```
kevincarrogan, line 4: I try not to include any of our `.env` files here if I can. I just usually skip the `pre-commit` check that complains here on my local machine instead of adding it to the exclude.

caseworker/urls.py
currycoder: We discussed this in slack and came to the conclusion that the narrow combination of `MOCK_SSO_ACTIVATE_ENDPOINTS` and `AUTHBROKER_URL` values is enough of a safety net.

kevincarrogan (approved):

PR #1051, opened 2023-01-13
LTD-3193 add serial number section to goods in case detail

MarkThomasLite:
Adding serial numbers (firearms) to the case detail section of review. This html is used by several others so code was added to html + custom tag.

### DIFF
--- a/caseworker/advice/templates/advice/case_detail.html
+++ b/caseworker/advice/templates/advice/case_detail.html
@@ -1,213 +1,233 @@
 {% load static advice_tags %}
 
-        {% with display_pv_grading=case.goods|is_case_pv_graded %}
-        <h2 class="govuk-heading-m">Case details</h2>
-        <details class="govuk-details" data-module="govuk-details">
-          <summary class="govuk-details__summary">
-              <span class="govuk-details__summary-text">Products</span>
-          </summary>
-          <div class="govuk-details__text">
+{% with display_pv_grading=case.goods|is_case_pv_graded %}
+    <h2 class="govuk-heading-m">Case details</h2>
+    <details class="govuk-details" data-module="govuk-details">
+        <summary class="govuk-details__summary">
+            <span class="govuk-details__summary-text">Products</span>
+        </summary>
+        <div class="govuk-details__text">
             <table id="table-nlr-products" class="govuk-table">
-              <tbody class="govuk-table__body">
-                {% for good in case.goods %}
-                <tr class="govuk-table__row">
-                  <td class="govuk-table__cell lite-!-no-border" colspan="2">
-                    <h3>{{ forloop.counter }}. {{ good.good.name }}</h3>
-                  </td>
-                </tr>
-                <tr class="govuk-table__row">
-                  <td class="govuk-table__cell govuk-!-font-weight-bold">Licence required</td>
-                  <td class="govuk-table__cell">
-                    {% if good.is_good_controlled is not None %}
-                      {{ good.is_good_controlled.value }}
-                    {% elif good.good.is_good_controlled is not None %}
-                        {{ good.good.is_good_controlled.value }}
-                    {% else %}
-                        Not added
-                    {% endif %}
-                </td>
-                </tr>
-                <tr class="govuk-table__row">
-                  <td class="govuk-table__cell govuk-!-font-weight-bold">Control list entries</td>
-                  <td class="govuk-table__cell">
-                    {% with clc=good|get_clc %}
-                        {% if clc %}
-                            {{ clc|join:', ' }}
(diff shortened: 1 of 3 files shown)


caseworker/advice/templatetags/advice_tags.py
```diff
@@ -89,6 +89,13 @@ def group_advice(context):
     context["grouped_advice"] = grouped_advice
     return context
 
+@register.filter()
+def format_serial_numbers(serial_numbers, quantity):
+    quantity = int(quantity)
+    response = [f"{index}. {sn}" for index, sn in enumerate(serial_numbers, start=1)]
+    if quantity > len(serial_numbers):
+        response += [f"{index + 1}." for index in range(len(serial_numbers), quantity)]
+    return response
```
antroy-madetech, line 99: Should be some unit tests around this logic.

caseworker/advice/templates/advice/case_detail.html
```diff
@@ -1,213 +1,233 @@
...
+                                            <summary class="govuk-details__summary">
+                                                <span class="govuk-details__summary-text">View serial numbers</span>
+                                            </summary>
+                                            <div class="govuk-details__text">
+                                                {{ good.good.firearm_details.serial_numbers|format_serial_numbers:good.quantity|join:"<br>" }}
+                                            </div>
+                                        </details>
+                                    {% endif %}
+                                </td>
+                            </tr>
```
antroy-madetech, line 100: There are probably view tests that can be written around this display logic.

caseworker/advice/templates/advice/case_detail.html
```diff
@@ -1,213 +1,233 @@
 {% load static advice_tags %}
 
-        {% with display_pv_grading=case.goods|is_case_pv_graded %}
-        <h2 class="govuk-heading-m">Case details</h2>
-        <details class="govuk-details" data-module="govuk-details">
-          <summary class="govuk-details__summary">
-              <span class="govuk-details__summary-text">Products</span>
-          </summary>
-          <div class="govuk-details__text">
+{% with display_pv_grading=case.goods|is_case_pv_graded %}
```
saruniitr, line 3: Is this run through a formatter? For existing templates we keep the formatting as is and it introduces lots of unnecessary changes like this one and changes history, please revert all formatting changes.

caseworker/advice/templates/advice/case_detail.html
```diff
@@ -1,213 +1,233 @@
...
+                                        {{ good.good.firearm_details.serial_numbers }}
+                                    {% else %}
+                                        <details class="govuk-details" data-module="govuk-details">
+                                            <summary class="govuk-details__summary">
+                                                <span class="govuk-details__summary-text">View serial numbers</span>
+                                            </summary>
+                                            <div class="govuk-details__text">
+                                                {{ good.good.firearm_details.serial_numbers|format_serial_numbers:good.quantity|join:"<br>" }}
+                                            </div>
+                                        </details>
```
saruniitr, line 97: We already have a template for this, please see if that can be reused here,
https://github.com/uktrade/lite-frontend/blob/master/caseworker/templates/goods/includes/serial_numbers.html

MarkThomasLite: the required formatting is different which is why I created the separate localized version in advice

caseworker/advice/templates/advice/case_detail.html
MarkThomasLite: I have reverted the formatting, inconsistent formatting across the html will lead to more confusion however which is why i attempted to address it

caseworker/advice/templates/advice/case_detail.html
```diff
@@ -1,213 +1,233 @@
...
+                            <td class="govuk-table__cell">{{ destination.country.name }}</td>
+                            {% if 'type_display_value' in destination %}
+                                <td class="govuk-table__cell">{{ destination.type_display_value }}</td>
+                            {% else %}
+                                <td class="govuk-table__cell">{{ destination.type }}</td>
+                            {% endif %}
+                            <td class="govuk-table__cell">{{ destination.name }}</td>
+                        </tr>
+                    {% endfor %}
+                </tbody>
```
saruniitr, line 133: this is again formatting change which is not necessary

caseworker/advice/templates/advice/case_detail.html
saruniitr: understand but those changes can be a separate PR on its own and if ok we can get those changes in.
Combining with other functional changes makes it difficult to see what are the actual changes vs formatting changes especially in templates.

saruniitr (approved):

PR #1052, opened 2023-01-13
Show countersigner comment in recommendation summary

hnryjmes:


unit_tests/caseworker/advice/views/test_view_ogd_advice_countersign.py
```diff
@@ -0,0 +1,143 @@
...
+from django.urls import reverse
+
+
+@pytest.fixture(autouse=True)
+def setup(mock_queue, mock_denial_reasons, mock_case):
+    yield
+
+
+@pytest.fixture
+def group_advice(MOD_team1_user, MOD_team2_user, MOD_ECJU_team_user, FCDO_team_user):
```
saruniitr, line 12: Can you add one advice from LU as well as they also countersign?

unit_tests/caseworker/advice/views/test_view_ogd_advice_countersign.py
```diff
@@ -0,0 +1,143 @@
...
+            "denial_reasons": [],
+            "end_user": "76de33f4-7834-4eb7-871a-66be3270fb59",
+            "footnote": "",
+            "good": None,
+            "goods_type": None,
+            "id": "217e7264-7f28-46ff-bf01-dc64f4432786",
+            "level": "user",
+            "note": "",
+            "proviso": None,
+            "text": "On behalf of the ECJU-FCDO Desk Officer for Buckinghamshire, their recommendation to refuse is as follows:\r\n\r\nThis licence would permit the export of miniature broccoli and associated accessories for stock. Ultimate end users are unknown.\r\n\r\nBased on the information available, alongside the unacceptable risk of diversion, I recommend to DIT that they refuse this licence.\r\n\r\nSarah Alcatraz ECJU-FCDO",
```
saruniitr, line 104: Please remove team/department references etc and make it very generic, can be a simple line of text.

unit_tests/caseworker/advice/views/test_view_ogd_advice_countersign.py
```diff
@@ -0,0 +1,143 @@
...
+    url = reverse("cases:advice_view", kwargs={"queue_pk": data_queue["id"], "pk": data_standard_case["case"]["id"]})
+    response = authorized_client.get(url)
+
+    soup = BeautifulSoup(response.content, "html.parser")
+
+    assert "Countersigned by Countersigner Team" in [h2.text for h2 in soup.find_all("h2")]
+    assert (
+        "agree recommendation to refuse. Head of Cases in ECJU-FCDO agreed the recommendation to refuse\r\nNartin Mario - ECJU - FCDO"
+        in [p.text for p in soup.find_all("p")]
+    )
```
saruniitr, line 127: Group advice is shown under links specific to each team which when expanded we can see the details so is it not possible to extract the html for specific team and then assert on elements within that html?

Currently it is not clear which team has given these comments.
Eg in the fixture if we include LU countersign advice also then we should be able to assert for both FCDO and LU.

unit_tests/caseworker/advice/views/test_view_ogd_advice_countersign.py
```diff
@@ -0,0 +1,143 @@
...
+
+    url = reverse("cases:advice_view", kwargs={"queue_pk": data_queue["id"], "pk": data_standard_case["case"]["id"]})
+    response = authorized_client.get(url)
+
+    soup = BeautifulSoup(response.content, "html.parser")
+
+    assert sum(["Countersigned by" in h2.text for h2 in soup.find_all("h2")]) != len(group_advice)
+    assert sum(["Countersigned by" in h2.text for h2 in soup.find_all("h2")]) == sum(
+        [x["countersigned_by"] is not None for x in group_advice]
+    )
```
saruniitr, line 143: As mentioned above if we can extract html specific to a team then we can assert absence of these values for the teams that didn't countersign.

hnryjmes: I think I prefer this way, because it is testing the implementation, rather than testing the test data

The response will not show the "Countersigned by" message for the teams that don't countersign because, for those teams, there is no "countersigned_by" field in the data which would trigger the conditional logic in the template

The way this test works it checks that the template is working correctly because the only way that last assertion can be true is if the number of times "Countersigned by" appears on the page is equal to the number of advices in the group advice with a countersigned_by field, and not simply the length of the group advice list (which it would be if there were no rendering logic in the template)

hnryjmes:
This is ready for review once the CI tests have passed

unit_tests/caseworker/advice/views/test_view_ogd_advice_countersign.py
saruniitr: Besides testing the presence of "countersigned_by" we also need to test whether it is from the correct team or not because not all the teams countersign and those they do also under certain conditions.

Some teams double countersign so we should be able to test that as well.
We will be adding more tests in this area as we change the countersigning process.

Currently this doesn't test for the team.

hnryjmes:
Ready for re-review

unit_tests/caseworker/advice/views/test_view_ogd_advice_countersign.py
```diff
@@ -0,0 +1,194 @@
...
+
+    fcdo_team_details = [d for d in list_of_details if "FCDO Team has refused" in d.find("span").text][0]
+    assert "Agree recommendation to refuse. Person Name - FCDO Countersigning Office" in [
+        p.text for p in fcdo_team_details.find_all("p")
+    ]
+
+    lu_team_details = [d for d in list_of_details if "LU Team has refused" in d.find("span").text][0]
+    assert "Agree recommendation to refuse. Person Name - LU Countersigning Office" in [
+        p.text for p in lu_team_details.find_all("p")
+    ]
```
saruniitr, line 167: I think you were previously testing for the user that countersigned which is removed now, can that be added again?
Essentially we want to check below,
- User details that countersigned
- countersign comments
- Team of the user that countersigned

unit_tests/caseworker/advice/views/test_view_ogd_advice_countersign.py
```diff
@@ -0,0 +1,194 @@
...
+
+    url = reverse("cases:advice_view", kwargs={"queue_pk": data_queue["id"], "pk": data_standard_case["case"]["id"]})
+    response = authorized_client.get(url)
+
+    soup = BeautifulSoup(response.content, "html.parser")
+
+    assert sum(["Countersigned by" in h2.text for h2 in soup.find_all("h2")]) != len(group_advice)
+    assert sum(["Countersigned by" in h2.text for h2 in soup.find_all("h2")]) == sum(
+        [x["countersigned_by"] is not None for x in group_advice]
+    )
```
saruniitr, line 183: We know the expected number of teams that countersign, can we check for that number directly?

unit_tests/caseworker/advice/views/test_view_ogd_advice_countersign.py
```diff
@@ -0,0 +1,194 @@
...
+):
+    data_standard_case["case"]["advice"] = group_advice
+
+    url = reverse("cases:advice_view", kwargs={"queue_pk": data_queue["id"], "pk": data_standard_case["case"]["id"]})
+    response = authorized_client.get(url)
+
+    soup = BeautifulSoup(response.content, "html.parser")
+
+    assert sum(["Countersigned by" in h2.text for h2 in soup.find_all("h2")]) != len(group_advice)
+    assert sum(["Countersigned by" in h2.text for h2 in soup.find_all("h2")]) == sum(
```
kevincarrogan, line 181: There shouldn't be `sum` here as you're not summing anything or you're relying on Falsey values being coerced into integers and I think this could be better signalled by writing this in a different way that wasn't relying on that coercion taking place.

unit_tests/caseworker/advice/views/test_view_ogd_advice_countersign.py
```diff
@@ -0,0 +1,194 @@
...
+        [x["countersigned_by"] is not None for x in group_advice]
+    )
+
+    list_of_details = soup.find_all("details", class_="govuk-details")
+
+    mod_team1_details = [d for d in list_of_details if "Approved by MoD Team1 User" in d.find("h2").text][0]
```
kevincarrogan, line 187: We're always grabbing the first thing but I can't really tell from the test why that's important.

hnryjmes: beautifulsoup always returns a list, so if there is one result, you still need to extract it from the list

hnryjmes:
I have rewritten the tests to make it a bit more obvious what's going on. 

For a specific countersigning team, the tests explicitly test the following:
- Countersign heading is shown
- Countersigner name is shown
- Countersign comment is shown

We also check that, for non countersigning teams:
- Countersign heading is not shown

BeautifulSoup was frustrating me so the assertions now use the string representation (i.e. raw html string) rather than their tag objects, which are very easy to make assertions on using standard python, and the logic should now be clearer to future users of this code

unit_tests/caseworker/advice/views/test_view_ogd_advice_countersign.py
```diff
@@ -0,0 +1,206 @@
...
+    response = authorized_client.get(url)
+
+    soup = BeautifulSoup(response.content, "html.parser")
+
+    list_of_details = soup.find_all("details", class_="govuk-details")
+
+    # Find details tag for specific team by label
+    fcdo_team_details = [d for d in list_of_details if "FCDO Team has refused" in str(d)]
+    # Check countersign heading
+    assert "Countersigned by" in "".join(str(fcdo_team_details))
```
saruniitr, line 162: We have the user details (name) that countersigned so please include that in the test,

`assert "Countersigned by Countersigner Team" in "".join(str(fcdo_team_details))`

This is required because when there is double countersigning we can differentiate by the user name otherwise this would still pass even if the user is not we are expecting.

Also worth changing the `countersigned_by` user names to be different from FCDO countersigned user but I will leave that with you.

With that I think it is good to go.

hnryjmes: I agree - good suggestion

hnryjmes:
We now check explicitly for which team has countersigned in the heading, e.g.

`assert "Countersigned by FCDO Countersigner Team" in "".join(str(fcdo_team_details))`

saruniitr (approved):

PR #1053, opened 2023-01-16
Ltd 3242 beis nuclear assessment flow

antroy-madetech:
Stay on Assessment page when submitting products.
If all products have been assessed, stay on Assessment page but show "Make recommendation" button.

### DIFF
--- a/caseworker/advice/templates/advice/assessed_products.html
+++ b/caseworker/advice/templates/advice/assessed_products.html
@@ -3,6 +3,9 @@
     <div class="govuk-grid-column-full">
         <table id="assessed-products" class="govuk-table">
             <h2 class="govuk-heading-m">Assessed products</h2>
+            {% if can_advise and buttons.make_recommendation %}
+                <a draggable="false" class="govuk-button govuk-button--primary govuk-!-margin-right-2" href="{% url 'cases:select_advice' queue_pk case.id %}">Make recommendation</a>
+            {% endif %}
             <a href="{% url 'cases:clear_trigger_list_assessments' queue_pk=queue_pk pk=case.id %}" class="govuk-button govuk-button--secondary" data-module="govuk-button">Clear assessments</a>
             <thead class="govuk-table__head">
                 <tr class="govuk-table__row">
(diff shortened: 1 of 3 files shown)


caseworker/advice/templates/advice/assessed_products.html
```diff
@@ -3,6 +3,9 @@
     <div class="govuk-grid-column-full">
         <table id="assessed-products" class="govuk-table">
             <h2 class="govuk-heading-m">Assessed products</h2>
+            {% if can_advise and buttons.make_recommendation %}
+                <a draggable="false" class="govuk-button govuk-button--primary govuk-!-margin-right-2" href="{% url 'cases:select_advice' queue_pk case.id %}">Make recommendation</a>
+            {% endif %}
```
saruniitr, line 8: We should not be showing "Make recommendation" button in this template.
It is already show in https://github.com/uktrade/lite-frontend/blob/master/caseworker/advice/templates/advice/view-advice.html#L39 when certain conditions are met. In this case when all trigger list products are reviewed.

caseworker/advice/views.py
```diff
@@ -623,7 +623,7 @@ class BEISProductAssessmentView(AdviceView, BEISNuclearMixin, BEISAssessmentBase
     form_class = BEISTriggerListAssessmentForm
 
     def get_success_url(self):
-        return reverse("cases:advice_view", kwargs=self.kwargs)
+        return self.request.path
```
saruniitr, line 626: We only need to stay in this page until all products are reviewed.
Once all are reviewed then we need to above page where "Make recommendation" is made available to give advice.

caseworker/advice/templates/advice/assessed_products.html
antroy-madetech: See the ticket. They now want the "Make recommendation" button here as well.

caseworker/advice/views.py
antroy-madetech: The ticket says: "Once all products are reviewed you are directed to a ‘Make a recommendation’ screen, we would like them to be directed to the Assessed Products table with the ‘Make recommendation’ and ‘Clear assessments’ options"

antroy-madetech:
Rework done on a new branch

PR #1054, opened 2023-01-17
LTD-3189 open queries count

Harriethw:
waiting on https://github.com/uktrade/lite-frontend/pull/1049  on the frontend and this API PR to merge so that the e2e tests can pass https://github.com/uktrade/lite-api/pull/1249#pullrequestreview-1253367975 

This PR adds the `count` of cases for the unselected tab on the queues view.

Screenshot 2023-01-19 at 10 35 20

### DIFF
--- a/caseworker/queues/services.py
+++ b/caseworker/queues/services.py
@@ -63,6 +63,12 @@ def get_cases_search_data(request, queue_pk, params):
     return data.json()
 
 
+def head_cases_search_count(request, queue_pk, params):
+    querystring = convert_parameters_to_query_params({"queue_id": queue_pk, **params})
+    response = client.head(request, f"/cases/{querystring}")
+    return response.headers["Resource-Count"]
+
+
 def put_queue(request, pk, json):
     data = client.put(request, f"/queues/{pk}/", json)
     return data.json(), data.status_code
--- a/caseworker/queues/views.py
+++ b/caseworker/queues/views.py
@@ -42,7 +42,7 @@ from caseworker.queues.services import (
     post_enforcement_xml,
 )
 from caseworker.users.services import get_gov_user
-from caseworker.queues.services import get_cases_search_data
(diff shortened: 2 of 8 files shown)


caseworker/queues/services.py
```diff
@@ -63,6 +63,11 @@ def get_cases_search_data(request, queue_pk, params):
     return data.json()
 
 
+def head_cases_search_count(request, queue_pk, params):
+    response = client.head(request, "/cases/" + f"?queue_id={queue_pk}&" + parse.urlencode(params, doseq=True))
```
currycoder, line 67: Minor thing but `convert_parameters_to_query_params()`  used elsewhere in this file may be better rather than repeating parameter conversion around the file

caseworker/queues/views.py
```diff
@@ -91,10 +104,13 @@ def open_queries_tabs(self):
         open_queries_url = "{}?{}".format(self.request.path, params_with_queries.urlencode())
         all_cases_url = "{}?{}".format(self.request.path, params_all_cases.urlencode())
 
+        unselected_tab_count = self.get_open_queries_count("True" if only_open_queries == "False" else "False")
```
currycoder, line 107: Could this perhaps be a dict rather than a single value?  Just thinking about making this an easy thing to add to when we have more than one unselected tab

Harriethw: we're going to need to refactor the context data quite a bit when we add another tab, which [I've already sketched out for this story](https://uktrade.atlassian.net/browse/LTD-3239?focusedCommentId=105600) - would it be ok to address that in that PR?

currycoder: Ok brilliant.  Yeah let's leave this for the coming refactor :raised_hands:

caseworker/queues/services.py
Harriethw: that's much tidier! thanks, have used.

currycoder (approved):

PR #1055, opened 2023-01-17
[LTD-3187] Queue view enhanced allocation

currycoder:
This change adjusts the cases search page so that the "Case Allocation" column shows all queues that a case is present on rather than just queues where there is an allocated caseworker.

There are also changes to generalise the "flag expander" javascript so that it can be used as a general "list expander" component anywhere in our app.  This generalised expander is applied to the Case Allocation column and replaces the existing flag expander.

Companion backend PR: https://github.com/uktrade/lite-api/pull/1248

**Note on release sequence**
The backend PR should be released first and then this PR can be released.  As long as they are released in this sequence there should be no backward/forward incompatibility errors.


caseworker/assets/javascripts/list-expander.js
```diff
@@ -0,0 +1,54 @@
...
+      return;
+    }
+    this.$expandButton.classList.add("expander__expand-button__visible");
+
+    for (let i = this.visibleElems; i < this.liElems.length; i++) {
+      let $li = this.liElems[i];
+      $li.classList.add("expander__expand-list__item__hidden");
+    }
+    this.$expandButton.querySelector("span").textContent =
+      "⌄ " + this.visibleElems + " of " + this.liElems.length;
```
currycoder, line 30: Replaced the "down chevron SVG" with a unicode character.  Would appreciate some guidance on how to continue to use the SVG without having to copy around it's html as per the original flag expander (https://github.com/uktrade/lite-frontend/blob/3c610e48095a5e1412ff28d622785e8f75e91fd2/caseworker/assets/javascripts/flag-expanders.js#L24)

caseworker/assets/javascripts/__tests__/list-expander.test.js
```diff
@@ -0,0 +1,97 @@
+import { fireEvent, getByLabelText, getByText } from "@testing-library/dom";
```
Harriethw, line 1: minor: looks like there's a couple of unused imports up here

caseworker/assets/javascripts/list-expander.js
kevincarrogan: I would place the SVG file in one of our media directories and then add it via CSS instead of having it as part of the HTML or the JS itself.

caseworker/assets/javascripts/list-expander.js
```diff
@@ -0,0 +1,54 @@
...
+    );
+  }
+
+  hideListElems() {
+    if (this.liElems.length <= this.visibleElems) {
+      return;
+    }
+    this.$expandButton.classList.add("expander__expand-button__visible");
+
+    for (let i = this.visibleElems; i < this.liElems.length; i++) {
```
kevincarrogan, line 25: You can use `for (let $li of this.visibleElems)` instead here.

caseworker/assets/javascripts/list-expander.js
```diff
@@ -0,0 +1,54 @@
+class ListExpander {
+  constructor($el) {
+    this.$el = $el;
+    this.$expandButton = this.$el.querySelector(".expander__expand-button");
+    this.$list = this.$el.querySelector(".expander__expand-list");
+    this.liElems = this.$list.querySelectorAll(".expander__expand-list__item");
+    this.visibleElems = parseInt(
+      this.$el.getAttribute("data-expander-visible-elems")
+    );
```
kevincarrogan, line 9: `parseInt` should be given a radix here so `parseInt(this.$el.getAttribute("data-expander-visible-elems"), 10)` because JS is quirky and sometimes does the wrong thing if you don't put in an explicit radix.

caseworker/assets/javascripts/__tests__/list-expander.test.js
```diff
@@ -0,0 +1,97 @@
...
+};
+
+const createComponent = (numItems, visibleElems) => {
+  let div = createElement(numItems, visibleElems);
+  new ListExpander(div).init();
+  return div;
+};
+
+describe("List expander", () => {
+  test("Elements below the line are not visible", async () => {
```
Harriethw, line 49: minor: I think with the exception of the last test these don't need to be async?

caseworker/assets/javascripts/__tests__/list-expander.test.js
```diff
@@ -0,0 +1,97 @@
...
+    let div = createComponent(2, 1);
+
+    const item1 = getByText(div, "Item 1");
+    expect(item1).not.toHaveClass("expander__expand-list__item__hidden");
```
kevincarrogan, line 53: I wouldn't necessarily make this change but you can also literally check for a thing being visible instead of coupling the test to a specific class https://github.com/testing-library/jest-dom#tobevisible.

caseworker/assets/styles/components/_expander.scss
```diff
@@ -0,0 +1,48 @@
+.expander {
+
+    &__expand-list {
+        &__item {
+	        &__animate {
+	        	animation: load-flag .2s;
+	        }
+
+            &__hidden {
+                display: none!important;
```
kevincarrogan, line 10: Only use `!important` if it's really necessary. I'm hoping that it's not actually necessary here.

caseworker/queues/views.py
```diff
@@ -79,6 +79,38 @@ def data(self):
     def filters(self):
         return self.data["results"]["filters"]
 
+    def _transform_destinations(self, case):
+        try:
+            destinations = case["destinations"]
+        except KeyError:
+            destinations = []
+
+        unique_destinations = [dict(t) for t in {tuple(destination["country"].items()) for destination in destinations}]
```
kevincarrogan, line 88: This is taking me a few reads to actually work out what it's doing, is there another way of writing this to make it a bit clearer.

caseworker/templates/includes/case-row-flags.html
```diff
@@ -1,15 +1,22 @@
...
+    			<li
+    				class="app-flag app-flag--{{ flag.colour }} expander__expand-list__item"
+    				{% if flag.label %}data-tooltip="{{ flag.label }}"{% endif %}>{{ flag.name }}</li>
+    		{% endfor %}
+    	{%endif%}
+    </ol>
+
+    <button class="expander__expand-button" type="button" aria-label="Show more">
+        <span></span>
+    </button>
```
kevincarrogan, line 20: Where possible we shouldn't add elements that are purely for JS. It would be better to just add these elements via the JS directly.

I reckon we could also put text in here instead of using `aria-label` and then hide it via CSS.

caseworker/assets/javascripts/__tests__/list-expander.test.js
```diff
@@ -0,0 +1,97 @@
...
+
+const createComponent = (numItems, visibleElems) => {
+  let div = createElement(numItems, visibleElems);
+  new ListExpander(div).init();
+  return div;
+};
+
+describe("List expander", () => {
+  test("Elements below the line are not visible", async () => {
+    let div = createComponent(2, 1);
```
Harriethw, line 50: really minor (sorry being nitpicky!): this could be a `const` instead of a let in the tests

caseworker/assets/javascripts/__tests__/list-expander.test.js
currycoder: This breaks my perception of how the tests work.  My understanding was that we were just instantiating small pieces of HTML in a virtual DOM rather than testing in actual web pages.  i.e. My thought was that the CSS isn't loaded for this component so we are just testing straight JS+HTML that has been formatted to the agreed contract.

caseworker/templates/includes/flags.html
```diff
@@ -1,9 +1,15 @@
...
-	<ul class="app-flags {% if list %}app-flags--list{% endif %} {{ classes }}">
+<div class="expander" data-expander-visible-elems="3">
+	<ul class="app-flags {% if list %}app-flags--list{% endif %} {{ classes }} expander__expand-list">
 		{% for flag in flags %}
-			<li data-label="{{ flag.label }}" class="app-flag {% if flag.label %}app-flag--label{% endif %} app-flag--{{ flag.colour }}"><span class="govuk-visually-hidden">{{ flag.label }}</span> {{ flag.name }}</li>
+			<li data-label="{{ flag.label }}" class="app-flag {% if flag.label %}app-flag--label{% endif %} app-flag--{{ flag.colour }} expander__expand-list__item"><span class="govuk-visually-hidden">{{ flag.label }}</span> {{ flag.name }}</li>
 		{% endfor %}
 	</ul>
+
+    <button class="expander__expand-button" type="button" aria-label="Show more">
```
Harriethw, line 9: minor suggestion: would it be worth having a shared fragment for this button?

caseworker/assets/javascripts/__tests__/list-expander.test.js
```diff
@@ -0,0 +1,97 @@
+import { fireEvent, getByLabelText, getByText } from "@testing-library/dom";
+import "@testing-library/jest-dom";
+import userEvent from "@testing-library/user-event";
+
+import { ListExpander } from "../list-expander";
+
+let form;
```
Harriethw, line 7: looks like this form var is also unused

caseworker/assets/javascripts/__tests__/list-expander.test.js
kevincarrogan: Yep. You have made a very valid point. Ignore me.

caseworker/assets/javascripts/list-expander.js
currycoder: I've ended up using `liElems.forEach()`

Harriethw (approved):

caseworker/queues/views.py
kevincarrogan: Oh. I hadn't noticed on my initial read through that this was just moved. Leave it as is then.

caseworker/assets/javascripts/list-expander.js
```diff
@@ -0,0 +1,61 @@
...
+    }
+    this.$expandButton.classList.add("expander__expand-button__visible");
+
+    const expander = this;
+    this.liElems.forEach(function ($li, currentIndex, listObj) {
+      if (currentIndex < expander.visibleElems) {
+        return;
+      }
+      $li.classList.add("expander__expand-list__item__hidden");
+    });
```
kevincarrogan, line 32: Instead of capturing `this` via `expander` you could just use a fat arrow function here and you'll have the correct `this`.

```
this.liElems.forEach(($li, currentIndex) => { console.log(this, "This is the outside scope this") })
```

kevincarrogan (approved):
lgtm

PR #1057, opened 2023-01-18
Ltd 3242 Fix BEIS Nuclear Assessment Workflow

antroy-madetech:


caseworker/advice/views.py
```diff
@@ -664,10 +670,10 @@ def get_context_data(self, **kwargs):
 
     def form_valid(self, form):
         data = {**form.cleaned_data}
-        selected_good_ids = data.pop("goods", [])
+        self.selected_good_ids = data.pop("goods", [])
```
saruniitr, line 673: I think this doesn't need to be in `self` now.

caseworker/advice/views.py
```diff
@@ -623,7 +624,12 @@ class BEISProductAssessmentView(AdviceView, BEISNuclearMixin, BEISAssessmentBase
     form_class = BEISTriggerListAssessmentForm
 
     def get_success_url(self):
-        return reverse("cases:advice_view", kwargs=self.kwargs)
+        if self.case:
+            del self.case
```
saruniitr, line 628: Please add a comment about why we need to do this

ui_tests/caseworker/conftest.py
```diff
@@ -447,7 +447,7 @@ def my_case_not_in_queue(driver, context):  # noqa
 @given("a queue has been created")  # noqa
 def create_queue(context, api_test_client):  # noqa
     api_test_client.queues.add_queue(f"queue {utils.get_formatted_date_time_y_m_d_h_s()}")
-    context.queue_id = api_test_client.context["queue_id"]
+    context.QUEUE_ID = api_test_client.context["queue_id"]
```
saruniitr, line 450: any particular reason for this change?

antroy-madetech: I think the IDE was being "smart". I'll roll these back

caseworker/advice/views.py
antroy-madetech: Done.

caseworker/advice/views.py
antroy-madetech: Done

PR #1061, opened 2023-01-24
add PR template

Harriethw:
## Aim 

add a suggested template to prompt more detail in a PR.

Should be auto available once merged into master. Documentation: https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/creating-a-pull-request-template-for-your-repository

### DIFF
--- a/docs/pull_request_template.md
+++ b/docs/pull_request_template.md
@@ -1,0 +1,12 @@
+### Aim
+
+<!--
+Enter a description of the changes in the PR here.
+
+Include any context that will help reviewers understand the reason for these changes.
+
+* Are there any dependencies on the API reviewers should be aware of?
+* Have you included any relevant screenshots in the JIRA ticket?
+-->
+
+[Link to story](insert Jira link here)


hnryjmes:
I'm happy with the template as it is, is this ready to leave draft status now?

docs/pull_request_template.md
```diff
@@ -0,0 +1,12 @@
+### Aim
```
saruniitr, line 1: Change description instead of Aim?

docs/pull_request_template.md
```diff
@@ -0,0 +1,12 @@
...
+
+<!--
+Enter a description of the changes in the PR here.
+Include any context that will help reviewers understand the reason for these changes.
+-->
+
+### Screenshots (if relevant)
```
saruniitr, line 8: We decided not to add screenshots so this is not required.

Should we ask about tests or any dependencies etc?

Harriethw: why do we not add screenshots for frontend changes? Personally I find them really useful to see instantly what's intended...

docs/pull_request_template.md
Harriethw: The reason that I put 'Aim' is to ensure the work and description explain what we want the outcome for the user/dev experience to be, rather than a list of technical changes. It's really helpful to have that kind of context when reviewing a PR imo, because the context is just as important as what's been added.

e.g. 
```
we need to add a tab on the cases list page for users to see cases assigned to them, so that they can quickly address any issues on them. This PR adds the tab and refactors the code so that we can more easily add more in the future.
```

docs/pull_request_template.md
Harriethw: re: asking about tests, dependencies. I feel test coverage should be implicit in what we expect, and there are codecov tools to regulate the basics of coverage.

could be a good one to ask about API dependencies though as this has been the case for my last couple of PRs.

docs/pull_request_template.md
```diff
@@ -0,0 +1,12 @@
+### Aim
+
+<!--
+Enter a description of the changes in the PR here.
+
+Include any context that will help reviewers understand the reason for these changes.
+
+* Are there any dependencies on the API reviewers should be aware of?
+* Have you included any relevant screenshots in the JIRA ticket?
```
hnryjmes, line 9: Slightly prefer `Jira` to `JIRA`

Harriethw: oh yeah it's not an acronym...so used to acronyms everywhere 😆

kevincarrogan (approved):

hnryjmes (approved):

PR #1062, opened 2023-01-25
LTD-3269-assign-case-worker-form 

depsiatwal:
This is a new form to support bulk case worker assignment form the queue page.

The current assign caseworker view only supports a single case. This is a new form/view which doesn't replace this page but a new form to plugin. The page is based on  django-forms unlike the old one which is lite-forms hence some common components moving to core to reuse. 

This is still work in progress and this page isn't accessible unless url is known. The next PR will integrate the whole flow.


caseworker/cases/services.py
```diff
@@ -320,6 +320,12 @@ def put_case_officer(request, pk, json):
     return data.json(), data.status_code
 
 
+def update_case_officer_on_cases(request, case_ids, user_id):
+    for case_id in case_ids:
+        client.put(request, f"/cases/{case_id}/case-officer/", {"gov_user_pk": user_id})
```
saruniitr, line 325: Few comments,
 - we are not checking the response
 - if it fails in between then we would've updated partially
 - multiple calls
May be best to have a separate endpoint that can accept list of case_ids so that everything can be updated in a single call.

depsiatwal: this is not active I can do this in 2 phases 
- error checking this end point 
- if time allows make a API change for single end point change

depsiatwal: Done with API update also :-)

caseworker/cases/services.py
```diff
@@ -320,6 +320,11 @@ def put_case_officer(request, pk, json):
     return data.json(), data.status_code
 
 
+def update_case_officer_on_cases(request, case_ids, user_id):
+    data = client.put(request, "/cases/cases-update-case-officer/", {"gov_user_pk": user_id, "case_ids": case_ids})
```
saruniitr, line 324: `response.raise_for_status()` required before accessing the response

caseworker/templates/queues/select_role_case_allocation.html
```diff
@@ -0,0 +1,20 @@
+{% extends 'layouts/base.html' %}
+{% load crispy_forms_tags %}
+
+{% block title %}
+	Which role do you want to allocate?
+{% endblock %}
+
+{% block body %}
+<div class="govuk-width-container">
+    <a href="">Back</a>
```
saruniitr, line 10: this is not pointing anywhere?

saruniitr (approved):

PR #1063, opened 2023-01-27
Build backend e2e image as part of CI

currycoder:
This change ensures that the circleci jobs for running our E2E tests build their own docker container for lite-api.

The new way for running these jobs is as follows:
- Pull down lite-api (with the specified API branch) and submodules.
- Build the liteapi docker container locally.
- Run the E2E test suite using this locally created docker container.

Note that we now have one circleci job per API branch (`dev`/`uat`/`master`); the idea is that for every lite-frontend PR we run the E2E suite against *all* of these API branches.  We will only require a pass for the applicable branch though. i.e. A PR to make a change to lite-frontend `dev` only requires that the jobs for E2E tests running against dev API pass.

The sequence for merging this should be as follows:
- Merge companion lite-api PR (which adjusts `Dockerfile.e2e` in that repository); https://github.com/uktrade/lite-api/pull/1261
    - Ensure this change is present across `dev`/`uat`/`master` lite-api branches.
- Remove the `Dockerfile.e2e` which is present in this PR.
- Merge this PR to `master` and then push this down to `uat`/`dev` branches.
- Adjust branch protection rules to ensure that `dev`/`uat`/`master` branches require the appropriate E2E jobs.


.circleci/config.yml
```diff
@@ -108,6 +108,18 @@ commands:
       - store_test_results:
           path: test-results
 
+  build_api_image:
+    steps:
+      - run: git clone git@github.com:uktrade/lite-api.git; cd lite-api; git checkout dev
```
kevincarrogan, line 113: Where does this checkout to? Are you currently in the `lite-frontend` codebase here? Could that mess anything up having this codebase embedded in the other one in some way?

Dockerfile-api.e2e
```diff
@@ -0,0 +1,31 @@
+FROM python:3.7-slim
```
currycoder, line 1: Note that this file will be deleted before we merge this PR.  It's just created here for convenience to work on these changes quickly.

.circleci/config.yml
currycoder: Yes that's what happens. I don't *think* that's an issue, but I can't quite work out how to check this out as a sibling of `lite-frontend`.  Happy to take suggestions here ofc

kevincarrogan (approved):

PR #1064, opened 2023-01-27
LTD-3239: Add a 'my cases' tab/filter

Harriethw:
### aim

[LTD-3239](https://uktrade.atlassian.net/jira/software/c/projects/LTD/boards/204?modal=detail&selectedIssue=LTD-3239)

The frontend for the cases/queues view is going to have several tabs that aim to filter out cases. We have already introduced the 'all cases' and 'open queries' tabs, but now that one more is being added ('my cases') we need to change the way the logic for filtering is handled (it's not either/or anymore).

[Together with this backend PR](https://github.com/uktrade/lite-api/pull/1260) I'm proposing:
 
_on the frontend_
1. instead of a param for each tab (e.g. `only_open_queries`) we have a param whose value is the selected tab e.g. `selected_tab=open_queries`
2. in the view, to get a count for each tab, we the tab names and make a HEAD request to get the count like we do currently, and save that in a dict for each tab, as well as its URL and whether it is selected.

_on the backend_
1. the view takes a look at the `selected_tab` value and passes the manager `True` for the name of the selected tab. e.g. `open_queries=True`
2. the manager can then apply the right filter. 

Would be great to get feedback on the structure/naming of this approach.

[LTD-3239]: https://uktrade.atlassian.net/browse/LTD-3239?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/caseworker/queues/views.py
+++ b/caseworker/queues/views.py
@@ -80,43 +80,42 @@ class Cases(LoginRequiredMixin, TemplateView):
 
         params["flags"] = self.request.GET.getlist("flags[]", [])
 
+        params["selected_tab"] = self.request.GET.get("selected_tab", CasesListPage.Tabs.ALL_CASES)
         # if the hidden param is not true
         # then cases with open queries are filtered out on team queue views
-        only_open_queries = self.request.GET.get("only_open_queries")
-        if only_open_queries == "True":
+        if (
+            params["selected_tab"] == CasesListPage.Tabs.MY_CASES
+            or params["selected_tab"] == CasesListPage.Tabs.OPEN_QUERIES
+        ):
             params["hidden"] = "True"
 
         return params
 
-    def get_open_queries_count(self, only_open_queries):
-        # we need to include all the existing params,
-        # except we want the opposite of the current value for only_open_queries
+    def _get_tab_url(self, tab_name):
+        params = self.request.GET.copy()
+        params["selected_tab"] = tab_name
+        return f"?{params.urlencode()}"
+
+    def _get_tab_count(self, tab_name):
         params = self.get_params()
-        params["only_open_queries"] = only_open_queries
-        count = head_cases_search_count(self.request, self.queue_pk, params)
-        return count
+        params["selected_tab"] = tab_name
+        if tab_name != CasesListPage.Tabs.ALL_CASES:
+            params["hidden"] = "True"
 
-    def open_queries_tabs(self):
-        only_open_queries = self.request.GET.get("only_open_queries") or "False"
+        return head_cases_search_count(self.request, self.queue_pk, params)
 
-        params_with_queries = self.request.GET.copy()
-        params_with_queries["only_open_queries"] = "True"
-        params_all_cases = self.request.GET.copy()
-        params_all_cases["only_open_queries"] = "False"
+    def _tab_data(self):
+        selected_tab = self.request.GET.get("selected_tab", CasesListPage.Tabs.ALL_CASES)
+        tab_data = {}
 
-        open_queries_url = "{}?{}".format(self.request.path, params_with_queries.urlencode())
-        all_cases_url = "{}?{}".format(self.request.path, params_all_cases.urlencode())
+        for tab in CasesListPage.Tabs:
+            tab_data[tab] = {
+                "count": self._get_tab_count(tab),
+                "is_selected": selected_tab == tab,
+                "url": self._get_tab_url(tab.value),
+            }
 
-        unselected_tab_count = self.get_open_queries_count("True" if only_open_queries == "False" else "False")
-
-        open_queries_tabs = {
-            "only_open_queries": only_open_queries,
-            "open_queries_url": open_queries_url,
-            "all_cases_url": all_cases_url,
-            "unselected_tab_count": unselected_tab_count,
-        }
-
-        return open_queries_tabs
+        return tab_data
 
     def get_context_data(self, *args, **kwargs):
 
@@ -151,7 +150,7 @@ class Cases(LoginRequiredMixin, TemplateView):
             "enforcement_check": Permission.ENFORCEMENT_CHECK.value in get_user_permissions(self.request),
             "updated_cases_banner_queue_id": UPDATED_CASES_QUEUE_ID,
             "show_updated_cases_banner": show_updated_cases_banner,
-            "open_queries_tabs": self.open_queries_tabs,
+            "tab_data": self._tab_data(),
         }
 
         return super().get_context_data(*args, **context, **kwargs)
--- a/caseworker/templates/queues/search-tabs.html
+++ b/caseworker/templates/queues/search-tabs.html
@@ -1,18 +1,18 @@
 <div class="lite-tabs__container">
-  <div class="lite-tabs">
-    <a href="{{ tab_data.all_cases.url }}" class="lite-tabs__tab  {% if tab_data.all_cases.is_selected %}lite-tabs__tab--selected{% endif %}" id="all-cases-tab">
-      {% if queue.is_system_queue %}
+	<div class="lite-tabs">
+		<a href="{{ tab_data.all_cases.url }}" class="lite-tabs__tab  {% if tab_data.all_cases.is_selected %}lite-tabs__tab--selected{% endif %}" id="all-cases-tab">
+			{% if queue.is_system_queue %}
 					All cases
 				{% else %}
 					Cases to review
 			{% endif %}
       ({{  tab_data.all_cases.count }})
     </a>
-    <a href="{{ tab_data.my_cases.url }}" class="lite-tabs__tab {% if tab_data.my_cases.is_selected %}lite-tabs__tab--selected{% endif %}" id="my-cases-tab">
+		<a href="{{ tab_data.my_cases.url }}" class="lite-tabs__tab {% if tab_data.my_cases.is_selected %}lite-tabs__tab--selected{% endif %}" id="my-cases-tab">
 				My cases ({{ tab_data.my_cases.count }})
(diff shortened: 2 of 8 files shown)


caseworker/queues/views.py
```diff
@@ -77,43 +77,43 @@ def get_params(self):
...
-        params["only_open_queries"] = only_open_queries
-        count = head_cases_search_count(self.request, self.queue_pk, params)
-        return count
-
-    def open_queries_tabs(self):
-        only_open_queries = self.request.GET.get("only_open_queries") or "False"
+    def _get_tab_url(self, tab_name):
+        params = self.request.GET.copy()
+        params["selected_tab"] = tab_name
+        return "{}?{}".format(self.request.path, params.urlencode())
```
kevincarrogan, line 94: I think you can leave off `self.request.path` here and just have the `?foo=bar` and it will work just the same.

caseworker/queues/views.py
```diff
@@ -77,43 +77,43 @@ def get_params(self):
...
         params["flags"] = self.request.GET.getlist("flags[]", [])
 
+        params["selected_tab"] = self.request.GET.get("selected_tab", CasesListPage.Tabs.ALL_CASES.value)
         # if the hidden param is not true
         # then cases with open queries are filtered out on team queue views
-        only_open_queries = self.request.GET.get("only_open_queries")
-        if only_open_queries == "True":
+        if (
+            params["selected_tab"] == CasesListPage.Tabs.MY_CASES.value
+            or params["selected_tab"] == CasesListPage.Tabs.OPEN_QUERIES.value
```
kevincarrogan, line 85: If you changed the definition of the `Tabs` enum to `class Tabs(str, Enum)` then you don't need to always prepend `.value`.

caseworker/queues/views.py
```diff
@@ -77,43 +77,43 @@ def get_params(self):
...
 
-        params_with_queries = self.request.GET.copy()
-        params_with_queries["only_open_queries"] = "True"
-        params_all_cases = self.request.GET.copy()
-        params_all_cases["only_open_queries"] = "False"
+    def _get_tab_count(self, tab_name):
+        params = self.get_params()
+        params["selected_tab"] = tab_name
+        if tab_name != CasesListPage.Tabs.ALL_CASES.value:
+            params["hidden"] = "True"
```
kevincarrogan, line 100: I don't really know what is being `hidden`. Is there a better string that we could use here to give an indicator of what's being hidden?

unit_tests/caseworker/queues/test_views.py
```diff
@@ -100,11 +98,8 @@ def mock_team_queue(requests_mock, data_queue):
 @pytest.fixture
 def mock_team_cases(requests_mock, data_cases_search):
     data_cases_search["results"]["filters"]["is_system_queue"] = False
-    encoded_params = parse.urlencode(
-        {"queue_id": queue_pk, "page": 1, "only_open_queries": True, "hidden": True}, doseq=True
-    )
-    url = client._build_absolute_uri(f"/cases/?{encoded_params}")
-    yield requests_mock.get(url=url, json=data_cases_search)
+    url = client._build_absolute_uri(f"/cases/")
+    yield requests_mock.get(url=re.compile(f"{url}.*"), json=data_cases_search)
```
kevincarrogan, line 102: This wildcard match feels like it might end up matching more than we actually want it to. It looks like it originally was matching against anything with params but now it's going to match anything that begins with `/cases/` which is a lot more than before.

unit_tests/caseworker/queues/test_views.py
```diff
@@ -209,68 +207,90 @@ def test_trigger_list_checkbox_visible_checked(authorized_client):
...
-    assert "/?only_open_queries=True" in open_queries_tab.attrs["href"]
-    assert "lite-tabs__tab--selected" in open_queries_tab.attrs["class"]
+    assert tab_text in selected_tab.get_text()
+    assert f"/?selected_tab={tab_name}" in selected_tab.attrs["href"]
+    assert "lite-tabs__tab--selected" in selected_tab.attrs["class"]
     assert mock_team_cases.last_request.qs == {
         "queue_id": [queue_pk],
         "page": ["1"],
-        "only_open_queries": ["true"],
+        "selected_tab": [f"{tab_name}"],
```
kevincarrogan, line 294: This doesn't need to be an f-string, you can just put `tab_name` here directly.

unit_tests/caseworker/queues/test_views.py
```diff
@@ -182,16 +178,18 @@ def test_cases_home_page_trigger_list_search(authorized_client, mock_cases_searc
...
 def test_cases_home_page_regime_entry_search(authorized_client, mock_cases_search):
-    url = reverse("queues:cases") + "?regime_entry=af8043ee-6657-4d4b-83a2-f1a5cdd016ed"
+    url = reverse("queues:cases") + "?regime_entry=af8043ee-6657-4d4b-83a2-f1a5cdd016ed"  # /PS-IGNORE
     authorized_client.get(url)
     assert mock_cases_search.last_request.qs == {
         "queue_id": ["00000000-0000-0000-0000-000000000001"],
         "page": ["1"],
-        "regime_entry": ["af8043ee-6657-4d4b-83a2-f1a5cdd016ed"],
+        "regime_entry": ["af8043ee-6657-4d4b-83a2-f1a5cdd016ed"],  # /PS-IGNORE
+        "selected_tab": ["all_cases"],
```
kevincarrogan, line 192: It looks like this is a default value that is repeated often, I wonder if we could extract it out so that if we changed this then we wouldn't have to find each test and update it.

unit_tests/caseworker/queues/test_views.py
```diff
@@ -209,68 +207,90 @@ def test_trigger_list_checkbox_visible_checked(authorized_client):
...
+def test_tabs_with_all_cases_default(authorized_client, mock_cases_search_head):
     response = authorized_client.get(reverse("core:index"))
     html = BeautifulSoup(response.content, "html.parser")
-    all_queries_tab = html.find(id="all-queries-tab")
-    open_queries_tab = html.find(id="open-queries-tab")
+    all_queries_tab = html.find(id="all-cases-tab")
 
     assert "All cases" in all_queries_tab.get_text()
-    assert "(2)" in all_queries_tab.get_text()
-    assert "(350)" in open_queries_tab.get_text()
```
kevincarrogan, line 220: Why have we decided not to check for the number being displayed?

caseworker/queues/views.py
Harriethw: I agree `hidden` is not the best named variable, but it felt out of the scope of this ticket to change it as it's a legacy param and there is code on frontend and backend around it...

unit_tests/caseworker/queues/test_views.py
Harriethw: it will only be triggered when it's called by a test that wants to use it though right? I wanted to make it purposefully broad because I check what it's been called with in the test (which could be various different combos of params)... before afaik it was checking with a set combo of params, and wouldn't kick if called with a variation of those params.

kevincarrogan: I _think_ a path match would work here still anyway without needing to put in the wildcard because you're just checking that the `/cases/` URL has been hit right?

unit_tests/caseworker/queues/test_views.py
Harriethw: hmm I took it out because the mock HEAD request to get the count was returning the same count, and it didn't seem valuable to test that it renders on the template because we check the correct params are sent to the API... but I guess still useful to have coverage that the template gets a count from the context

unit_tests/caseworker/queues/test_views.py
kevincarrogan: Hit with certain parameters that is and I believe a path match will still pick those up.

Harriethw: ah I see what you mean, I thought I needed the wildcard in to match the params...will give it a go without them

caseworker/queues/views.py
saruniitr: There is a checkbox in the UI to show hidden cases, it could be related to that.
When a case has open query then it is hidden.

Harriethw: Yep exactly, it is related to the checkbox and if it were NOT true we wouldn't get open queries back. I have put a comment in the past to try and clarify it as it's not intuitive:
``` 
# if the hidden param is not true
# then cases with open queries are filtered out on team queue views
```

caseworker/templates/queues/search-tabs.html
```diff
@@ -0,0 +1,18 @@
...
+				{% else %}
+					Cases to review
+			{% endif %}
+      ({{  tab_data.all_cases.count }})
+    </a>
+    <a href="{{ tab_data.my_cases.url }}" class="lite-tabs__tab {% if tab_data.my_cases.is_selected %}lite-tabs__tab--selected{% endif %}" id="my-cases-tab">
+				My cases ({{ tab_data.my_cases.count }})
+    </a>
+    <a href="{{ tab_data.open_queries.url }}" class="lite-tabs__tab {% if tab_data.open_queries.is_selected %}lite-tabs__tab--selected{% endif %}" id="open-queries-tab">
+				Open queries ({{  tab_data.open_queries.count }})
```
saruniitr, line 15: not sure if it is the rendering in github, formatting is off.
As this is a new file we can perhaps run a formatter on this?

saruniitr (approved):

PR #1076, opened 2023-01-31
Determine target branch based off current branch

kevincarrogan:


.circleci/config.yml
```diff
@@ -111,10 +111,17 @@ commands:
...
-        default: master
+        default: auto
         type: string
     steps:
-      - run: git clone git@github.com:uktrade/lite-api.git; cd lite-api; git checkout <<parameters.api_branch>>
+      - run:
+          name: Clone API
+          command: |
+            echo "Checking out"
+            python which_branch.py <<parameters.api_branch>>
```
currycoder, line 121: Is this needed anymore?

which_branch.py
```diff
@@ -0,0 +1,54 @@
...
+    return os.environ.get("CIRCLE_BRANCH", "dev")
+
+
+BASE_TARGET_MAP = {
+    "dev": "uat",
+    "uat": "master",
+}
+
+
+HOTFIX_BRANCH_RE = re.compile("^hotfix-(.*)-.*$")
```
currycoder, line 16: So we need to agree in our hotfix process to prefix branch names with "hotfix-" right?

currycoder: Ah actually "hotfix-{uat|master}"..?

currycoder: I like this as the complexity is around the exceptional case rather than the regular one

Harriethw (approved):

currycoder (approved):

PR #1078, opened 2023-02-01
LTD-3088-allocated-case-officer-assigned-user 

depsiatwal:
This connects the flow to add new LU case advisors and case advisor to multiple cases from the queue view. 

This contains multiple changes 
- New select role form 
- validation message changes 
- copy changes


caseworker/queues/forms.py
```diff
@@ -1,5 +1,6 @@
 from django import forms
 from django.urls import reverse_lazy
+from django.db import models
```
currycoder, line 3: Do we need this?  I wouldn't expect to see this anywhere in the FE TBH

exporter/core/organisation/views.py
```diff
@@ -98,7 +98,7 @@ def is_individual(self):
 
 
 class SelectOrganisation(LoginRequiredMixin, FormView):
-    form_class = SelectOrganisationForm
+    form_class = buttSelectOrganisationForm
```
currycoder, line 101: `butt`?

caseworker/queues/views.py
```diff
@@ -188,6 +188,30 @@ def init(self, request, **kwargs):
         self.success_url = reverse_lazy("queues:manage")
 
 
+class CaseAssignmentAllocateRole(LoginRequiredMixin, FormView):
```
currycoder, line 191: Could there be some tests for this page?

PR #1080, opened 2023-02-01
Amend Nuclear Cooperation Agreement question and error message

hnryjmes:
BEIS Nuclear have confirmed they need this amendment to the question text

Error message is also changing to match

### DIFF
--- a/caseworker/advice/forms.py
+++ b/caseworker/advice/forms.py
@@ -271,9 +271,9 @@ class BEISTriggerListFormBase(forms.Form):
         choices=NCA_CHOICES,
         coerce=coerce_str_to_bool,
         error_messages={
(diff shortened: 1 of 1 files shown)


caseworker/advice/forms.py
```diff
@@ -273,7 +273,7 @@ class BEISTriggerListFormBase(forms.Form):
         error_messages={
             "required": "Select yes if the Nuclear Cooperation Agreement applies to the product",
```
saruniitr, line 274: I wonder if we need to edit the error message also similar to the change below, could you get that clarified?

hnryjmes: good plan, I have commented on the ticket

hnryjmes: resolved

saruniitr (approved):

PR #1081, opened 2023-02-01
Frontend build

kevincarrogan:
This changes how we build the frontend assets both in development and on production.

The changes are:

  - Built assets are no longer committed to the repository and the previous ones have been removed
  - A docker container runs the `npm watch` command automatically
  - The frontend assets for production are built as part of the deployment process


depsiatwal (approved):
lgtm

.circleci/config.yml
```diff
@@ -36,6 +36,9 @@ image_python_browsers: &image_python_browsers
 image_redis: &image_redis
   image: circleci/redis:3.2-alpine
 
+image_node: &image_node
+  image: cimg/node:16.19
```
currycoder, line 40: We are jumping ahead some node versions in comparison to what our assets are currently built with locally.  Is that ok?

README.md
```diff
@@ -195,7 +195,13 @@ PIPENV_DOTENV_LOCATION=caseworker.env ENVIRONMENT=local pipenv run pytest -m "ru
 
 ## Javascript/SCSS
 
-Node version required is 16.5.0
+The Javascript/SCSS is automatically watched via the node docker service.
+
+The production assets are built via Jenkins on deployment.
```
currycoder, line 200: Where though?

manifest.yml
```diff
@@ -1,6 +1,7 @@
 ---
 applications:
   - buildpacks:
+      - https://github.com/cloudfoundry/nodejs-buildpack.git
```
currycoder, line 4: Ah is there something special where having this will ensure that the jenkins job does the build for you?  How does that work?

.circleci/config.yml
kevincarrogan: I've been chucking this on `demo` to check that everything is fine.

I've had to increase the version because the cloudfoundry buildpacks only have a limited set of versions they work with (the same as our Python version), this was the version that seemed to work across the board.

README.md
kevincarrogan: As in where are the production assets built?

manifest.yml
kevincarrogan: Yes. This is a cloudfoundry thing. It's the same as the Python buildpack thing really.

You specify the buildpacks and they know to look for things like pip files or package.json files. The extra bit of magic is the `heroku-postbuild` setting under `scripts` in the `package.json` file.

currycoder: ... Cloudfoundry runs `heroku-postbuild` for you?  The world of JS is fun.

Sooo just to fully trace this through, this will get run at a step when the machine running our application starts up and before CF sends traffic to it?  Just thinking about whether it's possible for the machine to receive traffic before the build is done.
Do we have this running anywhere at the moment? e.g. a manual deploy to one of our environments?

README.md
currycoder: Sorry the other comment has covered this.  Although it could be nice to expand just a touch more about *how* jenkins does this.  If that's easier with links then cool, but something more than "jenkins does it's thing" would be good

manifest.yml
currycoder: Great explainer, thanks

currycoder (approved):

PR #1082, opened 2023-02-02
small advice copy update

MarkThomasLite:


caseworker/tau/forms.py
```diff
@@ -289,7 +289,10 @@ def __init__(
         self.fields["goods"] = forms.MultipleChoiceField(
             choices=self.get_goods_choices(goods),
             widget=GoodsMultipleSelect(),
-            label="Select a product to begin. Or you can select multiple products to give them the same assessment.",
+            label=(
+                "Select a product to begin. Or you can select multiple products to give them the same assessment.\n"
+                "You will then be asked to make a recommendation for all products on this application."
+            ),
```
saruniitr, line 295: This should not be changed.
This is changing the TAU form whereas the content amendments required only for BEIS team.

We are using similar designs as TAU but the purpose is different.

saruniitr (changes requested):
Need to remove the changes done to TAU form which are not required.

saruniitr (approved):

PR #1084, opened 2023-02-02
Remove playwright

kevincarrogan:
The usage of playwright was an experiment to try to move away from Selenium but the end result was to stick with Selenium

We won't be attempting to move to playright any time soon so this removes any of the remnants of that experiment


depsiatwal (approved):
lgtm

PR #1085, opened 2023-02-03
LTD-3270-search-filter-on-case-office-allocation 

depsiatwal:
This add on-page filtering to the new form that allows case officer to be selected on multiple pages.


core/forms/templates/forms/filter-radios.html
```diff
@@ -0,0 +1,13 @@
+{% load svg static %}
+
+<div class="govuk-form-group">
+	<div class="lite-search__container" style="max-width: 400px; display: none;">
```
currycoder, line 4: I feel like the style shouldn't be inline here and should probably go in the relevant SCSS.  What do you think?

Harriethw (approved):
LGTM from what I can see! though I would second Brendan's comment about the inline style

core/forms/templates/forms/filter-radios.html
```diff
@@ -0,0 +1,13 @@
...
+	    <input type="search" class="govuk-input" id="filter-box" placeholder="{{ placeholder }}" autocomplete="off" autocorrect="off" autocapitalize="off" spellcheck="false">
+		<button class="lite-search__button lite-search__button--instant-search" type="submit">
+			{% svg 'search' %}
+			<span>Search</span>
+		</button>
+	</div>
+</div>
```
kevincarrogan, line 11: I think this is just a copy-and-paste from `filter-checkboxes.html` we could probably split this out into a shared include here.

depsiatwal: @kevincarrogan I don't know how to do this 
something we want to do as part of this PR ?

saruniitr (approved):

PR #1087, opened 2023-02-04
Slimline npm

kevincarrogan:
Due to an issue where some deployments ran out of disk space, this clears out any unnecessary files/data that are no longer required after the frontend asset build has completed.


depsiatwal (approved):
lgtm

.nvmrc
```diff
@@ -1 +0,0 @@
-12.16.1
```
currycoder, line 1: Isn't this sorta helpful for people running locally?

package.json
```diff
@@ -11,15 +11,15 @@
     "watch": "npm-run-all --parallel watch_exporter watch_caseworker",
     "build": "npm install --no-save && npm run build_all",
     "build_all": "export NODE_ENV=production && npm-run-all --serial build_exporter build_caseworker",
-    "heroku-postbuild": "npm run build",
+    "heroku-postbuild": "npm run build && rm -r node_modules",
```
currycoder, line 14: What sort of space savings is this granting us?

.nvmrc
kevincarrogan: No. Because it’s wrong.

package.json
kevincarrogan: Enough that I don’t have to bump the instance up to 8G. This allows it to build at the current spec without me making any changes.

.nvmrc
kevincarrogan: Most things seem to pick this up from the engines directive anyway. The change to have a node docker container also kind of solves this. 

I deleted this because it seems that the build would see this and install this version of node and then would later see a different version in the package file and install another version of node. 

Removing this means it only installs it once. I wasn’t sure if the downloading and installing left some stuff lingering around so remove this was another attempt to save some space.

PR #1091, opened 2023-02-06
[LTD-3234] Add case adviser removal functionality to case detail page

currycoder:
This change adds the ability for case adviser assignments to be removed from a case.  It depends on the lite-api PR: https://github.com/uktrade/lite-api/pull/1276

Screenshots can be found in a comment on the Jira ticket: https://uktrade.atlassian.net/browse/LTD-3234

### DIFF
--- a/caseworker/templates/case/slices/summary.html
+++ b/caseworker/templates/case/slices/summary.html
@@ -179,8 +179,10 @@
 				<ol class="govuk-list govuk-!-margin-0">
 					{% for user in users %}
 						<li>
-							<a href="{% url 'users:user' user.id %}?return_to={{ CURRENT_PATH }}&return_to_text={{ case.reference_code }}" class="govuk-link govuk-link--no-visited-state">{{ user|username }}</a>
-                            <a href="{% url 'cases:remove-case-assignment' queue.id case.id %}?assignment_id={{user.assignment_id}}" class="govuk-link govuk-link--no-visited-state app-case__summary-list__value__item-action">Remove</a>
+                            <div class="app-case__summary-list__value__fullwidth">
+							    <a href="{% url 'users:user' user.id %}?return_to={{ CURRENT_PATH }}&return_to_text={{ case.reference_code }}" class="govuk-link govuk-link--no-visited-state">{{ user|username }}</a>
+                                <a href="{% url 'cases:remove-case-assignment' queue.id case.id %}?assignment_id={{user.assignment_id}}" class="govuk-link govuk-link--no-visited-state app-case__summary-list__value__item-action">Remove</a>
+                            </div>
 						</li>
 					{% endfor %}
 				</ol>
--- a/caseworker/templates/case/slices/summary.html
+++ b/caseworker/templates/case/slices/summary.html
@@ -180,6 +180,7 @@
 					{% for user in users %}
 						<li>
 							<a href="{% url 'users:user' user.id %}?return_to={{ CURRENT_PATH }}&return_to_text={{ case.reference_code }}" class="govuk-link govuk-link--no-visited-state">{{ user|username }}</a>
+                            <a href="{% url 'cases:remove-case-assignment' queue.id case.id %}?assignment_id={{user.assignment_id}}" class="govuk-link govuk-link--no-visited-state app-case__summary-list__value__item-action">Remove</a>
 						</li>
 					{% endfor %}
 				</ol>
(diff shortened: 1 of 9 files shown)


caseworker/templates/case/slices/summary.html
```diff
@@ -188,11 +189,11 @@
...
-		<dd class="app-case__summary-list__actions">
-			<a id="link-change-assigned-users" class="govuk-link govuk-link--no-visited-state" href="{% url 'cases:assign_user' queue.id case.id %}">
-				{% lcs 'generic.CHANGE' %}
-			</a>
+            <div class="app-case__summary-list__value__actions">
+			    <a id="link-change-assigned-users" class="govuk-link govuk-link--no-visited-state" href="{% url 'cases:assign_user' queue.id case.id %}">
+			    	{% lcs 'cases.CasePage.DetailsTab.ADD_ASSIGNED_USER' %}
+			    </a>
+            </div>
 		</dd>
```
antroy-madetech, line 197: What are these changes for?

caseworker/queues/services.py
```diff
@@ -101,6 +101,11 @@ def put_queue_single_case_assignment(request, pk, json):
         return {"errors": {"queue": [AssignUserPage.QUEUE_ERROR_MESSAGE]}}, HTTPStatus.BAD_REQUEST
 
 
+def delete_queue_case_assignment(request, queue_id, assignment_id):
+    response = client.delete(request, f"/queues/{queue_id}/case-assignments/{assignment_id}")
+    return response.json(), response
+
```
antroy-madetech, line 107: Why do this rather than just returning the response? Doesn't look like the first part is actually used anywhere.

currycoder: Good question.  I did it to be consistent with the other service functions around it; e.g. there is a pattern here where these functions return a 2 item tuple of JSON + status...  Except I broke the pattern by returning response in the second item as that's better IMO :D

But yes, I could see us potentially using the JSON here in future so feel like it's a good thing to respond with.  What do you think?

caseworker/templates/case/slices/summary.html
currycoder: Probably easier to refer to the screenshots in the Jira for this one..

So if you take a look at these boxes on the case detail page in chrome debugger, you might notice that they generally have 2 full length columns; list of items on the left, available actions on the right.  But there wasn't an easy way to line these columns up in rows - e.g. see the remove link in the right column next to the correct adviser in the left column...

So I added a way for the column to be full width and have actions inline next to the items.  That's what this change accommodates.

unit_tests/caseworker/cases/views/test_case_assignments.py
```diff
@@ -0,0 +1,97 @@
...
+            {
+                "id": data_assignment["user"]["id"],
+                "first_name": data_assignment["user"]["first_name"],
+                "last_name": data_assignment["user"]["last_name"],
+                "email": data_assignment["user"]["email"],
+                "assignment_id": data_assignment["id"],
+            }
+        ]
+    }
+    yield requests_mock.get(url=url, json=data_standard_case)
```
kevincarrogan, line 43: Could just be `return` doesn't need to be a `yield`.

kevincarrogan (commented):
I feel there are a couple of missing tests/assertions here.

  - Testing for the error case when the API doesn't respond as expected.
  - Testing that the `messages` is populated (and maybe displayed).

unit_tests/caseworker/cases/views/test_case_assignments.py
```diff
@@ -0,0 +1,97 @@
...
+    mock_remove_assignment,
+    mock_queue,
+):
+
+    case = data_standard_case
+    url = reverse("cases:remove-case-assignment", kwargs={"queue_pk": data_queue["id"], "pk": case["case"]["id"]})
+    response = authorized_client.post(url, data={"assignment_id": str(data_assignment["id"])})
+    assert response.status_code == 302
+    assert response.url == f"/queues/{data_queue['id']}/cases/{data_standard_case['case']['id']}/"
+    assert mock_remove_assignment.called
```
kevincarrogan, line 96: Could we check that is was called with the right arguments?

kevincarrogan: Or are we using the URL in the mock to inform us of that?

currycoder: Welll, I'll check that the path used was as expected.  I agree it's better to make that check explicit than depend on the URL in the mock.

caseworker/queues/services.py
antroy-madetech: Personally think it better to just return the response. I can see why you'd return json and status, as you can't get either from each other. I think you are right that returning the response is more useful, but you can easily get the json from that so feels redundant returning it separately.

antroy-madetech (approved):
Comment left about the return values. Will leave it to you to decide - it's only a minor thing.

kevincarrogan (approved):

caseworker/queues/services.py
currycoder: Yeah you've sold me.  Have updated this - thanks!

PR #1092, opened 2023-02-07
add case activity to cases page

MarkThomasLite:
add recent activity column to cases page
-enhanced cases templates to include extra data from the api

compared getting data via api using several queries and this appeared to be the quickest when run on local environment.
We needed last 2 audit activities per case but the case_id could be an action object or target object. Alternatives tried:

using union to get correct results from the fb
using filter to get last 2 activities per action and target (total 4) and filtering down to 2
a few different methods of massaging the data after retrieving results

### DIFF
--- a/caseworker/templates/includes/case-row.html
+++ b/caseworker/templates/includes/case-row.html
@@ -57,6 +57,49 @@
 		<p class="govuk-tag govuk-tag--grey govuk-!-margin-0 govuk-!-margin-top-2">{{ case.status.value }}</p>
 	</td>
 
+    {# case updates #}
+	<td class="govuk-table__cell lite-mobile-hide">
+		<ul class="app-updates__list">
+			{% if case.activity_updates %}
+				{% for update in case.activity_updates %}
+					<li class="app-updates__item">
+						<div class="app-updates__action">
+							{% if forloop.first %}
+								Latest Action
+							{% else %}
+								Previous Action
+							{% endif %}
+						</div>
+					</li>
+					<li class="app-updates__item">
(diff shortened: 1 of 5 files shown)


.pre-commit-config.yaml
```diff
@@ -1,6 +1,6 @@
 exclude: (exporter/assets/built/|caseworker/assets/built/)
 default_language_version:
-  python: python3.7
+  python: python3.9
```
depsiatwal, line 3: was this python version intended could cause issues for some dev may be worth having a separate PR for this

MarkThomasLite: restored, I use this config locally as the 3.7 interpreter run during pre-commit has stopped working the past couple weeks

caseworker/templates/includes/case-row.html
```diff
@@ -57,6 +57,49 @@
...
+							{{ update.user.first_name }} {{ update.user.last_name }} {{ update.text|linebreaksbr }}
+						</div>
+						{% if activity.additional_text %}
+							<div class="app-updates__activity">
+								{{ update.additional_text }}
+							</div>
+						{% endif %}
+					</li>
+					<li class="app-updates__item">
+						<div class="govuk-hint govuk-!-margin-0 govuk-!-margin-bottom-2">
```
kevincarrogan, line 94: Is `govuk-hint` really the class we want to be using here? It's not really a hint.

kevincarrogan (approved):

kevincarrogan (approved):

saruniitr (approved):

PR #1095, opened 2023-02-07
[LTD-3294] add pagination to denials search

Harriethw:
## aim

[Ticket](https://uktrade.atlassian.net/jira/software/c/projects/LTD/boards/204?modal=detail&selectedIssue=LTD-3294) 

The Denials search was only returning the first 10 results, this adds pagination to the view. 
Relies on this API change for the e2es to pass: https://github.com/uktrade/lite-api/pull/1291

### DIFF
--- a/caseworker/cases/views/main.py
+++ b/caseworker/cases/views/main.py
@@ -642,10 +642,12 @@ class Denials(LoginRequiredMixin, TemplateView):
             search.append(party["address"])
             filter["country"].add(party["country"]["name"])
 
-        if search:
-            response = search_denials(request=self.request, search=search, filter=filter)
-            results = [item["_source"] for item in response.json()["hits"]["hits"]]
-        else:
-            results = []
+        total_pages = 0
+        results = []
 
-        return super().get_context_data(case=case, results=results, **kwargs)
+        if search:
+            response = search_denials(request=self.request, search=search, filter=filter).json()
+            results = response["results"]
+            total_pages = response["total_pages"]
+
+        return super().get_context_data(case=case, results=results, total_pages=total_pages, **kwargs)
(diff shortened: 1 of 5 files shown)


caseworker/cases/views/main.py
```diff
@@ -643,9 +643,11 @@ def get_context_data(self, **kwargs):
             filter["country"].add(party["country"]["name"])
 
         if search:
-            response = search_denials(request=self.request, search=search, filter=filter)
-            results = [item["_source"] for item in response.json()["hits"]["hits"]]
+            response = search_denials(request=self.request, search=search, filter=filter).json()
+            results = response["results"]
+            data = {"total_pages": response["total_pages"], "count": response["count"]}
```
Harriethw, line 648: the `pagination` component requires these keys to be in `data`

caseworker/cases/views/main.py
```diff
@@ -643,9 +643,11 @@ def get_context_data(self, **kwargs):
...
 
         if search:
-            response = search_denials(request=self.request, search=search, filter=filter)
-            results = [item["_source"] for item in response.json()["hits"]["hits"]]
+            response = search_denials(request=self.request, search=search, filter=filter).json()
+            results = response["results"]
+            total_pages = int(response["total_pages"])
         else:
+            total_pages = 0
             results = []
```
saruniitr, line 651: Could you consider moving these two above `if` then `else` can be removed?

caseworker/external_data/services.py
```diff
@@ -12,7 +12,8 @@ def upload_denials(request, data):
 
 
 def search_denials(request, search, filter):
-    data = {"search": search, **filter}
+    page = int(request.GET.get("page", 1))
```
saruniitr, line 15: ah you have just updated it.
I was wondering do we need the cast to `int` as it will be converted to `str` again in the query params?

saruniitr: Was it causing any issue if you don't cast it?

saruniitr (approved):

hnryjmes (approved):

PR #1099, opened 2023-02-09
LTD-3334-permissions-remove-top-actions 

depsiatwal:
This goes of the back of the banner functionality where a banner is displayed if they aren't the allocated case officer or assigned user. This PR goes one step further on that and disables some of the change links forcing the user to assign them selves first.


caseworker/templates/case/slices/summary.html
```diff
@@ -92,9 +95,11 @@
 		<dd class="app-case__summary-list__actions">
 			{% if permissible_statuses %}
 				{% if not is_terminal or 'REOPEN_CLOSED_CASES' in permissions %}
-					<a id="link-change-status" class="govuk-link govuk-link--no-visited-state" href="{% url 'cases:manage' queue.id case.id %}">
+					{% if can_user_change_case %}
```
currycoder, line 98: I feel like all of these should be narrowed down to the particular action in question.  Right now, the conditions for all of them are the same and boil down to `can_user_change_case` - but in the future we can make things tighter/more granular and only change the implementation of the rules themselves rather than needing to revisit the template.

For this one I would suggest; `can_user_change_case_status`

caseworker/templates/case/slices/summary.html
```diff
@@ -117,9 +122,11 @@
 			{% endif %}
 		</dd>
 		<dd class="app-case__summary-list__actions">
-			<a id="link-change-queues" class="govuk-link govuk-link--no-visited-state" href="{% url 'cases:move' queue.id case.id %}">
-				{% lcs 'generic.CHANGE' %}
-			</a>
+			{% if can_user_change_case %}
```
currycoder, line 125: For this one I would suggest; `can_user_change_case_queues`

caseworker/templates/case/slices/summary.html
```diff
@@ -213,9 +220,11 @@
 		</dd>
 		<dd class="app-case__summary-list__actions">
 			{% if not is_terminal %}
-				<a id="link-change-review-date" class="govuk-link govuk-link--no-visited-state" href="{% url 'cases:review_date' queue.id case.id %}">
-					{% lcs 'generic.CHANGE' %}
-				</a>
+				{% if can_user_change_case %}
```
currycoder, line 223: For this one I would suggest; `can_user_change_case_review_date`

caseworker/templates/case/slices/summary.html
depsiatwal: The name is generic because atm we only really have one condition check. 
I see what you're saying with the more granular approach and it makes sense in a full ACL type system. This is more 'generic permissions on the ui' breaking down would allow us flexibility but I feel this would  register more permissions then required and more complex code (including tests) to maintain etc etc. 

I feel we should create more granular permissions if and when needed and hopefully refactor along the way when required.

Pipfile.lock
```diff
@@ -263,7 +263,6 @@
                 "sha256:f9d0c5c045a3ca9bedfc35dca8526798eb91a07aa7a2c0fee134c6c6f321cbd7",
                 "sha256:ff6f3db31555657f3163b15a6b7c6938d08df7adbfc9dd13d9d19edad678f1e8"
             ],
-            "markers": "python_version >= '3.6'",
```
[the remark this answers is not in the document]
depsiatwal: This was due to a merge base conflict so I merged the pipfile hence then I was forced to update and create a new pipe.lock file. This branch was created of the back of an old PR I had hence the pipfile is probably = dev now not showing any changes,

caseworker/templates/case/slices/summary.html
currycoder: I get where you are coming from, but I feel like it's not going to end up being a tonne of work.  The tests that you have in this PR will work either way; the tests that would need adding to test the extra rule in isolation overlap almost entirely with the existing rule so I bet there's code to reuse there.

I personally feel like if we don't do this then we are going to lose a bunch of value in doing this and that value is cheap to realise.

enforcement_check.xml
```diff
@@ -0,0 +1,627 @@
+<?xml version="1.0" ?>
```
currycoder, line 1: howd-that-get-there

ui_tests/caseworker/features/enforcement.feature
```diff
@@ -10,6 +10,9 @@ Feature: I want to export and import XML for enforcement checking
     And I create an application with <name>,<product>,<part_number>,<clc_rating>,<end_user_name>,<end_user_address>,<consignee_name>,<consignee_address>,<country>,<end_use>
     And the status is set to "submitted"
     When I go to application previously created
+    And I click add case advisor
+    And filter by test user email
+    And I click the user and assign
```
currycoder, line 15: Could these all just be "And I assign myself to the case"?  Seems like they're always used in a sequence of 3, so could make sense to just make a single step

ui_tests/caseworker/step_defs/test_give_advice.py
```diff
@@ -217,3 +218,21 @@ def enter_case_note_text(driver, text, context):
...
+def filter_users_found(driver, internal_info):
+    case_officer_page = CaseOfficerPage(driver)
+    case_officer_page.search(internal_info["email"])
+
+
+@when("I click the user and assign")
+def click_user_and_assign(driver):
+    case_officer_page = CaseOfficerPage(driver)
+    case_officer_page.select_first_user()
+    functions.click_submit(driver)
```
currycoder, line 238: This is duplicated from `test_enforcement.py` - could it just be moved up in to a common helper file or even just a higher level `conftest.py` that all tests can use?

caseworker/templates/case/slices/summary.html
currycoder: We talked about this offline and agreed that this is OK as-is for now, but we should review as we continue to use permissions/rules more.

ui_tests/caseworker/features/enforcement.feature
depsiatwal: yeah that makes more sense done wasn't sure of the patterns here

ui_tests/caseworker/step_defs/test_give_advice.py
depsiatwal: as I result of above duplication removed also and places in conftest.py

enforcement_check.xml
depsiatwal: hahah result of test files that created removed

currycoder (approved):

saruniitr (approved):

PR #1101, opened 2023-02-09
LTD-3028-Filter out blank users

seijihg:
### Workflow

- [x] Ticket exists in Jira https://uktrade.atlassian.net/browse/LTD-3028

### Reviewing help

### Housekeeping

### Merging

### DIFF
--- a/caseworker/queues/views.py
+++ b/caseworker/queues/views.py
@@ -85,8 +85,18 @@ class Cases(LoginRequiredMixin, TemplateView):
 
         return response.json()
 
+    def _filter_user(self, user):
+        if user["pending"]:
+            return False
+        return True
+
     @property
     def filters(self):
+        gov_users = self.data["results"]["filters"]["gov_users"]
+        filtered_gov_users = list(filter(self._filter_user, gov_users))
+
+        self.data["results"]["filters"]["gov_users"] = filtered_gov_users
+
         return self.data["results"]["filters"]
 
     def get_params(self):
--- a/caseworker/queues/views.py
+++ b/caseworker/queues/views.py
@@ -85,15 +85,10 @@ class Cases(LoginRequiredMixin, TemplateView):
 
         return response.json()
 
-    def _filter_user(self, user):
-        if user["pending"]:
-            return False
-        return True
-
     @property
     def filters(self):
         gov_users = self.data["results"]["filters"]["gov_users"]
-        filtered_gov_users = list(filter(self._filter_user, gov_users))
+        filtered_gov_users = [gov_user for gov_user in gov_users if not gov_user["pending"]]
 
         self.data["results"]["filters"]["gov_users"] = filtered_gov_users
 
--- a/unit_tests/caseworker/teams/test_templates.py
+++ b/unit_tests/caseworker/teams/test_templates.py
@@ -5,7 +5,6 @@ from django.template.loader import render_to_string
 
 
 def test_pending_users_in_team_members(john_smith, gov_uk_user_id):
-    john_smith["pending"] = False
     context = {}
     context["users"] = [
         {
--- a/unit_tests/caseworker/teams/test_templates.py
+++ b/unit_tests/caseworker/teams/test_templates.py
@@ -1,0 +1,37 @@
+from bs4 import BeautifulSoup
+import re
+
+from django.template.loader import render_to_string
+
+
+def test_pending_users_in_team_members(john_smith, gov_uk_user_id):
+    john_smith["pending"] = False
+    context = {}
+    context["users"] = [
+        {
+            "id": gov_uk_user_id,
+            "email": "jane.doe@example.com",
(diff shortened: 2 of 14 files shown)


caseworker/queues/views.py
```diff
@@ -69,8 +69,18 @@ def data(self):
         data = get_cases_search_data(self.request, self.queue_pk, params)
         return data
 
+    def _filter_user(self, user):
+        if user["full_name"] == " ":
```
kevincarrogan, line 73: This doesn’t feel like we are actually solving the root cause of the bug here. 

Why are there users without a full name? Is there anything in particular about these users that means they don’t have a name set? Is the real but the fact that a users name isn’t being set?

caseworker/queues/views.py
```diff
@@ -69,8 +69,18 @@ def data(self):
         data = get_cases_search_data(self.request, self.queue_pk, params)
         return data
 
+    def _filter_user(self, user):
+        if user["pending"]:
```
kevincarrogan, line 73: You don't need an `if` statement here. If this is already a boolean then you can just return this value directly

```
def _filter_user(self, user):
    return user["pending"]
```

caseworker/queues/views.py
```diff
@@ -69,8 +69,18 @@ def data(self):
...
+            return False
+        return True
+
     @property
     def filters(self):
+        gov_users = self.data["results"]["filters"]["gov_users"]
+        filtered_gov_users = list(filter(self._filter_user, gov_users))
```
kevincarrogan, line 80: This probably should be a list comprehension. In fact you could probably just drop the method call as well and just do:

```
filtered_gov_users = [gov_user for gov_user in gov_users if not gov_user["pending"]]
```

kevincarrogan (commented):
There are other bits of code that are using `user.first_name` to check whether the user was pending or not. Could we also update those?

kevincarrogan:
> There are other bits of code that are using `user.first_name` to check whether the user was pending or not. Could we also update those?

By the way. Feel free to split that out in a separate PR if you want to but I think we should make sure we make that change as part of this ticket overall.

caseworker/queues/views.py
seijihg: jesus what kind of magic is that O_O

depsiatwal (commented):
is it possible to add some more tests here for the template. 

plenty of example in our code base using BeautifulSoup 

https://github.com/uktrade/lite-frontend/blob/3c610e48095a5e1412ff28d622785e8f75e91fd2/unit_tests/caseworker/advice/views/test_advice_view.py#L30

unit_tests/caseworker/teams/test_templates.py
```diff
@@ -0,0 +1,37 @@
+from bs4 import BeautifulSoup
+import re
+
+from django.template.loader import render_to_string
+
+
+def test_pending_users_in_team_members(john_smith, gov_uk_user_id):
+    john_smith["pending"] = False
```
Harriethw, line 8: minor/question: would it make sense to add the `pending` key to the fixture of `john_smith` so that the fixture matches the shape of the response we expect to see when requesting users?

unit_tests/caseworker/teams/test_templates.py
```diff
@@ -0,0 +1,37 @@
...
+            },
+            "role_name": "Super User",
+            "pending": True,
+        },
+        john_smith,
+    ]
+    html = render_to_string("teams/team.html", context)
+    soup = BeautifulSoup(html, "html.parser")
+    rows = soup.find_all(id=re.compile("^row-\d+"))
+    assert rows[0].find(class_="govuk-tag govuk-tag--blue").text.strip() == "Pending"
```
Harriethw, line 32: minor: would it be worth testing that you are also rendering the right data in the other cells for the `Pending` users? e.g. am I right in understanding you still render the email, status and link?

caseworker/queues/views.py
```diff
@@ -72,6 +72,11 @@ def data(self):
 
     @property
     def filters(self):
+        gov_users = self.data["results"]["filters"]["gov_users"]
```
Harriethw, line 75: is there any test coverage for this logic around filtering `gov_users`?

unit_tests/caseworker/teams/test_templates.py
seijihg: it does check below in line 34 :D

Harriethw: Isn't line 34 checking the next row, so for user John Smith, not for the 'Pending' user?

Harriethw (approved):

unit_tests/caseworker/teams/test_templates.py
seijihg: ah yes ur right!

depsiatwal (approved):
lgtm

PR #1102, opened 2023-02-10
LTD-3324 Update serial number message in case detail template

hnryjmes:
This change updates the display of serial numbers in the advice side panel to only show the list of numbers if there is at least one serial number present, otherwise it shows `Not added yet`.

The way this is done is by reusing the serial numbers template tag that is used elsewhere in the advice app.

Therefore the new if statement logic for this template satisfies the criteria for display given in the ticket, namely:
```
We’re currently showing serial numbers for all sub-firearm category products (so things like accessories where we don’t capture serial numbers on the application form') but it should just be:

firearms

ammunition

components for ammunition

components for firearms
```

### DIFF
--- a/caseworker/advice/templates/advice/case_detail.html
+++ b/caseworker/advice/templates/advice/case_detail.html
@@ -83,18 +83,7 @@
                   <tr class="govuk-table__row">
                     <td class="govuk-table__cell govuk-!-font-weight-bold">Serial numbers</td>
                     <td class="govuk-table__cell">
-                      {% if good.good.firearm_details.serial_numbers_available == "LATER" and not good.good.firearm_details.serial_numbers %}
-                        Not added yet
-                      {% else %}
-                        <details class="govuk-details" data-module="govuk-details">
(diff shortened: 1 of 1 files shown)


kevincarrogan (commented):
There's a template include that already encapsulates all the logic required here:

`caseworker/templates/goods/includes/serial_numbers.html`.

hnryjmes:
I don't think the whole include can be reused because the formatting is different here, at least that's what the discussion suggests on the PR when these were first made a few weeks ago

kevincarrogan:
I'm looking at the two templates side-by-side and I can't see anything that would make the formatting any different between the two except for maybe `govuk-!-margin-bottom-0` which could be solved in a different way.

This is also how we present this everywhere else so we should try to make the formatting consistent anyway, if there are some differences then we should either somehow wrap those differences into that include in some way or push back and suggest that this should be formatted the same across the application.

kevincarrogan (approved):

saruniitr (approved):

PR #1108, opened 2023-02-14
[LTD-3364] Fix queue view limit case updates lines

currycoder:
### Aim
This change fixes the case updates column for queue view as follows;
- Boldifies the "Last Action"/"Previous Action" spans.
- Makes the HTML more semantic by using one `<li>` per update.
- Shows `additional_text` for each update if it exists.
- Limits both `additional_text`/`text` for each update to show up to two lines.
- Adds tests.

**Update;**  There's also a fair chunk of overlap with the work ongoing here with some extra content tweaks in https://uktrade.atlassian.net/browse/LTD-3365 - so I've absorbed those changes in to the final commit.

Screenshots can be found here; https://uktrade.atlassian.net/browse/LTD-3364

### DIFF
--- a/unit_tests/caseworker/queues/test_views.py
+++ b/unit_tests/caseworker/queues/test_views.py
@@ -39,7 +39,7 @@ def setup(
 @pytest.fixture
 def data_cases_search(mock_case_statuses, data_case_types, gov_uk_user_id):
     return {
-        "count": 2,
+        "count": 1,
         "results": {
             "cases": [
                 {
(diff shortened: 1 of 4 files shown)


unit_tests/caseworker/queues/test_views.py
```diff
@@ -556,3 +584,53 @@ def test_queue_assignments(url, authorized_client):
...
+    ]
+    assert response.context["data"]["results"]["cases"][0]["activity_updates"] == expected_activity_updates
+
+    html = BeautifulSoup(response.content, "html.parser")
+    updates = [update.get_text() for update in html.select("li", {"class": "app-updates__item"})]
+    assert "LITE system" in updates[0]
+    assert "text line1" in updates[0]
+    assert "text line2" in updates[0]
+    assert "additional line1" in updates[0]
+    assert "additional line2" in updates[0]
```
antroy-madetech, line 634: Perhaps `assert "additional line3" not in updates[0]` as well to prove the truncation.

unit_tests/caseworker/queues/test_views.py
```diff
@@ -556,3 +584,53 @@ def test_queue_assignments(url, authorized_client):
...
+            "additional_text": "",
+        },
+    ]
+    assert response.context["data"]["results"]["cases"][0]["activity_updates"] == expected_activity_updates
+
+    html = BeautifulSoup(response.content, "html.parser")
+    updates = [update.get_text() for update in html.select("li", {"class": "app-updates__item"})]
+    assert "LITE system" in updates[0]
+    assert "text line1" in updates[0]
+    assert "text line2" in updates[0]
```
antroy-madetech, line 632: As below.

antroy-madetech (commented):
Couple of assertions required in tests, but that's it.

Harriethw (approved):
thank you for the helpful context and screenshots!

kevincarrogan:
We appear to be truncating the text without indicating to the user that we're truncating it. Could this lead to some ambiguity e.g.

"We can delete this.
It will be fine.
UNLESS IT'S TUESDAY AND THEN THE WORLD WILL EXPLODE."

They would never read that last line and there would be no indication that there was anything more to read.

kevincarrogan (approved):

currycoder (commented):
**Update;** There's also a fair chunk of overlap with the work ongoing here with some extra content tweaks in https://uktrade.atlassian.net/browse/LTD-3365 - so I've absorbed those changes in to the final commit.

PR #1109, opened 2023-02-14
Dependabot/pip/cryptography 39.0.1

kevincarrogan:
This solves the dependabot issue of requiring us to upgrade our version of cryptography.

It seems that we don't explicitly use the `PyOpenSSL` nor `cryptography` packages explicitly on this repo but they're subdependencies of `selenium-wire`.

Trying to upgrade to a higher version of both `PyOpenSSL` and `cryptography` caused there to be package version conflicts so in the end I bumped the version of `selenium` and `selenium-wire` so that they pulled in newer versions of `PyOpenSSL` and `cryptography` themselves.

The only major change I needed to get the e2e tests working was to switch away from the deprecated `find_element_by_*` methods.


tests_common/functions.py
```diff
@@ -14,13 +14,13 @@ def click_submit(driver: WebDriver):
...
 def click_finish_button(driver: WebDriver):
-    element = driver.find_element_by_css_selector("button[value='finish']")
+    element = driver.find_element(by=By.CSS_SELECTOR, value="button[value='finish']")
     driver.execute_script("arguments[0].scrollIntoView();", element)
     driver.execute_script("arguments[0].click();", element)
 
 
 def click_back_link(driver: WebDriver):
-    driver.find_element_by_id("back-link").click()
+    driver.find_element(by=By.ID, value="back-link").click()
```
hnryjmes, line 23: now this is the syntax I know and love 😄 haven't seen the old style for a while

Pipfile
```diff
@@ -59,6 +59,7 @@ sigauth = "~=4.3.0"
 whitenoise = "~=5.3.0"
 django-formtools = "~=2.3.0"
 crispy-forms-gds = "~=0.2.4"
+django-crispy-forms = "~=1.14.0"
```
hnryjmes, line 62: so without this pin, this would have been changed to a non-backwards-compatible version as a result of upgrading selenium?

hnryjmes: sounds like a stupid question but I'm just making sure I'm following the order of these commits precisely

kevincarrogan: It didn't upgrade as a direct result. `django-crispy-forms` isn't a dependency or related to `selenium` but when we install a new package pipenv will upgrade other packages as a result and this ended up upgrading `django-crispy-forms` to a backward incompatible version.

hnryjmes: right, that makes sense

hnryjmes (approved):

currycoder (approved):

PR #1110, opened 2023-02-14
[LTD-3354] show correct tab count for 'all_cases' tab

Harriethw:
### Aim

Screenshot 2023-02-13 at 12 34 49

There is some logic on both the front [and backend](https://github.com/uktrade/lite-api/blob/dev/api/cases/views/search/views.py#L114) around whether cases with open queries should be `hidden` on a queue view. With the introduction of different tabs adding different rules this logic is getting a bit confusing.

This PR is to put in a quick a fix for the issue where the `cases to review/all cases` tab count was including cases that should be hidden when the user is on a different tab, and hopefully clarifies the rules around when we should be 'hiding' open queries. 

There's logic in two places on the API around the `include_hidden` param - maybe the API should handle it more, and the frontend only pass 'true' for hidden if it's on the form? 

I would suggest as this maybe isn't a priority the next time we touch this we make time to refactor a bit, or we create a ticket to refactor. At the least the param could be renamed, `hidden` or `include_hidden` is a bit confusing and might be better if the boolean was the other way round, could be e.g. `exclude_open_queries`. 

[Link to story](https://uktrade.atlassian.net/jira/software/c/projects/LTD/boards/204?modal=detail&selectedIssue=LTD-3354)

### DIFF
--- a/unit_tests/caseworker/queues/test_views.py
+++ b/unit_tests/caseworker/queues/test_views.py
@@ -267,7 +267,7 @@ def mock_team_queue(requests_mock, data_queue):
 
 
 @pytest.fixture
-def mock_team_cases(requests_mock, data_cases_search):
+def mock_team_cases_search(requests_mock, data_cases_search):
     data_cases_search["results"]["filters"]["is_system_queue"] = False
     url = client._build_absolute_uri(f"/cases/")
     yield requests_mock.get(url=url, json=data_cases_search)
@@ -424,8 +424,7 @@ def test_case_assignment_select_role(authorized_client, mock_gov_user, user_role
     assert response.url == reverse(f"queues:{expected_url_name}", kwargs={"pk": queue_pk}) + url_params
 
 
-def test_with_all_cases_default(authorized_client, mock_cases_search, mock_cases_search_head):
-
+def test_tabs_with_all_cases_default(authorized_client, mock_cases_search, mock_cases_search_head):
     response = authorized_client.get(reverse("core:index"))
     html = BeautifulSoup(response.content, "html.parser")
     all_queries_tab = html.find(id="all-cases-tab")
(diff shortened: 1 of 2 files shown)


unit_tests/caseworker/queues/test_views.py
```diff
@@ -505,11 +508,51 @@ def test_tabs_on_team_queue(authorized_client, mock_team_cases, mock_team_queue,
...
+    assert mock_team_cases.last_request.qs == {
         "hidden": ["true"],
+        "page": ["1"],
+        "queue_id": [queue_pk],
+        "selected_tab": ["all_cases"],
     }
+    head_request_history = [x.qs for x in mock_cases_search_head.request_history]
+    tabs_with_hidden_param = ("all_cases", "my_cases", "open_queries")
+    for tab in tabs_with_hidden_param:
+        assert {
```
hnryjmes, line 550: I am assuming that this assertion here checks the functionality which, as a result of the changes in this PR, has the consequences that it makes the count correct in the tab. Would you or have you considered having another test assertion that explicitly tests the count, just to correspond with what you had in the ticket (`When I am on the ‘open queries’ tab on a team queue // Then the count on the ‘cases to review’ tab should NOT include “hidden” cases`)? I don't know if that would make the code better but that sort of explicit assertion on the count shown in the tab was the first thing I looked for after reading the ticket.

Harriethw: I haven't checked the count that's rendered in the template because I guess what's interesting to me is the contract between the API, because these aren't integration tests, so I'm checking that the frontend is sending the right flag, and then there are tests on the API that cover when that flag is sent over...
We could look at maybe checking it on e2es but then the issue would be if the test data changes, the count would change so it might be a bit brittle 🤔

kevincarrogan (approved):

hnryjmes (approved):

PR #1115, opened 2023-02-16
Hide the 'Not allocated' message for unassigned case advisers on LU Post/Pre-circ

Harriethw:
### Aim

When I am on the Licensing Unit pre-circulation OR Licensing Unit post-circulation queues
And I’m looking at the list of cases
And a case has an assigned case officer
And no assigned case advisers 
Then I only see the information for the assigned case officer
And NOT the information for case advisers 

When I am on any other queue
Then I see the information for BOTH the case advisers and case officer

[Link to story](https://uktrade.atlassian.net/jira/software/c/projects/LTD/boards/204)

### DIFF
--- a/unit_tests/caseworker/queues/test_views.py
+++ b/unit_tests/caseworker/queues/test_views.py
@@ -635,7 +635,7 @@ def test_queue_assignments(url, authorized_client):
     assert "Initial Queue" in li_elems[0]
     assert "Joe Smith" in li_elems[1]
     assert "Initial Queue" in li_elems[1]
-    assert "Not Allocated" in li_elems[2]
+    assert "Not allocated" in li_elems[2]
     assert "Another Queue" in li_elems[2]
 
 
--- a/unit_tests/caseworker/queues/test_views.py
+++ b/unit_tests/caseworker/queues/test_views.py
@@ -11,6 +11,8 @@ import os
 
 from core import client
 
+from caseworker.cases.helpers.case import LU_POST_CIRC_FINALISE_QUEUE_ALIAS, LU_PRE_CIRC_REVIEW_QUEUE_ALIAS
+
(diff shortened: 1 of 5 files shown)


unit_tests/caseworker/queues/test_views.py
```diff
@@ -294,6 +296,16 @@ def mock_team_queue(requests_mock, data_queue):
     yield requests_mock.get(url=re.compile(f"{url}.*/"), json=data_queue)
 
 
+@pytest.fixture()
+def mock_queue_with_alias_factory(requests_mock, data_queue):
+    def _mock_queue_with_alias(alias):
+        data_queue["alias"] = alias
+        url = client._build_absolute_uri("/queues/")
+        yield requests_mock.get(url=re.compile(f"{url}.*/"), json=data_queue)
```
currycoder, line 304: I think this can just be `return` not `yield`

unit_tests/caseworker/queues/test_views.py
```diff
@@ -677,3 +689,20 @@ def test_activity_updates(url, authorized_client):
...
+        (LU_PRE_CIRC_REVIEW_QUEUE_ALIAS),
+    ),
+)
+def test_assignments_hidden_alias(authorized_client, mock_team_cases_search, mock_queue_with_alias_factory, alias):
+    mock_queue_with_alias_factory(alias)
+    url = client._build_absolute_uri(f"/queues/{queue_pk}")
+    response = authorized_client.get(url)
+    html = BeautifulSoup(response.content, "html.parser")
+    li_elems = [elem.get_text() for elem in html.select("li", {"class": "app-assignments__item"})]
+    assert "Not allocated" not in li_elems[2]
```
currycoder, line 707: Could this assert that the queue that we do expect is present? e.g. the one which has an assignment

currycoder (approved):
Some minor comments, but looks good

kevincarrogan (approved):

PR #1116, opened 2023-02-16
 docker-file-add-python-buffered 

depsiatwal:
This change fixes an issue where by python/django code is cached on my machine. Every time I make a change a restart is required this appears to fix that issue


kevincarrogan (approved):

currycoder (approved):
Any ideas what's up with the E2E tests on CI?

PR #1119, opened 2023-02-18
LTD-3335-permission-tau-advice-move-case-forward

depsiatwal:
### Aim

[Link to story](https://uktrade.atlassian.net/browse/LTD-3232)


caseworker/core/rules.py
```diff
@@ -18,3 +18,4 @@ def is_user_assigned(user, case):
 
 
 rules.add_rule("can_user_change_case", is_user_case_adviser | is_user_assigned)
+rules.add_rule("can_user_move_case_forward", is_user_case_adviser | is_user_assigned)
```
currycoder, line 21: Can this be tested directly like the other rules? [unit_tests/caseworker/cases/test_permissions.py](https://github.com/uktrade/lite-frontend/pull/1094/files#diff-11440c48a76ac77cb2a6726ab6d9bf2a3bce6553910e3449543e46197f0cf62c)

currycoder (commented):
I think there are other "Move case forward" buttons;
https://github.com/uktrade/lite-frontend/blob/6fd31b9075f1b0188cb8efaeec8bc533c4f4540c/caseworker/advice/views.py#L204
https://github.com/uktrade/lite-frontend/blob/6fd31b9075f1b0188cb8efaeec8bc533c4f4540c/caseworker/advice/views.py#L545

There's also "I'm done" buttons;
https://github.com/uktrade/lite-frontend/blob/10689ba609d457ede0bd29a1484c5dd39dc507ab/caseworker/templates/layouts/case.html#L144

I feel like these should probably all be covered too.

depsiatwal:
> I think there are other "Move case forward" buttons;
> 
> https://github.com/uktrade/lite-frontend/blob/6fd31b9075f1b0188cb8efaeec8bc533c4f4540c/caseworker/advice/views.py#L204
> 
> 
> https://github.com/uktrade/lite-frontend/blob/6fd31b9075f1b0188cb8efaeec8bc533c4f4540c/caseworker/advice/views.py#L545
> 
> There's also "I'm done" buttons;
> 
> https://github.com/uktrade/lite-frontend/blob/10689ba609d457ede0bd29a1484c5dd39dc507ab/caseworker/templates/layouts/case.html#L144
> 
> I feel like these should probably all be covered too.

yeah they're a different style and handles very differently was going to cover under a different PR but guess can try and extend this one

caseworker/core/rules.py
depsiatwal: These have all been extended the forms actually use the same base view and template

depsiatwal:
> > I think there are other "Move case forward" buttons;
> > https://github.com/uktrade/lite-frontend/blob/6fd31b9075f1b0188cb8efaeec8bc533c4f4540c/caseworker/advice/views.py#L204
> > 
> > https://github.com/uktrade/lite-frontend/blob/6fd31b9075f1b0188cb8efaeec8bc533c4f4540c/caseworker/advice/views.py#L545
> > 
> > There's also "I'm done" buttons;
> > https://github.com/uktrade/lite-frontend/blob/10689ba609d457ede0bd29a1484c5dd39dc507ab/caseworker/templates/layouts/case.html#L144
> > 
> > I feel like these should probably all be covered too.
> 
> yeah they're a different style and handles very differently was going to cover under a different PR but guess can try and extend this one

These have all been extended the forms actually use the same base view and template

unit_tests/caseworker/core/test_permissions.py
```diff
@@ -0,0 +1,166 @@
...
+                    ]
+                },
+                "case_officer": {"id": "fake_id"},
+            },
+            False,
+        ),
+    ),
+)
+def test_can_user_change_case_permission(data, mock_gov_user, expected_result):
+    assert rules.test_rule("can_user_change_case", mock_gov_user["user"], data) == expected_result
```
currycoder, line 166: Looks to me like this is duplicated from above?  Was this meant to be a test for the new rule?

unit_tests/caseworker/tau/test_home.py
```diff
@@ -289,6 +290,7 @@ def test_move_case_forward(
     """
     gov_user["user"]["team"]["name"] = team_name
     gov_user["user"]["team"]["alias"] = team_alias
+    rules.set_rule("can_user_move_case_forward", lambda: True)
```
currycoder, line 293: Does this get reset between testcases?  Could it be better to just adjust the user so that they satisfy the condition rather than fiddling with the internals of the rules?

unit_tests/caseworker/tau/test_home.py
```diff
@@ -393,3 +395,40 @@ def test_precedents_starting_point(
...
+    is_user_case_advisor,
+):
+    gov_user["user"]["team"]["name"] = "TAU"
+    gov_user["user"]["team"]["alias"] = views.TAU_ALIAS
+    requests_mock.get(
+        client._build_absolute_uri("/gov-users/2a43805b-c082-47e7-9188-c8b3e1a83cb0"),
+        json=gov_user,
+    )
+
+    rules.set_rule("can_user_move_case_forward", lambda: is_user_case_advisor)
```
currycoder, line 421: Same comment as above

unit_tests/caseworker/cases/test_permissions.py
```diff
@@ -147,3 +50,36 @@ def test_permission_summary_change_links(
...
+    data_queue,
+    authorized_client,
+    data_standard_case,
+    id_element_name,
+    is_user_case_advisor,
+    id_element_name_visible,
+):
+    url = reverse("cases:case", kwargs={"queue_pk": data_queue["id"], "pk": data_standard_case["case"]["id"]})
+    # We are changing rule here since mocking doesn't seem straight forward
+    rules.set_rule("can_user_move_case_forward", lambda: is_user_case_advisor)
```
currycoder, line 74: Same comment on this line as below RE: mocking permissions

unit_tests/caseworker/advice/views/test_view_my_advice_view.py
```diff
@@ -208,3 +218,41 @@ def test_view_security_approvals(
...
+):
+    url = reverse(
+        "cases:view_my_advice",
+        kwargs={"queue_pk": "f458094c-1fed-4222-ac70-ff5fa20ff649", "pk": data_standard_case["case"]["id"]},
+    )
+    mock_gov_user["user"]["team"]["alias"] = FCDO_TEAM
+    advice_with_consignee[0]["user"]["team"]["alias"] = FCDO_TEAM
+    data_standard_case["case"]["advice"] = advice_with_consignee
+    case_id = data_standard_case["case"]["id"]
+    rules.set_rule("can_user_move_case_forward", lambda: is_user_case_advisor)
```
currycoder, line 244: Same comment as elsewhere RE: mocking permission rules

unit_tests/caseworker/tau/test_home.py
currycoder: I guess it's OK to mock it in this way, but I'd like to work out how the rules get reset between testcases - any ideas?

caseworker/core/rules.py
currycoder: Pls see my comment below in `unit_tests/caseworker/core/test_permissions.py` - I still don't think this is tested

unit_tests/caseworker/core/test_permissions.py
```diff
@@ -0,0 +1,166 @@
...
+
+
+@pytest.fixture(autouse=True)
+def setup():
+    # This is required because we need to set back the rules for testing the actual rule
+    # Some tests may have overridden this rule which has a global effect
+    rules.set_rule("can_user_change_case", caseworker_rules.is_user_case_adviser | caseworker_rules.is_user_assigned)
+    rules.set_rule(
+        "can_user_move_case_forward", caseworker_rules.is_user_case_adviser | caseworker_rules.is_user_assigned
+    )
```
currycoder, line 17: Ah I see..  Yeah I'm not a huge fan of the mocking where we reset the rules all over the place in testcases since it means that the tests don't actually test how things work in the application for real but instead how we mocked out the testcases.  There's room to easily go out of sync between the way we mock the rules and the way our app sets the rules.

I think I'd prefer avoiding `rules.set_rule()` across the board in testcases and just setting up the test data to have the correct characteristics to pass/fail the permissions check as required.

I can't think of a way to genuinely mock the rules for view tests..

unit_tests/caseworker/advice/views/test_view_my_advice_view.py
```diff
@@ -166,6 +174,8 @@ def test_move_case_forward(
     advice_with_and_without_consignee[0]["user"]["team"]["alias"] = FCDO_TEAM
     data_standard_case["case"]["advice"] = advice_with_and_without_consignee
     case_id = data_standard_case["case"]["id"]
+    rules.set_rule("can_user_move_case_forward", lambda: True)
```
currycoder, line 177: Same comment as elsewhere RE: mocking permission rules

caseworker/core/rules.py
depsiatwal: renamed test and test these both at the same time

unit_tests/caseworker/core/test_permissions.py
depsiatwal: renamed test and added new check

unit_tests/caseworker/core/test_permissions.py
depsiatwal: absolutely agree i've now banned the set_rule method

currycoder (approved):

caseworker/tau/views.py
```diff
@@ -222,6 +222,7 @@ def get_context_data(self, **kwargs):
             "cle_suggestions_json": get_cle_suggestions_json(self.unassessed_goods),
             "organisation_documents": self.organisation_documents,
             "is_tau": self.caseworker["team"]["alias"] == TAU_ALIAS,
+            "user": self.caseworker,
```
Harriethw, line 225: would it be worth putting the `user` in the `TAUMixIn` context, so that all the TAU views have it? Otherwise on e.g. the edit view the rule doesn't have access to the `user` and I see the case warning banner even if I'm allocated to the case

Harriethw: I can move when I get round to this ticket though https://uktrade.atlassian.net/jira/software/c/projects/LTD/boards/204?modal=detail&selectedIssue=LTD-3372

caseworker/advice/templates/advice/view_my_advice.html
```diff
@@ -52,7 +52,10 @@ <h1 class="govuk-heading-xl">View recommendation</h1>
             </div>
             <div class="govuk-grid-column-one-quarter">
               {% if my_advice and advice_completed and buttons.move_case_forward %}
-                {% crispy form %}
+                {% test_rule 'can_user_move_case_forward' caseworker case as can_user_move_case_forward %}
```
saruniitr, line 55: just for my understanding, `caseworker`, `case` gets passed as arguments to `can_user_move_case_forward()`?

When including other templates we pass arguments as `with var=xxx ...`, wondering if similar is possible here for redability?

caseworker/advice/templates/advice/view_my_advice.html
```diff
@@ -52,7 +52,10 @@ <h1 class="govuk-heading-xl">View recommendation</h1>
...
+                    {% if can_user_move_case_forward %}
+                        {% crispy form %}
+                    {% endif %}
```
saruniitr, line 58: If I understand correctly, if the user doesn't have permissions then we don't want to show 'Move case forward' button correct?

Is it not better to hide the button rather than the form?
I mean if the user doesn't have permission ensure that `buttons.move_case_forward` is `False` then we don't need to change this template correct?

depsiatwal: at the moment we just changing template as it's then a consistent pattern also same is happening with other show/hide items

caseworker/advice/templates/advice/view_my_advice.html
depsiatwal: these were already passed in didn't really change it 
consistent with other templates

caseworker/tau/views.py
depsiatwal: possibly but with everything with lite min is best don't want to make too many changes it normally ends up breaking tests etc. balance is do minimum and then change what's required

saruniitr (approved):

PR #1120, opened 2023-02-20
LTD-3294: enable duplicate keys in pagination

Harriethw:
### Aim

We encountered a bug in the denials search, where duplicate keys were
being removed in the pagination as it was converting the params to a
regular dict. This commit uses a QueryDict which allows for duplicate
params.

[Link to story](https://uktrade.atlassian.net/jira/software/c/projects/LTD/boards/204?modal=detail&selectedIssue=LTD-3356)

### DIFF
--- a/lite_forms/templatetags/custom_tags.py
+++ b/lite_forms/templatetags/custom_tags.py
@@ -1,7 +1,9 @@
 import json
 from html import escape
 from json import JSONDecodeError
+import urllib.parse as urlparse
 
+from django.http import QueryDict
 from django.template.defaulttags import register
 from django.urls import reverse
 from django.utils.safestring import mark_safe
@@ -192,14 +194,10 @@ def make_list(*args):
 
 @register.filter
 def pagination_params(url, page):
-    import urllib.parse as urlparse
-    from urllib.parse import urlencode
-
-    url_parts = list(urlparse.urlparse(url))
-    query_params = dict(urlparse.parse_qsl(url_parts[4]))
-    query_params.pop("page", 1)
-    query = {"page": page, **query_params}
-    url_parts[4] = urlencode(query)
+    url_parts = urlparse.urlparse(url)
+    query_dict = QueryDict(url_parts.query, mutable=True)
+    query_dict["page"] = page
+    url_parts = url_parts._replace(query=query_dict.urlencode())
 
     return urlparse.urlunparse(url_parts)
 
--- a/unit_tests/lite_forms/templates/test_custom_tags.py
+++ b/unit_tests/lite_forms/templates/test_custom_tags.py
@@ -15,16 +15,24 @@ def test_file_type(filename, expected):
 
 
 @pytest.mark.parametrize(
-    "url, expected",
+    "url, page, expected",
     [
         (
             "https://www.example.com/?page=1&example_param=foo&example_param=bar",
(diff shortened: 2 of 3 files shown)


seijihg (approved):

unit_tests/lite_forms/templates/test_custom_tags.py
```diff
@@ -12,3 +12,19 @@
...
+        (
+            "https://www.example.com/?page=1&example_param=foo&example_param=bar",
+            "https://www.example.com/?page=2&example_param=foo&example_param=bar",
+        ),
+        ("https://www.example.com/?page=3&example_param=bar", "https://www.example.com/?page=2&example_param=bar"),
+        ("https://www.example.com/?example_param=bar", "https://www.example.com/?example_param=bar&page=2"),
+    ],
+)
+def test_pagination_params(url, expected):
+    page = 2
```
kevincarrogan, line 29: What’s this doing here? Why are we always setting page to 2?

lite_forms/templatetags/custom_tags.py
```diff
@@ -192,14 +194,11 @@ def make_list(*args):
...
 def pagination_params(url, page):
-    import urllib.parse as urlparse
-    from urllib.parse import urlencode
-
     url_parts = list(urlparse.urlparse(url))
-    query_params = dict(urlparse.parse_qsl(url_parts[4]))
-    query_params.pop("page", 1)
-    query = {"page": page, **query_params}
-    url_parts[4] = urlencode(query)
+    query_string = url_parts[4]
```
kevincarrogan, line 198: I know you didn’t originally write this function but I think this could be made more readable by specifying `url_parts.query` instead.

unit_tests/lite_forms/templates/test_custom_tags.py
Harriethw: There didn't seem anything to test around what the `page` variable passed to the method is , we just want to check that it's set by checking that the `page` in the query param changes, so I didn't see we'd get any value out of changing it in the test... this method is called by another part of the code so it doesn't need to worry about checking it's the right type. But I can add a few different values if you think it would give more coverage

lite_forms/templatetags/custom_tags.py
Harriethw: done!
I had to use the `_replace` method because the `urlparse` method returns a named tuple, [if I understand correctly?](https://docs.python.org/3/library/urllib.parse.html) - is it ok to use "internal" methods like that in some cases?

unit_tests/lite_forms/templates/test_custom_tags.py
Harriethw: added a few different variations now

lite_forms/templatetags/custom_tags.py
kevincarrogan: Yeah. The documentation for urlparse suggests that's the right method to use so it's all good.

unit_tests/lite_forms/templates/test_custom_tags.py
kevincarrogan: Ok. I just wasn't sure what this variable was trying to tell me or why setting it to `2` was significant. Sounds like the extra tests would help clarify that though.

PR #1124, opened 2023-02-21
[LTD-3233] Improve case assignment queue selection form wizard

currycoder:
### Aim

This change solves the following jira stories;
No longer require a queue to be selected when adding a case adviser from within the case; https://uktrade.atlassian.net/browse/LTD-3233
"Allocate yourself" button/flow errors if initiated from "all cases" queue; https://uktrade.atlassian.net/browse/LTD-3373
A couple of content change stories;
https://uktrade.atlassian.net/browse/LTD-3363
https://uktrade.atlassian.net/browse/LTD-3369

Given these goals ^ it made sense to revisit the flow for allocating advisers.

Previously, we had two lite-forms views for allocating a user from a case;
https://github.com/uktrade/lite-frontend/pull/1124/files#diff-c9c4b119484dc74cf22b4b5251882c1bf1c6941110385fb394d90416d7d80513L108-L109

We also had a lite-form view for allocating a user from a queue;
https://github.com/uktrade/lite-frontend/pull/1124/files#diff-41d28b1d73b40b9cb4d9fbeb4476fb3a7ee14f02727a3ee544b01d6d55f132ffL18

The changes in this PR achieve the following;
- ~~Make our patterns for FormWizard from exporter app usable in caseworker.~~ FormWizard relocated in dependency PR: https://github.com/uktrade/lite-frontend/pull/1136
- Replace the above lite-form views with a single new general purpose formwizard view for allocating a user.
    - The formwizard conditionally shows the final step for selecting a queue if the queue ID in the URL is for a system queue; otherwise the queue in the URL is used to create the case assignment.
- Back links/success URLs are handled so that the user is taken back to the URL that they started the journey from.


caseworker/queues/services.py
```diff
@@ -79,12 +79,11 @@ def get_queue_case_assignments(request, pk):
     return data.json(), data.status_code
 
 
-def put_queue_case_assignments(request, pk, _):
-    case_ids = request.GET.getlist("cases")
-    json = {"case_assignments": [], "remove_existing_assignments": True, "note": request.POST.get("note")}
+def put_queue_case_assignments(request, queue_id, case_ids, user_ids, note):
+    data = {"case_assignments": [], "remove_existing_assignments": False, "note": note}
```
currycoder, line 83: Note that `"remove_existing_assignments": False` is an intentional change.  From a UI perspective, the flows that use this function are additive to it makes sense to avoid blatting all the existing users - especially since we now have a mechanism for removing users (added as part of case alloc).

depsiatwal (approved):
wow i'be been through the whole `PR and this is nicely executed can't fault even when i'm trying to.

it's nice to have the wizard ported to caseworker and I didn't like the divergence with the assign user which this solves nicely done :-)

caseworker/queues/forms.py
```diff
@@ -172,3 +173,102 @@ class RoleChoices(models.TextChoices):
...
+        required=False,
+    )
+
+    def __init__(self, request, team_id, *args, is_next_step=False, **kwargs):
+        self.request = request
+        self.team_id = team_id
+        if is_next_step:
+            self.Layout.SUBMIT_BUTTON_TEXT = "Save and continue"
+        else:
+            self.Layout.SUBMIT_BUTTON_TEXT = "Save and return"
```
kevincarrogan, line 205: The submit button always feels kind of wrong being in here when it's in a wizard context but I know there's nothing we can really do about that.

kevincarrogan: Although it seems like we always override the default here so maybe we can just set "Save and return" as the default value and only change it if `is_next_step` is `True`?

caseworker/queues/forms.py
```diff
@@ -172,3 +173,102 @@ class RoleChoices(models.TextChoices):
...
+            (
+                TextChoice(
+                    Choice(
+                        user["id"],
+                        user.get("first_name") + " " + user.get("last_name")
+                        if user.get("first_name")
+                        else user["email"],
+                    ),
+                    hint=user["email"],
+                )
```
kevincarrogan, line 223: I don't think you need to use `TextChoice` here you're find to just use `Choice`.

caseworker/queues/forms.py
```diff
@@ -172,3 +173,102 @@ class RoleChoices(models.TextChoices):
...
+        self.declared_fields["queue"].choices = self.get_queue_choices()
+        super().__init__(*args, **kwargs)
+
+    def get_queue_choices(self):
+        queues, _ = get_users_team_queues(self.request, self.user_id, False)
+        return [
+            (
+                TextChoice(
+                    Choice(queue_id, queue_name),
+                )
```
kevincarrogan, line 264: You can just use `Choice` here. You might actually just be able to return a tuple.

caseworker/queues/views.py
```diff
@@ -352,6 +330,82 @@ def get_context_data(self, *args, **kwargs):
         return super().get_context_data(*args, **context, **kwargs)
 
 
+SELECT_USERS = "SELECT_USERS"
+SELECT_QUEUE = "SELECT_QUEUE"
```
kevincarrogan, line 334: On the exporter side we've been following the pattern of wrapping these up in a class e.g.

```
class CaseAssignmentsCaseAssigneeSteps:
    SELECT_USERS = "SELECT_USERS"
    ...
```

Don't really know if this is better or worse but it's slightly inconsistent with that pattern.

caseworker/queues/views.py
```diff
@@ -352,6 +330,82 @@ def get_context_data(self, *args, **kwargs):
...
+    def get_queue_id(self):
+        return str(self.kwargs["pk"])
+
+    def get_context_data(self, form, **kwargs):
+        context = super().get_context_data(form, **kwargs)
+
+        context["back_link_url"] = self.get_back_link_url()
+        context["title"] = form.Layout.TITLE
+
+        return context
```
kevincarrogan, line 387: We really should bump this up into the base class at some point, I think we do this everywhere (not one for now really).

caseworker/templates/case/slices/summary.html
```diff
@@ -199,7 +199,7 @@
 				</span>
 			{% endfor %}
             <div class="app-case__summary-list__value__actions">
-			    <a id="link-change-assigned-users" class="govuk-link govuk-link--no-visited-state" href="{% url 'cases:assign_user' queue.id case.id %}">
+			    <a id="link-change-assigned-users" class="govuk-link govuk-link--no-visited-state" href="{% url 'queues:case_assignments_assign_user' queue.id %}?cases={{case.id}}&return_to={{CURRENT_PATH}}">
```
kevincarrogan, line 202: Is `CURRENT_PATH` nicely sanitised anywhere?

core/forms/templates/forms/filter_checkboxes.html
```diff
@@ -2,4 +2,4 @@
 
 {% include "./_filter_search_container.html" %}
 
-<script type="text/javascript" src="{% static '/javascripts/filter-checkbox-list.js' %}" nonce="{{ request.csp_nonce }}"></script>
+<script type="text/javascript" src="{% static 'javascripts/filter-checkbox-list.js' %}" nonce="{{ request.csp_nonce }}"></script>
```
kevincarrogan, line 5: I thought we'd made this fix before? Has this been missed being merged back?

unit_tests/caseworker/queues/test_views.py
```diff
@@ -723,3 +723,266 @@ def test_activity_updates(url, authorized_client):
...
+        "queues:case_assignments_assign_user",
+        kwargs={
+            "pk": data_queue["id"],
+        },
+    )
+    url = f"{url_base}?cases={case['case']['id']}"
+    response = authorized_client.get(url)
+    assert response.status_code == 200
+    context = response.context
+    assert context["form"]
```
kevincarrogan, line 749: We could even assert what form this is.

unit_tests/caseworker/queues/test_views.py
```diff
@@ -723,3 +723,266 @@ def test_activity_updates(url, authorized_client):
...
+    case = data_standard_case
+    url_base = reverse("queues:case_assignments_assign_user", kwargs={"pk": data_queue["id"]})
+    url = f"{url_base}?cases={case['case']['id']}"
+    data = {
+        "case_assignments_case_assignee-current_step": "SELECT_USERS",
+        "SELECT_USERS-users": [user_id],
+        "SELECT_USERS-note": "foobar",
+    }
+    # POST step 1
+    response = authorized_client.post(url, data)
```
kevincarrogan, line 789: Because this is such a common pattern we have a helper this which helps build the post data

https://github.com/uktrade/lite-frontend/blob/63d3e7b478ba95a347fc670cbdbf6058d809821e/unit_tests/exporter/conftest.py#L350.

Stops you having to write what step you are posting to and adding the form prefix.

caseworker/queues/views.py
currycoder: Agreed, but yeah..

caseworker/templates/case/slices/summary.html
currycoder: I've popped a `|urlencode` on this and others throughout this template

core/forms/templates/forms/filter_checkboxes.html
currycoder: The one we did was for `filter-radios.js` - this pre-emptively fixes the same thing for `filter-checkbox-list.js`

unit_tests/caseworker/queues/test_views.py
currycoder: I've pushed this up in to `unit_tests/conftest.py` and used as suggested.

kevincarrogan (approved):

PR #1127, opened 2023-02-21
Ltd 3372 case allocation banner

Harriethw:
### Aim

The “allocate yourself” banner was showing on case sub-pages, for TAU, advice, and notes, even after you have allocated yourself to a case.  
This was happening because the rule in the `case` template checking whether that banner should render needed the `user` in the context.

This PR adds the `user` to the different pages contexts. I'm not sure that I've covered all the different places it might need to be in... I'm also wondering if there were a more centralised way to add the user in the context.

[Link to story](https://uktrade.atlassian.net/jira/software/c/projects/LTD/boards/204?modal=detail&selectedIssue=LTD-3372)

### DIFF
--- a/caseworker/activities/views.py
+++ b/caseworker/activities/views.py
@@ -13,6 +13,7 @@ from caseworker.cases.services import (
 )
 from caseworker.cases.views.main import CaseTabsMixin
 from caseworker.queues.services import get_queue
+from caseworker.users.services import get_gov_user
 
 
 class NotesAndTimeline(LoginRequiredMixin, CaseTabsMixin, TemplateView):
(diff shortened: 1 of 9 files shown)


currycoder:
> I'm also wondering if there were a more centralised way to add the user in the context.

Great question.  Django has a thing called context processors which allow a more global way to inject things in to template context...  And if we look at the ones already set up for the caseworker app it seems like there might already be one that fits the bill for this; https://github.com/uktrade/lite-frontend/blob/dev/conf/caseworker.py#L51

Harriethw:
closing to open another branch

PR #1138, opened 2023-02-22
LTD-3324 Fix serial numbers display in select advice view

hnryjmes:
### Aim

[LTD-3396](https://uktrade.atlassian.net/browse/LTD-3396)

This change fixes the serial numbers display bug by passing the correct object into the template.

In this context we should be using `good.firearm_details` and not `good.good.firearm_details` as the former relates to `good_on_application` and the latter relates to `good`.

Also, two tests are added to make sure the display is correct: when there are serial numbers present, they should display, and when they are not present and `serial_numbers_available` is `LATER`, then the message `Not added yet` should display in place of the serial numbers.

[LTD-3396]: https://uktrade.atlassian.net/browse/LTD-3396?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/caseworker/advice/templates/advice/case_detail.html
+++ b/caseworker/advice/templates/advice/case_detail.html
@@ -79,22 +79,11 @@
                     {% endif %}
                   </td>
                 </tr>
-                {% if good.good.firearm_details %}
+                {% if good.firearm_details %}
                   <tr class="govuk-table__row">
                     <td class="govuk-table__cell govuk-!-font-weight-bold">Serial numbers</td>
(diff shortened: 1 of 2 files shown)


depsiatwal (approved):
nicely done

antroy-madetech (approved):
lgtm

kevincarrogan (approved):

PR #1140, opened 2023-02-23
ARS Prefix critical fixes

antroy-madetech:
### Aim
Fixed these bugs:
1. Could not edit a prefix to remove it.
2. No error displayed if an invalid value was added for prefix.

The latter was also solved for report summary.

[JIRA-3395](https://uktrade.atlassian.net/browse/LTD-3395)


caseworker/tau/forms.py
```diff
@@ -305,23 +305,43 @@ def clean(self):
 
     def validate_report_summary_subject(self, cleaned_data):
         subject_id = cleaned_data.get(REPORT_SUMMARY_SUBJECT_KEY)
+        subject_name_as_typed = self.data.get("_report_summary_subject")
+
+        def invalid_subject():
+            self.add_error(REPORT_SUMMARY_SUBJECT_KEY, "Enter a valid report summary subject")
```
kevincarrogan, line 311: I think just writing this out each time is more readable as seeing `invalid_subject` in the code later doesn't really tell me what's going on.

kevincarrogan: If we want to make the error message consistent then extracting that out as a variable/constant is probably better.

caseworker/tau/forms.py
```diff
@@ -305,23 +305,43 @@ def clean(self):
...
         if not subject_id:
+            if subject_name_as_typed:
+                invalid_subject()
             return
 
         try:
-            get_report_summary_subject(self.request, subject_id)
+            actual_name = get_report_summary_subject(self.request, subject_id)[REPORT_SUMMARY_SUBJECT_KEY]["name"]
+            if "_report_summary_subject" in self.data:
+                if subject_name_as_typed != actual_name:
```
kevincarrogan, line 321: I think this can just be composed using `and` instead of two `if` statements.

kevincarrogan (approved):

depsiatwal (approved):
lgtm

PR #1141, opened 2023-02-23
LTD-3372: use `current_user` to check assessment permissions

Harriethw:
### aim

The “allocate yourself” banner was showing on case sub-pages, for TAU, advice, and notes, even after you have allocated yourself to a case.  
This was happening because the rules `can_user_change_case` and `can_user_move_case_forward` are checked in several templates to decide whether to render buttons/forms and require a `user` key to be present in the context, which some didn't have.

This PR uses the already-present `current_user` from the caseworker `context_processor`, so we don't need to add or use a different value in these templates.

## feedback

I don't know how useful the context-checking tests really are, but it was a quick way for me to check every view when developing. I've added some frontend-y tests to `test_permissions.py` - possibly have been optimised with parameters at the expense of readability...  

[Link to story](https://uktrade.atlassian.net/jira/software/c/projects/LTD/boards/204?modal=detail&selectedIssue=LTD-3372)

### DIFF
--- a/caseworker/tau/views.py
+++ b/caseworker/tau/views.py
@@ -222,7 +222,6 @@ class TAUHome(LoginRequiredMixin, TAUMixin, FormView):
             "cle_suggestions_json": get_cle_suggestions_json(self.unassessed_goods),
             "organisation_documents": self.organisation_documents,
             "is_tau": self.caseworker["team"]["alias"] == TAU_ALIAS,
-            "user": self.caseworker,
         }
 
     def get_goods(self, good_ids):
--- a/unit_tests/caseworker/cases/test_permissions.py
+++ b/unit_tests/caseworker/cases/test_permissions.py
@@ -1,12 +1,12 @@
 import pytest
-from django.urls import reverse
+import re
+from core import client
 
-import rules
 from bs4 import BeautifulSoup
 from django.urls import reverse
 
 
-mock_gov_user_id = "2a43805b-c082-47e7-9188-c8b3e1a83cb0"
+mock_gov_user_id = "2a43805b-c082-47e7-9188-c8b3e1a83cb0"  # /PS-IGNORE
 
 
 @pytest.fixture(autouse=True)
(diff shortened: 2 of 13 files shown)


unit_tests/caseworker/cases/test_permissions.py
```diff
@@ -1,12 +1,13 @@
 import pytest
 from django.urls import reverse
+from core import client
+import re
```
Harriethw, line 4: i am going to rearrange the imports but cba to wait for CI to run again 🙃

unit_tests/caseworker/cases/test_permissions.py
```diff
@@ -90,3 +91,59 @@ def test_permission_move_case_forward_done_button(
...
+    mock_precedents_api,
+    data_standard_case,
+    view_name,
+    is_user_case_officer,
+    is_user_assigned,
+):
+    if is_user_case_officer:
+        data_standard_case["case"]["case_officer"] = mock_gov_user["user"]
+    elif is_user_assigned:
+        print("in is user assigned")
```
Harriethw, line 138: see above, will remove this print statement!

unit_tests/caseworker/tau/test_home.py
```diff
@@ -124,6 +130,9 @@ def test_home_content(
         "test-report-summary",
     ]
 
+    # test the user is passed through to check rule "can_user_change_case"
```
depsiatwal, line 133: might be worth adding where current user comes from i.e context processor 
for reference

depsiatwal (approved):
nicely done just small comment about adding a note somewhere to explain how current user gets set.

caseworker/tau/views.py
```diff
@@ -222,7 +222,6 @@ def get_context_data(self, **kwargs):
             "cle_suggestions_json": get_cle_suggestions_json(self.unassessed_goods),
             "organisation_documents": self.organisation_documents,
             "is_tau": self.caseworker["team"]["alias"] == TAU_ALIAS,
-            "user": self.caseworker,
```
currycoder, line 225: Just checking, is this needed for anything else in this template?

unit_tests/caseworker/cases/test_permissions.py
```diff
@@ -90,3 +90,58 @@ def test_permission_move_case_forward_done_button(
...
+        ["cases:advice_view", False, False],
+        ["cases:tau:home", False, False],
+        ["cases:case", True, False],
+        ["cases:activities:notes-and-timeline", True, False],
+        ["cases:advice_view", True, False],
+        ["cases:tau:home", True, False],
+        ["cases:case", False, True],
+        ["cases:activities:notes-and-timeline", False, True],
+        ["cases:advice_view", False, True],
+        ["cases:tau:home", False, True],
```
currycoder, line 118: Brilliant!

currycoder (approved):

caseworker/tau/views.py
Harriethw: good question, and not that I can see - it [was added recently](https://github.com/uktrade/lite-frontend/commit/d472b66d0af32d5d3aafbca5b224844373b714df#diff-6766bbcc80abd25b3350dce2f181d4341e3c13c477e428de6dc455f17c5a4f5e) for the sake of the rule afaik

PR #1145, opened 2023-02-23
Fix to allow user to clear the prefix during edit

antroy-madetech:
### Aim
Fixed these bugs:
1. Could not edit a prefix to remove it.
2. No error displayed if an invalid value was added for prefix.

The latter was also solved for report summary.

[JIRA-3395](https://uktrade.atlassian.net/browse/LTD-3395)


kevincarrogan (approved):

depsiatwal (approved):
lgtm

PR #1146, opened 2023-02-24
LTD-3194: UI changes for LU countersign

saruniitr:
### Aim

UI changes that allow LU to countersign using 'Recommendations and decision' tab.
Countersigner can now accept/reject original outcome providing comments for their decision.
This is for both first and second countersign which may be required in some Cases.

Dependencies
- Requires API changes from https://github.com/uktrade/lite-api/pull/1322
- Feature flags set in both API and FE


hnryjmes (approved):

depsiatwal (approved):
lgtm

PR #1154, opened 2023-02-28
Ltd 3421 report summary validation

kevincarrogan:
### Aim

This updates the validation for report summary to take into account whether the assessment is being marked as not requiring CLEs.

If no CLEs are required then the report summary subject can be blank.

[LTD-3421](https://uktrade.atlassian.net/browse/LTD-3421)

[LTD-3421]: https://uktrade.atlassian.net/browse/LTD-3421?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/caseworker/assets/javascripts/tau/ars.js
+++ b/caseworker/assets/javascripts/tau/ars.js
@@ -54,7 +54,6 @@ const initAutocompleteField = (summaryFieldType, summaryFieldPluralised) => {
     showNoOptionsFound: true,
     autoselect: true,
     showAllValues: true,
-    required: "subject" == summaryFieldType,
     confirmOnBlur: true,
   });
 };
--- a/caseworker/tau/forms.py
+++ b/caseworker/tau/forms.py
@@ -23,6 +23,7 @@ from ..report_summary.services import get_report_summary_prefix, get_report_summ
 
 REPORT_SUMMARY_SUBJECT_KEY = "report_summary_subject"
 REPORT_SUMMARY_PREFIX_KEY = "report_summary_prefix"
+MISSING_REPORT_SUMMARY_SUBJECT = "Enter a report summary subject"
 INVALID_REPORT_SUMMARY_SUBJECT = "Enter a valid report summary subject"
 INVALID_REPORT_SUMMARY_PREFIX = "Enter a valid report summary prefix"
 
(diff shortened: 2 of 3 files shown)


caseworker/tau/forms.py
```diff
@@ -305,10 +306,14 @@ def clean(self):
 
         return cleaned_data
 
-    def validate_report_summary_subject(self, cleaned_data):
+    def validate_report_summary_subject(self, cleaned_data, has_no_cle_entries):
         subject_id = cleaned_data.get(REPORT_SUMMARY_SUBJECT_KEY)
         subject_name_as_typed = self.data.get("_report_summary_subject")
 
+        if not has_no_cle_entries and not subject_id:
```
currycoder, line 313: The double negative is making me double take.  Can we make it `has_cle_entries` to make this particular bit easier to grok

caseworker/assets/javascripts/tau/ars.js
```diff
@@ -54,7 +54,6 @@ const initAutocompleteField = (summaryFieldType, summaryFieldPluralised) => {
     showNoOptionsFound: true,
     autoselect: true,
     showAllValues: true,
-    required: "subject" == summaryFieldType,
```
currycoder, line 57: Do I understand this correctly that we are removing the JS validation of required/not required and deferring it to the backend?

caseworker/tau/forms.py
kevincarrogan: This is how the rest of the code refers to it which is why I've not changed it. It was hurting my head too but I didn't want to change everywhere that this was used.

caseworker/assets/javascripts/tau/ars.js
kevincarrogan: It's not actually JS validation that is kicking in here. The component adds the `required` attribute to the input which is against the GDS style guide.

currycoder (approved):

caseworker/tau/forms.py
```diff
@@ -257,7 +255,10 @@ def clean(self):
             self.add_error("does_not_have_control_list_entries", self.MESSAGE_NO_CLC_REQUIRED)
         # report summary is required when there are CLEs
 
-        self.validate_report_summary_subject(cleaned_data)
+        # We pass in has_no_cle_entries here because throwing the errors above will clear out the value from
+        # cleaned_data so if we try to get it in the below method we won't have the original value which we want to be
+        # able to set the right validation for the report summary subject
+        self.validate_report_summary_subject(cleaned_data, not has_no_cle_entries)
```
saruniitr, line 261: thanks for the comment, was wondering why it needs to be passed as argument.

saruniitr (approved):

PR #1155, opened 2023-02-28
LTD-3432-banner-persisting-bug 

depsiatwal:
Hotfix into UAT which is blocking deployment for case allocation.

When we have multiple assigned users from a queue on a case the banner was persisting. This is because the code assumed only one list and return on the first iteration. We now loop through all assigned users for muliple queue until we have ex the exhausted the complete list

LTD-](https://uktrade.atlassian.net/browse/LTD-)](https://uktrade.atlassian.net/browse/LTD-3432

### DIFF
--- a/unit_tests/caseworker/core/test_permissions.py
+++ b/unit_tests/caseworker/core/test_permissions.py
@@ -14,7 +14,10 @@ mock_gov_user_id = "2a43805b-c082-47e7-9188-c8b3e1a83cb0"  # /PS-IGNORE
             {
                 "fake queue": [
                     {"id": mock_gov_user_id},
-                ]
+                ],
+                "fake queue 2": [
+                    {"id": "12345zyz"},
+                ],
             },
             True,
         ),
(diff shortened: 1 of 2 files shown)


unit_tests/caseworker/core/test_permissions.py
```diff
@@ -14,7 +14,10 @@
             {
                 "fake queue": [
                     {"id": mock_gov_user_id},
-                ]
+                ],
+                "fake queue 2": [
+                    {"id": "12345zyz"},
+                ],
```
currycoder, line 20: Could there be an extra parameter set here where the `mock_gov_user_id` is assigned to the second queue but not the first?  I think that would have failed before this change

kevincarrogan, line 20: How does this prove the code change? It looks like we're missing a test where the user id is actually on the second queue and not the first.

currycoder (approved):

PR #1157, opened 2023-03-01
Multiple countersignatures for LU

antroy-madetech:
### Aim

Update the various advice views to display LU countersignatures where appropriate. This should be anywhere final advice is displayed if this is an LU case.

[LTD-3413](https://uktrade.atlassian.net/browse/LTD-3413)

[LTD-3413]: https://uktrade.atlassian.net/browse/LTD-3413?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/caseworker/advice/templates/advice/advice_details.html
+++ b/caseworker/advice/templates/advice/advice_details.html
@@ -84,7 +84,20 @@
 </table>
 <div class="govuk-grid-row">
     <div class="govuk-grid-column-three-quarters">
-        {% if advice.0.countersigned_by %}
+        {% if is_lu_countersigning %}
+            <div id="countersignatures">
+                {% for countersign_advice in ordered_countersign_advice %}
+                <div class="govuk-!-padding-4 govuk-!-margin-bottom-2" style="border: 1px solid black;">
+                    <div class="govuk-grid-row">
+                        <div class="govuk-grid-column-three-quarters">
+                            <h2 class="govuk-heading-m">Countersigned by {{ countersign_advice.countersigned_user|full_name }}</h2>
+                            <p class="govuk-body">{{ countersign_advice.reasons }}</p>
+                        </div>
+                    </div>
+                </div>
+            {% endfor %}
+            </div>
+        {% elif advice.0.countersigned_by %}
         <div class="govuk-!-padding-4 govuk-!-margin-bottom-2" style="border: 1px solid black;">
             <div class="govuk-grid-row">
                 <div class="govuk-grid-column-three-quarters">
--- a/caseworker/advice/views.py
+++ b/caseworker/advice/views.py
@@ -1,17 +1,13 @@
 from http import HTTPStatus
 
+import sentry_sdk
 from django.conf import settings
 from django.http import Http404, HttpResponseRedirect
-from django.views.generic import FormView, TemplateView
-from django.urls import reverse
 from django.shortcuts import redirect
+from django.urls import reverse
 from django.utils.functional import cached_property
+from django.views.generic import FormView, TemplateView
 from requests.exceptions import HTTPError
-import sentry_sdk
-
-from core import client
-from core.constants import SecurityClassifiedApprovalsType, OrganisationDocumentType
-from core.decorators import expect_status
 
 from caseworker.advice import forms, services, constants
 from caseworker.advice.forms import BEISTriggerListAssessmentForm, BEISTriggerListAssessmentEditForm
@@ -21,7 +17,10 @@ from caseworker.core.helpers import get_organisation_documents
 from caseworker.core.services import get_denial_reasons
 from caseworker.tau.summaries import get_good_on_application_tau_summary
 from caseworker.users.services import get_gov_user
+from core import client
 from core.auth.views import LoginRequiredMixin
+from core.constants import SecurityClassifiedApprovalsType, OrganisationDocumentType
+from core.decorators import expect_status
 
 
 class CaseContextMixin:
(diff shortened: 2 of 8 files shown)


caseworker/advice/templates/advice/advice_details.html
```diff
@@ -84,7 +84,21 @@ <h2 class="govuk-heading-m app-bg-colour--red govuk-!-padding-2">
 </table>
 <div class="govuk-grid-row">
     <div class="govuk-grid-column-three-quarters">
-        {% if advice.0.countersigned_by %}
+        {% if is_lu_countersigning %}
+            {# pp([(ca['reasons'],ca['countersigned_user']['first_name']) for ca in  context['case']['countersign_advice']]) #}
```
saruniitr, line 88: debug code

caseworker/advice/views.py
```diff
@@ -95,14 +95,21 @@ def get_context_data(self, **kwargs):
...
             **self.get_context(case=self.case),
             "case": self.case,
             "queue_pk": self.kwargs["queue_pk"],
             "caseworker": self.caseworker,
+            "is_lu_countersigning": (is_in_lu_team and settings.FEATURE_LU_POST_CIRC_COUNTERSIGNING),
+            "ordered_countersign_advice": self.ordered_countersign_advice(),
         }
 
+    def ordered_countersign_advice(self):
+        countersign_advice = [cs for cs in self.case.get("countersign_advice", []) if cs.get("advice", {}).get("good")]
```
saruniitr, line 110: this can be an issue in the case of multiple products on the application. There will be one advice object for each product.
I think it is enough to just sort based on order and pick first one from each of those.

unit_tests/caseworker/conftest.py
```diff
@@ -611,6 +616,82 @@ def advice_for_countersign(current_user):
...
+            "last_name": "Bob",
+            "role_name": "Adviser",
+            "status": "Active",
+            "team": lu_team,
+        },
+    }
+
+
+@pytest.fixture
+def first_countersignature():
```
saruniitr, line 652: I already defined a fixture for this https://github.com/uktrade/lite-frontend/blob/dev/unit_tests/caseworker/advice/test_services.py#L72

If you want to use this one then I think it should take `advice` as argument and include that many countersignatures. Currently it assumes 3 advice items.

We also have user fixtures defined.

caseworker/advice/views.py
saruniitr: Can you check in the case where there are multiple products on the application?

antroy-madetech: I think it would show multiple. I've refactored and fixed this so the tests have multiple goods and is now grouping by order and picking one sig from each.

unit_tests/caseworker/conftest.py
antroy-madetech: Refactored into a function that is more flexible.

caseworker/advice/views.py
```diff
@@ -95,14 +95,25 @@ def get_context_data(self, **kwargs):
...
+            "ordered_countersign_advice": self.ordered_countersign_advice(),
         }
 
+    def ordered_countersign_advice(self):
+        def order_fn(case):
+            return case["order"]
+
+        ordered_advice = sorted(self.case.get("countersign_advice", []), key=order_fn)
+        countersign_advice = [list(g)[0] for k, g in itertools.groupby(ordered_advice, key=order_fn)]
+        return sorted(countersign_advice, key=order_fn, reverse=True)
```
saruniitr, line 115: I think this is making it more complicated than necessary.

What do you think of something like this?

```
from collections import defaultdict
results = defaultdict(list)

for item in self.case.get("countersign_advice", []):
    # here we can probably use the order value itself as the key instead of append "order_"
    results[f"order_{item['order']}"].append(item)

After this we have a dict with keys are order and list of countersignatures as a values for each of that order.
In the template we can just loop through the keys with higher order first and pick first element from the list
```

antroy-madetech: Changed to something I think is simpler than the previous options.

caseworker/advice/views.py
```diff
@@ -95,14 +94,21 @@ def get_context_data(self, **kwargs):
...
             **self.get_context(case=self.case),
             "case": self.case,
             "queue_pk": self.kwargs["queue_pk"],
             "caseworker": self.caseworker,
+            "is_lu_countersigning": (is_in_lu_team and settings.FEATURE_LU_POST_CIRC_COUNTERSIGNING),
+            "ordered_countersign_advice": self.ordered_countersign_advice(),
         }
 
+    def ordered_countersign_advice(self):
+        countersignature_per_order = {cs["order"]: cs for cs in self.case.get("countersign_advice", [])}
```
saruniitr, line 109: There will be multiple countersign_advice objects and this overwrites the data, we don't want to lose data here.

caseworker/advice/views.py
```diff
@@ -95,14 +94,25 @@ def get_context_data(self, **kwargs):
...
+            "ordered_countersign_advice": self.ordered_countersign_advice(),
         }
 
+    def ordered_countersign_advice(self):
+        """
+        Return a single countersignature per order value in order to filter out duplicates
+        for display in the templates
+        """
+        one_countersignature_per_order_value = {cs["order"]: cs for cs in self.case.get("countersign_advice", [])}
+        return sorted(one_countersignature_per_order_value.values(), key=lambda cs: cs["order"], reverse=True)
```
hnryjmes, line 114: does type coercion to integer happen here or is the `cs["order"]` variable guaranteed to be an integer?

hnryjmes (approved):
looks good. should also be easily adaptable if or when we support other teams doing double countersigning.

saruniitr (approved):

PR #1159, opened 2023-03-02
LTD-3407: return user to `return_to` url in allocate case journey

Harriethw:
### Aim

When a user clicks the 'Allocate case' button from a case detail page,
we would like them to be returned to that case after they have completed
the journey to assign either a case officer or a case adviser.

This commit passes the `return_to` param from the button through the
different forms used.

[LTD-3407](https://uktrade.atlassian.net/browse/LTD-3407)

[LTD-3407]: https://uktrade.atlassian.net/browse/LTD-3407?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/caseworker/queues/views.py
+++ b/caseworker/queues/views.py
@@ -340,7 +340,8 @@ class CaseAssignmentsCaseOfficer(LoginRequiredMixin, SuccessMessageMixin, FormVi
         return update_case_officer_on_cases(self.request, case_ids, user_id)
 
     def get_success_url(self):
-        return reverse("queues:cases", kwargs={"queue_pk": self.kwargs["pk"]})
+        default_success_url = reverse("queues:cases", kwargs={"queue_pk": self.kwargs["pk"]})
+        return self.request.GET.get("return_to", default_success_url)
 
     def get_context_data(self, *args, **kwargs):
         context = {
--- a/unit_tests/caseworker/queues/test_views.py
+++ b/unit_tests/caseworker/queues/test_views.py
@@ -474,6 +474,12 @@ def test_case_assignment_case_office_no_user_selected(authorized_client, mock_go
     }
 
 
+@pytest.fixture
+def post_to_step_user_assignment_with_return_to(post_to_step_factory, user_assignment_url):
+    user_assignment_url = user_assignment_url + f"&return_to={example_return_to_url}"
+    return post_to_step_factory(user_assignment_url)
+
+
 def test_case_assignment_case_adviser(
     mock_gov_users,
     mock_team_queue,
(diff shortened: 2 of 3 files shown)


caseworker/queues/views.py
```diff
@@ -340,7 +340,8 @@ def update_case_officer(self, user_id):
         return update_case_officer_on_cases(self.request, case_ids, user_id)
 
     def get_success_url(self):
-        return reverse("queues:cases", kwargs={"queue_pk": self.kwargs["pk"]})
+        default_success_url = reverse("queues:cases", kwargs={"queue_pk": self.kwargs["pk"]})
+        return self.request.GET.get("return_to", default_success_url)
```
kevincarrogan, line 344: This currently trusts any URL that we are given in the parameter as something we can redirect to, which is a potential security issue.

Django has a function you can use to ensure that we're sticking to our domain and scheme.

You can see an example of this being used here https://github.com/uktrade/lite-frontend/blob/943b03ec6a624f8779e3101d58f8ca7192761be9/core/cookies/views.py#L12.

unit_tests/caseworker/queues/test_views.py
```diff
@@ -18,13 +18,17 @@
 
 queue_pk = "59ef49f4-cf0c-4085-87b1-9ac6817b4ba6"
 
+gov_user_id = "1f288b81-2c26-439f-ac32-2a43c8b1a5cb"
```
saruniitr, line 21: Can the below fixture be used instead?
https://github.com/uktrade/lite-frontend/blob/dev/unit_tests/caseworker/conftest.py#L37
or
https://github.com/uktrade/lite-frontend/blob/dev/unit_tests/caseworker/conftest.py#L167

caseworker/queues/views.py
Harriethw: good shout... is it possible Django does some kind of validation itself? I can't find anything in the docs but when I pass `http://www.not-valid.com` it kicks out a Forbidden error even without using that method to check it

unit_tests/caseworker/queues/test_views.py
Harriethw: nope seems like it has to use the `mock_gov_users` in order to mock the right endpoints

caseworker/queues/views.py
Harriethw: `www.example.com` _does_ pass the `url_has_allowed_host_and_scheme` test... I'm guessing because of [how the utility works?](https://docs.djangoproject.com/en/4.1/_modules/django/utils/http/)
And I don't understand why I would get a forbidden error on the `POST` request by changing the `return_to` from `www.example.com` to `https://www.example.com` without adding any url validation 🤔 
I'm breaking for lunch now but can dig deeper after.

Harriethw: we discovered the middleware `ValidateReturnToMiddleware` and decided to just put unit tests covering the endpoints in this PR.

kevincarrogan (approved):

saruniitr (approved):

PR #1161, opened 2023-03-02
Remove flags when countersigning form submitted

hnryjmes:
### Aim 

This change modifies the view and view tests for LU countersigning to ensure that the correct LU flags are removed on submission of the form.

[LTD-3195](https://uktrade.atlassian.net/browse/LTD-3195)

[LTD-3195]: https://uktrade.atlassian.net/browse/LTD-3195?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


caseworker/advice/services.py
```diff
@@ -384,6 +393,33 @@ def countersign_decision_advice(request, case, queue_id, caseworker, formset_dat
     response.raise_for_status()
 
 
+def remove_lu_countersign_flags(request, case, queue_id, caseworker, order, outcome_accepted):
+    # TODO: refactor to move this logic somewhere else
+
+    if order == FIRST_COUNTERSIGN and outcome_accepted == OUTCOME_ACCEPTED_YES:
```
saruniitr, line 399: Instead of `order` we should be using the `queue_alias` because `order` will be usually `1` for other queues also.
If it is "Licensing manager countersigning" queue then take order as 1
If it is "Senior licensing manager countersigning" queue then take order as 2

caseworker/advice/services.py
```diff
@@ -359,6 +364,10 @@ def countersign_decision_advice(request, case, queue_id, caseworker, formset_dat
     if queue_alias == LU_SR_LICENSING_MANAGER_QUEUE:
         order = SECOND_COUNTERSIGN
 
+    # for LU countersigning, we expect only one form in the formset
+    outcome_accepted = formset_data[0]["outcome_accepted"]
+    remove_lu_countersign_flags(request, case, queue_id, caseworker, order, outcome_accepted)
+
```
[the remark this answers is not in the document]
hnryjmes: good point. when I come around to actually writing the service function again I'll look at calling it from the view. presumably in `ReviewCountersignDecisionAdviceView` after `services.countersign_decision_advice` is called but before the `HttpResponseRedirect` is returned.

caseworker/advice/services.py
hnryjmes: I will make sure it is based on the queue when I write this function again

depsiatwal (approved):
all looks good to me

hnryjmes:
closing as we are not doing this approach anymore, we are going with a different approach instead

PR #1164, opened 2023-03-03
Report Summary Improvements

antroy-madetech:
Remove 'blur' on autoselected item.
Added proper copy. Fixed pointer.
Minimise the "empty" option if the prefix or subject dropdowns are empty.

[LTD-3416](https://uktrade.atlassian.net/browse/LTD-3416)

[LTD-3416]: https://uktrade.atlassian.net/browse/LTD-3416?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/caseworker/assets/styles/overrides/_autocomplete.scss
+++ b/caseworker/assets/styles/overrides/_autocomplete.scss
@@ -6,3 +6,11 @@
 .lite-autocomplete__option {
     min-height: 2ex;
 }
+.lite-autocomplete__option--focused {
+    div.govuk-body {
+        color: white;
+    }
+}
+.lite-autocomplete__hint {
+    padding-left: 3px;
+}
--- a/caseworker/assets/styles/overrides/_autocomplete.scss
+++ b/caseworker/assets/styles/overrides/_autocomplete.scss
@@ -9,9 +9,7 @@
     border: none;
 }
 .lite-autocomplete__option:has(.govuk-body:empty)  {
-    padding-bottom: 0;
-    padding-top: 0;
-    min-height: 0;
+    display: none;
 }
 .lite-autocomplete__option--focused {
     div.govuk-body {
--- a/caseworker/assets/styles/overrides/_autocomplete.scss
+++ b/caseworker/assets/styles/overrides/_autocomplete.scss
@@ -3,8 +3,15 @@
         color: #fff;
     }
 }
-.lite-autocomplete__option {
-    min-height: 2ex;
+// The following selectors hide the blank option if it is the only
+// option in the dropdown.
+.lite-autocomplete__menu:has(.govuk-body:empty)  {
+    border: none;
+}
+.lite-autocomplete__option:has(.govuk-body:empty)  {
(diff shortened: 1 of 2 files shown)


caseworker/assets/styles/overrides/_autocomplete.scss
```diff
@@ -3,6 +3,21 @@
...
     }
 }
-.lite-autocomplete__option {
-    min-height: 2ex;
+// The following selectors hide the blank option if it is the only
+// option in the dropdown.
+.lite-autocomplete__menu:has(.govuk-body:empty)  {
+    border: none;
+}
+.lite-autocomplete__option:has(.govuk-body:empty)  {
```
hnryjmes, line 11: is this the css thing you mentioned before? i.e. you were concerned about if we needed to support firefox? https://caniuse.com/css-has

antroy-madetech: Yes, this is the css thing. Firefox doesn't support :has

antroy-madetech: Colin and Jessica have agreed that it's fine not to support Firefox and IE in this instance. They'll raise a future ticket at some point if the IE/Firefox support is deemed necessary.

antroy-madetech: However it doesn't look great, as that blank suggestion only appears if you clear the field, so it's the sole option that happens to be empty.

antroy-madetech: So this fix makes it look like that option isn't there, but still lets you clear the field by collapsing the option LI and removing the UL's border.

kevincarrogan (approved):

caseworker/assets/styles/overrides/_autocomplete.scss
```diff
@@ -3,6 +3,21 @@
...
+    padding-bottom: 0;
+    padding-top: 0;
+    min-height: 0;
```
saruniitr, line 14: Basically to remove this empty blue box, we are setting the height to 0?
Is this the standard way of achieving this in css?

antroy-madetech: It's _a_ way of doing it. I suppose I could also do "display:none". Not sure if that would break the component though. I can test it easily enough. It's more the :has css selector I was worried about (as it is unsupported on IE and Firefox), but Colin and Jessica are fine with that for now.

hnryjmes: Yeah maybe just test display:none out and see if that works, it might make the code more readable for anyone working on this in the future :)

antroy-madetech: display:none works fine.

hnryjmes (approved):

PR #1167, opened 2023-03-06
Ltd 3414 edit countersign decision

MarkThomasLite:
Add EditCountersignDecisionAdviceView reusing ReviewCountersignDecisionAdviceView and updating the existing decision(s).


caseworker/advice/services.py
```diff
@@ -204,6 +205,18 @@ def get_countersigners_decision_advice(case, caseworker):
     return countersigned_by
 
 
+def get_decision_advices_by_countersigner(case, caseworker):
+    """Get users countersign advice on this case with accept/reject decision."""
+    advices = []
+    for advice in case.countersign_advice:
+        if (
+            advice["countersigned_user"]["team"]["id"] == caseworker["team"]["id"]
+            and advice["countersigned_user"]["id"] == caseworker["id"]
```
hnryjmes, line 214: is there ever a situation where the user ids would match but the team ids would NOT match?

caseworker/advice/services.py
```diff
@@ -383,6 +396,27 @@ def countersign_decision_advice(request, case, queue_id, caseworker, formset_dat
     response.raise_for_status()
 
 
+def update_countersign_decision_advice(request, case, caseworker, form_data):
+    # not tracking id through process so currently unable to identify advice individually
+    # retrieve countersign decision advices for countersigner and updating them with the same form data
+    countersign_advices = get_decision_advices_by_countersigner(case, caseworker)
+    if len(countersign_advices) == 0:
+        return {"error": "countersigner decision advice not found"}, status.HTTP_400_BAD_REQUEST
```
hnryjmes, line 404: just a question, not a suggestion - does this error get shown to the user? or is it just shown in logging?

caseworker/advice/templates/advice/view_countersign.html
```diff
@@ -38,6 +38,10 @@ <h1 class="govuk-heading-xl">View recommendation</h1>
                 <a role="button" draggable="false" class="govuk-button govuk-button--secondary" href="{% url 'cases:countersign_edit' queue_pk case.id %}">
                     Edit recommendation
                 </a>
+                {% elif lu_can_edit and buttons.edit_recommendation %}
+                <a role="button" draggable="false" class="govuk-button govuk-button--secondary" href="{% url 'cases:countersign_decision_edit' queue_pk case.id %}">
+                    Edit countersigning decision
```
saruniitr, line 43: As per designs this is "Edit countersigning comment"

caseworker/advice/views.py
```diff
@@ -412,6 +412,9 @@ def get_context_data(self, **kwargs):
         advice_to_countersign = services.get_advice_to_countersign(self.case.advice, self.caseworker)
         context["advice_to_countersign"] = advice_to_countersign.values()
         context["can_edit"] = self.can_edit(advice_to_countersign)
+        context["lu_can_edit"] = self.caseworker_id in services.get_countersigners_decision_advice(
+            self.case, self.caseworker
+        )
```
saruniitr, line 417: This should be `False` if `FEATURE_LU_POST_CIRC_COUNTERSIGNING` is not enabled as this is an existing view.
We cannot raise an error in `dispatch` because it is existing view.

unit_tests/caseworker/advice/test_services.py
```diff
@@ -208,3 +217,106 @@ def test_get_countersign_advice_tab_context(
...
+    data_standard_case["case"]["advice"] = advice
+    advices = get_decision_advices_by_countersigner(Case(data_standard_case["case"]), current_user)
+    assert len(advices) == 1
+
+
+def put_request(rf, client, url, data=None):
+    request = rf.put(url, data if data else {})
+    request.session = client.session
+    request.requests_session = requests.Session()
+    return request
```
saruniitr, line 243: leftover?

caseworker/advice/services.py
MarkThomasLite: was thinking more around security if you move away from LU team

unit_tests/caseworker/advice/test_services.py
```diff
@@ -208,3 +217,106 @@ def test_get_countersign_advice_tab_context(
...
+
+
+def setup_requests_mock(requests_mock, client):
+    requests_mock.requests_session = requests.Session()
+    requests_mock.session = client.session
+    requests_mock.headers = {}
```
saruniitr, line 249: Not sure this is required?

caseworker/advice/services.py
MarkThomasLite: it doesn't get shown to the user, also shouldn't ever happen but added it to protect the api

unit_tests/caseworker/advice/test_services.py
MarkThomasLite: Yeah it shouldn't be yet without it the tests weren't running due to missing attributes

caseworker/advice/templates/advice/view_countersign.html
MarkThomasLite: Thought that may be an old design as you can change your decision not just comment

caseworker/advice/views.py
```diff
@@ -451,6 +458,43 @@ def get_success_url(self):
...
+
+    def post(self, request, *args, **kwargs):
+        context = self.get_context_data()
+        advice = context["advice_to_countersign"]
+        formset = forms.get_formset(self.form_class, len(advice), data=request.POST)
+        if formset.is_valid():
+            # single form item returned currently so using it to update decisions
+            services.update_countersign_decision_advice(
+                request, self.case, self.caseworker, form_data=formset.cleaned_data[0]
+            )
```
saruniitr, line 492: I think the correct way is to group the countersigned advice by the user (currently it is taking `advice_to_countersign` in line 486 which is different.
The `len(countersigned_advice)` will also be `1` as the same user countersigns all Advice instances and send `formset.cleaned_data` completely to the service function. In the service function we can use the grouped countersign advice and prepare payload for the API call.

It will be the same as in `ReviewCountersignDecisionAdviceView`

caseworker/advice/services.py
kevincarrogan: We shouldn't return things like this pretending that we're a response from the API.

If we think this won't happen then just throw an exception instead. It's actually even more confusing returning a 400 here as later we have `raise_for_status` which would actually always make a 400, if it were returned from the API, return an exception.

hnryjmes (approved):

caseworker/advice/views.py
```diff
@@ -451,6 +458,43 @@ def get_success_url(self):
...
+        return context
+
+    def post(self, request, *args, **kwargs):
+        context = self.get_context_data()
+        countersign_advice = services.get_decision_advices_by_countersigner(self.case, self.caseworker)
+        formset = forms.get_formset(self.form_class, len(countersign_advice), data=request.POST)
+        if formset.is_valid():
+            # single form item returned currently so using it to update decisions
+            services.update_countersign_decision_advice(
+                request, self.case, self.caseworker, form_data=formset.cleaned_data[0]
```
saruniitr, line 491: There should not be a need to send first item from the `cleaned_data` as it will only have one entry.

caseworker/advice/views.py
```diff
@@ -451,6 +458,43 @@ def get_success_url(self):
...
+        return super().dispatch(request, *args, **kwargs)
+
+    def get_data(self):
+        advices = services.get_decision_advices_by_countersigner(self.case, self.caseworker)
+        # given that there is currently no functionality to countersign individual pieces of advice
+        # the value of outcome_accepted on any single piece of advice is the same for the whole case
+        advice = advices[0]
+        reason_field = "approval_reasons" if advice.get("outcome_accepted") else "rejected_reasons"
+        data = [{"outcome_accepted": advice.get("outcome_accepted"), reason_field: advice.get("reasons")}]
+        return data
```
saruniitr, line 475: is this still getting used?

caseworker/advice/views.py
```diff
@@ -451,6 +458,43 @@ def get_success_url(self):
...
+    def get_context(self, **kwargs):
+        context = super().get_context()
+        advice = context["advice_to_countersign"]
+        data = self.get_data()
+        context["formset"] = forms.get_formset(self.form_class, len(advice), initial=data)
+        return context
+
+    def post(self, request, *args, **kwargs):
+        context = self.get_context_data()
+        countersign_advice = services.get_decision_advices_by_countersigner(self.case, self.caseworker)
```
saruniitr, line 486: I think the issue is this function is returning list of countersign objects submitted by this user, instead we need to group those objects by user which `get_countersigners_decision_advice()` returns.

Then `len(countersign_advice)` will be `1` as the same user countersigned all those and `formset.cleaned_data` can be sent completely.

In the function `update_countersign_decision_advice()` we can loop through those objects and update the data accordingly.

caseworker/advice/templates/advice/view_countersign.html
```diff
@@ -34,7 +34,11 @@ <h2 class="govuk-error-summary__title" id="error-summary-title">
             <br>
             <div class="govuk-grid-column-two-thirds">
                 <h1 class="govuk-heading-xl">View recommendation</h1>
-                {% if can_edit and buttons.edit_recommendation %}
+                {% if lu_can_edit and buttons.edit_recommendation %}
```
antroy-madetech, line 37: Is "edit_recommendation" the right variable for this? Should there be a separate "edit_countersignature" variable?

caseworker/advice/views.py
```diff
@@ -443,6 +447,9 @@ def get_context(self, **kwargs):
         advice = services.get_advice_to_countersign(self.case.advice, self.caseworker)
         context["formset"] = forms.get_formset(self.form_class, len(advice))
         context["advice_to_countersign"] = advice.values()
+        context["lu_can_edit"] = self.caseworker_id in services.get_countersigners_decision_advice(
+            self.case, self.caseworker
+        )
```
antroy-madetech, line 452: Feature flag not required here because dispatch will 404 if flag not present?

antroy-madetech (approved):
Couple of comments/questions. Otherwise looks good.

PR #1172, opened 2023-03-07
Stop crispy forms from generating its own form tag

kevincarrogan:
This was producing malformed HTML (a form within a form) that eventually caused the submit button to be rendered in the wrong place after we moved it out of the crispy form


antroy-madetech (approved):
Just a comment on the comment.

Harriethw (approved):

PR #1173, opened 2023-03-07
expand other recommendations if LU team

MarkThomasLite:
in other recommendations if you belong to the LU team the recommendations should be expanded, for any other team it should still be minimised


caseworker/advice/templates/advice/group-advice.html
```diff
@@ -2,7 +2,11 @@
 {% load get_denial_reason_display_values from advice_tags %}
 {% for team_advice in grouped_advice %}
     {% for user_advice in team_advice.advice %}
+    {% if is_lu_countersigning %}
+	<details class="govuk-details" data-module="govuk-details" open>
+    {% else %}
 	<details class="govuk-details" data-module="govuk-details">
+    {% endif %}
```
saruniitr, line 9: This is not the correct place and these are not the correct changes.
We need to change here https://github.com/uktrade/lite-frontend/blob/dev/caseworker/advice/templates/advice/view-advice.html#L97
Currently `hide_advice` is `True` for LU which we want to unhide as well now.

saruniitr (approved):

hnryjmes (approved):

PR #1175, opened 2023-03-08
Http error middleware

kevincarrogan:
### Aim

There are many cases where we make a service call and then check for an invalid response by running `raise_for_status`. When this fails we then get a HTTPError thrown and this ends up in Sentry.

The problem with the above is that in some cases the API is returning a 400 _and_ some text in the payload that would be important for us to know about, however just surfacing the HTTPError ends up hiding the response we are getting from the API.

This adds a middleware that adds a logger call of the response text which will end up in the breadcrumbs of the Sentry exception when we receive these errors.

### DIFF
--- a/unit_tests/core/middleware/test_http_error.py
+++ b/unit_tests/core/middleware/test_http_error.py
@@ -1,0 +1,28 @@
+from requests import HTTPError
+from requests.models import Response
+
+from core.middleware import HttpErrorHandlerMiddleware
+
+
+def test_http_error_handler_non_http_error_exception(rf, mocker, caplog):
+    get_response = mocker.MagicMock()
+    http_error_handler = HttpErrorHandlerMiddleware(get_response)
+
+    request = rf.get("/")
+    not_http_error = Exception()
+    assert http_error_handler.process_exception(request, not_http_error) is None
+    assert not caplog.records
+
+
+def test_http_error_handler_with_http_error(rf, mocker, caplog):
+    get_response = mocker.MagicMock()
+    http_error_handler = HttpErrorHandlerMiddleware(get_response)
+
+    response = Response()
+    response.status_code = 400
+    response._content = b'{"error": "this is an error"}'
+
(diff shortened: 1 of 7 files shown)


currycoder:
I don't really have anything to fault in the code for this but I just wanted to ask about the following statement;

> This adds a middleware that adds a logger call of the response text which will end up in the breadcrumbs of the Sentry exception when we receive these errors.

How does this work?  I would have thought that sentry would just show the error/warning log messages raised by our app and no info ones.  Does it have some concept of info messages that occur during a request and do something special with them to attach them to error events..?

kevincarrogan:
> I don't really have anything to fault in the code for this but I just wanted to ask about the following statement;
> 
> > This adds a middleware that adds a logger call of the response text which will end up in the breadcrumbs of the Sentry exception when we receive these errors.
> 
> How does this work? I would have thought that sentry would just show the error/warning log messages raised by our app and no info ones. Does it have some concept of info messages that occur during a request and do something special with them to attach them to error events..?

https://docs.sentry.io/platforms/python/guides/logging/

Specifically this bit:

level (default INFO): The Sentry Python SDK will record log records with a level higher than or equal to level as breadcrumbs. Inversely, the SDK completely ignores any log record with a level lower than this one. If a value of None occurs, the SDK won't send log records as breadcrumbs.

currycoder:
Brilliant - thanks for explaining.  Yeah sounds like just the lite touch that we need

currycoder (approved):

seijihg (approved):

unit_tests/core/middleware/test_http_error.py
```diff
@@ -0,0 +1,28 @@
+from requests import HTTPError
+from requests.models import Response
+
+from core.middleware import HttpErrorHandlerMiddleware
+
+
+def test_http_error_handler_non_http_error_exception(rf, mocker, caplog):
```
seijihg, line 7: spelling `non` instead of `none`

PR #1178, opened 2023-03-09
e2e tests using seleniarm for arm devices

hnryjmes:
outdated, please see Tomos's comment below

> I am looking for someone with an ARM machine (e.g. Mac M1) to test this chromium spike. It seems like it could work but I feel like I'm fixing quite specific errors to my setup now and I want someone else to try it at this stage to see if they see the same errors as me.
> 
> At the moment, I am using this method to run the tests:
> 
> 1. Make sure the frontend is running in docker (`docker-compose up --build`)
> 2.  Open a shell for caseworker (`docker exec -it $(docker ps -aqf "name=lite-frontend-caseworker") /bin/bash`
> 3. Run the frontend tests directly without make (`PIPENV_DOTENV_LOCATION=caseworker.env pipenv run pytest --headless ./ui_tests/caseworker`)
> 
> Currently the error I'm getting is the following:
> `E       selenium.common.exceptions.WebDriverException: Message: unknown error: net::ERR_PROXY_CONNECTION_FAILED`
> 
> I think this is something to do with `selenium-wire` and this issue seems relevant: https://github.com/wkeeling/selenium-wire/issues/424
> 
> (incidentally, @wkeeling, the author of selenium-wire, used to work on this team)
> 
> Please could someone test this spike out and see if you get to the same point as me, thanks!
> 
> To understand a bit more about seleniarm and why it exists, you can the first bit of the documentation here which should explain a little: https://github.com/seleniumhq-community/docker-seleniarm
>


docker-compose.yml
```diff
@@ -39,6 +39,16 @@ services:
...
+  selenium:
+    image: seleniarm/standalone-chromium:latest
+    ports:
+      - "4444:4444"
+    expose:
+      - 4444
+    shm_size: '2gb'
+    networks:
+      - lite
+
```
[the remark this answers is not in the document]
hnryjmes: One option would be that we could have the tests create a temporary docker shared volume just for the selenium and the other lite containers to use, and then remove it once we don't need it anymore (e.g. screenshots or other files have been moved elsewhere) at the end of the test run

docker-compose.yml
```diff
@@ -2,14 +2,15 @@ version: '3.4'
 
 services:
   caseworker:
-    platform: linux/amd64
+    # platform: linux/amd64
```
hnryjmes, line 5: I think we should have a team conversation about whether we need this at all, commenting it out is probably not the long term solution. The new Docker update for Mac with Rosetta built in also possibly makes this unnecessary now

Maybe someone else will have an opinion, but I'm also happy having this line commented out as the default as long as others agree

tests_common/fixtures/driver.py
```diff
@@ -1,43 +1,41 @@
-import base64
-
-from seleniumwire import webdriver
```
hnryjmes, line 3: did we verify the assumption that removing seleniumwire won't break anything?

Tllew: seems to work

Tllew (approved):

antroy-madetech (approved):
Works for me

PR #1181, opened 2023-03-09
Ltd 3442 countersign returned cases

antroy-madetech:
### Aim
This ticket has two parts:
1) Display a "countersignature rejected" warning on the Recommendations and Decisions tab when this is on the Post Circ to Finalise queue and there are rejected countersignatures present.
2) Enable the "Countersignatures required" warning on the same page for ManPads and AP Landmines flags.

[LTD-3442](https://uktrade.atlassian.net/browse/LTD-3442)


caseworker/advice/views.py
```diff
@@ -113,6 +113,16 @@ def ordered_countersign_advice(self):
...
 
+    def rejected_countersign_advice(self):
+        """
+        Return a single rejected countersignature per order value (if there are any) in order to filter out duplicates
+        for display in the templates
+        """
+        one_countersignature_per_order_value = {
+            cs["order"]: cs for cs in self.case.get("countersign_advice", []) if not cs["outcome_accepted"]
+        }
+        return sorted(one_countersignature_per_order_value.values(), key=lambda cs: cs["order"], reverse=True)
```
currycoder, line 124: Do we expect there to be duplicates?  Why wouldn't we want to show all of the rejections..?

currycoder: Also, say that a case is rejected from countersign once...  What prevents this from always showing up in the future?  Does the rejected advice get deleted/replaced somehow?

caseworker/advice/templates/advice/view-advice.html
```diff
@@ -56,9 +56,17 @@ <h2 class="govuk-error-summary__title" id="error-summary-title">
...
                 <div class="govuk-grid-column-three-quarters">
-
-                    {% if lu_countersign_required %}
-                        <div class="govuk-warning-text">
+                    {% if rejected_lu_countersignatures %}
+                        <div id="rejected-countersignature" class="govuk-warning-text">
+                            <span class="govuk-warning-text__icon" aria-hidden="true">!</span>
+                            <strong class="govuk-warning-text__text">
+                                <span class="govuk-warning-text__assistive">Warning</span>
+                                This case has been returned for editing by countersigner {{ rejected_lu_countersignatures.0.countersigned_user|full_name }}.
```
hnryjmes, line 64: there is a comma after 'editing' in the original design, so replace this line with:
`                                This case has been returned for editing, by countersigner {{ rejected_lu_countersignatures.0.countersigned_user|full_name }}.`

currycoder (commented):
A couple of initial questions for my understanding

caseworker/advice/templates/advice/view-advice.html
antroy-madetech: I suppose I should have left in the comma. Thought it read better without :-)

caseworker/advice/views.py
antroy-madetech: So the way countersignatures work is that they are attached to a good. So the first countersigner would end up adding their countersignature to each good in the case. The countersignatures have an "order" which dictates whether they are first or second signatures. So what this code is doing is picking an arbitrary signature for each "order" since all for each order will be identical. 

Not sure where deletion of rejection countersignatures gets done. It's out of scope for this ticket, @saruniitr - do you know where that gets handled? Future work or one of the other tickets?

caseworker/advice/templates/advice/view-advice.html
antroy-madetech: Added back in

caseworker/advice/views.py
saruniitr: We create `Advice` objects for each entity on the application (products, parties) and there will be a corresponding `CountersignAdvice` for each of those objects.
Currently we either approve all or reject all but when we make changes for FCDO that will change. They can approve for certain destinations and reject for other destinations.

In this case we are using the returned data to drive other logic so taking one value is not correct. It is not only used for display purposes in templates but changes user actions so not comfortable with the fact that we are overwriting data and using that for decision making.

They are all going to be accepted/rejected now which is an implementation detail and can change so it is not advisable to add changes that depend on that.

If the idea is to check if there are any rejected countersignatures then we can just loop through all of `case.countersign_advice` and return everything. In the template anyway we are taking one item from that list.

saruniitr: Multiple rejections are possible but my understanding is they are not common so we will handle that as a follow up

caseworker/advice/views.py
```diff
@@ -594,21 +604,35 @@ def get_context(self, **kwargs):
...
+            }
+            case_flag_ids = {flag["id"] for flag in self.case.all_flags}
+            lu_countersign_required = user_team_alias == services.LICENSING_UNIT_TEAM and bool(
+                lu_countersign_flags.intersection(case_flag_ids)
+            )
+
+        finalise_case = (
+            user_team_alias == services.LICENSING_UNIT_TEAM
+            and not lu_countersign_required
+            and not rejected_lu_countersignatures
```
saruniitr, line 626: These changes and the template changes need to be under feature flag

unit_tests/caseworker/advice/views/test_consolidate.py
```diff
@@ -59,6 +68,77 @@ def advice(current_user):
...
+FLAG_MAP = {
+    GREEN_COUNTRIES_ID: {
+        "id": GREEN_COUNTRIES_ID,
+        "name": "Green Countries",
+        "level": "Destination",
+        "colour": "green",
+        "label": "Green countries",
+        "priority": 20,
+        "alias": None,
+    },
```
saruniitr, line 84: this is not relevant for countersigning

caseworker/advice/views.py
```diff
@@ -594,21 +604,35 @@ def get_context(self, **kwargs):
         if user_team_alias in [services.LICENSING_UNIT_TEAM, services.MOD_ECJU_TEAM]:
             consolidated_advice = services.get_consolidated_advice(self.case.advice, user_team_alias)
         nlr_products = services.filter_nlr_products(self.case["data"]["goods"])
+        rejected_lu_countersignatures = self.rejected_countersign_advice()
 
-        lu_countersign_flags = {services.LU_COUNTERSIGN_REQUIRED, services.LU_SR_MGR_CHECK_REQUIRED}
-        case_flag_aliases = {flag["alias"] for flag in self.case.all_flags}
-        lu_countersign_required = user_team_alias == services.LICENSING_UNIT_TEAM and bool(
-            lu_countersign_flags.intersection(case_flag_aliases)
+        if rejected_lu_countersignatures:
```
saruniitr, line 609: This assumes countersign is required in all cases, shouldn't this be coupled with `lu_countersign_required` and only then check for rejected countersignatures?

saruniitr: We also need to consider the case where countersign required but no rejected countersignatures (all approved), in this case we should be able to finalise.

caseworker/advice/views.py
antroy-madetech: Ah - OK. I will sort that.

caseworker/advice/views.py
antroy-madetech: I was thinking that if countersignatures aren't required, then there won't be any countersignatures (and so no rejected ones). I can put in an explicit check for lu_countersign_required though to be sure.

I'll add in a test around the finalise case to ensure that still works.

caseworker/advice/templates/advice/view-advice.html
hnryjmes: I sort of agree with you but I'm not the person you have to convince :)

hnryjmes (approved):
looks good to me, meaning, while I haven't personally verified every line in this I don't see anything that looks broken or that might be a risk, and it appears to well covered by the tests

saruniitr (approved):

PR #1186, opened 2023-03-14
LU Countersignature Rework.

antroy-madetech:
New message box for rejected countersignature detail. 
Refactoring to support this. 
Also removal of rejected countersignatures from the display of countersignatures in the advice view further down.
In addition, as per comments, we only show the accepted countersignatures if there were no rejections (i.e. the case where the caseworker accepts the advice but the supervisor rejects it will prevent the accepted countersignatures from being displayed to avoid confusion)

Detail in this ticket:

[LTD-3482](https://uktrade.atlassian.net/browse/LTD-3482)

[LTD-3482]: https://uktrade.atlassian.net/browse/LTD-3482?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/caseworker/advice/templates/advice/view-advice.html
+++ b/caseworker/advice/templates/advice/view-advice.html
@@ -98,7 +98,7 @@
                     {% endif %}
 
                     <br><br>
-                    {% include "advice/advice_details.html" with advice=consolidated_advice team=True outcome_accepted_countersignatures_to_display="ACCEPTED" %}
+                    {% include "advice/advice_details.html" with advice=consolidated_advice team=True %}
                     <br>
                 </div>
 
--- a/caseworker/advice/templates/advice/view-advice.html
+++ b/caseworker/advice/templates/advice/view-advice.html
@@ -56,14 +56,24 @@
         <div class="govuk-width-container">
             <div class="govuk-grid-row govuk-!-margin-top-6">
                 <div class="govuk-grid-column-three-quarters">
-                    {% if is_lu_countersigning and rejected_lu_countersignatures %}
+                    {% if is_lu_countersigning and rejected_lu_countersignature %}
                         <div id="rejected-countersignature" class="govuk-warning-text">
                             <span class="govuk-warning-text__icon" aria-hidden="true">!</span>
                             <strong class="govuk-warning-text__text">
                                 <span class="govuk-warning-text__assistive">Warning</span>
-                                This case has been returned for editing, by countersigner {{ rejected_lu_countersignatures.0.0.countersigned_user|full_name }}.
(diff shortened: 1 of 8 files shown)


caseworker/advice/templates/advice/countersignature.html
```diff
@@ -0,0 +1,21 @@
+{% for countersign_advice in ordered_countersign_advice %}
+    {% if countersign_advice.outcome_accepted and not outcome_accepted_countersignatures_to_display == "REJECTED" %}
```
saruniitr, line 2: What is the need for `outcome_accepted_countersignatures_to_display`?
Countersignatures is considered complete if `outcome_accepted=True` and the requirement is to not show incomplete countersignatures, so the condition can simply be,
`{% if countersign_advice.outcome_accepted %}`

unit_tests/caseworker/conftest.py
```diff
@@ -821,58 +821,71 @@ def final_advice(current_user, lu_team):
...
-            "id": "654165",
-            "first_name": "Testy",
-            "last_name": "McTest",
-            "team": {
-                "id": "809eba0f-f197-4f0f-949b-9af309a844fb",
-                "name": "LU Team",
-                "alias": LICENSING_UNIT_TEAM,
-                "part_of_ecju": False,
-                "is_ogd": True,
+def countersignatures(good_with_advice_count=2, end_user=True, ultimate_end_user=True, accepted=[True]):
```
saruniitr, line 824: We need to base this on the application entities rather than fixing these here.
If the intention is to return countersignatures it needs to take the `case` and associated `Advice` of final level (count matching to number of entities) and then return countersignatures for specified order, outcome etc.
I am not convinced for the need to change parties etc because `CountersignAdvice` refers to the `Advice` which is either for a product/party

caseworker/advice/templates/advice/view-advice.html
```diff
@@ -88,7 +98,7 @@ <h2 class="govuk-error-summary__title" id="error-summary-title">
                     {% endif %}
 
                     <br><br>
-                    {% include "advice/advice_details.html" with advice=consolidated_advice team=True %}
+                    {% include "advice/advice_details.html" with advice=consolidated_advice team=True outcome_accepted_countersignatures_to_display="ACCEPTED" %}
```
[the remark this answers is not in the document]
saruniitr: rest of the changes looks good to me

saruniitr (approved):

Harriethw (approved):

PR #1187, opened 2023-03-16
LTD-3503: Use put method to update consolidated advice by LU

saruniitr:
### Aim

We added new put method in API to support updating of final level Advice. This is the consolidated advice by LU.
Previously updates were supported by that involved deleting existing objects and creating new ones in place of them instead of actually updating original objects. Previous method has unintended effects in terms of cascade deletes.

In this PR, FE is updated the use the new method for updating consolidated advice.
This is required for LU countersigning work hence restricted to LU team only to limit regressions (if any). It can easily be expanded to other advice levels/teams.

[LTD-3503](https://uktrade.atlassian.net/browse/LTD-3503)

[LTD-3503]: https://uktrade.atlassian.net/browse/LTD-3503?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/caseworker/advice/services.py
+++ b/caseworker/advice/services.py
@@ -349,7 +349,7 @@ def post_refusal_advice(request, case, data, level="user-advice"):
     return response.json(), response.status_code
 
 
-def update_approval_advice(request, case, caseworker, data, level):
+def update_advice(request, case, caseworker, advice_type, data, level):
     user_team_alias = caseworker["team"]["alias"]
     if user_team_alias != LICENSING_UNIT_TEAM:
         raise NotImplementedError(f"Implement approval advice update for {user_team_alias}")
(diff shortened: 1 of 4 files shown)


caseworker/advice/services.py
```diff
@@ -349,6 +349,48 @@ def post_refusal_advice(request, case, data, level="user-advice"):
...
+            "text": data["refusal_reasons"],
+            "denial_reasons": data["denial_reasons"],
+        }
+        for advice in consolidated_advice
+    ]
+    response = client.put(request, f"/cases/{case['id']}/{level}/", json)
+    response.raise_for_status()
+    return response.json(), response.status_code
+
+
```
antroy-madetech, line 393: Only thing different between the two methods as far as I can see is the proviso/denial_reasons in the request json. This could be easily either parameterised or the majority of the methods extracted.

unit_tests/caseworker/advice/views/test_consolidate_edit.py
```diff
@@ -26,6 +24,124 @@ def setup(mock_queue, mock_case, mock_denial_reasons, mock_post_team_advice, moc
...
+            "ultimate_end_user_id": None,
+            "consignee_id": None,
+            "third_party_id": None,
+            "proviso": "some conditions",
+            "pv_grading": None,
+            "collated_pv_grading": None,
+            "countersign_comments": "",
+            "countersigned_by_id": None,
+        },
+    ]
```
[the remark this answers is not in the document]
saruniitr: the fixture in the file `test_consolidate.py` doesn't work. We need specifically for LU consolidate so all of the level should be `final` and user also from LU.

caseworker/advice/services.py
saruniitr: refactored to a common function

unit_tests/caseworker/advice/views/test_consolidate_edit.py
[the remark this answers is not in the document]
saruniitr: I have updated it based on the subjects of the Case.

antroy-madetech (approved):

caseworker/advice/views.py
```diff
@@ -631,6 +631,30 @@ def get_form_kwargs(self):
...
+                )
+            if isinstance(form, forms.RefusalAdviceForm):
+                services.update_advice(
+                    self.request, self.case, self.caseworker, self.advice_type, form.cleaned_data, level
+                )
+        except HTTPError as e:
+            errors = e.response.json()["errors"]
+            form.add_error(None, errors)
+            return super().form_invalid(form)
+        return HttpResponseRedirect(self.get_success_url())
```
antroy-madetech, line 656: Looks like this needs a test to satisfy coverage.

saruniitr: yep, test added now

PR #1189, opened 2023-03-21
Added filter and new logic for getting the countersignature

antroy-madetech:
### Aim


caseworker/advice/templatetags/advice_tags.py
```diff
@@ -57,6 +59,19 @@ def countersigned_user_team(advice):
     return f"{advice['countersigned_by']['team']['name']}"
 
 
+@register.filter
+def countersignatures_for_advice(case, advice):
```
saruniitr, line 63: Could you please add a brief description of what is happening here?

saruniitr (approved):

caseworker/advice/views.py
```diff
@@ -102,17 +102,8 @@ def get_context_data(self, **kwargs):
             "queue_pk": self.kwargs["queue_pk"],
             "caseworker": self.caseworker,
             "is_lu_countersigning": (is_in_lu_team and settings.FEATURE_LU_POST_CIRC_COUNTERSIGNING),
-            "ordered_countersign_advice": self.ordered_countersign_advice(),
         }
 
-    def ordered_countersign_advice(self):
```
hnryjmes, line 108: unless I'm missing something, I don't think my PR does use this function, so you should be able to delete it if you want.

if we need it again we can get it from git

antroy-madetech: No - I agree. I checked and you aren't using it :+1:

PR #1190, opened 2023-03-21
wip

hnryjmes:
### Aim


caseworker/advice/views.py
```diff
@@ -670,13 +685,26 @@ def get_context(self, **kwargs):
             finalise_case = user_team_alias == services.LICENSING_UNIT_TEAM and not lu_countersign_required
 
         return {
-            **super().get_context(**kwargs),
-            "consolidated_advice": consolidated_advice,
-            "nlr_products": nlr_products,
             "finalise_case": finalise_case,
             "lu_countersign_required": lu_countersign_required,
             "rejected_lu_countersignature": rejected_lu_countersignature,
+        }
```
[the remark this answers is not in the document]
antroy-madetech: Unit tests required around all of these.
View tests will need adding/updating to assert that the finalise button is shown/not shown where appropriate.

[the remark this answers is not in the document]
hnryjmes: updated. the new logic for lu_countersign_required is now covered by a unit test.

unit_tests/caseworker/advice/views/test_consolidate.py
```diff
@@ -949,6 +952,24 @@ def test_case_returned_info_for_second_countersignature_rejection(
     assert not soup.find(id="finalise-case-button")
 
 
+@pytest.mark.parametrize(("is_lu_team", "rejected_lu_countersignature"), [(True, False), (True, True)])
```
antroy-madetech, line 955: There are actually 4 cases here. TF, TT, FT, FF

unit_tests/caseworker/advice/views/test_consolidate.py
```diff
@@ -949,6 +952,24 @@ def test_case_returned_info_for_second_countersignature_rejection(
...
+    view_instance.case = Case(data_standard_case)
+    view_instance.case.all_flags = [
+        FLAG_MAP[k] for k in [LU_COUNTERSIGN_REQUIRED_ID, LU_SR_MGR_CHECK_REQUIRED_ID, MANPADS_ID, AP_LANDMINE_ID]
+    ]
+    lu_countersign_required = view_instance.get_lu_countersign_required(is_lu_team, rejected_lu_countersignature)
+
+    if rejected_lu_countersignature:
+        assert not lu_countersign_required
+    else:
+        assert lu_countersign_required
```
antroy-madetech, line 970: If you expand the test parameters to take in an "expected_result" param, this can just be assert lu_countersign_required is expected_result.

antroy-madetech: Actually there are a lot more than just the 4 cases to test, as the logic depends on flags etc. Would be worth expanding to cover the important cases.

unit_tests/caseworker/advice/views/test_consolidate.py
```diff
@@ -963,7 +984,12 @@ def test_finalise_button_shown_if_no_rejected_countersignatures(
     data_standard_case["case"]["countersign_advice"] = countersignatures_for_advice(
         consolidated_advice, accepted=[True, True]
     )
-    data_standard_case["case"]["all_flags"] = [FLAG_MAP[GREEN_COUNTRIES_ID]]
+    data_standard_case["case"]["all_flags"] = [
+        FLAG_MAP[LU_COUNTERSIGN_REQUIRED_ID],
+        FLAG_MAP[LU_SR_MGR_CHECK_REQUIRED_ID],
+        FLAG_MAP[MANPADS_ID],
+        FLAG_MAP[AP_LANDMINE_ID],
+    ]
```
antroy-madetech, line 993: Looks like this tests that finalise is shown when all 4 relevant flags are set. Probably should be parameterised to test each flag independantly, and a test to confirm that it isn't shown if none of these are set (unless there's another test for this already present?)

saruniitr (approved):
One comment I leave it with you

unit_tests/caseworker/advice/conftest.py
```diff
@@ -9,3 +10,19 @@ def with_lu_countersigning_enabled(settings):
 @pytest.fixture
 def with_lu_countersigning_disabled(settings):
     settings.FEATURE_LU_POST_CIRC_COUNTERSIGNING = False
+
+
+@pytest.fixture
+def countersign_advice(data_standard_case, advice_for_countersign, current_user):
```
antroy-madetech, line 16: Doesn't look like this is used anywhere

unit_tests/caseworker/advice/views/test_consolidate.py
```diff
@@ -989,6 +992,95 @@ def test_finalise_button_shown_if_no_rejected_countersignatures(
...
+    requests_mock.get(
+        client._build_absolute_uri("/gov-users/2a43805b-c082-47e7-9188-c8b3e1a83cb0"),
+        json=gov_user,
+    )
+
+    response = authorized_client.get(view_consolidate_outcome_url)
+    assert response.status_code == 200
+
+    soup = BeautifulSoup(response.content, "html.parser")
+    assert (soup.find(id="finalise-case-button") is not None) is expected_value_finalise_case
```
antroy-madetech, line 1081: may be cleaner as `bool(soup.find(id="finalise-case-button")) is expected_value_finalise_case`

antroy-madetech (approved):
Couple of minor comments to address, otherwise approved.

PR #1192, opened 2023-03-23
New designs implemented

antroy-madetech:
### Aim
Previously we were not showing countersignatures where there is rejected advice in favour of a warning message that there is a rejection. This is fine for most use cases, but it is required that the countersigner can see any rejected advice (for example any rejected advice that they have made) in addition to any previous approved countersignatures.

This PR implements the updated designs in the ticket.

[LTD-3521](https://uktrade.atlassian.net/browse/LTD-3521)

[LTD-3521]: https://uktrade.atlassian.net/browse/LTD-3521?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/caseworker/advice/views.py
+++ b/caseworker/advice/views.py
@@ -83,6 +83,12 @@ class CaseContextMixin:
             if (dest["id"], self.caseworker["team"]["id"]) not in advised_on.items()
         }
 
+    def get_rejected_lu_countersignature(self):
+        if settings.FEATURE_LU_POST_CIRC_COUNTERSIGNING:
+            return self.rejected_countersign_advice()
+        else:
+            return None
+
     def get_context(self, **kwargs):
         return {}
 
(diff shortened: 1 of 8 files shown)


caseworker/advice/views.py
```diff
@@ -418,6 +418,7 @@ def get_context_data(self, **kwargs):
         )
         context["denial_reasons_display"] = self.denial_reasons_display
         context["current_tab"] = "cases:countersign_view"
+        context["show_rejected_countersignatures"] = True
```
hnryjmes, line 421: could you explain a bit what this does?

hnryjmes (approved):
Looks good, just a small comment but not critical

saruniitr (approved):

PR #1195, opened 2023-03-24
LTD-3515: Ignore invalid countersignatures

saruniitr:
### Aim

Countersignatures when they are created initially by countersigning managers are `valid`.
If a countersigning manager rejects countersigning then the Case is sent back to caseworker to edit their recommendation. Depending on the comments they update their recommendation and send it again for countersigning. At this point (after editing) we mark all previous countersignatures as `invalid` because it will start the journey again.
So we need to ignore invalid countersignatures otherwise we end up showing warnings etc even for invalid countersignatures.

 - Refactored a method that generates countersignatures to make it clear on what is happening and easy to parameterize in other tests
 - Add tests that check for presence warning message that the case is sent back when countersignatures are `valid` and the usual warning about case requiring countersignatures when they are `invalid`
 - Use `mock_gov_lu_user` where possible instead of mocking it in every test

[LTD-3515](https://uktrade.atlassian.net/browse/LTD-3515)

### DIFF
--- a/caseworker/advice/templates/advice/advice_details.html
+++ b/caseworker/advice/templates/advice/advice_details.html
@@ -89,7 +89,7 @@
                 <div class="countersignatures">
                     {% for countersign_advice in case|countersignatures_for_advice:advice  %}
                         {%  if countersign_advice.0.outcome_accepted %}
-                        <div class="govuk-!-padding-4 govuk-!-margin-bottom-2 countersigned-by">
+                        <div class="govuk-!-padding-4 govuk-!-margin-bottom-2 countersigned-by" style="border: 1px solid black;">
                             <div class="govuk-grid-row">
                                 <div class="govuk-grid-column-three-quarters">
                                     <h2 class="govuk-heading-m">{% if countersign_advice.0.order == 1 %}Countersigned{% elif countersign_advice.0.order == 2 %}Senior countersigned{% endif %} by {{ countersign_advice.0.countersigned_user|full_name }}</h2>
--- a/caseworker/advice/views.py
+++ b/caseworker/advice/views.py
@@ -121,7 +121,7 @@ class CaseContextMixin:
         rejection (case will be returned to edit the advice once a rejection has occurred.
         """
         for cs in self.case.get("countersign_advice", []):
-            if not cs["outcome_accepted"]:
+            if cs["valid"] is True and cs["outcome_accepted"] is False:
                 return cs
         return None
 
--- a/unit_tests/caseworker/advice/views/test_countersign_view.py
+++ b/unit_tests/caseworker/advice/views/test_countersign_view.py
@@ -1,16 +1,12 @@
-import copy
-from unittest.mock import patch
-
 import pytest
 
-from copy import deepcopy
+from unittest.mock import patch
 
 from bs4 import BeautifulSoup
 from django.urls import reverse
 
-from caseworker.advice.services import LICENSING_UNIT_TEAM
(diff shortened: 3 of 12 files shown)


caseworker/advice/templates/advice/advice_details.html
```diff
@@ -89,7 +89,7 @@ <h2 class="govuk-heading-m app-bg-colour--red govuk-!-padding-2">
                 <div class="countersignatures">
                     {% for countersign_advice in case|countersignatures_for_advice:advice  %}
                         {%  if countersign_advice.0.outcome_accepted %}
-                        <div class="govuk-!-padding-4 govuk-!-margin-bottom-2 countersigned-by">
+                        <div class="govuk-!-padding-4 govuk-!-margin-bottom-2 countersigned-by" style="border: 1px solid black;">
```
saruniitr, line 92: noticed during testing that this was missing

unit_tests/caseworker/conftest.py
```diff
@@ -891,55 +891,62 @@ def final_advice(current_user, lu_team):
     }
 
 
-def countersignatures_for_advice(all_advice, accepted=[True]):
-    def first_countersignature(advice):
+def countersignatures_for_advice(all_advice, data):
```
saruniitr, line 894: same method as before, just made it more clear on what is happening and easy to parameterize in various tests

caseworker/advice/templates/advice/advice_details.html
antroy-madetech: (prob need to rebuild your assets)

caseworker/advice/views.py
```diff
@@ -121,7 +121,7 @@ def rejected_countersign_advice(self):
         rejection (case will be returned to edit the advice once a rejection has occurred.
         """
         for cs in self.case.get("countersign_advice", []):
-            if not cs["outcome_accepted"]:
+            if cs["valid"] is True and cs["outcome_accepted"] is False:
```
antroy-madetech, line 124: `if cs["valid"] and not cs["outcome_accepted"]:` would be easier to read IMO.

hnryjmes (approved):
looks good. appreciate some of the cleaning up you have done here as well.

unit_tests/caseworker/advice/views/test_countersign_view.py
```diff
@@ -158,7 +168,11 @@ def test_lu_rejected_senior_countersignature(
...
     data_standard_case["case"]["countersign_advice"] = countersignatures_for_advice(
-        [final_advice], accepted=[True, False]
+        [final_advice],
+        [
+            {"order": FIRST_COUNTERSIGN, "outcome_accepted": True},
+            {"order": SECOND_COUNTERSIGN, "outcome_accepted": False},
+        ],
     )
     requests_mock.get(client._build_absolute_uri(f"/cases/{case_id}"), json=data_standard_case)
     mock_get_gov_user.return_value = (
```
antroy-madetech, line 178: We should also test this with some `"valid": False` countersignatures in there as well to prove out the negative logic. Looks like you already catered for this in the rework of the countersignatures_for_advice function.

caseworker/advice/templates/advice/advice_details.html
saruniitr: yep, that was it. Rebuilt assets and it is fine without this change, reverting it.

unit_tests/caseworker/advice/views/test_countersign_view.py
saruniitr: In case of `valid=False` we cannot really assert on anything because there won't be any warning.
So it will be just a check to see there is no warning, is that ok?

antroy-madetech (approved):
LGTM

PR #1197, opened 2023-03-27
Added additional text to all Spire links

Tllew:
Added hidden div to all instances of the SPIRE URL, 
Needed to convert some p tags to span to ensure styling stayed the same

This was for accessibility

can be checked here:
http://localhost:8300/licences
http://localhost:8300/

[LTD-2081](https://uktrade.atlassian.net/browse/LTD-2081)

[LTD-2081]: https://uktrade.atlassian.net/browse/LTD-2081?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


exporter/core/templates/goods/forms/common/help_with_security_gradings.html
```diff
@@ -20,7 +20,11 @@
...
 
-<p class="govuk-body">If the product was developed without UK or overseas government support, you should apply for a
+<span class="govuk-body">If the product was developed without UK or overseas government support, you should apply for a
     private venture grading using <a class="govuk-link" target="_blank"
-        href="https://www.spire.trade.gov.uk/">SPIRE</a>. The grading will be provided by MOD.
-</p>
+        href="https://www.spire.trade.gov.uk/">SPIRE</a>
+        <div class="govuk-visually-hidden">
+            &nbsp;(the online export licensing system)
+        </div>
```
kevincarrogan, line 28: What was the reason that this isn't a `span`? I feel like having a `div` inside a `span` is against the HTML spec.

kevincarrogan: I think the ticket also recommended putting this visually hidden information inside the anchor tag. Is there a reason why this has changed?

Tllew: This was originally a <div>, the others were originally <p> tags, I think I missed the fact the <a> ended immediately after the SPIRE text, I'll fix that now, I'll have another look at if we can change some of the other values from spans to div while retaining the text font etc

Tllew:
Added additional text to all Spire links, 
also made some of the code blocks more consistent with the rest of their files (using divs instead of p/span tags)

This was for accessibility

can be checked here:
http://localhost:8300/licences
http://localhost:8300/

[LTD-2081](https://uktrade.atlassian.net/browse/LTD-2081)

currycoder (approved):

PR #1202, opened 2023-03-29
Added govuk-country-and-territory-autocomplete to match functionality

Tllew:
Searching a country was returning as thought it was doing a "in" function in python rather than a "startswith".

[this](https://github.com/alphagov/govuk-country-and-territory-autocomplete) was used to replicate the "startswith" functionality and allow for regional/territorial querying.

To test:
go to the exporter,
make a new application,
add end user information,
URL should look like this:
* http://localhost:8300/applications/<APPLICATION_ID>/end-user/set/

[LTD-3206](https://uktrade.atlassian.net/browse/LTD-3206)

[LTD-3206]: https://uktrade.atlassian.net/browse/LTD-3206?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


core/forms/templates/forms/autocomplete_country.html
```diff
@@ -0,0 +1,31 @@
+{% load static %}
+
+<select name="{{ widget.name }}"{% include "django/forms/widgets/attrs.html" %}>{% for group_name, group_choices, group_index in widget.optgroups %}{% if group_name %}
+  <optgroup label="{{ group_name }}">{% endif %}{% for option in group_choices %}
+  {% include option.template_name with widget=option %}{% endfor %}{% if group_name %}
+  </optgroup>{% endif %}{% endfor %}
+</select>
+
+<script type="text/javascript" src="/assets/javascripts/location-autocomplete.min.js" nonce="{{ widget.attrs.nonce }}"></script>
```
kevincarrogan, line 9: I don't think this needs a nonce. This also seems to not use the `static` tag.

core/forms/templates/forms/autocomplete_country.html
```diff
@@ -0,0 +1,31 @@
...
+<script type="text/javascript" nonce="{{ widget.attrs.nonce }}">
+	$(document).ready(function() {
```
kevincarrogan, line 11: We are trying to move away from using jquery. Can this be done in a different way? Could this also maybe be a script instead of inline?

core/forms/templates/forms/autocomplete_country.html
```diff
@@ -0,0 +1,31 @@
...
+		var selectElement = document.getElementById('{{ widget.attrs.id }}');
+
+        openregisterLocationPicker({
+            url: '/assets/location-autocomplete-graph.json',
+            defaultValue: '{{ widget.value.0 }}',
+			displayMenu: 'overlay',
+			selectElement: selectElement,
+			cssNamespace: 'lite-autocomplete',
+			onConfirm: (query => {
+				const requestedOption = [].filter.call(selectElement.options, option => (option.textContent || option.innerText) === query)[0]
```
kevincarrogan, line 21: What is this doing?

core/assets/location-autocomplete-graph.json
kevincarrogan, line 1: How was this file generated? What happens if we need to regenerate it in future? Could this possibly actually be dynamic and be based on the country list we have in LITE?

core/forms/templates/forms/autocomplete_country.html
```diff
@@ -0,0 +1,31 @@
...
+				if (requestedOption) {
+					requestedOption.selected = true
+					var changeEvent = new CustomEvent('autocomplete {{ widget.name }} change', { detail: requestedOption.value });
+					document.dispatchEvent(changeEvent);
```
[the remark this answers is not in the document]
Tllew: it's a direct copy of what exists in the autocomplete, happy to do some testing and remove but it will take some time to investigate as the original code isn't clear why it's sending a dispatch for this event.

core/forms/templates/forms/autocomplete_country.html
Tllew: This is a direct copy of what is currently in the autopicker

core/assets/location-autocomplete-graph.json
Tllew: it comes from the country package in node_modules and is used as a required argument during initialisation:
`url: '/assets/location-autocomplete-graph.json'`

I'll take a look and see if we can specify our own json file.

PR #1209, opened 2023-03-30
Ltd 3514 prefix removal

kevincarrogan:
### Aim

Fix the ability to remove an ARS entry. This works for both prefix and suffix.

### DIFF
--- a/caseworker/assets/javascripts/tau/ars.js
+++ b/caseworker/assets/javascripts/tau/ars.js
@@ -2,7 +2,11 @@ import "fetch-polyfill";
 import accessibleAutocomplete from "accessible-autocomplete";
 import debounce from "lodash.debounce";
 
-const initAutocompleteField = (summaryFieldType, summaryFieldPluralised) => {
+const initAutocompleteField = (
+  summaryFieldType,
+  summaryFieldPluralised,
+  getDefaultValue
+) => {
   const originalInput = document.querySelector(
     `#report_summary_${summaryFieldType}`
   );
@@ -46,10 +50,7 @@ const initAutocompleteField = (summaryFieldType, summaryFieldPluralised) => {
       }
     },
     // Check the following is actually required:
-    defaultValue:
-      "subject" == summaryFieldType
-        ? originalInput.dataset.name
-        : originalInput.dataset.name || "",
+    defaultValue: getDefaultValue(originalInput),
     showNoOptionsFound: true,
     autoselect: true,
     confirmOnBlur: true,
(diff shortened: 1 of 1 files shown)


caseworker/assets/javascripts/tau/ars.js
```diff
@@ -39,26 +44,31 @@ const initAutocompleteField = (summaryFieldType, summaryFieldPluralised) => {
...
-  initAutocompleteField("subject", "subjects");
+  initAutocompleteField(
+    "prefix",
+    "prefixes",
+    (originalInput) => originalInput.dataset.name || ""
+  );
+  initAutocompleteField(
+    "subject",
+    "subjects",
+    (originalInput) => originalInput.dataset.name
```
currycoder, line 70: My read of this is that subjects cannot be empty string...  Is that the case?

caseworker/assets/javascripts/tau/ars.js
```diff
@@ -2,14 +2,19 @@ import "fetch-polyfill";
...
+  getDefaultValue
+) => {
   const originalInput = document.querySelector(
     `#report_summary_${summaryFieldType}`
   );
   const autocompleteContainer = document.createElement("div");
   autocompleteContainer.id = `report_summary_${summaryFieldType}_container`;
   originalInput.parentElement.appendChild(autocompleteContainer);
   originalInput.style = "display:none";
+  let nameInput;
```
currycoder, line 17: What is the purpose of this and `originalInput`?  In general I don't find this component to be particularly easy to read, so a little explainer on this would be helpful for my review.

kevincarrogan: Original input is what’s rendered on the page from Django. This is then progressively enhanced with this component which generates its own set of inputs. 

nameInput is me now checking the input that this component generates so I can look at the value to make sure that we can wipe out the hidden input that we use to submit the form. 

The hidden original input is what takes the id value to submit as we don’t want to use the name.

caseworker/assets/javascripts/tau/ars.js
kevincarrogan: This is when we have the edit page where we give the value to the component to display the initial value. Subject is mandatory so it will have a value.

kevincarrogan: This logic isn’t changing. I’ve just moved it.

currycoder (approved):

hnryjmes (approved):

PR #1213, opened 2023-03-31
LTD-3536: Show LU countersigning comments in other recommendations

saruniitr:
### Aim

Show countersignatures when advice details are viewed under "Other recommendations".
We only want to show approved comments because if it is rejected then we consider Countersigning is not completed and are not shown to caseworker.

Add test that checks for the presence of countersignatures of both orders.

[LTD-3536](https://uktrade.atlassian.net/browse/LTD-3536)

[LTD-3536]: https://uktrade.atlassian.net/browse/LTD-3536?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


caseworker/advice/templates/advice/group-advice.html
```diff
@@ -56,7 +57,20 @@ <h2 class="govuk-!-font-weight-bold govuk-body govuk-!-padding-2 app-advice-reco
                     </tbody>
                 </table>
                 {% with decision_advice.advice.0.advice as advice %}
-                {% if advice.countersigned_by %}
+                {% if is_lu_countersigning and case|countersignatures_for_advice:advice %}
```
antroy-madetech, line 60: I don't think the "and case|..." clause is necessary here, as if countersignatures_for_advice is empty the for loop will not output anything

antroy-madetech, line 60: I don't think the "and case|..." clause is necessary here, as if countersignatures_for_advice is empty the for loop will not output anything

antroy-madetech (approved):
Minor comment to clean up the template slightly. Otherwise LGTM

unit_tests/caseworker/advice/conftest.py
```diff
@@ -10,3 +11,167 @@ def with_lu_countersigning_enabled(settings):
...
+            "goods_type": None,
+            "country": None,
+            "end_user": None,
+            "ultimate_end_user": None,
+            "consignee": None,
+            "third_party": None,
+            "countersigned_by": None,
+            "countersign_comments": "",
+        },
+    ]
```
antroy-madetech, line 168: I wonder if all these cases can come from a base dict that is updated with the things that differ in each case. It's hard to see what is different between the cases this way.

antroy-madetech (approved):
Minor comment on the test fixture. Approved since you have just moved the fixture, but would be nice to address this if you have time.

hnryjmes (approved):

PR #1220, opened 2023-04-06
Re-enable and fix failing E2E tests

saruniitr:
### Aim

We temporarily disabled failing E2E tests to merge routing changes that activates Countersign rules.
They are all merged now so we can re-enable them.

The refuse test is still skipped because we need additional fix which we are doing it as a follow-up so it will be re-enabled later.

One of the exporter test is also skipped at the moment but we are adding fix in this PR (in progress) so will update the PR once that is ready.


kevincarrogan (approved):

antroy-madetech (approved):
LGTM

PR #1233, opened 2023-04-17
Ltd 3371 ncsc exporter application content changes

Tllew:
### Aim
change content when feature flag for ncsc is enabled

[LTD-3371](https://uktrade.atlassian.net/browse/LTD-3371)

[LTD-3371]: https://uktrade.atlassian.net/browse/LTD-3371?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


caseworker/tau/templates/tau/software_related_to_firearms_table.html
```diff
@@ -21,7 +21,11 @@
         </tr>
       {% endif %}
       <tr class="govuk-table__row">
+        {% if FEATURE_C7_NCSC_ENABLED %}
+        <td class="govuk-table__header">Does the product include cryptography or other information security features?</td>
+        {% else %}
```
saruniitr, line 26: Is it possible to do the same in forms similar to the below one to follow a consistent pattern?

exporter/core/templates/goods/forms/common/help_with_security_features.html
```diff
@@ -1,3 +1,9 @@
+{% if FEATURE_C7_NCSC_ENABLED %}
+<p class="govuk-body">
+    Information security features include cryptography, authentication, and cryptanalytic functions. They are often found in communication, wireless or internet-based products, digital forensics and surveillance tools.
+</p>
+{% else %}
```
saruniitr, line 5: same for this one, please see if this can be done in forms.
We haven't added feature flags in templates so far.

unit_tests/exporter/goods/views/test_component_product_details.py
```diff
@@ -77,7 +77,11 @@ def test_component_accessory_product_details_context(
         ("pv-grading-issuing-authority", "Government entity", "Name and address of the issuing authority"),
         ("pv-grading-details-reference", "GR123", "Reference"),
         ("pv-grading-details-date-of-issue", "20 February 2020", "Date of issue"),
-        ("uses-information-security", "No", "Does the product include security features to protect information?"),
+        (
+            "uses-information-security",
+            "No",
+            "Does the product include security features to protect information?",
+        ),
```
saruniitr, line 84: may be still in progress, there doesn't seem to be any tests with the flag enabled.

Tllew: Yeah I meant to open this as a draft

caseworker/tau/templates/tau/software_related_to_firearms_table.html
Tllew: resolved

exporter/core/templates/goods/forms/common/help_with_security_features.html
Tllew: This is the only one I wasn't sure about, to change this to not have a templated change would be to reduce help_with_security_features to just a p tag and move the rest to the form, which can be done but I'm not sure it's what we want

depsiatwal (approved):
I think this  looks good just need to fix lint.

exporter/core/templates/goods/forms/common/help_with_security_features.html
saruniitr: You are currently including `FEATURE_C7_NCSC_ENABLED` in `context`, instead you could add the `p` content in the `context` which can be switched based on the flag at https://github.com/uktrade/lite-frontend/pull/1233/files#diff-8689765ddff197b9e69a81ef74bafe8e599b418c3b1df17e9d2e1556db47290fR655

does that work?

saruniitr:
```
<p class="govuk-body">
    {{ content }}
</p>
```

exporter/goods/forms/common.py
```diff
@@ -664,7 +673,7 @@ def get_layout_fields(self):
             ),
             HTML.details(
                 "Help with security features",
-                render_to_string("goods/forms/common/help_with_security_features.html"),
+                render_to_string("goods/forms/common/help_with_security_features.html", context=ctx),
```
hnryjmes, line 676: I think you could replace this with `"<p class="govuk-body"> Information security features include cryptography, authentication, and cryptanalytic functions. They are often found in communication, wireless or internet-based products, digital forensics and surveillance tools. </p>"` and feature switch in this file instead

PR #1238, opened 2023-04-19
Fix paging for licences page

kevincarrogan:
### Aim

This fixes the paging on the licences page.

The issue was that the page parameter wasn't being passed through to the API correctly.

[LTD-3590](https://uktrade.atlassian.net/browse/LTD-3590)

[LTD-3590]: https://uktrade.atlassian.net/browse/LTD-3590?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/exporter/licences/views.py
+++ b/exporter/licences/views.py
@@ -82,7 +82,11 @@ class ListOpenAndStandardLicences(AbstractListView):
         filter_bar = filters.get_licences_filters(
             licence_type=self.object_name, control_list_entries=self.control_list, countries=self.countries
         )
-        licences = get_licences(self.request, licence_type="licence")
+        licences = get_licences(
+            self.request,
+            licence_type="licence",
+            **self.params,
+        )
 
         context = super().get_context_data(data=licences, filters=filter_bar, **kwargs)
         context.update(
--- a/unit_tests/exporter/licences/test_views.py
+++ b/unit_tests/exporter/licences/test_views.py
@@ -121,13 +121,13 @@ def data_list_licences():
 
 @pytest.fixture(autouse=True)
 def mock_list_licences(requests_mock, data_list_licences):
-    url = _build_absolute_uri("/licences/?page=1")
+    url = _build_absolute_uri("/licences/")
     return requests_mock.get(url=url, json=data_list_licences)
 
 
 @pytest.fixture(autouse=True)
 def mock_list_no_licence_required(data_list_no_licence_required, requests_mock):
-    url = _build_absolute_uri("/licences/nlrs/?page=1")
+    url = _build_absolute_uri("/licences/nlrs/")
     return requests_mock.get(url=url, json=data_list_no_licence_required)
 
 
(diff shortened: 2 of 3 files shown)


exporter/licences/views.py
```diff
@@ -82,7 +82,11 @@ def get_context_data(self, **kwargs):
         filter_bar = filters.get_licences_filters(
             licence_type=self.object_name, control_list_entries=self.control_list, countries=self.countries
         )
-        licences = get_licences(self.request, licence_type="licence")
+        licences = get_licences(
+            self.request,
+            licence_type="licence",
+            **self.params,
```
currycoder, line 88: I think if we called this page with a querystring `?foo=bar` we will end up 500'ing as the call through to `get_licences` will error with TypeError as the kwarg `foo` would be unexpected.  Probably unlikely for now, but I feel like this can be more resilient

hnryjmes (approved):

exporter/licences/views.py
kevincarrogan: Yep. I was just keeping it consistent with how these other views worked. It seems that the real bug here was that this view just never passed in the params and the others did so it felt like it was just an accidental omission that I wanted to resolve in the quickest way possible.

unit_tests/exporter/licences/test_views.py
```diff
@@ -144,40 +144,97 @@ def client(authorized_client, mock_control_list_entries, mock_countries, mock_pv
...
+    response = client.get(f"{list_open_standard_licences_url}?page=2")
+    assert response.status_code == 200
+    assert len(mock_list_licences.request_history) == 1
+    assert mock_list_licences.last_request.qs == {
+        "licence_type": ["licence"],
+        "page": ["2"],
+    }
+
+
+def test_open_general_licences(client, data_list_open_general_licences, mock_list_open_general_licences):
```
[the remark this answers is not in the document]
kevincarrogan: Yep. For now I'm just sticking to resolving this particular bug, I only added these tests just to make sure that it wasn't in any of these other pages and seemed like a quick thing to do. I do realise that some of these tests are redundant though as these pages aren't used.

saruniitr (approved):

exporter/licences/views.py
currycoder: While I understand the rationale, it's difficult for me to thumbs up when I don't know exactly what GET params we are now passing in and whether that will cause 500s (or if there's some filtering combination on the page that will).  If we don't want to spend too much time making it explicit/verbose all the way down, as a quick fix could it be possible to make the `get_licences` function accept `**kwargs` to avoid the possibility of the exception?

kevincarrogan: The tests already exercise all of the available filters and checking all of the current filters on the page doesn't result in an error.

PR #1241, opened 2023-04-20
Allocate me button

antroy-madetech:
### Aim
In order to speed up the flow when wanting to allocate yourself to a case, this PR adds a button to the bar next to "Allocate case" button to allocate yourself directly to the case without an intermediate page flow.

[LTD-3586](https://uktrade.atlassian.net/browse/LTD-3586)

[LTD-3586]: https://uktrade.atlassian.net/browse/LTD-3586?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/caseworker/queues/views/case_assignments.py
+++ b/caseworker/queues/views/case_assignments.py
@@ -1,6 +1,5 @@
 from http import HTTPStatus
 
-from django.views import View
 from django.views.generic import FormView
 from django.contrib.messages.views import SuccessMessageMixin
 from django.shortcuts import redirect
@@ -45,17 +44,19 @@ class CaseAssignmentAllocateRole(LoginRequiredMixin, FormView):
         return super().get_context_data(*args, **context, **kwargs)
 
 
-class CaseAssignmentAllocateToMe(LoginRequiredMixin, View):
-    def post(self, request, *args, **kwargs):
-        queue_id = request.POST.get("queue_id", None)
-        user_id = request.POST.get("user")
-        case_id = request.POST.get("case")
+class CaseAssignmentAllocateToMe(LoginRequiredMixin, FormView):
+    template_name = "core/form.html"
+    form_class = forms.CaseAssignmentsAllocateToMeForm
 
-        put_queue_case_assignments(request, queue_id, [case_id], [user_id], "Allocated self to the case")
-        messages.success(request, f"You have been successfully added as case adviser")
+    def form_valid(self, form):
+        data = form.cleaned_data
+        put_queue_case_assignments(
+            self.request, data["queue_id"], [data["case_id"]], [data["user_id"]], "Allocated self to the case"
+        )
+        messages.success(self.request, f"You have been successfully added as case adviser")
+        self.success_url = data["return_to"]
 
-        default_success_url = reverse("queues:cases", kwargs={"queue_pk": queue_id})
-        return redirect(request.POST.get("return_to", default_success_url))
+        return super().form_valid(form)
 
 
 class CaseAssignmentsCaseOfficer(LoginRequiredMixin, SuccessMessageMixin, FormView):
--- a/caseworker/queues/views/case_assignments.py
+++ b/caseworker/queues/views/case_assignments.py
@@ -52,7 +52,7 @@ class CaseAssignmentAllocateToMe(LoginRequiredMixin, FormView):
         data = form.cleaned_data
         self.allocate_to_me(data)
         messages.success(self.request, f"You have been successfully added as case adviser")
-        self.success_url = self.request.headers.get("Referer", data["return_to"])
+        self.success_url = data["return_to"]
 
         return super().form_valid(form)
 
--- a/caseworker/queues/views/case_assignments.py
+++ b/caseworker/queues/views/case_assignments.py
@@ -50,13 +50,21 @@ class CaseAssignmentAllocateToMe(LoginRequiredMixin, FormView):
 
     def form_valid(self, form):
         data = form.cleaned_data
-        put_queue_case_assignments(
-            self.request, data["queue_id"], [data["case_id"]], [data["user_id"]], "Allocated self to the case"
-        )
+        self.allocate_to_me(data)
         messages.success(self.request, f"You have been successfully added as case adviser")
         self.success_url = self.request.headers.get("Referer", data["return_to"])
 
         return super().form_valid(form)
+
+    @expect_status(
+        HTTPStatus.OK,
+        "Error allocating user as case advisor",
+        "Unexpected error allocating you as case advisor",
+    )
+    def allocate_to_me(self, data):
+        return put_queue_case_assignments(
+            self.request, data["queue_id"], [data["case_id"]], [data["user_id"]], "Allocated self to the case"
+        )
 
 
 class CaseAssignmentsCaseOfficer(LoginRequiredMixin, SuccessMessageMixin, FormView):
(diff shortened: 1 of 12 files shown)


caseworker/queues/views/case_assignments.py
```diff
@@ -44,6 +46,33 @@ def get_context_data(self, *args, **kwargs):
         return super().get_context_data(*args, **context, **kwargs)
 
 
+class CaseAssignmentAllocateTo(LoginRequiredMixin, View):
```
kevincarrogan, line 49: You could think of this as a FormView instead and you might be able to get rid of some of the code.

I kind of also feel this view should _just_ be doing the "allocate me" thing instead of also being a jumping off point to the other view.

caseworker/templates/layouts/case.html
```diff
@@ -118,10 +118,25 @@ <h1 id="heading-reference-code" class="app-case-header__heading">
...
+                <form action="{% url 'queues:case_assignment_assign_to_me' queue.id %}" method="post" style="display: inline;">
+                    <span class="app-case-warning-banner__icon" aria-hidden="true">!</span>
+                    <span class="app-case-warning-banner__text">
+                        You need to allocate yourself or someone else to this case to work on it
+                        <button type="submit" name="action" value="allocate" class="app-case-warning-banner__action" id="assign_to_me_button"
+                                style="border: 0px; display: inline"
+                        >Allocate case</button>
+                        <input type="hidden" name="queue_id" value="{{ queue.id }}" />
+                        <input type="hidden" name="return_to" value="{{ CURRENT_PATH  }}" />
+                        <input type="hidden" name="case" value="{{ case.id }}" />
```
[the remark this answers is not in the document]
antroy-madetech: Changed back and form now crispy

caseworker/queues/views/case_assignments.py
antroy-madetech: FormView implemented

caseworker/tau/views.py
```diff
@@ -26,7 +26,8 @@
 from .services import get_first_precedents
 from .summaries import get_good_on_application_tau_summary
 from .utils import get_cle_suggestions_json
-
+from ..cases.helpers.case import CaseworkerMixin
```
currycoder, line 29: I realise this is mainly for consistency with above, but normally we do absolute imports rather than relative ones.

caseworker/queues/views/case_assignments.py
```diff
@@ -44,6 +44,21 @@ def get_context_data(self, *args, **kwargs):
...
+    template_name = "core/form.html"
+    form_class = forms.CaseAssignmentsAllocateToMeForm
+
+    def form_valid(self, form):
+        data = form.cleaned_data
+        put_queue_case_assignments(
+            self.request, data["queue_id"], [data["case_id"]], [data["user_id"]], "Allocated self to the case"
+        )
+        messages.success(self.request, f"You have been successfully added as case adviser")
+        self.success_url = self.request.headers.get("Referer", data["return_to"])
```
currycoder, line 57: Would the `return_to` not suffice here?

caseworker/queues/views/case_assignments.py
```diff
@@ -44,6 +44,21 @@ def get_context_data(self, *args, **kwargs):
         return super().get_context_data(*args, **context, **kwargs)
 
 
+class CaseAssignmentAllocateToMe(LoginRequiredMixin, FormView):
+    template_name = "core/form.html"
+    form_class = forms.CaseAssignmentsAllocateToMeForm
+
+    def form_valid(self, form):
+        data = form.cleaned_data
+        put_queue_case_assignments(
```
currycoder, line 53: I believe that this view will *always* return a success message, even if the API call resulted in a 500.  Below in `CaseAssignmentsCaseAssignee` we have handled this with a `expect_status` method decorator on a helper function.  I believe that will produce a warning message for a user if the call to the backend did not result in a 200.  It might be worth doing that here and adding a test which covers the API error case.

unit_tests/caseworker/cases/views/test_case_details.py
```diff
@@ -99,3 +100,43 @@ def test_case_details_with_no_destinations(
...
+    is_system_queue,
+    is_user_assigned,
+    authorized_client,
+    data_queue,
+    data_standard_case,
+    mock_gov_user,
+    assign_user_to_case,
+):
+    data_queue["is_system_queue"] = is_system_queue
+    if is_user_assigned:
```
currycoder, line 124: To improve readability, could this test be broken in to two?  `test_case_assign_me_button_user_already_assigned` `test_case_assign_me_button_user_not_already_assigned`  This will reduce the conditional logic within the testcases themselves which could be good.

currycoder (commented):
It's looking good!  A few minor comments which it would be good to remedy.

caseworker/queues/views/case_assignments.py
antroy-madetech: I believe I did this because the return_to is set by the view when the user visits the case page. However a few of the tabs just reveal a different view and change the URL without reloading the page. e.g. Documents page. In these cases return_to is the queues/xyz/case/fgh whereas the page the user is looking at is actually queues/xyz/case/fgh/documents

Using referer returns the user to the correct tab once they are allocated. It just falls back to return_to if referer isn't available for some reason (privacy settings etc).

currycoder: I do prefer the latter suggestion here ^
> or alternatively when the JS is updating the URL bar and flipping the content you could update the return_to field value instead.

For better or worse return_to is something that we ourselves manage/effect and is already sanitised.  It's our established pattern.  At least that's my take on this.

caseworker/templates/layouts/case.html
```diff
@@ -120,7 +121,21 @@ <h1 id="heading-reference-code" class="app-case-header__heading">
 			<div id="allocation-warning" class="app-case-warning-banner">
 				<span class="app-case-warning-banner__icon" aria-hidden="true">!</span>
 				<span class="app-case-warning-banner__text">
-					You need to allocate yourself or someone else to this case to work on it <a class="app-case-warning-banner__action" href="{% url 'queues:case_assignment_select_role' queue.id %}?&cases={{ case.id }}&return_to={{ CURRENT_PATH }}">Allocate case</a>
+					You need to allocate yourself or someone else to this case to work on it
+                    <a id="allocate-case-link" class="app-case-warning-banner__action"
+                       href="{% url 'queues:case_assignment_select_role' queue.id %}?&cases={{ case.id }}&return_to={{ CURRENT_PATH }}"
+                    >Allocate case</a>
+                    {% if not queue.is_system_queue %}
```
currycoder, line 128: Minor; you could just do `{% if allocate_to_me_form %}` instead of duplicating the system queue logic between template and view.

caseworker/queues/views/case_assignments.py
```diff
@@ -44,6 +44,29 @@ def get_context_data(self, *args, **kwargs):
...
+        self.allocate_to_me(data)
+        messages.success(self.request, f"You have been successfully added as case adviser")
+        self.success_url = self.request.headers.get("Referer", data["return_to"])
+
+        return super().form_valid(form)
+
+    @expect_status(
+        HTTPStatus.OK,
+        "Error allocating user as case advisor",
+        "Unexpected error allocating you as case advisor",
```
currycoder, line 62: I think we try to use spelling "adviser" everywhere that we can

currycoder (approved):
Just some minor comments which I will leave with you

kevincarrogan (approved):

PR #1244, opened 2023-04-21
[LTD-3594] [Spike] Add basic queue view column customisation options

currycoder:
### Aim
WIP

[LTD-3594](https://uktrade.atlassian.net/browse/LTD-3594)

[LTD-3594]: https://uktrade.atlassian.net/browse/LTD-3594?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/caseworker/assets/javascripts/customiser.js
+++ b/caseworker/assets/javascripts/customiser.js
@@ -1,11 +1,11 @@
-class TableCustomiser {
+class Customiser {
   constructor($el) {
     this.$el = $el;
     let $headers = this.$el.querySelectorAll("th");
     let $rows = this.$el.querySelectorAll("tr");
     this.customisableColumns = {};
     $headers.forEach(($header, columnIndex, obj) => {
-      if ($header.classList.contains("table-customiser__static")) {
+      if ($header.classList.contains("customiser__static")) {
         return;
       }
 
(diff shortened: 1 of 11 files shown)


caseworker/assets/javascripts/customiser.js
```diff
@@ -0,0 +1,128 @@
...
+    this.showHideElems();
+    this.storePreferences();
+  }
+
+  storePreferences() {
+    let preferences = {};
+    for (const [key, value] of Object.entries(this.toggleableElems)) {
+      preferences[key] = value.visible;
+    }
+    window.localStorage.setItem(
```
kevincarrogan, line 101: Instead of grabbing local storage from window here maybe we could inject this into this class at instantiation time. 

This will make testing easier (you can just pass in a mock instead of mocking the one on window). But it also gives us the ability to swap this out in future more easily by passing in a local storage like thing.

caseworker/templates/includes/case-row.html
```diff
@@ -57,42 +58,19 @@
...
-							<div class="app-updates__activity">
-								{{ update.additional_text|linebreaksbr }}
-							</div>
-						{% endif %}
-					</li>
-				{% endfor %}
-			{% endif %}
-		</ul>
+	{# destinations #}
+	<td class="govuk-table__cell lite-mobile-hide customiser__additional__destinations">
```
kevincarrogan, line 62: I would probably use more data elements here instead of using the class name to deal with so wouldn’t have `customer__additional__destinations`

PR #1254, opened 2023-05-03
Remove LU post circ feature switch in lite frontend

hnryjmes:
- Code which referenced `FEATURE_LU_POST_CIRC_COUNTERSIGNING` has been removed
- References to the fixture `with_lu_countersigning_enabled` have been removed from test arguments
- Tests which used `with_lu_countersigning_disabled` have been removed entirely as they are no longer needed


MarkThomasLite (approved):

caseworker/advice/views.py
```diff
@@ -84,10 +83,7 @@ def unadvised_countries(self):
         }
 
     def get_rejected_lu_countersignature(self):
```
saruniitr, line 85: Do we still need this method now?

Could we use `self.rejected_countersign_advice()` in place of this method?

caseworker/advice/views.py
```diff
@@ -485,9 +477,6 @@ def get_success_url(self):
 
 class EditCountersignDecisionAdviceView(ReviewCountersignDecisionAdviceView):
     def dispatch(self, request, *args, **kwargs):
-        if not settings.FEATURE_LU_POST_CIRC_COUNTERSIGNING:
-            raise Http404("LU Countersigning feature not enabled")
-
```
saruniitr, line 490: Please remove `dispatch()` method completely

caseworker/advice/views.py
```diff
@@ -415,7 +411,7 @@ def get_form_kwargs(self):
         kwargs = super().get_form_kwargs()
 
         is_in_lu_team = self.caseworker["team"]["alias"] == services.LICENSING_UNIT_TEAM
-        is_lu_countersigning = ((is_in_lu_team and settings.FEATURE_LU_POST_CIRC_COUNTERSIGNING),)
+        is_lu_countersigning = (is_in_lu_team,)
         if is_in_lu_team:
             rejected_lu_countersignature = self.get_rejected_lu_countersignature()
             if rejected_lu_countersignature and is_lu_countersigning:
```
antroy-madetech, line 417: I think `is_lu_countersigning` will always be equal to `is_in_lu_team` now, so if we are in this block `is_lu_countersigning` will be True, and so the `and` condition can go?

caseworker/advice/views.py
```diff
@@ -677,44 +666,38 @@ class ViewConsolidatedAdviceView(AdviceView, FormView):
     form_class = forms.MoveCaseForwardForm
 
     def get_lu_finalise_case(self, lu_countersign_required, rejected_lu_countersignature):
-        if settings.FEATURE_LU_POST_CIRC_COUNTERSIGNING:
-            finalise_case = (not lu_countersign_required) and (not rejected_lu_countersignature)
-        else:
-            finalise_case = not lu_countersign_required
+        finalise_case = (not lu_countersign_required) and (not rejected_lu_countersignature)
         return finalise_case
```
antroy-madetech, line 670: This method can probably be inlined - it is only used once, and just to set a variable.

antroy-madetech (commented):
Couple of small ones

caseworker/advice/views.py
hnryjmes: updated

caseworker/advice/views.py
hnryjmes: updated

hnryjmes:
> For some reason not able to add comment inline so adding it here.
> 
> At line https://github.com/uktrade/lite-frontend/pull/1254/files#diff-a84e0d397fc7d7cb3b6f58747a8354a7df56357dcd706b4f425eebf5f27a744aR669, I think the method `get_lu_finalise_case()` can be removed as well
> 
> Where it is used, perhaps can be simplified as below?
> 
> `finalise_case = not (lu_countersign_required and rejected_lu_countersignature)`

I have updated this and simplified it in the most recent set of commits

I had to dig out the De Morgan's laws wikipedia page but I believe `not (X or Y)` is equivalent to `(not X) and (not Y)` so have updated accordingly :)

antroy-madetech (approved):

saruniitr (approved):

PR #1255, opened 2023-05-03
[LTD-3626] Add a customisable queue view

currycoder:
### Aim

This change achieves the following;
- Adds the following columns to the queue view; products, CLEs, report summaries, regimes, total value, queries, denial matches, intended end use
- Adds a "customiser" javascript component which is responsible for making certain HTML elements toggleable and persisting user customisation preferences via local storage.
- Applies the customiser to the queue view so that visible columns are customisable by caseworkers.

This depends on a companion API PR: https://github.com/uktrade/lite-api/pull/1429

**Missing additions**
- ~~The E2E tests need updating to ensure that the customiser works as this is probably classed as "critical functionality" given that it governs the queue view.  Subtask; https://uktrade.atlassian.net/browse/LTD-3649~~
- ~~The customiser JS should be extended in some way to fire GA events so that we can measure caseworker engagement with this. Subtask; https://uktrade.atlassian.net/browse/LTD-3650~~
- Both analytics and E2E tests are implemented now ^

A gif of the customiser in action can be seen here; [LTD-3626](https://uktrade.atlassian.net/browse/LTD-3626)

[LTD-3626]: https://uktrade.atlassian.net/browse/LTD-3626?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


caseworker/queues/views/cases.py
```diff
@@ -180,10 +183,30 @@ def _transform_queue_assignments(self, case):
...
+            "total_value": Decimal(),
+        }
+        for good in case["goods"]:
+            goods_summary["cles"].update(good["cles"])
+            goods_summary["regimes"].update(good["regimes"])
+            if good["report_summary_subject"]:
+                report_summary = good["report_summary_subject"]
+                if good["report_summary_prefix"]:
+                    report_summary = f"{good['report_summary_prefix']} {report_summary}"
+                goods_summary["report_summaries"].add(report_summary)
```
saruniitr, line 200: I think we already expose `report_summary` in the serializer which is a combination of prefix/subject (this should be there in `good` already).
Is this added because that is not available?

caseworker/templates/queues/cases.html
```diff
@@ -61,41 +61,55 @@ <h1 class="govuk-heading-l">
 
 	{% include 'filters.html' %}
 
-	{% include 'queues/search-tabs.html' %}
+    <div class="customiser" data-customiser-spec='{"options_label": "Customise table columns", "identifier": "queue-view", "analytics_prefix": "qv", "toggleable_elements": [{"label": "Destinations", "key": "destinations", "default_visible": true}, {"label": "Case allocation", "key": "case_allocation", "default_visible": true}, {"label": "Case updates", "key": "case_updates", "default_visible": true}, {"label": "Products", "key": "products"}, {"label": "Control list entries", "key": "control_list_entries"}, {"label": "Report summaries", "key": "report_summaries"}, {"label": "Regimes", "key": "regimes"}, {"label": "Total value", "key": "total_value"}, {"label": "Queries", "key": "queries"}, {"label": "Denial matches", "key": "denial_matches"}, {"label": "Intended end use", "key": "intended_end_use"}, {"label": "Flags", "key": "flags", "default_visible": true}]}'>
```
saruniitr, line 64: Is it better loading this from context?

saruniitr, line 64: Is it better loading this from context?

caseworker/assets/javascripts/customiser.js
```diff
@@ -0,0 +1,149 @@
...
+    this.countToggleableElements = 0;
+    this.spec.toggleable_elements.forEach((toggleable_details, index, obj) => {
+      const key = toggleable_details.key;
+      const label = toggleable_details.label;
+
+      let visible = false;
+      if (toggleable_details.hasOwnProperty("default_visible")) {
+        visible = toggleable_details.default_visible;
+      }
+      let $elems = $el.querySelectorAll("[data-customiser-key=" + key + "]");
```
antroy-madetech, line 17: Could use template literals (backticks) here perhaps.

antroy-madetech: (and in other similar places)

caseworker/assets/javascripts/customiser.js
```diff
@@ -0,0 +1,149 @@
...
+  }
+
+  loadPreferences() {
+    if (!window.localStorage.getItem(this.localStorageKey)) {
+      return;
+    }
+    const preferences = JSON.parse(
+      window.localStorage.getItem(this.localStorageKey)
+    );
+    for (const [key, value] of Object.entries(preferences)) {
```
antroy-madetech, line 135: Could use `for const key in preferences` here for readability

caseworker/assets/javascripts/customiser.js
```diff
@@ -0,0 +1,149 @@
...
+      if (key in this.toggleableElems) {
+        this.toggleableElems[key].visible = value;
```
antroy-madetech, line 137: But value would need to be preferences[key] here

antroy-madetech (approved):
Couple of unimportant things to address (or not as it suits)

caseworker/queues/views/cases.py
currycoder: Which serializer Arun?  I wasn't able to find it on the case search API so added it to the new `GoodOnApplicationSummarySerializer` over here; https://github.com/uktrade/lite-api/pull/1429/files#diff-abecc9cc1e7851e8f229e15d4841224ed730e6c68967861bafb339c066162048R173

caseworker/assets/javascripts/customiser.js
currycoder: I was wondering what the benefit might be and then I realised that template literals support variables without needing to do the `+` concatenation dance :facepalm:

caseworker/assets/javascripts/customiser.js
currycoder: I think I'll keep this as-is; I feel like we could probably argue readability either way.  Pls let me know if you strongly disagree

caseworker/templates/queues/cases.html
currycoder: Good question.  I feel like because this is purely a spec for the presentation of the data, it might be OK here rather than being another level removed and in the python view.  I am open to being convinced though!

caseworker/assets/javascripts/customiser.js
currycoder: Remedied this

saruniitr (approved):

saruniitr (approved):

PR #1259, opened 2023-05-11
Update CLE filter id in the unit test

saruniitr:
### Aim

Case filters now use django-form components so id is slightly different


hnryjmes (approved):
If the tests are passing then that would suggest this works

kevincarrogan (approved):

PR #1260, opened 2023-05-11
Ltd 3618 mention case notes

Tllew:
### Aim
Added core mention functionality to case notes, able to post a mention and select the user.

[LTD-3618](https://uktrade.atlassian.net/browse/LTD-3618)

[LTD-3618]: https://uktrade.atlassian.net/browse/LTD-3618?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


caseworker/activities/views.py
```diff
@@ -76,4 +84,38 @@ def get_context_data(self, **kwargs):
...
+        response, status_code = post_case_notes(request, self.case_id, post_data)
+
+        if status_code != 201:
+            return error_page(request, response.get("errors")["text"][0])
+
+        return redirect(
+            "cases:activities:notes-and-timeline",
+            pk=self.case_id,
+            queue_pk=self.queue_id,
+        )
```
kevincarrogan, line 121: This feels like this could be a `FormView` and this would just be the `success_url`.

unit_tests/caseworker/activities/test_views.py
```diff
@@ -38,18 +38,24 @@ def notes_and_timelines_url(data_queue, data_standard_case):
...
-    authorized_client,
-    notes_and_timelines_url,
-):
+def test_notes_and_timelines_view_flag_on_status_code(authorized_client, notes_and_timelines_url, requests_mock):
+    requests_mock.get(
+        client._build_absolute_uri(f"/gov-users/"),
+        json={
+            "results": [],
+        },
+    )
```
kevincarrogan, line 47: Do we need to write some tests where this _does_ return some results?

caseworker/activities/views.py
depsiatwal: LGTM but has this been addressed ?

depsiatwal (approved):
lgtm

caseworker/assets/javascripts/case-notes.js
```diff
@@ -79,15 +89,50 @@ class CaseNote {
...
     this.toggleSubmitButtonEnabled();
   }
 }
 
 const initCaseNotes = () => {
   document
-    .querySelectorAll("[data-module=case-note]")
-    .forEach(($el) => new CaseNote($el).init());
+    .querySelectorAll("#case_notes")
+    .forEach(($el) =>
```
kevincarrogan, line 114: Using `data-module` here means that this can be used in multiple different contexts. It looks like we're now making this specific to this one exact form.

The `forEach` is also confusing now as it implies that there could be more that one of these on a page but using an `id` suggests that's not the case and the call to `new` also implies that there's really only ever one of these on the page.

caseworker/users/services.py
```diff
@@ -8,33 +8,52 @@
 from caseworker.core.constants import SUPER_USER_ROLE_ID
 
 
-def get_gov_users(request, params=None, convert_to_options=False):
+def get_gov_users(request, params=None, convert_to_options=False, convert_to_choices=False):
```
kevincarrogan, line 11: Having two boolean parameters that are mutually exclusive is confusing. What's happens if someone sets `True` to both here?

caseworker/users/services.py
```diff
@@ -8,33 +8,52 @@
...
-                value = email
-                description = None
 
-            converted.append(Option(key=user.get("id"), value=value, description=description))
+def convert_users_to_choices(data):
+    choices = []
+    # Hide users without emails (eg system users)
+    users = [user for user in data if user["email"]]
+    for user in users:
+        display = user.get("email")
```
kevincarrogan, line 31: We know that the user will have an email so don't need to use `get`.

caseworker/users/services.py
```diff
@@ -8,33 +8,52 @@
...
+        if user.get("first_name"):
```
kevincarrogan, line 32: Why are there some users without a first_name? I think @seijihg did some work on this at some point so would be worth finding out what we're really skipping here.

unit_tests/caseworker/activities/test_forms.py
```diff
@@ -0,0 +1,201 @@
...
+    url = client._build_absolute_uri(f"/organisations/{organisation_pk}/sites/")
+    yield requests_mock.post(
+        url=url, json={"site": {"name": "Test site", "id": "00000000-0000-0000-0000-000000000001"}}
+    )
+
+
+@pytest.fixture()
+def mock_validate_site(requests_mock, organisation_pk):
+    url = client._build_absolute_uri(f"/organisations/{organisation_pk}/sites/")
+    yield requests_mock.post(url=url, json={})
```
kevincarrogan, line 28: Why are we mocking things to do with `site` in these tests? It doesn't seem obvious to me what should be calling these endpoints.

unit_tests/caseworker/activities/test_forms.py
```diff
@@ -0,0 +1,201 @@
...
+    ),
+)
+def test_notes_and_timeline(data, valid_status, mock_request, requests_mock, mock_gov_users):
+    url = "/gov-users/"
+    json = {
+        "results": mock_gov_users,
+    }
+    requests_mock.get(client._build_absolute_uri(url), json=json)
+    form = forms.NotesAndTimelineForm(data=data, request=mock_request)
+    assert form.is_valid() == valid_status
```
kevincarrogan, line 72: Could we also check what errors are expected when the form isn't valid?

unit_tests/caseworker/activities/test_views.py
```diff
@@ -133,45 +153,129 @@ def test_notes_and_timelines_searching_by_team(
...
+    response = authorized_client.get(notes_and_timelines_url)
+    users = []
+    for user in mock_gov_users:
+        value = (
+            f'{user["first_name"]} {user["last_name"]} ({user["team"]["name"]})'
+            if user["first_name"]
+            else user["email"]
+        )
+        users.append((user["id"], value))
```
kevincarrogan, line 184: I think I'd prefer to see something explicit here to make sure that we're checking all of the different paths that display the users details instead of relying on `mock_gov_users`.

kevincarrogan (approved):

PR #1276, opened 2023-05-18
[LTD-3676] Rationalise approval screen

currycoder:
### Aim
This change hides some supplemental fields for the "Recommend an approval" page by placing them within a GDS "details" component.  This adheres to the designs in the Jira here; https://uktrade.atlassian.net/browse/LTD-3676

[LTD-3676](https://uktrade.atlassian.net/browse/LTD-3676)

[LTD-3676]: https://uktrade.atlassian.net/browse/LTD-3676?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/caseworker/advice/forms.py
+++ b/caseworker/advice/forms.py
@@ -5,9 +5,9 @@ from django.forms.formsets import formset_factory
 from django.utils.html import format_html
 
 from crispy_forms_gds.helper import FormHelper
-from crispy_forms_gds.layout import Field, Layout, Submit, HTML
+from crispy_forms_gds.layout import Field, Layout, Submit
 
-from core.forms.layouts import ConditionalRadios, ConditionalRadiosQuestion
+from core.forms.layouts import ConditionalRadios, ConditionalRadiosQuestion, ExpandingFieldset
 from core.forms.utils import coerce_str_to_bool
 
 from caseworker.advice import services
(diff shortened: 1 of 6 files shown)


caseworker/advice/forms.py
```diff
@@ -263,19 +269,11 @@ def __init__(self, countries, *args, **kwargs):
...
-            "proviso",
-            "instructions_to_exporter",
-            "footnote_details",
-            HTML.details(
-                "What is a footnote?",
-                "Footnotes explain why products to a destination have been approved or refused. "
-                "They will be publicly available in reports and data tables.",
-            ),
-            Submit("submit", "Submit recommendation"),
+            parent_layout,
```
kevincarrogan, line 276: Does the nesting of layouts cause there to be extra elements generated on the page or does it just flatten all out in the end?

caseworker/advice/forms.py
```diff
@@ -87,26 +90,30 @@ class GiveApprovalAdviceForm(forms.Form):
...
         picklist_attrs={"target": "proviso", "type": "proviso", "name": "licence condition"},
         label="Add a licence condition (optional)",
-        help_text="Choose a licence condition from the template list",
+        help_link_text="Choose a licence condition from the template list",
+        min_rows=1,
         required=False,
     )
     instructions_to_exporter = forms.CharField(
-        widget=forms.Textarea(attrs={"rows": "10"}),
+        widget=forms.Textarea(attrs={"rows": "1"}),
```
kevincarrogan, line 105: Is this small and then we expand it or is it meant to remain small?

caseworker/advice/forms.py
currycoder: I believe that this is a way to *compose* layouts rather than nest them.  But I will double check before/after and get back to you

caseworker/advice/forms.py
currycoder: I think the user can choose to expand it using the normal capability that text areas have.  Our overall intention here is to just reduce the textarea to one line in height though - according to the designs.  Does this answer your question?

caseworker/advice/forms.py
currycoder: Yep, these are all composed together at the same level without actually being nested

depsiatwal (approved):
lgtm

kevincarrogan (approved):

PR #1278, opened 2023-05-18
LTD-3622-show-user-mentions-list-2

depsiatwal:
https://uktrade.atlassian.net/browse/LTD-3622

New screen to support user mentions.

For those evil eyed you will noticed this isn't Feature flagged. This is because of my new view of having smart feature flags where necessary (or i'm just being lazy). So my argument is no FF is needed because this is internal users only with no link you would have to guess the URL and even then the page would be empty because the creation is feature flagged.

### DIFF
--- a/caseworker/templates/users/mentions.html
+++ b/caseworker/templates/users/mentions.html
@@ -29,7 +29,7 @@
 			</thead>
 			<tbody class="govuk-table__body">
 				{% for user_case_note_mention in user_mentions.mentions %}
-					<tr class="govuk-table__row" id="row-{{ forloop.counter }}">
+					<tr class="govuk-table__row" id="mentions-row-{{ forloop.counter }}">
 						<td class="govuk-table__header" scope="row">
 								{{ user_case_note_mention.reference_code }}
 						</td>
--- a/caseworker/templates/users/mentions.html
+++ b/caseworker/templates/users/mentions.html
@@ -1,0 +1,53 @@
+{% extends 'layouts/base.html' %}
+
+{% block back_link %}
+{% endblock %}
+
+{% block title %}
+	Mentions
+{% endblock %}
+
+{% block body %}
+	<div class="lite-app-bar">
+		<div class="lite-app-bar__content">
+			<h1 class="govuk-heading-l">
+				Mentions
+			</h1>
+		</div>
+	</div>
+
+
+		<table class="govuk-table">
+			<thead class="govuk-table__head">
+				<tr class="govuk-table__row">
+					<th class="govuk-table__header" scope="col">Case</th>
+					<th class="govuk-table__header" scope="col">Tags</th>
+					<th class="govuk-table__header" scope="col">Date</th>
+					<th class="govuk-table__header" scope="col">Case note mention</th>
+					<th class="govuk-table__header" scope="col"></th>
+				</tr>
+			</thead>
+			<tbody class="govuk-table__body">
+				{% for user_case_note_mention in user_mentions.mentions %}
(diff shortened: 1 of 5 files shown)


Tllew (approved):
looks good

caseworker/templates/users/mentions.html
```diff
@@ -0,0 +1,53 @@
...
+				{% for user_case_note_mention in user_mentions.mentions %}
+					<tr class="govuk-table__row" id="row-{{ forloop.counter }}">
+						<td class="govuk-table__header" scope="row">
+								{{ user_case_note_mention.reference_code }}
+						</td>
+						<td class="govuk-table__cell">{% if user_case_note_mention.is_urgent %} <p class="warning-text">URGENT</p> {% endif %}<br>{% if user_case_note_mention.is_accessed %} <p class="warning-text">NEW</p> {% endif %}</td>
+						<td class="govuk-table__cell">{{ user_case_note_mention.created_at|to_datetime|date:"d F Y" }}</td>
+						<td class="govuk-table__cell">
+							<div>
+								{% if user_case_note_mention.case_note_user.team %}<span class="app-activity__item__user">{{ user_case_note_mention.case_note_user.team.name }}:</span> {% endif %}{{ user_case_note_mention.case_note_user.first_name }} {{ user_case_note_mention.case_note_user.last_name }}
```
Tllew, line 40: this looks a little dense, is it worth throwing this in the view.py?

caseworker/users/views/users.py
```diff
@@ -139,3 +141,11 @@ def post(self, request, **kwargs):
         put_gov_user(request, str(kwargs["pk"]), json={"status": request.POST["status"]})
 
         return redirect("/users/")
+
+
+class UserCaseNoteMentions(LoginRequiredMixin, TemplateView):
+    def get(self, request, **kwargs):
+        data, _ = get_user_case_note_mentions(request)
```
currycoder, line 148: This doesn't seem to do any response code checking

unit_tests/caseworker/users/test_views.py
```diff
@@ -0,0 +1,22 @@
...
+    }
+    requests_mock.get(
+        client._build_absolute_uri("/cases/user-case-note-mentions/"),
+        json=mentions_data,
+    )
+
+    url = reverse("users:user_case_note_mentions")
+    response = authorized_client.get(url)
+    assert response.status_code == 200
+    assert response.context["user_mentions"] == mentions_data
```
currycoder, line 22: Do we want to also check the mentions appear through beautiful soup?  We have been doing this for the past few months for frontend tests

currycoder:
I also see in the designs that users appear to be able to send mentions to entire teams; is that covered here or will it be something for the future?

caseworker/users/views/users.py
depsiatwal: you mean have it as a cached property ?

depsiatwal: added as cached property

unit_tests/caseworker/users/test_views.py
depsiatwal: good point done

caseworker/users/views/users.py
currycoder: I don't see how `cached_property` helps.  We still aren't checking that the API response was 2XX.  If it's not we should raise an exception, could even just be something like https://github.com/uktrade/lite-frontend/blob/ab5775efd50a818bf397a1bb61c2156a9abf5298/caseworker/cases/services.py#L18

depsiatwal: ahh ok miss-understood changed

currycoder (approved):

PR #1282, opened 2023-05-22
Add delete and display bookmark functionality

antroy-madetech:
### Aim
This PR encompasses 
- Adding a "Save filter" button to the case filter panel which will save the filter with a default name "New unnamed filter"
- Adding a Saved filter expandable panel which will have the filters displayed to click on.
- By each saved filter, an Edit button which will overlay an input field to change the name of the filter and Save and Cancel buttons.
- By each saved filter, a delete button for deleting the filter.

[LTD-3665](https://uktrade.atlassian.net/browse/LTD-3665)

[LTD-3665]: https://uktrade.atlassian.net/browse/LTD-3665?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


caseworker/assets/javascripts/bookmarks.js
```diff
@@ -0,0 +1,14 @@
+import { enableToggle, enableSwap } from "./toggle";
+
+enableToggle("bookmarks", false, "show-bookmarks-link", "hide-bookmarks-link");
+
+var bookmarkNameElements = document.querySelectorAll('[id^="bookmark-"]');
+var nodes = Array.from(bookmarkNameElements);
```
kevincarrogan, line 6: You should use `let` or `const`.

caseworker/assets/javascripts/bookmarks.js
```diff
@@ -0,0 +1,14 @@
...
+
+nodes.forEach((bookmark) => {
```
kevincarrogan, line 8: I think you don't need to do the `Array.from` if you use `for ... of`.

caseworker/assets/javascripts/case-filters.js
```diff
@@ -70,52 +73,69 @@ export default function initFlagsFiltersField() {
   );
 }
 
+function filterIsPopulated(filterGroupName) {
+  let filterGroup = document.getElementById(filterGroupName);
+  let inputs = Array.from(filterGroup.getElementsByTagName("input"));
+  let selects = Array.from(filterGroup.getElementsByTagName("select"));
```
kevincarrogan, line 79: You might be able to use a different selector here to just get both at once e.g. `filterGroup.querySelectorAll("input, select")`.

Then you could use `for ... of` later too.

caseworker/assets/javascripts/case-filters.js
```diff
@@ -70,52 +73,69 @@ export default function initFlagsFiltersField() {
...
-            ($(this).attr("type") == "checkbox" && $(this).attr("checked")))
-        ) {
-          $filters.show();
-          $filters.prev().find("#show-filters-link").hide();
-          $filters.prev().find("#hide-filters-link").show();
-          $(this).parents(".govuk-details").attr("open", "");
-        }
-      });
-  });
+    let ifPopulated = filterIsPopulated("advanced-filter-fields");
```
kevincarrogan, line 109: `isPopulated` could be a better variable name here.

caseworker/assets/javascripts/case-filters.js
```diff
@@ -70,52 +73,69 @@ export default function initFlagsFiltersField() {
...
+}
+
 function showHideFilters() {
-  $(".case-filters").each(function () {
-    var $filters = $(this).parent();
-    $filters.hide();
-    $filters.prev().find("#show-filters-link").show();
-    $filters.prev().find("#hide-filters-link").hide();
+  let expand = filterIsPopulated("basic-filter-fields");
+  if (expand) {
```
kevincarrogan, line 101: Could we return early here by checking the reverse?

caseworker/assets/javascripts/toggle.js
```diff
@@ -0,0 +1,38 @@
...
+export function enableSwap(
+  visibleElementId,
+  hiddenElementId,
+  swapLinkId,
+  revertLinkId
+) {
+  let visibleElement = document.getElementById(visibleElementId);
+  let hiddenElement = document.getElementById(hiddenElementId);
+  let swapLink = document.getElementById(swapLinkId);
+  let revertLink = document.getElementById(revertLinkId);
```
kevincarrogan, line 18: You should use `const` here instead of `let`.

caseworker/assets/styles/_case-filters.scss
```diff
@@ -42,7 +42,7 @@
     }
 }
 
-.basic-filter-fields {
+#basic-filter-fields {
```
kevincarrogan, line 45: We should try not to style things using id selectors where possible as this ends up with a really high specificity level.

caseworker/bookmarks/templates/bookmarks/bookmarks.html
```diff
@@ -0,0 +1,42 @@
...
+                            {{ bookmark.filter_description }}
+                        </div>
+                    </div>
+                    <div class="govuk-grid-column-one-third">
+                        <input type="hidden" name="id" value="{{ bookmark.id }}"/>
+                        <input type="hidden" name="return_to" value="{{ return_to }}"/>
+                        <input type="button" id="bookmark-{{ bookmark.id }}-name-edit-cancel" value="Cancel" class="bookmark-submit" class="toggle-hidden"/>
+                        <input type="submit" id="bookmark-{{ bookmark.id }}-name-save" name="submit" value="Save" class="bookmark-submit"/>
+                        <input type="button" id="bookmark-{{ bookmark.id }}-name-edit-button" value="Edit" class="bookmark-submit"/>
+                        <input type="submit" name="submit" value="Delete" class="bookmark-submit"/>
```
kevincarrogan, line 35: Could this be a Django form instead?

kevincarrogan, line 35: We should use `button` where possible instead of `input`.

caseworker/bookmarks/views.py
```diff
@@ -0,0 +1,77 @@
...
+        return super().form_valid(form)
+
+    @expect_status(
+        HTTPStatus.CREATED,
+        "Error adding bookmark",
+        "Unexpected error adding bookmark",
+    )
+    def add_bookmark(self, user_id, return_to, name, description):
+        data = {"url": return_to, "name": name, "description": description, "user_id": user_id}
+        response = client.post(self.request, f"/bookmarks", data)
```
kevincarrogan, line 35: We usually wrap this in a "service" function call.

caseworker/bookmarks/templates/bookmarks/bookmarks.html
```diff
@@ -0,0 +1,42 @@
+{% load crispy_forms_tags crispy_forms_gds %}
+{% load svg static %}
+
+<div id="bookmarks" class="govuk-width-container toggle-hidden">
+    <p class="govuk-body govuk-!-font-weight-bold bookmark-headline">Select a filter you want to display within queue view</p>
+    <div>
+        {% for bookmark in bookmarks.user %}
```
kevincarrogan, line 7: This looks a lot like a formset to me https://docs.djangoproject.com/en/4.2/topics/forms/formsets/.

core/assets/javascripts/filter-bar.js
kevincarrogan, line 1: We should just delete this instead of commenting it out.

antroy-madetech: Was just keeping it around while I got the replacement working correctly.

caseworker/bookmarks/services.py
```diff
@@ -0,0 +1,67 @@
...
+from caseworker.users.services import get_gov_user
+from core import client
+
+
+class BookmarkException(Exception):
+    pass
+
+
+def fetch_bookmarks(request):
+    bookmarks = client.get(request, "/bookmarks/").json()
```
currycoder, line 17: Would suggest some response code checking here so that we can fail early

caseworker/bookmarks/views.py
```diff
@@ -0,0 +1,77 @@
...
+
+    @expect_status(
+        HTTPStatus.OK,
+        "Error updating bookmark",
+        "Unexpected error updating bookmark",
+    )
+    def edit_bookmark(self, bookmark_id, name):
+        data = {"id": bookmark_id, "name": name}
+        response = client.put(self.request, f"/bookmarks", data)
+        return response.json(), response.status_code
```
currycoder, line 77: Same comment as Kevin's above, might be good to do the requests for this in `service.py`

package.json
```diff
@@ -5,9 +5,9 @@
...
-    "build_caseworker": "parcel build ./caseworker/assets/javascripts/{main,head,cookie-policy-form,search-cases,search-products,beis,tau,tau-edit,case-filters}.js --dist-dir caseworker/assets/built --public-url /assets/",
+    "build_caseworker": "parcel build ./caseworker/assets/javascripts/{main,head,bookmarks,cookie-policy-form,search-cases,search-products,beis,tau,tau-edit,case-filters}.js --dist-dir caseworker/assets/built --public-url /assets/",
     "watch_exporter": "parcel watch --port 8400 ./exporter/assets/javascripts/{main,head}.js ./exporter/assets/javascripts/cookie-policy-form.js --dist-dir exporter/assets/built --public-url /assets/",
-    "watch_caseworker": "parcel watch --port 8401 ./caseworker/assets/javascripts/{main,head,cookie-policy-form,search-cases,search-products,beis,tau,tau-edit,case-filters}.js --dist-dir caseworker/assets/built --public-url /assets/",
+    "watch_caseworker": "parcel watch --port 8401 ./caseworker/assets/javascripts/{main,head,bookmarks,cookie-policy-form,search-cases,search-products,beis,tau,tau-edit,case-filters}.js --dist-dir caseworker/assets/built --public-url /assets/",
```
currycoder, line 10: Oh my, hadn't seen this before...

caseworker/bookmarks/templates/bookmarks/bookmarks.html
antroy-madetech: Moving this into another ticket as this one is dragging on as it is, and moving this into a form doesn't look trivial...

antroy-madetech: Moving this into another ticket as this one is dragging on as it is, and moving this into a form doesn't look trivial...

unit_tests/caseworker/conftest.py.orig
```diff
@@ -0,0 +1,2110 @@
+import copy
```
currycoder, line 1: I don't think this should be here

caseworker/queues/views/forms.py.orig
```diff
@@ -0,0 +1,257 @@
+from crispy_forms_gds.fields import DateInputField
```
currycoder, line 1: What's happened to cause all these ".orig" files to show up?

core/forms/templates/forms/expanding_fieldset.html.orig
```diff
@@ -0,0 +1,22 @@
+<fieldset {% if fieldset.css_id %}id="{{ fieldset.css_id }}"{% endif %}
```
currycoder, line 1: Another .orig

caseworker/queues/views/forms.py.orig
antroy-madetech: Ergh. Rebase and the mergetool creates them. Then I think the IDE has been auto-adding them all.

caseworker/bookmarks/views.py
```diff
@@ -0,0 +1,92 @@
...
+        "Unexpected error deleting filter",
+    )
+    def delete_bookmark(self, bookmark_id):
+        response = services.delete_bookmark(self.request, bookmark_id)
+        return response.json(), response.status_code
+
+
+class RenameBookmark(LoginRequiredMixin, FormView):
+    template_name = "core/form.html"
+    form_class = forms.DeleteBookmark
```
currycoder, line 71: Should this be the DeleteBookmark form?

caseworker/bookmarks/forms.py
```diff
@@ -0,0 +1,9 @@
+from django import forms
+from django.forms.widgets import HiddenInput
+
+
+class DeleteBookmark(forms.Form):
+    id = forms.CharField(widget=HiddenInput)
+    return_to = forms.CharField(widget=HiddenInput)
+    name = forms.CharField(required=False)
+    submit = forms.CharField()
```
currycoder, line 9: Does this form need these two fields?  I'm not really sure what `submit` would do..?

caseworker/assets/javascripts/toggle.js
```diff
@@ -0,0 +1,38 @@
+export function enableToggle(elementId, startVisible, showLinkId, hideLinkId) {
+  if (startVisible) {
+    enableSwap(elementId, null, showLinkId, hideLinkId);
+  } else {
+    enableSwap(null, elementId, showLinkId, hideLinkId);
+  }
+}
+
+export function enableSwap(
```
currycoder, line 9: It would be great to see some docstrings that explain what these functions do and how to use them for folks that haven't seen them before.

caseworker/bookmarks/services.py
```diff
@@ -0,0 +1,107 @@
...
+from django.utils.http import urlencode
+
+from caseworker.users.services import get_gov_user
+from core import client
+
+
+def fetch_bookmarks(request, filter_data):
+    response = client.get(request, "/bookmarks/")
+    if response.status_code >= 300:
+        # Not important enough to break the page, so return an empty set of bookmarks.
```
currycoder, line 15: Do we want to log the exception somehow despite not breaking the page?  I think we can call `logger.exception` to do that such that the exception still makes it to sentry without raising an error to the user; https://github.com/uktrade/lite-api/blob/ac6dde8a7892346147c42341b389110ca691ab70/api/external_data/management/commands/ingest_sanctions.py#L123

caseworker/queues/views/forms.py
```diff
@@ -100,21 +101,25 @@ def __init__(self, request, queue, filters_data, *args, **kwargs):
             choices=case_type_choices,
             label="Filter by type",
             widget=forms.Select(attrs={"id": "case_type"}),
+            required=False,
```
currycoder, line 104: How come these changes were necessary?

core/assets/javascripts/filter-bar.js
```diff
@@ -1,38 +1,38 @@
...
-$(".lite-filter-toggle-link")
-  .unbind()
-  .click(function () {
-    var $filters = $(this).parent().next();
-    $filters.toggle();
-    $(this).parent().find("> *").toggle();
-
-    return false;
-  });
+// function tryShowFilterBar() {
```
currycoder, line 1: Please can this file be totally removed?

unit_tests/caseworker/bookmarks/test_views.py
```diff
@@ -0,0 +1,142 @@
+import json
```
currycoder, line 1: I feel like we should add some E2E tests for this eventually..  Could this be created as a follow-up ticket?

caseworker/assets/javascripts/bookmarks.js
```diff
@@ -0,0 +1,29 @@
...
+      .addEventListener("keydown", (evt) => {
+        if (evt.key === "Escape") {
+          document.getElementById(`${id}-edit-cancel`).click();
+        }
+      });
+  }
+}
+
+document.getElementById("show-bookmarks-link").addEventListener("click", () => {
+  document.getElementById("bookmarks").scrollIntoView();
```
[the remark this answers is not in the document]
antroy-madetech: Not a requirement, and yes it is a bit jarring. However without it, if you have the filters open (e.g. you have clicked on a saved filter, so the filter panel opens up) and then click on Saved Filters, it appears that nothing happens. Because the Saved filters are below the filters themselves. I thought I'd put the scrolling in just to prompt UCD to have a think about the right way to do this. Happy to take it out, as the apparent non-action of the button will also work as a prompt to UCD

caseworker/assets/javascripts/case-filters.js
```diff
@@ -70,52 +73,72 @@ export default function initFlagsFiltersField() {
...
-          $filters.prev().find("#show-filters-link").hide();
-          $filters.prev().find("#hide-filters-link").show();
-          $(this).parents(".govuk-details").attr("open", "");
-        }
-      });
-  });
+  let expandBasic = filterIsPopulated("basic-filter-fields");
+  let expandAdvanced = filterIsPopulated("advanced-filter-fields");
+
+  console.log("Basic: " + expandBasic + "; Adv: " + expandAdvanced);
```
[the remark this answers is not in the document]
antroy-madetech: No.

caseworker/queues/views/forms.py
antroy-madetech: Hmm. Not sure. I'll test and find out.

unit_tests/caseworker/bookmarks/test_views.py
antroy-madetech: Will do.

caseworker/assets/javascripts/bookmarks.js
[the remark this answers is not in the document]
currycoder: Ok no worries, let's leave it for now as we are likely to rework the frontend presentation imminently anyway

caseworker/assets/javascripts/case-filters.js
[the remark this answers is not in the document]
currycoder: OK so could it be removed? :)

caseworker/queues/views/forms.py
currycoder: We discussed offline and it seems this is ok

currycoder (approved):

seijihg:
We could use some JS tests since we have been trying to add them whenever JS is involved but yea looking good.

seijihg (approved):

PR #1287, opened 2023-05-23
Ltd 3621 mention notification header

Tllew:
### Add mention notification in header

adds Mentions(1) when a user is mentioned, where 1 is the count of the mentions they're mentioned in that have not been accessed.

[LTD-3621](https://uktrade.atlassian.net/jira/software/c/projects/LTD/boards/204?modal=detail&selectedIssue=LTD-3621)

[LTD-3621]: https://uktrade.atlassian.net/browse/LTD-3621?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


caseworker/core/context_processors.py
```diff
@@ -52,6 +54,10 @@ def lite_menu(request):
         notifications = get_menu_notifications(request)
         notification_data = notifications["notifications"]
         has_notifications = notifications["has_notifications"]
+        # this can default to 0 and we can remove the "is not False" in the template when removing the feature flag
+        mentions = False
```
depsiatwal, line 58: why not pass in the feature flag into the template that way when we remove the FF all we need to do is search for the instances and remove IF statements. 

This minimised the impact on actual logic and code changes and hence regression.

caseworker/core/services.py
```diff
@@ -234,3 +234,10 @@ def get_menu_notifications(request):
         request.cached_get_menu_notifications = client.get(request, "/gov-users/notifications/")
     response = request.cached_get_menu_notifications
     return response.json()
+
+
+def get_mention_count(request):
+    if not hasattr(request, "cached_get_mention_count"):
+        request.cached_get_mention_count = client.get(request, "/cases/user-case-note-mentions/")
```
depsiatwal, line 241: cached ?
is this sticky what happens when they're rest ?

depsiatwal (approved):
lgtm

caseworker/assets/styles/components/_header.scss
```diff
@@ -8,7 +8,7 @@ $header-link-active-colour: $govuk-focus-colour;
 
 .app-header__container {
 	display: grid;
-	grid-template-columns: auto 1fr auto auto auto auto auto;
+	grid-template-columns: auto 1fr auto auto auto auto auto auto;
```
currycoder, line 11: Just curious, what does this achieve?

caseworker/core/context_processors.py
```diff
@@ -98,7 +102,12 @@ def lite_menu(request):
         ]
     else:
         pages = []
-    return {"LITE_MENU": [x for x in pages if x is not None], "MENU_NOTIFICATIONS": has_notifications}
+    return {
+        "LITE_MENU": [x for x in pages if x is not None],
+        "MENU_NOTIFICATIONS": has_notifications,
+        "MENTIONS_COUNT": mentions,
+        "FEATURE_MENTIONS_ENABLED": settings.FEATURE_MENTIONS_ENABLED,
```
currycoder, line 109: Would these be better off in a dedicated additional context processor function?

caseworker/core/services.py
currycoder, line 241: Please can there be some response code checking? e.g. fail loudly on non 2XX responses `response.raise_for_status()` would probably do the trick

caseworker/core/context_processors.py
```diff
@@ -47,11 +49,13 @@ def export_vars(request):
 
 def lite_menu(request):
     has_notifications = False
+    mentions = False
     if "lite_api_user_id" in request.session:
         permissions = get_user_permissions(request)
         notifications = get_menu_notifications(request)
         notification_data = notifications["notifications"]
         has_notifications = notifications["has_notifications"]
+        mentions = get_mention_count(request)
```
currycoder, line 58: Does this need feature flagging?  Right now the API call will be made regardless of frontend presentation

caseworker/templates/layouts/base.html
```diff
@@ -61,6 +61,18 @@
 				</span>
 			</a>
 			<span></span>
+			{% if FEATURE_MENTIONS_ENABLED %}
+			<a id="link-mentions" class="app-header__link" href="{% url 'users:user_case_note_mentions' %}" data-tooltip="Mentions">
+				{% if MENTIONS_COUNT > 0 %}
+					<b>
+						Mentions ({{ MENTIONS_COUNT }})
```
currycoder, line 68: Is this count new mentions only or a count of all mentions?

unit_tests/caseworker/cases/views/test_case_filters.py
```diff
@@ -80,7 +80,7 @@ def test_case_filters(authorized_client, requests_mock, mock_cases, mock_cases_h
     assert isinstance(response.context["form"], CasesFiltersForm)
 
     history = requests_mock.request_history
-    case_search_request = history[-4]
+    case_search_request = history[-5]
```
currycoder, line 83: Is there a better way?  This seems a bit naff

unit_tests/caseworker/users/test_views.py
```diff
@@ -1,15 +1,80 @@
...
+                        "case_id": "4966212d-5b52-4a6d-9e06-e589ab9dc221",  # /PS-IGNORE
+                        "is_accessed": False,
+                    },
+                ]
+            },
+            3,
+        ),
+    ),
+)
+def test_user_case_note_mention_count(data, count, authorized_client, requests_mock):
```
currycoder, line 57: Can we have another testcase where the mentions API call returns 5XX?

caseworker/core/context_processors.py
Tllew: It's required on all pages (as it is in the header)

caseworker/assets/styles/components/_header.scss
Tllew: allows the mentions section to fit in nicely in the header (due to the grid)

caseworker/core/context_processors.py
Tllew: it will be made regardless once the FF is removed, it's just not rendered now

caseworker/templates/layouts/base.html
Tllew: only new

Tllew: only new, the spec says it should only show new mentions, otherwise it should be the word Mentions as a linkl

Tllew: renamed variable to new_mentions to make things clearer

caseworker/core/context_processors.py
```diff
@@ -98,7 +100,22 @@ def lite_menu(request):
...
-    return {"LITE_MENU": [x for x in pages if x is not None], "MENU_NOTIFICATIONS": has_notifications}
+    return {
+        "LITE_MENU": [x for x in pages if x is not None],
+        "MENU_NOTIFICATIONS": has_notifications,
+    }
+
+
+def new_mentions(request):
+    new_mentions = False
+    if "lite_api_user_id" in request.session:
```
kevincarrogan, line 111: What's this for?

caseworker/core/context_processors.py
```diff
@@ -98,7 +100,22 @@ def lite_menu(request):
...
         pages = []
-    return {"LITE_MENU": [x for x in pages if x is not None], "MENU_NOTIFICATIONS": has_notifications}
+    return {
+        "LITE_MENU": [x for x in pages if x is not None],
+        "MENU_NOTIFICATIONS": has_notifications,
+    }
+
+
+def new_mentions(request):
+    new_mentions = False
```
kevincarrogan, line 110: This variable starts life as a Boolean and then becomes an integer, we should be consistent with the type here.

caseworker/core/context_processors.py
```diff
@@ -98,7 +100,22 @@ def lite_menu(request):
...
+    if "lite_api_user_id" in request.session:
+        mentions, _ = get_mentions(request)
+        if mentions.get("mentions"):
```
kevincarrogan, line 113: Are we expecting there to _not_ be a `mentions` item?

caseworker/core/services.py
```diff
@@ -234,3 +234,10 @@ def get_menu_notifications(request):
         request.cached_get_menu_notifications = client.get(request, "/gov-users/notifications/")
     response = request.cached_get_menu_notifications
     return response.json()
+
+
+def get_mentions(request):
+    if not hasattr(request, "cached_mentions"):
+        request.cached_mentions = client.get(request, "/cases/user-case-note-mentions/")
+    response = request.cached_mentions
```
kevincarrogan, line 242: Why are we doing this caching? Do we anticipate that `get_mentions` is going to be called multiple times for a given request?

caseworker/templates/layouts/base.html
```diff
@@ -61,6 +61,18 @@
 				</span>
 			</a>
 			<span></span>
+			{% if FEATURE_MENTIONS_ENABLED %}
+			<a id="link-mentions" class="app-header__link" href="{% url 'users:user_case_note_mentions' %}" data-tooltip="Mentions">
```
kevincarrogan, line 65: The tooltip doesn't seem necessary here.

caseworker/templates/layouts/base.html
```diff
@@ -61,6 +61,18 @@
...
+				{% if NEW_MENTIONS_COUNT > 0 %}
+					<b>
```
kevincarrogan, line 67: We shouldn't use `<b>` tags as they are considered deprecated `<strong>` or some other semantically relevant element would be better.

unit_tests/caseworker/users/test_views.py
```diff
@@ -1,7 +1,84 @@
...
+    requests_mock.get(
+        client._build_absolute_uri("/cases/user-case-note-mentions/"),
+        json=data,
+    )
+
+    url = reverse("users:user_case_note_mentions")
```
kevincarrogan, line 63: This could be any URL right because we display the mentions on every page? Maybe we should check a couple of URLs to make sure we know that this is appearing in the context for more than one (and when I first read this test I thought it _only_ applied to pages related to mentions).

unit_tests/caseworker/users/test_views.py
```diff
@@ -1,7 +1,84 @@
...
+        client._build_absolute_uri("/cases/user-case-note-mentions/"),
+        status_code=500,
+        json=data,
+    )
+
+    url = reverse("users:user_case_note_mentions")
+    with pytest.raises(HTTPError) as exc_info:
+        response = authorized_client.get(url)
+    exception = exc_info.value
+    assert exception.response.status_code == 500
```
kevincarrogan, line 81: Considering we are calling the mentions count on _every_ page, do we want to have such a catastrophic failure if we just can't load the mentions count i.e. do we want the whole application to be non-functional just because we might not be able to get the mentions count?

caseworker/core/context_processors.py
```diff
@@ -98,7 +100,22 @@ def lite_menu(request):
...
+            new_mentions = len([mention for mention in mentions["mentions"] if not mention.get("is_accessed")])
+    return {
+        "NEW_MENTIONS_COUNT": new_mentions,
```
saruniitr, line 116: does this need to be in uppercase?

unit_tests/caseworker/cases/views/test_case_filters.py
saruniitr: if it is the last request you can just use `requests_mock.request_history.pop()`

saruniitr (approved):
few minor comments to consider

unit_tests/caseworker/cases/views/test_case_filters.py
Tllew: it's not otherwise we could just do -1

caseworker/core/context_processors.py
Tllew: only if the api was to fail

caseworker/core/services.py
Tllew: it will be called on every page load as it's in the header, this seems consistent with how some of the other calls in context_processor are used

unit_tests/caseworker/users/test_views.py
Tllew: I realise why this feels wrong now, it's because the reverse url is to 'users:user_case_note_mentions" which calls the same url and will actually fail, I'll change this url and it will fail gracefully returning 0 for the mention_count

Tllew: I realise why this felt wrong now, it's because the reverse url is to 'users:user_case_note_mentions" which calls the same url and will actually fail, I'll change this url and it will fail gracefully returning 0 for the mention_count.

caseworker/core/context_processors.py
kevincarrogan: I don’t think you’d even get this far if it failed as the .json call would break first.

caseworker/core/context_processors.py
Tllew: it fit in with the other variables in the context_processor

caseworker/core/context_processors.py
Tllew: I was just following the pattern above but I'll change this to a boolean

Tllew: I was just following the pattern above but I'll change this to an integer.

caseworker/core/context_processors.py
Tllew: seems to be a simple way to see if the user is authenticated,

lite-api returns a lite_api_user_id to authenticated users

kevincarrogan: I suppose I don't really understand why we're doing that check at all. I don't think we do this check for anything else really so I don't understand why these are special cases. Also, if we want to find out if the user is logged in there's a Django way of finding that out based on the current user.

caseworker/core/services.py
kevincarrogan: But this is only cached per request so it's going to get called every single page load anyway. If it's consistent then lets keep it but I just don't really know what it's adding.

PR #1288, opened 2023-05-24
 LTD-3699-is-new-reset 

depsiatwal:
This resets the 'is new' marker for case not mentions. When a user views their mention on these pages it's marked as accessed. The next page loads won't show the 'new' tag. 

This is in conjunction with the API change 
https://github.com/uktrade/lite-api/pull/1454

https://uktrade.atlassian.net/browse/LTD-3699


depsiatwal:
API https://github.com/uktrade/lite-api/pull/1454

Tllew (approved):

caseworker/activities/views.py
```diff
@@ -69,6 +75,18 @@ def get_team_filters(self):
         ]
         return team_filters
 
+    def get(self, request, *args, **kwargs):
+        response = super().get(request, *args, **kwargs)
+        if "mentions" in list(request.GET.keys()):
+            my_unread_mentions = [
+                {"id": m["id"], "is_accessed": True}
```
saruniitr, line 82: shouldn't `is_accessed=False` as these are unread?

depsiatwal: No other way round these are setting that your have read the message hence is_accessed=True from default False.

saruniitr (approved):

PR #1308, opened 2023-06-05
Ltd 3745 filter description enrichment

antroy-madetech:
### Aim


caseworker/bookmarks/services.py
```diff
@@ -69,11 +71,31 @@ def enhance_bookmark(bookmark, filter_data):
             bookmark_filter[f"{key}_1"] = m
             bookmark_filter[f"{key}_2"] = y
 
+    for key in list(bookmark_filter.keys()):
+        if key.startswith("_id_"):
+            del bookmark_filter[key]
```
saruniitr, line 76: what is the reason to delete this?

antroy-madetech: The bookmark_filter dict is just used to create the URL. So this just cleans up the unnecessary entries from the url. It still works with those entries present, so this isn't strictly necessary I guess, but it makes for a cleaner bookmark URL.

caseworker/bookmarks/services.py
```diff
@@ -82,11 +104,18 @@ def _ordered_filter(data):
...
     ]
-    filters = OrderedDict(sorted({k: data[k] for k in data.keys() if data[k] and k not in keys_to_remove}.items()))
-    return filters
+
+    filters = {k: data[k] for k in data.keys() if data[k] and k not in keys_to_remove}
+
+    # Add in _id_ prefixed data to preserve country and regime names.
+    for key in raw_filters:
+        if key.startswith("_id_") and key[4:] in filters:
+            filters[key] = raw_filters[key]
```
currycoder, line 113: I think this whole thing of adding `_id_` keys in at one place and removing them from the other is a little confusing - the average dev maintaining this probably wouldn't know that they need to go to some javascript to figure out how it all hangs together; https://github.com/uktrade/lite-frontend/blame/8953082843f840d1d6e3f74f1adca53726905d79/caseworker/assets/javascripts/case-filters.js#L26

antroy-madetech: Agree with the _id_ thing. It's existing behaviour from the "accessible autocomplete" fields we use in the filters. Only reason I'm using them is to avoid doing an unnecessary lookup for that data, as it's otherwise not available. 

Have tried to document what's going on better though.

currycoder (approved):

PR #1319, opened 2023-06-08
Ensure failed get_queue() call does not prevent user logins

currycoder:
### Aim

Hotfix to resolve issue where logged out users would get a 500 error page when going to a queue page e.g. /queues/00000000-0000-0000-0000-000000000001/

Prior to this fix, the middleware would not handle an exception raised by `get_queue()` when the requesting user was logged out.

### DIFF
--- a/caseworker/queues/middleware.py
+++ b/caseworker/queues/middleware.py
@@ -11,4 +11,7 @@ class RequestQueueMiddleware:
 
     def process_view(self, request, view_func, view_args, view_kwargs):
         if view_kwargs.get("queue_pk"):
-            request.queue = get_queue(request, view_kwargs["queue_pk"])
+            try:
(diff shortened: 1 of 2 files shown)


caseworker/queues/middleware.py
```diff
@@ -11,4 +11,7 @@ def __call__(self, request):
 
     def process_view(self, request, view_func, view_args, view_kwargs):
         if view_kwargs.get("queue_pk"):
-            request.queue = get_queue(request, view_kwargs["queue_pk"])
+            try:
```
depsiatwal, line 14: is it not better to check if user is logged in 
under the session

depsiatwal (commented):
just commented may be as follow up and this could be quick fix

kevincarrogan (approved):

PR #1328, opened 2023-06-14
Bugfix for issue where saving a filter with flags will break the UI

antroy-madetech:
### Aim

Flags are a multi-selection in the filters, and this wasn't originally taken into account causing an issue when trying to render filters that were saved with flags. This fixes the issue.

[LTD-3816](https://uktrade.atlassian.net/browse/LTD-3816)

[LTD-3816]: https://uktrade.atlassian.net/browse/LTD-3816?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/caseworker/bookmarks/services.py
+++ b/caseworker/bookmarks/services.py
@@ -67,7 +67,7 @@ def enrich_bookmark_for_display(bookmark, filter_data, all_flags):
 
         return out
     except Exception as ex:  # pylint: disable=broad-except
-        logger.exception(f"Error enriching bookmark: {str(ex)}")
+        logger.exception("Error enriching bookmark: %s", str(ex))
         return None
 
 
--- a/caseworker/bookmarks/services.py
+++ b/caseworker/bookmarks/services.py
@@ -97,7 +97,9 @@ def description_from_filter(bookmark_filter, filter_data, all_flags):
 
     _change_flag_ids_to_comma_seperated_names(filter_dict, all_flags)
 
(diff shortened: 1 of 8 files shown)


saruniitr (approved):
one comment to consider, otherwise looks good to me

seijihg (approved):

caseworker/bookmarks/services.py
```diff
@@ -20,7 +20,7 @@ def fetch_bookmarks(request, filter_data):
 
     bookmarks = response.json()
     for bookmark in bookmarks["user"]:
-        enrich_bookmark_for_display(bookmark, filter_data)
+        enrich_bookmark_for_display(bookmark, filter_data, all_flags)
```
currycoder, line 23: To make this extra bulletproof, can we please put a `try`/`except` block around this?  Would recommend doing that and a `logger.error()` so we see the issue in sentry

currycoder:
In addition, this PR appears to be pulling in latest dev changes in to uat - we just want to hotfix I think.

caseworker/bookmarks/services.py
antroy-madetech: Actually have done this in the cases.py view so that any issues in fetching the bookmarks will not cause frontend page errors.

currycoder (approved):

PR #1329, opened 2023-06-15
Timeout tau batching spike

Tllew:
### Aim
frontend creates a current_objects dict:

{goods.goods.id : goods.id }

this allows the goods.id to be the "current object" which allows the api to get the correct current_object based on the goods.goods.id

### DIFF
--- a/caseworker/tau/views.py
+++ b/caseworker/tau/views.py
@@ -17,7 +17,7 @@ from caseworker.cases.services import get_case
 from caseworker.cases.views.main import CaseTabsMixin
 from caseworker.core.services import get_control_list_entries
 from caseworker.core.helpers import get_organisation_documents
-from caseworker.cases.services import post_review_good, post_review_goods, post_review_goods_json
+from caseworker.cases.services import post_review_good
 from caseworker.regimes.enums import Regimes
 from caseworker.regimes.services import get_regime_entries
 from caseworker.users.services import get_gov_user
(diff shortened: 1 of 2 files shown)


caseworker/tau/views.py
```diff
@@ -237,33 +237,41 @@ def get_goods(self, good_ids):
...
-            # when sending to the backend
-            del payload["mtcr_entries"]
-            del payload["wassenaar_entries"]
-            del payload["nsg_entries"]
-            del payload["cwc_entries"]
-            del payload["ag_entries"]
-            del payload["regimes"]
-
-            post_review_good(self.request, case_id=self.kwargs["pk"], data=payload)
+            current_objects[good["good"]["id"]] = good["id"]
```
saruniitr, line 256: The keys of `current_objects` is id of `Good` whereas the values are ids of `GoodOnApplication` correct?

There could be multiple `GoodOnApplication` with the same underlying `Good` so the values should really be a list.

Tllew: this dict is {goodonapplication.id : good.id} which allows the api to process it correctly by having a lookup as it does:

goodonapplication.filter(good_id__in=goods)
which is a list of goodonapplication.ids 

the api then needs the current_object which should be the correct good.id, which would be messed up if you don't have a map like this

this is explained in my comments in lite-api draft PR

kevincarrogan (approved):

saruniitr (approved):

PR #1342, opened 2023-06-22
[LTD-3857] Ensure assigned queues filter is only applied for system queue searches

currycoder:
### Aim

Ensure that `assigned_queues` filter is stripped out for searches on user team queues; this should only be possible for system queue searches.  The form does not show this field for system queues, but it was possible for a user to search by it on non-system queues by using a saved filter.  This has now been remedied.

[LTD-3857](https://uktrade.atlassian.net/browse/LTD-3857)

[LTD-3857]: https://uktrade.atlassian.net/browse/LTD-3857?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/unit_tests/caseworker/queues/views/test_cases.py
+++ b/unit_tests/caseworker/queues/views/test_cases.py
@@ -47,6 +47,26 @@ def mock_cases_search(requests_mock, data_cases_search, queue_pk):
 
 
 @pytest.fixture
+def mock_cases_search_team_queue(requests_mock, data_cases_search):
+    encoded_params = parse.urlencode({"page": 1, "flags": []}, doseq=True)
+    url = client._build_absolute_uri(f"/cases/?queue_id={queue_pk}&{encoded_params}")
+    return requests_mock.get(url=url, json=data_cases_search)
+
+
+@pytest.fixture
+def mock_team_queue(requests_mock):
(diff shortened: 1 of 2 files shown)


unit_tests/caseworker/queues/views/test_cases.py
```diff
@@ -46,6 +46,26 @@ def mock_cases_search(requests_mock, data_cases_search, queue_pk):
...
+def mock_team_queue(requests_mock):
+    data = {
+        "id": queue_pk,
+        "alias": None,
+        "name": "Some team",
+        "is_system_queue": False,
+        "countersigning_queue": None,
+    }
+    url = client._build_absolute_uri("/queues/")
+    yield requests_mock.get(url=re.compile(f"{url}.*/"), json=data_queue)
```
kevincarrogan, line 66: You've got a `yield` here and a `return` in the fixture above. Is there a reason there's a difference?

currycoder: Have replaced this with a `return` :+1:

kevincarrogan (approved):

PR #1351, opened 2023-06-26
Regime and CLE multi-select 

antroy-madetech:
### Aim
Add multi-select functionality to regime and CLE filters.

[LTD-3828](https://uktrade.atlassian.net/browse/LTD-3828)
[LTD-3829](https://uktrade.atlassian.net/browse/LTD-3829)

[LTD-3828]: https://uktrade.atlassian.net/browse/LTD-3828?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[LTD-3829]: https://uktrade.atlassian.net/browse/LTD-3829?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


caseworker/core/services.py
```diff
@@ -183,25 +183,9 @@ def get_control_list_entries(request, convert_to_options=False, include_parent=F
...
-                key=regime_entry["pk"],
-                value=regime_entry["name"],
-                description=regime_entry["name"],
-            )
-        )
-
-    return regime_entries_cache
+def get_regime_entries(request):
+    data = client.get(request, "/static/regimes/entries/")
+    return [{"id": regime["pk"], "name": regime["name"]} for regime in sorted(data.json(), key=lambda r: r["name"])]
```
currycoder, line 188: Just talking this through, but it's safe to change the contract for this function because you've removed any remaining callers which expect a list of `Option` objects, right?

currycoder: The old function did some batshit things with a default list in order to introduce caching such that regimes would only be looked up once per python worker.  I don't think we want to go that far, but could we want to do request-level caching for this?

e.g. similar to this pattern https://github.com/uktrade/lite-frontend/blob/dev/caseworker/core/services.py#L232-L236

caseworker/queues/views/cases.py
```diff
@@ -53,6 +54,14 @@ def data(self):
     def all_flags(self):
         return get_flags(self.request, disable_pagination=True)
 
+    @cached_property
+    def all_cles(self):
+        return get_control_list_entries(self.request, include_parent=True)
+
+    @cached_property
+    def all_regimes(self):
+        return get_regime_entries(self.request)
```
currycoder, line 63: I've noticed locally that the speed for the queue view is now much slower than before. For a modest queue, the time for the queue view web request on this new branch is 4.5S for me locally vs 2.3S locally on dev branch.  Comparing with `dev` I think that the Regime entries list endpoint is a big culprit - there's quite an explosion of queries.  You can see this for yourself if you turn on django silk on the api.  That endpoint can definitely be optimised with some `select_related()` calls; https://docs.djangoproject.com/en/4.2/ref/models/querysets/#select-related

caseworker/core/services.py
antroy-madetech: Exactly. The caching stuff was never used. The only place this method is now used doesn't use it so I cleaned up.
I'm pretty sure the method calling this caches.

antroy-madetech: Oh - yes, see below the all_regimes cached property in cases.py

caseworker/queues/views/cases.py
antroy-madetech: I wonder if we get this in as is for now and raise a separate ticket to optimise this?

currycoder: We discussed offline and landed on adding select_related to the API PR to remedy this

currycoder (approved):

seijihg (approved):

PR #1353, opened 2023-06-27
LTD-3758 Type ahead changed to custom sort/filter

Tllew:
### Aim
Have the tokenfield sort by startswith without excluding things that don't start with

[LTD-3768](https://uktrade.atlassian.net/browse/LTD-3758)

[LTD-3768]: https://uktrade.atlassian.net/browse/LTD-3768?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


core/assets/javascripts/lite-tokenfield-starts-with.js
```diff
@@ -0,0 +1,26 @@
...
+        const bStarts = bName.toLowerCase().startsWith(val);
+        if (aStarts && bStarts) return aName.localeCompare(bName);
+        if (aStarts && !bStarts) return -1;
+        if (!aStarts && bStarts) return 1;
+        return aName.localeCompare(bName);
+      });
+  };
+}
+
+export default window.TokenField = LiteTokenFieldCustomSearch;
```
kevincarrogan, line 26: I don't like this side effect here of writing to `window.TokenField`. If we include this version and the other version of token field on the same page then they're going to overwrite each other.

antroy-madetech (approved):
LGTM

caseworker/assets/javascripts/case-notes.js
```diff
@@ -1,4 +1,4 @@
-import { progressivelyEnhanceMultipleSelectField } from "core/multi-select";
+import { progressivelyEnhanceMultipleSelectField } from "core/multi-select-starts-with";
```
kevincarrogan, line 1: I think having this named the same thing as the other one from the other file is going to cause confusion. It won't be obvious that these are different without looking at the imports.

core/assets/javascripts/__tests__/multi-select-starts-with.test.js
```diff
@@ -0,0 +1,158 @@
...
+// make a document with a basic dropdown to instantiate and filter with tokenfield
+
+const createElements = () => {
+  document.body.innerHTML = `
+    <select
+    name="mentions"
+    aria-describedby="id_mentions_hint"
+    class="selectmultiple"
+    id="id_mentions"
+    multiple=""
```
kevincarrogan, line 11: We should remove any attributes here that aren't related to the tests themselves.

core/assets/javascripts/__tests__/multi-select-starts-with.test.js
kevincarrogan, line 1: These tests appear to know a lot about the implementation and are essentially calling a non-public method.

We should write these tests at a higher level to test that the outcome is valid not testing a private method.

core/assets/javascripts/multi-select-starts-with.js
kevincarrogan, line 1: This seems to replicate a lot of what's in the other multiselect file and means these could too easily go out of sync with each other.

core/assets/javascripts/__tests__/multi-select-starts-with.test.js
Tllew: I think the name of the file is wrong here, this should be lite-tokenfield-starts-with.test.js

kevincarrogan (approved):

PR #1354, opened 2023-06-27
Added Report Summary filter

antroy-madetech:
### Aim

Add Report Summary filter to the Product filter tab

[LTD-3831](https://uktrade.atlassian.net/browse/LTD-3831)

[LTD-3831]: https://uktrade.atlassian.net/browse/LTD-3831?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/unit_tests/caseworker/queues/views/test_cases.py
+++ b/unit_tests/caseworker/queues/views/test_cases.py
@@ -152,6 +152,7 @@ def test_cases_home_page_view_context(authorized_client):
         "max_total_value",
         "control_list_entry",
         "regime_entry",
+        "report_summary",
         "submitted_from",
         "submitted_to",
         "finalised_from",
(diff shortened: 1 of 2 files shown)


unit_tests/caseworker/queues/views/test_cases.py
```diff
@@ -153,6 +153,7 @@ def test_cases_home_page_view_context(authorized_client):
         "country",
         "control_list_entry",
         "regime_entry",
+        "report_summary",
```
currycoder, line 156: Could I suggest adding an extra test case along these lines; https://github.com/uktrade/lite-frontend/blob/05d585f76b7be12047ae30ea58ded550f82a9cf8/unit_tests/caseworker/queues/views/test_cases.py#L319-L330

It tests that the input is present in the HTML and that using it will pass things through to the API call

PR #1374, opened 2023-07-05
[LTD-3793] Quick summary back button

currycoder:
### Aim

Fix up back links from pages that should link back to the case detail tab instead of the new default "quick summary" tab.

[LTD-3793](https://uktrade.atlassian.net/browse/LTD-3793)

[LTD-3793]: https://uktrade.atlassian.net/browse/LTD-3793?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/caseworker/cases/forms/change_status.py
+++ b/caseworker/cases/forms/change_status.py
@@ -21,6 +21,8 @@ def change_status_form(queue, case, statuses):
                 ],
             ),
         ],
-        back_link=BackLink(url=reverse("cases:case", kwargs={"queue_pk": queue["id"], "pk": case["id"]})),
+        back_link=BackLink(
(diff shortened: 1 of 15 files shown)


kevincarrogan (commented):
Is there any way of consolidating this so we don't have to update this in so many places at once?

currycoder:
@kevincarrogan `browser.back()`.  All the way through updating this I've been feeling like the coupling we introduce where these pages need to know about their predecessors is pretty non-ideal; so I do feel like we need a bigger discussion about this button and what value maintaining these links ourselves actually gives us.

kevincarrogan (approved):

Tllew (approved):

PR #1379, opened 2023-07-06
Ltd 3855 saved filters form based part 2

kevincarrogan:
### Aim

Use the case filters form itself to generate a description for the bookmark.

[LTD-3855](https://uktrade.atlassian.net/browse/LTD-3855)

[LTD-3855]: https://uktrade.atlassian.net/browse/LTD-3855?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/caseworker/bookmarks/services.py
+++ b/caseworker/bookmarks/services.py
@@ -115,10 +115,10 @@ class BookmarkEnricher:
             out["url"] = self._url_from_bookmark(bookmark_filter)
 
             return out
-        except Exception as ex:  # pylint: disable=broad-except
-            logger.exception("Error enriching bookmark")
+        except Exception:  # pylint: disable=broad-except
             if settings.DEBUG:
                 raise
+            logger.exception("Error enriching bookmark")
             return None
 
     def _description_from_filter(self, bookmark_filter):
--- a/caseworker/bookmarks/services.py
+++ b/caseworker/bookmarks/services.py
@@ -39,9 +39,6 @@ def add_bookmark(request, data):
     user, _ = get_gov_user(request)
     user_id = user["user"]["id"]
 
-    for k, v in filter_to_save.items():
-        if type(v) == date:
-            filter_to_save[k] = v.strftime("%d-%m-%Y")
     data = {
         "filter_json": filter_to_save,
         "name": bookmark_name,
@@ -177,6 +174,9 @@ def enrich_filter_for_saving(data):
             continue
         if isinstance(value, Decimal):
             filters[key] = str(value)
+            continue
+        if isinstance(value, date):
+            filters[key] = value.strftime("%d-%m-%Y")
             continue
         filters[key] = value
 
--- a/caseworker/bookmarks/services.py
+++ b/caseworker/bookmarks/services.py
@@ -40,9 +40,8 @@ def add_bookmark(request, data):
     user_id = user["user"]["id"]
 
     for k, v in filter_to_save.items():
-        if isinstance(v, date):
+        if type(v) == date:
             filter_to_save[k] = v.strftime("%d-%m-%Y")
-
     data = {
         "filter_json": filter_to_save,
         "name": bookmark_name,
@@ -170,11 +169,15 @@ def enrich_filter_for_saving(data):
         "return_to",
     ]
 
-    filters = {key: val for key, val in data.items() if val and key not in keys_to_remove}
-
-    # Ensure that Decimal values are cast to string ready for json serialization
-    for key, value in filters.items():
+    filters = {}
+    for key, value in data.items():
+        if not value:
+            continue
+        if key in keys_to_remove:
+            continue
         if isinstance(value, Decimal):
             filters[key] = str(value)
+            continue
+        filters[key] = value
 
     return OrderedDict(sorted(filters.items()))
--- a/caseworker/bookmarks/services.py
+++ b/caseworker/bookmarks/services.py
@@ -47,20 +47,20 @@ def add_bookmark(request, data):
         "name": bookmark_name,
         "user_id": user_id,
     }
-    response = client.post(request, f"/bookmarks", data)
+    response = client.post(request, "/bookmarks", data)
 
     return response
 
 
 def delete_bookmark(request, bookmark_id):
     data = {"id": bookmark_id}
-    response = client.delete(request, f"/bookmarks", data)
+    response = client.delete(request, "/bookmarks", data)
     return response
 
 
 def rename_bookmark(request, bookmark_id, name):
     data = {"id": bookmark_id, "name": name}
-    return client.put(request, f"/bookmarks", data)
+    return client.put(request, "/bookmarks", data)
 
 
 def get_label_from_field(field):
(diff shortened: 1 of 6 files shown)


caseworker/bookmarks/services.py
```diff
@@ -1,66 +1,106 @@
...
+
+from caseworker.queues.views.forms import CasesFiltersForm
 from caseworker.users.services import get_gov_user
 from core import client
 
 logger = logging.getLogger(__name__)
 
 
-def fetch_bookmarks(request, filter_data, all_flags, all_regimes, bookmark_base_url):
+def fetch_bookmarks(request, filter_data, all_flags, all_cles, all_regimes, queue, bookmark_base_url):
```
currycoder, line 20: You mentioned you want to get rid of this "threading of parameters" in a follow-up, how were you intending on doing that?

caseworker/bookmarks/services.py
```diff
@@ -75,34 +115,31 @@ def _enrich_bookmark_for_display(self, bookmark):
...
-            if isinstance(filter_dict[filter_key], list):
-                filter_dict[filter_key] = ", ".join(sorted(str(item) for item in filter_dict[filter_key]))
-
-        return ", ".join(
-            [f"{k.capitalize().replace('_', ' ')}: {str(v).replace('_', ' ')}" for (k, v) in filter_dict.items()]
+        # We split the dates here because the form is expecting data as if it
+        # were originally posted from the form but we store the date as a single
+        # value so we need to split it out again
+        data = self._split_dates(bookmark_filter)
+        form = CasesFiltersForm(
```
currycoder, line 129: I wasn't really expecting the form to be instantiated over here but rather for the description to just get passed in to this module for saving.  This does avoid the trickiness of instantiating the form on filters submit, though.  Out of interest, did you consider doing that?

unit_tests/caseworker/bookmarks/test_services.py
```diff
@@ -11,41 +13,95 @@ def new_bookmarks(filter):
...
-        ({"status": "ogd_advice"}, "Status: OGD Advice"),
-        ({"team_advice_type": "no_licence_required"}, "Team advice type: No Licence Required"),
-        ({"final_advice_type": "not_applicable"}, "Final advice type: Not Applicable"),
+        ({"case_officer": GOV_USER_ID}, "Licensing Unit case officer: John Smith"),
+        ({"assigned_user": GOV_USER_ID}, "Case adviser: John Smith"),
+        ({"status": "ogd_advice"}, "Case status: OGD Advice"),
+        ({"control_list_entry": ["ML1", "ML1a"]}, "Control list entry: ML1, ML1a"),
+        ({"submitted_from": "12-12-2010"}, "Submitted after: 12-12-2010"),
+        ({"is_trigger_list": True}, "Trigger list: True"),
+        ({"countries": ["DE"]}, "Country: Germany"),
```
currycoder, line 46: Could `assigned_queues` be added here?

currycoder (commented):
Minor comments and questions

unit_tests/caseworker/bookmarks/test_services.py
kevincarrogan: In the follow up the separate fields aren't really going to matter as much so there might not be much value to adding it now.

caseworker/bookmarks/services.py
kevincarrogan: We can get there eventually. This keeps most of the code the same but just stops a lot of the hardcoded retrieval of fields and just allows it to come from the form itself.

caseworker/bookmarks/services.py
kevincarrogan: This code will eventually have no idea about a form or instantiate it itself so it won't need to have all of these parameters.

currycoder (approved):

caseworker/bookmarks/services.py
kevincarrogan: Although, actually. The description won't ever be passed in because we're working it out once we get the filters back from the API so we're having to work it out there.

Shifting it to when we save the bookmark has the problem where things might change on the form e.g. a field label, and the saved data will be stuck at a point in time. Doing it this way means that the description should change inline with any form changes.

currycoder: I agree that is a problem but TBH I feel like having a "fixed description" is kind of minor vs doing `n` bookmark description constructions (and `n` form instantiations) for every hit on the search page.

currycoder: Say that we rename a field, I think right now that field would drop out of the form description entirely.  Not sure if this is preferable to it appearing but not working.  I suppose if we really thought about it, we could pair a field name change with a data migration on bookmarks.

kevincarrogan: There are some things that will only be solvable via migrations and we'll have to respond to those by watching the exceptions thrown in Sentry and changing accordingly, or being proactive and making sure we make migrations when we change the form.

What this solves is some of the smaller, but probably more frequent, changes where we change a field label or an option label.

caseworker/bookmarks/services.py
```diff
@@ -1,66 +1,106 @@
...
     data = {"id": bookmark_id, "name": name}
-    return client.put(request, f"/bookmarks", data)
+    return client.put(request, "/bookmarks", data)
+
+
+def get_label_from_field(field):
+    return field.label
+
+
+def get_readable_value_from_field(field, value):
```
depsiatwal, line 67: As this is more of a helper as opposed to service api calls is it cleaner to put in it's owner helper file

caseworker/bookmarks/services.py
```diff
@@ -1,66 +1,106 @@
...
+    if isinstance(field, forms.ChoiceField):
+        return dict(field.choices)[value]
+
+    if isinstance(field, DateInputField):
+        return value.strftime("%d-%m-%Y")
+
+    return value
+
+
+def get_description_from_form(form):
```
depsiatwal, line 82: same as above

caseworker/bookmarks/services.py
```diff
@@ -1,66 +1,106 @@
...
+        label = get_label_from_field(field)
+        readable_value = get_readable_value_from_field(field, field_value)
+        labelled_values.append((label, readable_value))
+
+    return ", ".join(f"{k}: {v}" for k, v in sorted(labelled_values, key=lambda v: v[0]))
 
 
 class BookmarkEnricher:
-    def __init__(self, filter_data, all_flags, all_regimes, bookmark_base_url):
+    def __init__(self, request, filter_data, all_flags, all_cles, all_regimes, queue, bookmark_base_url):
```
depsiatwal, line 97: again is service best place for this ?

depsiatwal (commented):
LGTM just wonder if it can be structured a little cleaner

caseworker/bookmarks/services.py
kevincarrogan: It's specifically used in this services file. It's not a generic helper.

kevincarrogan:
> LGTM just wonder if it can be structured a little cleaner

Yeah. I kind of don't disagree. At the moment this is mostly trying to move away from hardcoded values to build the form description, I already have two more PRs that follow on after this that do more tidying (they don't really move this out though).

depsiatwal (approved):
lgtm

PR #1382, opened 2023-07-07
[LTD-3952] Denials match back button

currycoder:
### Aim
Achieves two things;
- Ensures that the back button on the denial detail page returns to the case detail tab rather than the quick summary tab.
- Improves the logic for whether the filters sidebar is visible by default on the queue view page; this should only show if the caseworker has applied some filtering. Previously, it would show if any GET parameters were present; so a user would go to the second page and the filters sidebar would pop open. 

[LTD-3952](https://uktrade.atlassian.net/browse/LTD-3952)

[LTD-3952]: https://uktrade.atlassian.net/browse/LTD-3952?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/caseworker/queues/views/cases.py
+++ b/caseworker/queues/views/cases.py
@@ -295,6 +295,13 @@ class Cases(LoginRequiredMixin, CaseDataMixin, FormView):
         sanitised_url = self._strip_param_from_url(url, "csrfmiddlewaretoken")
         return redirect(sanitised_url)
 
+    def is_filters_visible(self):
+        # when this view instantiates the form on submission, we can do better by using form.is_bound
+        # until then we must interrogate GET parameters
+        params_to_ignore = set(["selected_tab", "page"])
+        all_params = set(self.request.GET.keys())
+        return len(all_params - params_to_ignore) > 0
+
     def get_context_data(self, *args, **kwargs):
         try:
             updated_queue = [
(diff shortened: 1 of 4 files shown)


kevincarrogan (approved):

caseworker/queues/views/cases.py
```diff
@@ -295,6 +295,13 @@ def post(self, request, *args, **kwargs):
         sanitised_url = self._strip_param_from_url(url, "csrfmiddlewaretoken")
         return redirect(sanitised_url)
 
+    def is_filters_visible(self):
+        # when this view instantiates the form on submission, we can do better by using form.is_bound
+        # until then we must interrogate GET parameters
+        params_to_ignore = set(["selected_tab", "page"])
+        all_params = set(self.request.GET.keys())
+        return len(all_params - params_to_ignore) > 0
+
```
saruniitr, line 304: Not sure if the behaviour is updated recently when filters are moved to side panel - if any filters are selected then filters section is automatically expanded in JS.
Is that not the case with side panel which is why this is required?

currycoder: Exactly Arun, the filters have been relocated under a GDS details component.  So this makes sure that the functionality you are describing works as before (which I think was some custom JS)

saruniitr: nice, good to see this moved to Python land.

PR #1384, opened 2023-07-10
Run case filters through form validation on post

kevincarrogan:
### Aim

Display errors when filter information is invalid.

[LTD-3957](https://uktrade.atlassian.net/browse/LTD-3957)


caseworker/queues/views/cases.py
```diff
@@ -286,14 +285,14 @@ def get_form_kwargs(self):
         kwargs["initial"]["return_to"] = self.get_return_url()
         return kwargs
 
-    def post(self, request, *args, **kwargs):
```
currycoder, line 289: When I have a bookmark saved and I apply filters with an invalid date value (e.g. day = "foo"), I get a 500 error;

```
TypeError at /
DeclarativeFieldsMetaclass object got multiple values for keyword argument 'data'
```

kevincarrogan: Fixed this now.

Tllew (approved):
Looks good to me

PR #1396, opened 2023-07-16
Minio

kevincarrogan:
### Aim


unit_tests/exporter/applications/views/test_documents.py
```diff
@@ -0,0 +1,46 @@
...
+            "document_pk": document_id,
+        },
+    )
+    response = authorized_client.get(url)
+
+    assert response.status_code == 200
+    assert isinstance(response, StreamingHttpResponse)
+    assert response.headers["Content-Type"] == "application/doc"
+    assert response.headers["Content-Disposition"] == 'attachment; filename="fakefile.doc"'
+    assert reduce(lambda x, y: x + y, response.streaming_content) == b"test"
```
stuaxo, line 46: Everything looks good, though the reduce bit at the end feels like a slightly fancy way to concatenate data from the streaming_content (if I understand it correctly) -- not a blocker.

stuaxo:
NIce - I like that this moves the repeated s3 stuff into it's own abstraction.

PR #1399, opened 2023-07-19
Added new selectable column for End users

antroy-madetech:
### Aim

This ticket adds a new customisable column for displaying the end users and ultimate end users for a case.

[LTD-3960](https://uktrade.atlassian.net/browse/LTD-3960)


depsiatwal (commented):
LGTM are there  any tests for any of the other configurable columns ?

caseworker/templates/includes/case-row.html
```diff
@@ -164,6 +164,25 @@
 		</ol>
 	</td>
 
+    {# Users #}
+    <td class="govuk-table__cell lite-mobile-hide" data-customiser-key="users">
+        <ol class="govuk-list destinations__list app-users__list">
```
kevincarrogan, line 169: Why does this have a class of destinations list?

caseworker/templates/includes/case-row.html
```diff
@@ -164,6 +164,25 @@
...
+            {% for user in case.endusers %}
+                <li class="app-cles__item">
+                    {% if user.type == "end_user"  %}
+                    <div>
```
kevincarrogan, line 173: Is this div necessary?

caseworker/templates/includes/case-row.html
antroy-madetech: Just to keep it consistent with all the other columns. I'll try without to see if it still looks OK, but regimes, users, products etc all have it.

caseworker/templates/includes/case-row.html
antroy-madetech: No. Left over from an earlier version. Will remove.

caseworker/templates/includes/case-row.html
kevincarrogan: I'm not necessarily saying that we should remove this if it needs to be there for styling purposes, but if we're using this class in places that aren't literally the destinations list then we should rename this class.

antroy-madetech: I've removed it, and I can't see any visual difference. Happy to investigate where this class is used and rename to something more suitable if you want though

kevincarrogan (approved):

saruniitr (approved):

PR #1400, opened 2023-07-19
LTD-3938-add-case-advice-column

depsiatwal:
https://uktrade.atlassian.net/browse/LTD-3958


caseworker/templates/includes/recommandations.html
```diff
@@ -0,0 +1,23 @@
+<ul class="case-rec__list">
+    {% if case.advice %}
+        {% for team_id, team_advice in case.advice.items %}
+            {% for advice in team_advice %}
+            {% if advice.type.key in 'approve, proviso, refuse' %}
+            <li class="case-rec__item">
```
Tllew, line 6: formatting here might make it a little easier to track, it will be quite nested though

Tllew (approved):
Looks good to me

caseworker/assets/styles/components/_case_recommandations.scss
```diff
@@ -0,0 +1,19 @@
...
+        list-style: none;
+        margin: 0;
+        padding: 0;
+    }
+
+    &__item {
+        margin-bottom: govuk-spacing(2);
+    }
+
+    &__recommandation {
```
kevincarrogan, line 12: recommendation is spelt recommendation not recommandation.

caseworker/assets/styles/components/_case_recommandations.scss
```diff
@@ -0,0 +1,19 @@
+.case-rec {
```
kevincarrogan, line 1: Lets just use the full word here instead of abbreviations.

caseworker/assets/styles/components/_case_recommandations.scss
kevincarrogan: This spelling is wrong in multiple places so check in more than one place.

caseworker/templates/includes/recommandations.html
```diff
@@ -0,0 +1,23 @@
+<ul class="case-rec__list">
```
kevincarrogan, line 1: When using BEM we typically want something at the top level that wraps everything and then it's the elements that would get the `__{element_name}`.

Ideally there should be a top level element just called `.case-rec`.

caseworker/templates/includes/recommandations.html
```diff
@@ -0,0 +1,23 @@
...
+    {% if case.advice %}
+        {% for team_id, team_advice in case.advice.items %}
+            {% for advice in team_advice %}
+            {% if advice.type.key in 'approve, proviso, refuse' %}
```
kevincarrogan, line 5: This feels kind of weird. If the key was a substring of any of those three this would also pass (the chances of this are low but it would be better if this were more explicit).

caseworker/assets/styles/components/_case_recommandations.scss
depsiatwal: that's my bad spelling :-( 
corrected

caseworker/assets/styles/components/_case_recommandations.scss
depsiatwal: corrected

caseworker/templates/includes/case-row.html
```diff
@@ -122,7 +122,10 @@
 			{% endif %}
 		</ul>
 	</td>
-
+	{# case recommondation #}
+	<td class="govuk-table__cell lite-mobile-hide" data-customiser-key="case_recommondation">
```
Tllew, line 126: this should be `class="govuk-table__cell lite-mobile-hide case-recommendation"`

Tllew: also spelling on case_recommondation

caseworker/templates/includes/case-row.html
```diff
@@ -122,7 +122,10 @@
 			{% endif %}
 		</ul>
 	</td>
-
+	{# case recommondation #}
```
Tllew, line 125: spelling

caseworker/templates/includes/case-row.html
```diff
@@ -122,7 +122,10 @@
...
+	<td class="govuk-table__cell lite-mobile-hide case-recommendation" data-customiser-key="case_recommondation">
```
Tllew: customiser-key spelling

caseworker/templates/queues/cases.html
```diff
@@ -76,7 +76,7 @@ <h1 class="govuk-heading-l">
...
+            <div class="customiser" data-customiser-spec='{"options_label": "Customise queue view", "identifier": "queue-view", "analytics_prefix": "qv", "options_hint": "Select columns to show", "toggleable_elements": [{"label": "Destination", "key": "destination", "default_visible": true}, {"label": "Case allocation", "key": "case_allocation", "default_visible": true}, {"label": "Case recommondation", "key": "case_recommondation", "default_visible": true}, {"label": "Case updates", "key": "case_updates", "default_visible": true}, {"label": "Products", "key": "products"}, {"label": "Control list entry", "key": "control_list_entry"}, {"label": "Report summary", "key": "report_summary"}, {"label": "Regime", "key": "regime"}, {"label": "Total value", "key": "total_value"}, {"label": "Queries", "key": "queries"}, {"label": "Denial matches", "key": "denial_matches"}, {"label": "Intended end use", "key": "intended_end_use"}, {"label": "Flags", "key": "flags", "default_visible": true}]}'>
```
Tllew, line 79: spelling

caseworker/templates/queues/cases.html
```diff
@@ -102,6 +102,7 @@ <h1 class="govuk-heading-l">
 	            	    			<th class="govuk-table__header app-table__header--skeleton" scope="col">Case</th>
 	            	    			<th class="govuk-table__header app-table__header--skeleton" data-customiser-key="destination" scope="col">Destination</th>
 	            	    			<th class="govuk-table__header app-table__header--skeleton" data-customiser-key="case_allocation" scope="col">Case allocation</th>
+									<th class="govuk-table__header app-table__header--skeleton" data-customiser-key="case_recommondation" scope="col">Case recommondation</th>
```
Tllew, line 105: spelling for data-customiser-key and text

caseworker/templates/includes/case-row.html
depsiatwal: ahhh corrected

caseworker/templates/includes/recommandations.html
depsiatwal: not sure key is very specific to DB and has finite values so either we leave or have a messy statement for each condition

kevincarrogan: Why does it have to be messy? Pass through an actual list in the context and match it as an item in a list.

kevincarrogan (approved):

PR #1401, opened 2023-07-20
LTD-3964-refusal-reasons

seijihg:
I could not find a way to override a class in Django form for the labels. The crispy form would just override it so I just change it inside .scss instead. Any suggestion is welcome.

### Workflow

- [x] Ticket exists in Jira https://uktrade.atlassian.net/browse/LTD-3964

### Reviewing help

- [ ] Includes screenshot(s) - ideally before and after, but at least after

### Housekeeping

### Merging

### DIFF
--- a/caseworker/assets/javascripts/advice.js
+++ b/caseworker/assets/javascripts/advice.js
@@ -1,10 +1,10 @@
-import { progressivelyEnhanceMultipleSelectField } from "core/multi-select";
+/* global $ */
 
-let text_for = "text";
-let text_label = $("label[for=" + text_for + "]");
+let textFor = "text";
+let textLabel = $("label[for=" + textFor + "]");
 
 $(document).ready(function () {
-  text_label.append('<span class="lite-form-optional">(optional)</span>');
+  textLabel.append('<span class="lite-form-optional">(optional)</span>');
   addOrRemoveOptional();
 });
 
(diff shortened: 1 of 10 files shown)


caseworker/assets/javascripts/advice.js
```diff
@@ -11,14 +14,24 @@ $(".govuk-radios--conditional").on("click", function () {
...
+
+const initDenialReasons = () => {
+  const denialReasonField = document.getElementById("id_denial_reasons");
+
+  if (!denialReasonField) return;
+
+  progressivelyEnhanceMultipleSelectField(denialReasonField);
+};
+
+export { initDenialReasons };
```
kevincarrogan, line 37: There’s no need to export this function. 

This file looks like it’s included on the page that needs it so just run the function here.

caseworker/assets/javascripts/main.js
```diff
@@ -21,6 +21,7 @@ import initDestinationsList from "./show-hide-destinations.js";
 import initMentionUsers, { initCaseNotes } from "./case-notes";
 import { initExpanders } from "./list-expander";
 import { initCustomisers } from "./customiser";
+import { initDenialReasons } from "./advice.js";
```
kevincarrogan, line 24: Importing this on every page means that all of the advice logic is being run for every page which I do think I’d what we want.

caseworker/assets/styles/components/_advice.scss
```diff
@@ -216,3 +216,11 @@
 		margin-left: auto;
 	}
 }
+
+.advice-refusal-labels {
+  .govuk-label {
```
kevincarrogan, line 221: Why are we having to override so much don’t stuff here?

caseworker/assets/javascripts/advice.js
seijihg: yea it seems like I have put it in the wrong place this file is not running during review and combine. Where the changes are.

caseworker/assets/javascripts/main.js
seijihg: I cant just import it in review_consolidate.html since it dependant on different modules outside of this JS file and at the time is not loaded yet. Unless I am missing something :D (like always)

`{% block javascript %}
    <script src="{% static 'javascripts/refusal-advice.js' %}" type="module"></script>
{% endblock %}`

kevincarrogan (approved):

antroy-madetech (approved):
LGTM

PR #1403, opened 2023-07-24
LTD-3965-replace-refusal-reason-to-note

seijihg:
### Aim

Waiting for Duane to come back if we are just replacing the refusal to note and functionality wise is the same. If it is, there will be no changes needed to be made additionally.

https://uktrade.atlassian.net/browse/LTD-3965

### DIFF
--- a/caseworker/advice/forms.py
+++ b/caseworker/advice/forms.py
@@ -177,6 +177,12 @@ class ConsolidateApprovalForm(GiveApprovalAdviceForm):
 
 
 class RefusalAdviceForm(forms.Form):
+    refusal_reasons = forms.CharField(
+        widget=forms.Textarea(attrs={"rows": "7"}),
+        label="Enter the refusal note as agreed in the refusal meeting",
+        error_messages={"required": "Enter the refusal meeting note"},
+    )
+
     def _group_denial_reasons(self, denial_reasons):
         grouped = defaultdict(list)
         for item in denial_reasons:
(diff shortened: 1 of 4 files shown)


kevincarrogan (commented):
It would be good to see some tests around this. At the very least something that checks the error value coming back if this isn't input.

caseworker/advice/forms.py
```diff
@@ -173,11 +173,10 @@ def __init__(self, denial_reasons, *args, **kwargs):
...
         )
-        self.fields["refusal_reasons"] = PicklistCharField(
-            picklist_attrs={"target": "refusal_reasons", "type": "standard_advice", "name": "standard advice"},
-            label="What are your reasons for this refusal?",
-            help_link_text="Choose a refusal reason from the template list",
-            error_messages={"required": "Enter a reason for refusing"},
+        self.fields["refusal_reasons"] = forms.CharField(
+            widget=forms.Textarea(attrs={"rows": "7"}),
+            label="Enter the refusal note as agreed in the refusal meeting",
+            error_messages={"required": "Enter the meeting refusal note"},
```
kevincarrogan, line 179: This would be better outside of the init and put into the declaration of the form instead.

saruniitr, line 179: Could you check if it is by any chance "Enter the refusal meeting note".

seijihg: done and fixed!

saruniitr (approved):

kevincarrogan (approved):

PR #1408, opened 2023-07-25
LTD-3963 Licence condition -> Radio textarea

Tllew:
* added radiotextfield to proviso for approvals
* fixed issues with bold labels 
* changed class on some fields to more closely reflect the figma designs 
* fixed the input field retaining on the tokenfield

[LTD-3963](https://uktrade.atlassian.net/browse/LTD-3963)

[LTD-3963]: https://uktrade.atlassian.net/browse/LTD-3963?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


antroy-madetech (approved):
LGTM

caseworker/assets/styles/components/_advice.scss
```diff
@@ -218,9 +218,19 @@
...
+		font-weight:700;
+	}
+	.tokenfield{
+		max-width: 100%;
+	}
+}
+
+.govuk-radios{
+	&__label{
+		font-weight:400 !important;
```
kevincarrogan, line 234: We really shouldn't use `!important` this is usually a sign that there's a better way to do this.

unit_tests/caseworker/advice/views/test_consolidate.py
```diff
@@ -408,6 +408,55 @@ def test_approval_reasons_manual(
...
+                {"name": "ALLCAPSNOSPACES", "text": "This is all caps text"},
+            ]
+        },
+    )
+    response = authorized_client.get(url + "approve/")
+    assert response.status_code == 200
+    form = response.context["form"]
+    assert isinstance(form, forms.GiveApprovalAdviceForm)
+    # this is built mock_approval_reason
+    # must be == and not is, because it's a django choicefield not a list
```
kevincarrogan, line 443: I don't really understand this comment. If you wanted to check two lists you wouldn't really want to use `is` anyway.

caseworker/assets/styles/components/_advice.scss
kevincarrogan: You’re also setting this globally for all these types of labels which is a really good reason not to use `!important`.

unit_tests/caseworker/advice/views/test_consolidate.py
Tllew: removed, it was because I generally compare with is where possible

Tllew: removed, it was because I generally compare with is where possible

caseworker/assets/styles/components/_advice.scss
Tllew:
```<label class="govuk-label govuk-radios__label" for="id_approval_radios_1">
          No clear concerns
        </label>```
Let's have a chat about this, I've moved it into the `.advice-refusal-labels` but the refusals flow uses .govuk-label: font-weight:700; this is what was making me think we should try to use govuk-label--s for those items specifically

kevincarrogan (approved):

PR #1411, opened 2023-07-27
Seperates the forms

seijihg:
### Aim
Updates the form for it to be specific to ReviewConsolidate so it wont affect AdviceView forms. 
Updates to the post helper with conditions for saving down refusal_note.

https://uktrade.atlassian.net/browse/LTD-3965


depsiatwal (commented):
generally LGTM looks like more tests are needed

caseworker/advice/forms.py
```diff
@@ -178,7 +178,7 @@ def __init__(self, team_alias, *args, **kwargs):
         )
 
 
-class RefusalAdviceForm(forms.Form):
+class BaseRefusalForm(forms.Form):
```
saruniitr, line 181: `refusal_note` is only to be shown for LU so is it not an option to return a different form based on the team?

caseworker/advice/views.py
```diff
@@ -544,6 +544,13 @@ def is_advice_approve_only(self):
     def get_form(self):
         form_kwargs = self.get_form_kwargs()
 
+        if (
+            self.kwargs.get("advice_type") == "refuse"
```
saruniitr, line 548: we added constants for advice types, please update to use constants everywhere,
https://github.com/uktrade/lite-frontend/blob/dev/caseworker/advice/constants.py#L24

caseworker/advice/views.py
```diff
@@ -615,6 +622,7 @@ def setup(self, request, *args, **kwargs):
         # to debug when this goes south.
         sentry_sdk.set_context("caseworker", self.caseworker)
         sentry_sdk.set_context("advice", {"advice": self.case.advice})
+        self.advices_by_team = services.filter_advice_by_team(team_advice, user_team_alias)
```
saruniitr, line 625: We are doing the same in the next line so what is the need for this?

ui_tests/caseworker/features/give_advice.feature
```diff
@@ -323,9 +323,9 @@ Feature: I want to record my user advice and any comments and conditions relatin
     And I click refuse
     And I click continue
     And I select refusal criteria "1a, 1b, 1c, 1d, 1e, 1f"
-    And I enter "reason for this refusal" as the reasons for refusal
+    And I enter "refusal meeting note" as the refusal note
```
saruniitr, line 326: instead of "refusal note" at the end, can it be changed as "refusal meeting note" ?

caseworker/advice/views.py
seijihg: I needed to get all the advices so I could filter out the one with `is_refusal_note AND refuse`
The next line was giving just a first advice.

caseworker/advice/views.py
```diff
@@ -626,9 +634,23 @@ def get_approval_data(self):
...
 
     def get_refusal_data(self):
+        # Filtered advices for refusal note so specifically for LU
+        filtered_advices = [
+            advice
+            for advice in self.advices_by_team
+            if advice["is_refusal_note"] and advice["type"].get("key") == "refuse"
+        ]
+        refusal_note = ""
+        denial_reasons = [r for r in self.advice["denial_reasons"]]
```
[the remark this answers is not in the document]
seijihg: we are having two options here one is for:
 - MOD_ECJU and 
 - one for LU so 

if there is no `is_refusal_note` then --> old option (MOD_ECJU) since they are sharing the view.
if there is `is_refusal_note` --> new option

If we are LU at this point we may have advices for both OLD LU `refusal_reasons` and NEW LU `refusal_note` so `filtered_advices` we make sure we are showing the correct `refusal_note` values not the old LU reasons.

caseworker/advice/views.py
saruniitr: All advice instances will have same data for `text`, `type` etc so you can take anyone which is why we are only taking the first one in the next line.

caseworker/advice/forms.py
seijihg: https://github.com/uktrade/lite-frontend/blob/969364b842d4a0020b12ddae2c57b19a0abea259/caseworker/advice/views.py#L549

this line is doing exactly that? or is there another way inside form?

saruniitr: Correct - so is it not an option to have a separate form `LUConsolidateRefusalForm()` just like `RefusalAdviceForm()` instead of changing it to `BaseRefusalForm()`?
This form inheritance may be helpful if that is useful for all teams but instead it is only for one team.

seijihg: class LUConsolidateRefusalForm(BaseRefusalForm) --> it is doing exactly that. It is inheriting the BaseRefusalForm.

So both LUConsolidateRefusalForm and RefusalAdviceForm are using it :D

caseworker/advice/views.py
seijihg: yea actually this part can be refactor and simplified now after confirming with Colin. the old refusal reasons can be overwritten with note and no need for seperation

caseworker/advice/views.py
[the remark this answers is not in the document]
seijihg: no longer need this part refactored!

caseworker/advice/services.py
```diff
@@ -325,13 +326,21 @@ def post_approval_advice(request, case, data, level="user-advice"):
 
 
 def post_refusal_advice(request, case, data, level="user-advice"):
+    if data.get("refusal_note"):
```
Tllew, line 329: Personal preference here really but I like:
```
text = data["refusal_reasons"]
is_refusal_note = False
if data.get("refusal_note"):
    text = data["refusal_note"]
    is_refusal_note = True
```

caseworker/advice/services.py
```diff
@@ -325,13 +326,21 @@ def post_approval_advice(request, case, data, level="user-advice"):
...
+
     json = [
         {
             "type": "refuse",
-            "text": data["refusal_reasons"],
+            "text": text,
             "footnote_required": False,
             subject_name: subject_id,
             "denial_reasons": data["denial_reasons"],
+            "is_refusal_note": is_refusal_note,
```
Tllew, line 343: This doesn't quite feel right, is there backend logic that inserts text into a different field or is this just when it flags is_refusal?

could potentically be a different endpoint for posting refusal_notes?

caseworker/advice/forms.py
Tllew: I've made some comments on slack around this, I think we should split it to two forms too, it'll just be easier to understand.

ideally if the transform of the data can be done in the higher level, then a lot of the repeated code becomes redundant anyway.

caseworker/advice/services.py
seijihg: no, everything is happening on FE. if form has `text = data["refusal_note"]` we send this otherwise we send `text = data["refusal_reasons"]`. they both are using text field in BE. `is_refusal_note` is just a boolean to know it is a `Advice` with `refusal_note`

caseworker/advice/views.py
```diff
@@ -544,11 +545,18 @@ def is_advice_approve_only(self):
     def get_form(self):
         form_kwargs = self.get_form_kwargs()
 
-        if self.kwargs.get("advice_type") == "refuse":
+        if (
```
Tllew, line 548: if (self.kwargs.get("advice_type") == "refuse"):
            denial_reasons = get_denial_reasons(self.request)

            if self.caseworker["team"]["alias"] == services.LICENSING_UNIT_TEAM:
                return forms.LUConsolidateRefusalForm(denial_reasons=denial_reasons, **form_kwargs)
            return forms.RefusalAdviceForm(denial_reasons=denial_reasons, **form_kwargs)

caseworker/advice/services.py
```diff
@@ -380,11 +389,21 @@ def update_advice(request, case, caseworker, advice_type, data, level):
             for advice in licenceable_products_advice
         ]
     elif advice_type == AdviceType.REFUSE:
+        # If refusal_note exists then we will convert the old Advice to refusal note from refusal reason
+        # ATM this function is only being used by LU final-level for updating.
+        # OGDs use post_refusal_advice function for both Create and Update
+        # So essentially we dont need this part of the code since it will always be converted to note from now on.
+        # Just putting here for the future compatibility in case we get refusal_reasons from form. This would make sure things work.
+        is_refusal_note = False
```
currycoder, line 397: I don't feel like this function should be responsible for data transformation.

Right now we have this issue where we are changing the forms in use (which validate data submitted to the frontend) and we are passing that validated data straight through to this `update_advice()` function and `post_refusal_advice()` function above.  The problem with that is that it makes this function tightly coupled to the form; change the form in some way and you *must* change this function.

I feel like it might be better to transform the data before it gets here; that way it hopefully becomes clearer what we are sending to the API and the contract for this function becomes clearer as well (without the need for big caveat comments about calling this from only a certain part of the code)

caseworker/advice/views.py
seijihg: moved choices to view

caseworker/advice/services.py
seijihg: moved refusal note transformation outside of the methods

caseworker/advice/views.py
Tllew: the if statement structure can still be simplified to the above

Tllew (approved):
One comment but otherwise looks good

currycoder (approved):

caseworker/advice/views.py
seijihg: done updated

PR #1418, opened 2023-07-31
LTD-4042: Consider NLR products when LU edits their recommendation

saruniitr:
### Aim

Currently when LU edits their recommendation (final-level) then we are updating all products as either approve/refuse depending on their decision. This includes NLR products also which means if application has any NLR products they will also be recorded as approve/refuse after they edit recommendation.

This is a problem as it is final level advice because they are no longer treated as NLR and they will be added to the licence document also.

Fix this by filtering NLR products and recording the type as NLR correctly for all NLR products. We have existing tests so update them to include NLR products.

[LTD-4042](https://uktrade.atlassian.net/browse/LTD-4042)

[LTD-4042]: https://uktrade.atlassian.net/browse/LTD-4042?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/caseworker/advice/services.py
+++ b/caseworker/advice/services.py
@@ -4,6 +4,7 @@ from requests.exceptions import HTTPError
 
 from core import client
 from caseworker.advice import constants
+from caseworker.advice.constants import AdviceType
 
 # Queues
 BEIS_CHEMICAL_CASES_TO_REVIEW = "BEIS_CHEMICAL_CASES_TO_REVIEW"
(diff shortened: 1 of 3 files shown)


caseworker/advice/services.py
```diff
@@ -360,6 +360,8 @@ def update_advice(request, case, caseworker, advice_type, data, level):
 
     team_advice = filter_advice_by_level(case.advice, ["final"])
     consolidated_advice = filter_advice_by_team(team_advice, user_team_alias)
+    licenceable_products_advice = [item for item in consolidated_advice if item["type"]["key"] != "no_licence_required"]
+    nlr_products_advice = [item for item in consolidated_advice if item["type"]["key"] == "no_licence_required"]
```
kevincarrogan, line 364: Is there an enum or constant that could be used instead of `no_licence_required`?

unit_tests/caseworker/advice/views/test_consolidate_edit.py
```diff
@@ -254,15 +254,24 @@ def test_edit_advice_get(
...
     authorized_client,
     requests_mock,
     data_standard_case,
     url,
     consolidated_advice,
 ):
+    if include_nlr:
+        # Mark last item as NLR
+        consolidated_advice[-1]["type"] = {"key": "no_licence_required", "value": "No licence required"}
+
```
kevincarrogan, line 274: Instead of if statements in tests could this be better presented as two distinct tests?

saruniitr: I wanted to avoid duplication as the test will be the same as this one with a mix of licenceable and NLR products

kevincarrogan: I kind of feel that this is just much harder to read and understand what the test is doing. I think duplication between tests is fine if it makes it clear what's happening.

currycoder (approved):
Approved pending Kevin's comments

unit_tests/caseworker/advice/views/test_consolidate_edit.py
saruniitr: separated the test that includes NLR products

kevincarrogan (approved):

PR #1422, opened 2023-08-01
Update RegisterNameForm page title

hnryjmes:
### Aim


kevincarrogan:
I feel that there's a lot of cases where the following:

```
    def get_context_data(self, *args, **kwargs):
        context = super().get_context_data(*args, **kwargs)
        context["form_title"] = self.form_class.Layout.TITLE
        return context
```

Could be put in a higher up base class and it would only need to be added once instead of added in multiple different views.

unit_tests/exporter/goods/views/test_single_form_views.py
```diff
@@ -467,3 +520,61 @@ def test_good_component_view(authorized_client, requests_mock, good_pk):
...
+    def mock_get_category_type(self):
+        return "group3_software"
+
+    def test_good_software_technology_view_get_form_title(self, instance):
+        instance.get_category_type = self.mock_get_category_type
+
+        expected_title = CreateGoodForm.TechnologySoftware.TITLE + get_category_display_string("group3_software")
+        actual_title = instance.get_form_title()
+
+        assert expected_title == actual_title
```
kevincarrogan, line 580: We don't tend to use class based tests.

Is there a particular reason why this is a class based test?

exporter/goods/forms/goods.py
```diff
@@ -1141,6 +1140,8 @@ def clean(self):
 
 
 class FirearmsActConfirmationForm(forms.Form):
+    rfd_form_title = "Is the product covered by section 5 of the Firearms Act 1968?"
+    non_rfd_form_title = CreateGoodForm.FirearmGood.FirearmsActCertificate.TITLE
```
kevincarrogan, line 1144: I think these constants should be upper cased.

unit_tests/exporter/goods/views/test_single_form_views.py
```diff
@@ -467,3 +520,61 @@ def test_good_component_view(authorized_client, requests_mock, good_pk):
...
+        return_value="Goods Starting Point",
+    ):
+        response = authorized_client.get(url)
+    soup = BeautifulSoup(response.content, "html.parser")
+    assert soup.title.string.strip() == "Where will the products begin their export journey? - LITE - GOV.UK"
+
+
+class TestGoodSoftwareTechnologyView:
+    """
+    This is needed to keep codecov happy
```
seijihg, line 564: this comment is just so funny =))

exporter/goods/views.py
```diff
@@ -599,10 +647,16 @@ class EditFirearmActDetailsView(LoginRequiredMixin, GoodCommonMixin, FormView):
...
+        This determines which title to show in both
+        the form itself and the html title
+        """
+        return has_valid_rfd_certificate(self.application)
+
     def get_form_kwargs(self):
         kwargs = super().get_form_kwargs()
-        kwargs["is_rfd"] = has_valid_rfd_certificate(self.application)
-
+        kwargs["is_rfd"] = self.is_rfd()
```
saruniitr, line 659: is there any reason for this change?

saruniitr: is it to make the condition below readable?

saruniitr (approved):
With Kevin's comments addressed it looks good to me

unit_tests/exporter/goods/views/test_single_form_views.py
hnryjmes: it's one of those codecov things. I moved some logic out into a function so it could be reused in two places thereby ensuring consistent behaviour between two titles

but in doing that I no longer got coverage for the lines, codecov was complaining about the new lines even though they were old lines that I moved

kevincarrogan: But codecov shouldn't be related to this being a class based test.

exporter/goods/forms/goods.py
hnryjmes: updated

exporter/goods/views.py
hnryjmes: you're right, I initially thought this is needed to make it reusable, but it's no different to just using that function twice. have reverted back that change

unit_tests/exporter/goods/views/test_single_form_views.py
hnryjmes: I've rewritten this test now to do the same thing but in a single function

hnryjmes: probably it was a class-based test because I was copying code without thinking about it enough

kevincarrogan (approved):

PR #1433, opened 2023-08-07
 LTD-3971-select-one-inform-letter-template

depsiatwal:
New form to allow an inform letter to be generated


caseworker/advice/templates/advice/decision_documents.html
```diff
@@ -39,7 +39,7 @@ <h2 class="govuk-heading-m app-bg-colour--light-grey govuk-!-padding-2">
 			</strong>
 		  </td>
 		  <td class="govuk-table__cell">
-		    <a href="{% url 'cases:finalise_document_template' queue_id case.id decision %}" class="govuk-button" data-module="govuk-button" id="generate-document-{{ decision }}">
+		    <a href={% if data.value == 'Inform letter' %} "{% url 'cases:finalisation_letters_select_inform_template' queue_id case.id %}" {% else %}  "{% url 'cases:finalise_document_template' queue_id case.id decision %}" {% endif %} class="govuk-button" data-module="govuk-button" id="generate-document-{{ decision }}">
```
kevincarrogan, line 42: I'm presuming the plan is to eventually have these all run off of the same view instead of this `if`?

caseworker/cases/urls.py
```diff
@@ -128,4 +129,15 @@
...
+    path(
+        "letters/select-inform-template/",
+        SelectInformTemplate.as_view(),
+        name="finalisation_letters_select_inform_template",
+    ),
+    path(
+        "letters/select-edit-text/<uuid:paragraph_id>/",
+        EditLetterText.as_view(),
+        name="inform_edit_text",
+    ),
```
kevincarrogan, line 142: Could we try to keep the naming of the url, the view and the url name consistent where possible?

caseworker/cases/views/finalisation/forms.py
```diff
@@ -0,0 +1,57 @@
...
+        self.fields["select_template"].choices = inform_pargraphs
+
+        self.helper = FormHelper()
+        self.helper.layout = Layout(
+            "select_template",
+            Submit("submit", "Continue"),
+        )
+
+    def get_layout_fields(self):
+        return "select_template"
```
kevincarrogan, line 32: Is this actually being called? I don't think a string is the right return value.

caseworker/cases/views/finalisation/forms.py
```diff
@@ -0,0 +1,57 @@
...
+    text = forms.CharField(
+        widget=forms.Textarea(attrs={"rows": "30"}),
+        error_messages={"required": "Edit text is Required"},
+        label="Add a case note",
+    )
+
+    def __init__(self, *args, **kwargs):
+        text = kwargs.pop("text")
+        super().__init__(*args, **kwargs)
+        self.fields["text"].initial = text
```
kevincarrogan, line 48: Why are we setting initial like this? `initial` is a regular `kwargs` for a form so why are we setting this separately?

caseworker/cases/views/finalisation/letters.py
```diff
@@ -0,0 +1,100 @@
...
+    def letter_picklist_to_choices(self, picklist):
+        picklist_choices = []
+        for value in picklist:
+            picklist_choices.append([value["id"], value["name"]])
+        return picklist_choices
+
+    def get_inform_letter_template_id(self, templates):
+        for t in templates:
+            if t["name"] == "Inform letter":
+                return t["id"]
```
kevincarrogan, line 33: This is just going to return `None` if it doesn't find it. Could this be an exception instead so it's more obvious when it fails?

caseworker/cases/views/finalisation/letters.py
```diff
@@ -0,0 +1,100 @@
...
+
+    def letter_picklist_to_choices(self, picklist):
+        picklist_choices = []
+        for value in picklist:
+            picklist_choices.append([value["id"], value["name"]])
+        return picklist_choices
+
+    def get_inform_letter_template_id(self, templates):
+        for t in templates:
+            if t["name"] == "Inform letter":
```
kevincarrogan, line 32: Checking for the name like this as a hardcoded value feels a bit brittle. Is there another way of achieving this?

caseworker/cases/views/finalisation/letters.py
```diff
@@ -0,0 +1,100 @@
...
+
+
+class EditLetterText(LoginRequiredMixin, FormView):
+    template_name = "case/edit-letter.html"
+    form_class = LetterEditTextForm
+
+    def get_picklist_text(self, id, picklist):
+        for x in picklist:
+            if x["id"] == id:
+                return x["text"]
```
kevincarrogan, line 62: Maybe a `KeyError` could be thrown if this doesn't find anything?

caseworker/cases/views/finalisation/letters.py
```diff
@@ -0,0 +1,100 @@
...
+)
+
+
+TEXT = "text"
+
+
+class SelectInformTemplate(LoginRequiredMixin, FormView):
+    template_name = "core/form.html"
+    form_class = SelectInformLetterTemplateForm
+    success_url = ""
```
kevincarrogan, line 22: This doesn't need to be set as you override `get_success_url`.

caseworker/cases/views/finalisation/letters.py
```diff
@@ -0,0 +1,100 @@
...
+
+    def get_picklist_text(self, id, picklist):
+        for x in picklist:
+            if x["id"] == id:
+                return x["text"]
+
+    def get_inform_letter_template_id(self, templates):
+        for t in templates:
+            if t["name"] == "Inform letter":
+                return t["id"]
```
kevincarrogan, line 67: I feel there's a bit of overlap in this view and the `SelectInformTemplate` view. Could some of this be extracted out and reused?

caseworker/cases/views/finalisation/letters.py
```diff
@@ -0,0 +1,100 @@
...
+            self.request, case_id, template=self.template_id, text=quote(text), addressee=addressee
+        )
+
+        if status_code == 400:
+            return generate_document_error_page()
+
+        return render(
+            self.request,
+            "generated-documents/preview.html",
+            {"preview": preview["preview"], TEXT: text, "addressee": addressee, "kwargs": self.kwargs},
```
kevincarrogan, line 99: Why is `TEXT` a constant here? I don't see it reused anywhere else.

caseworker/templates/case/edit-letter.html
```diff
@@ -0,0 +1,25 @@
...
+
+{% block two_thirds %}
+	{% if errors %}
+		{% include "forms-errors.html" %}
+	{% endif %}
+
+	<form action="{{ form_action_url }}" method="post" enctype="multipart/form-data">
+		{% csrf_token %}
+
+		{% error_summary form %}
```
kevincarrogan, line 17: There's `error_summary` here and `{% include "forms-errors.html" %}` above. Aren't they doing the same thing?

caseworker/templates/case/edit-letter.html
```diff
@@ -0,0 +1,25 @@
...
+{% block back_link %}
+	<a href="{{ back_link_url }}" id="back-link" class="govuk-back-link">{{ back_link_text|default:"Back" }}</a>
+{% endblock %}
+
+{% block two_thirds %}
+	{% if errors %}
+		{% include "forms-errors.html" %}
+	{% endif %}
+
+	<form action="{{ form_action_url }}" method="post" enctype="multipart/form-data">
```
kevincarrogan, line 14: You don't need the `enctype` here.

caseworker/templates/generated-documents/new/select_template.html
```diff
@@ -0,0 +1,21 @@
...
+		<div class="lite-app-bar__content">
+			<h1 class="govuk-heading-l">{% block title %}Select Template{% endblock %}</h1>
+		</div>
+		<div class="lite-app-bar__controls">
+		</div>
+		<div class="govuk-grid-column-full" style="text-align: right;">
+			<form action="" method="POST">
+				{% csrf_token %}
+				<input class="govuk-button" type="submit" value="Select">
+			</form>
```
kevincarrogan, line 15: What's this form doing at the moment? Could this be a Django form?

caseworker/templates/generated-documents/new/select_template.html
```diff
@@ -0,0 +1,21 @@
...
+		</div>
+	</div>
+	<div class="app-letter-preview__page" id="preview">
+	</div>
```
kevincarrogan, line 19: Should there be something here for the preview?

caseworker/templates/generated-documents/new/select_template.html
```diff
@@ -0,0 +1,21 @@
...
+
+{% block body %}
+<div class="app-letter-preview__container">
+	<div class="lite-app-bar">
+		<div class="lite-app-bar__content">
+			<h1 class="govuk-heading-l">{% block title %}Select Template{% endblock %}</h1>
+		</div>
+		<div class="lite-app-bar__controls">
+		</div>
+		<div class="govuk-grid-column-full" style="text-align: right;">
```
kevincarrogan, line 11: Don't use `style`.

caseworker/templates/teams/markdown-help-details.html
kevincarrogan, line 1: Isn't there already a template with all of this in already?

caseworker/cases/views/finalisation/forms.py
Tllew: text comes from the kwargs, I don't think it's available before the init

caseworker/advice/templates/advice/decision_documents.html
currycoder: As mentioned over in the lite-api PR, the show/hide of this whole table should probably not be based on whether there is an "inform letter" decision document; https://github.com/uktrade/lite-api/pull/1532#discussion_r1294483764

Looks to me like a feature flag is unavoidable at this stage, but if we do it it should be on the frontend rather than hacked in to a backend API response.

So my suggestion is; remove the feature flag from the backend and move it over to the logic which determines whether this table is shown or not on the frontend.

caseworker/cases/urls.py
Tllew: went with:
select-inform-template
select-edit-text

felt reasonable

caseworker/templates/teams/markdown-help-details.html
Tllew: Yeah but it doesn't use the govuk-details formatting,

the other version is like a 1/3rds the sits in the right hand side of the page

this is meant to be a details button that expands to the information

caseworker/advice/templates/advice/decision_documents.html
Tllew: added a FF, this is still required as they use different urls, in the future I think we should pass the decision to select-template and have that do the work, which will be possible as it's a part of the inheritance work I've done in letters.py.

Currently they other views are in lite forms and I think that change will be our next steps

currycoder: @Tllew I'm missing it; where is the feature flag?

caseworker/cases/views/finalisation/forms.py
kevincarrogan: I mean. To set the initial value of a form you use the `initial` kwarg, you don't need a completely separate kwarg just to set the initial value for this one field.

```
form = LetterEditTextForm(initial={"text": "My initial value for text"})
```

https://docs.djangoproject.com/en/4.2/ref/forms/api/#initial-form-values

caseworker/templates/teams/markdown-help-details.html
kevincarrogan: Ok. I can see that we're using this in a slightly different format but the other template is still mostly identical. Could we extract some of it into includes so that we don't have to remember to update this intwo places at once each time?

kevincarrogan (approved):

PR #1434, opened 2023-08-07
LTD-4068: Add accessible name for the button that expands list of flags

saruniitr:
### Aim

As per recommended solution a hidden span is added to explain the purpose of the button.

[LTD-LTD-4068](https://uktrade.atlassian.net/browse/LTD-LTD-4068)


caseworker/assets/styles/components/_expander.scss
kevincarrogan, line 1: We don’t need to add this new class. There is already one called `govuk-visually-hidden` that does the same job.

currycoder (approved):

PR #1435, opened 2023-08-07
[LTD-4056] View to send generated letters

currycoder:
### Aim

This change adds a new view to the caseworker frontend which calls the new API endpoint for sending generated documents to exporters (https://github.com/uktrade/lite-api/pull/1531)

This is not yet visible/callable by lite users; it will be wired up when we are ready to release the inform letter functionality.

[LTD-4056](https://uktrade.atlassian.net/browse/LTD-4056)

[LTD-4056]: https://uktrade.atlassian.net/browse/LTD-4056?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/caseworker/cases/views/generate_document.py
+++ b/caseworker/cases/views/generate_document.py
@@ -1,9 +1,10 @@
 from http import HTTPStatus
 from urllib.parse import quote
 
+from django.contrib import messages
 from django.shortcuts import redirect, render
-from django.urls import reverse_lazy
-from django.views.generic import TemplateView
+from django.urls import reverse_lazy, reverse
+from django.views.generic import TemplateView, View
 
 from caseworker.cases.forms.generate_document import (
     select_template_form,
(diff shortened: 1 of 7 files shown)


depsiatwal (approved):
lgtm

assuming this isn't linked any where atm

caseworker/cases/views/generate_document.py
```diff
@@ -168,6 +170,22 @@ def post(self, request, queue_pk, pk, tpk):
...
+        if response.ok:
+            messages.success(
+                self.request, f"Document sent to {case['data']['organisation']['name']}, {case['reference_code']}"
+            )
+        else:
+            messages.error(
+                self.request,
+                f"An error occurred when sending the document. Please try again later",
+            )
+        return redirect(reverse_lazy("queues:cases", kwargs={"queue_pk": queue_pk}))
```
kevincarrogan, line 186: You don’t need to use reverse lazy here you could use reverse. In fact I think you can just use redirect on its own as it will take a route name directly.

kevincarrogan (approved):

stuaxo:
Making an HTTP call in a view and not async via something like Celery feels like something with the potential to block, so might be worth looking at in future.

PR #1447, opened 2023-08-10
Ltd 4061 dac selected state

kevincarrogan:
## Aim

This is the first step to having the "Select all" checkboxes on the queue page be toggled via a checkbox input as opposed to a button pretending to be a checkbox.

I'm taking this as an opportunity to reuse the component we already have that select checkboxes, although currently it assumes that there will be a button that handles this interaction whereas we want it to be a checkbox for the queue.

This initial pull request makes the SelectAll component only aware of checkboxes so that it can interact with other components when required.

### DIFF
--- a/core/assets/javascripts/select-all.js
+++ b/core/assets/javascripts/select-all.js
@@ -2,9 +2,9 @@ export const SELECT_ALL_BUTTON_TEXT = "Select all";
 export const DESELECT_ALL_BUTTON_TEXT = "Deselect all";
 
 class SelectAll {
-  constructor($selectAllButton, $checkboxes) {
-    this.$selectAllButton = $selectAllButton;
+  constructor($checkboxes, $selectAllButton) {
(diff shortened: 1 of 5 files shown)


core/assets/javascripts/select-all.js
```diff
@@ -2,35 +2,20 @@ export const SELECT_ALL_BUTTON_TEXT = "Select all";
...
 
 class SelectAll {
-  constructor($selectAllButton, $checkboxes) {
-    this.$selectAllButton = $selectAllButton;
+  constructor($checkboxes, allSelectedCallback) {
     this.$checkboxes = $checkboxes;
+    this.allSelectedCallback = allSelectedCallback;
 
-    this.isAllSelected = false;
+    this.isAllSelected = null;
```
saruniitr, line 9: next step is to use this component to replace the buttons on case summary page?

saruniitr (approved):
changes looks good to me

PR #1458, opened 2023-08-16
Ltd 4110 exporter journey

kevincarrogan:
### Aim

Add a form for exporters to be able to appeal a refusal.

This is currently behind a feature flag and just includes the smallest possible amount to produce a page and a form that will post to the API.

Subsequent tickets will further pad out all the other requirements.

[LTD-4114](https://uktrade.atlassian.net/browse/LTD-4114)


exporter/applications/rules.py
```diff
@@ -0,0 +1,11 @@
...
+
+from django.conf import settings
+
+
+@rules.predicate
+def is_appeal_feature_flag_set(request):
+    return settings.FEATURE_FLAG_APPEALS
+
+
+rules.add_rule("can_user_appeal_case", is_appeal_feature_flag_set)
```
depsiatwal, line 11: ahh nice using rules for FF :-)

exporter/applications/views/common.py
```diff
@@ -371,3 +378,63 @@ def init(self, request, **kwargs):
...
+    def get_application_url(self):
+        return reverse(
+            "applications:application",
+            kwargs={"pk": self.kwargs["case_pk"]},
+        )
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+
+        context["form_title"] = self.form_class.Layout.TITLE
```
depsiatwal, line 407: why in context ?
and not using django forms to set title ?

depsiatwal (commented):
nicely done just a small question

exporter/applications/views/common.py
kevincarrogan: There isn't a way to set the title using Django forms.

conf/exporter.py
```diff
@@ -76,6 +76,7 @@
 
 FEATURE_FLAG_ONLY_ALLOW_SIEL = env.bool("FEATURE_FLAG_ONLY_ALLOW_SIEL", True)
 FEATURE_FLAG_ALLOW_CLC_QUERY_AND_PV_GRADING = env.bool("FEATURE_FLAG_ALLOW_CLC_QUERY_AND_PV_GRADING", False)
+FEATURE_FLAG_APPEALS = env.bool("FEATURE_FLAG_APPEALS", False)
```
currycoder, line 79: Going back to our discussion around reducing the use of feature flags...  Could we do this without one and just pop it on a hidden URL?  Or do you feel that sort of approach is more dicey for the exporter application?

exporter/applications/urls.py
```diff
@@ -498,6 +498,7 @@
     path("<uuid:pk>/notes/", common.Notes.as_view(), name="notes"),
     path("<uuid:pk>/withdraw/", common.WithdrawApplication.as_view(), name="withdraw"),
     path("<uuid:pk>/surrender/", common.SurrenderApplication.as_view(), name="surrender"),
+    path("<uuid:case_pk>/appeal/", common.AppealApplication.as_view(), name="appeal"),
```
currycoder, line 501: `application_pk`?

exporter/applications/rules.py
currycoder: Yeah this is a great idea; we could do some cool things with an allow list of users etc in future

exporter/applications/views/common.py
```diff
@@ -371,3 +378,63 @@ def init(self, request, **kwargs):
...
+    )
+    def post_appeal(self, request, application_pk, data):
+        return post_appeal(request, application_pk, data)
+
+    def form_valid(self, form):
+        self.post_appeal(
+            self.request,
+            self.application["id"],
+            form.cleaned_data,
+        )
```
currycoder, line 438: Remind me, will a success message be created automatically here?

conf/exporter.py
kevincarrogan: I would prefer to put this behind a feature switch as there will be other elements in future functionality of this that will be switched on and not be able to just be hidden.

exporter/applications/urls.py
kevincarrogan: Other URLs in this file used `case_pk` so I stuck with the convention.

currycoder (approved):

saruniitr (approved):

PR #1460, opened 2023-08-18
LTD-2427 - MockSSO support for exporter.

stuaxo:
Enable exporter to use MockSSO: 
- Reduce friction for new starters
- As a side effect may run unit tests a little faster

General:
- [x] Move mock_sso to core.

Exporter:
- [x] Add conf settings for mock_sso
- [x] Configure pytest to use mock_sso

CI:
- [ ] Use self as authbroker.

[LTD-3427](https://uktrade.atlassian.net/browse/LTD-3427)


stuaxo:
Pushing the CI config change a separate commit, since pre-commit wasn't happy about that, can at least verify things pass here first.

unit_tests/core/mock_sso/test_views.py
```diff
@@ -1,8 +1,13 @@
+import pytest
 from django.conf import settings
 from django.urls import reverse
-from django.test import TestCase
 
+skip_mock_sso_tests = pytest.mark.skipif(
+    settings.MOCK_SSO_ACTIVATE_ENDPOINTS is False, reason="MOCK_SSO_ACTIVATE_ENDPOINTS is not True"
+)
```
kevincarrogan, line 7: I don't think we should skip these tests at all, these should always be run.

stuaxo: MOCK_SSO_ACTIVATE_ENDPOINTS isn't always set (it's set when we run tests for caseworker, exporter but not core).

It's definitely more dangerous to disable the tests though - I should probably either:
 - ensure the endpoint is enabled when running the core tests (there isn't an exporter or caseworker config to do it here, so I could set an environment variable)

Alternately: In the tests themselves if MOCK_SSO_ACTIVATE_ENDPOINTS is not set, I could do something like just verify the endpoint is not available.

Initially I tried having a test, where I changed the MOCK_SSO_ACTIVATE_ENDPOINTS setting and checked if the endpoints were present, this didn't work as expected which may have been due to Django already having imported urls before my tests, so changing the setting didn't have an effect.

stuaxo: Have to work out why https://app.circleci.com/pipelines/github/uktrade/lite-frontend/12245/workflows/5bad0022-a616-46e8-a86e-fa12ec078497/jobs/64281 mock_sso isn't enabled while running the core tests.

stuaxo: Bit annoying, failure that only hits on CI.

unit_tests/exporter/conftest.py
```diff
@@ -24,6 +24,13 @@ def pytest_configure(config):
     if not os.environ.get("PIPENV_DOTENV_LOCATION"):
         load_dotenv(dotenv_path=DEFAULT_ENVFILE, override=True)
 
+    # Force mock_sso django application to be activated for test environments;
+    # must be activated up front for mock_sso django app urls to be added
+    settings.MOCK_SSO_ACTIVATE_ENDPOINTS = True
+    settings.MOCK_SSO_USER_EMAIL = "test@example.net"
+    settings.MOCK_SSO_USER_FIRST_NAME = "test"
+    settings.MOCK_SSO_USER_LAST_NAME = "user"
+
```
kevincarrogan, line 33: Do we really need to set this? I'm not sure any unit tests are actually using this except for the tests that actively test the mock sso endpoint so maybe these settings could go with those tests instead of at this high level.

unit_tests/core/mock_sso/test_views.py
```diff
@@ -38,7 +47,10 @@ def test_mock_token(client):
     assert response.json() == {"access_token": "DUMMYTOKEN", "token_type": "Bearer"}
 
 
-def test_mock_api_user_me(client):
+def test_mock_api_user_me(settings, client):
+    settings.MOCK_SSO_ACTIVATE_ENDPOINTS = True
+    reload_urlconf()
```
kevincarrogan, line 52: This seems to be set for each test, could this be a good use of `@pytest.fixture(autouse=True)`?

kevincarrogan:
Have you tried running the end-to-end tests with mock sso turned on? I think on circleci we probably have the authbroker set explicitly so I'd be curious when you run the end-to-end tests with it being set to localhost.

I was expecting that we'd need to make some changes to the end-to-end test scenarios to not have to go through the current OAuth flow where it goes off to gov.uk login.

unit_tests/exporter/conftest.py
stuaxo: Good call, this was copied from the caseworker example, but I could probably take both out and mixup the relevant tests.

stuaxo: Having said that... I'm a fan of unit tests being setup so they can't hit external end points by default, and at least enabling the endpoints is one way of doing this.

I believe there are pytest plugins that can block sockets (except where wanted), that is another angle.

ci.exporter.env
```diff
@@ -5,7 +5,7 @@ ALLOWED_HOSTS=*
 PERMISSIONS_FINDER_URL=
 GTM_ID=
 FEEDBACK_URL=mailto:
-AUTHBROKER_URL=<EXPORTER_AUTHBROKER_URL>
+AUTHBROKER_URL=http://localhost:8300
```
Tllew, line 8: I think this has been dockerhost in a lot of places to avoid issues with linux

Tllew (approved):
minor comment but looks good

ci.exporter.env
stuaxo: Is this to do with `host.docker.internal` not being available by default on Linux?

If so, it is there now - but needs to explicitly adding - can you see if this PR looks OK?  I haven't tested it on non-Linux like OSX or WSL -
https://github.com/uktrade/lite-frontend/pull/1515

kevincarrogan: I think `localhost` would have been fine here. Also, what does `dockerhost` point to?

`host.docker.internal` is so that the container can talk to ports on the local machine. In some cases I've ended up doing that where we want to communicate between docker containers and for some reason doing it natively wasn't working i.e. between `api` and `frontend`.

In this particular case this variable has nothing to do with container to container communication so setting it to `localhost` should be fine.

stuaxo: Yep, localhost makes sense here since we are on the same server (in future we may want to consider moving all mock sso stuff outside the main codebase)

PR #1461, opened 2023-08-18
Ltd 4117 appeal documents

kevincarrogan:
### Aim


core/common/forms.py
```diff
@@ -40,3 +40,18 @@ def get_layout_actions(self):
...
+
+
+class MultipleFileField(forms.FileField):
+    def __init__(self, *args, **kwargs):
+        kwargs.setdefault("widget", MultipleFileInput())
+        super().__init__(*args, **kwargs)
+
+    def clean(self, data, initial=None):
+        single_file_clean = super().clean
+        result = [single_file_clean(d, initial) for d in data]
```
seijihg, line 56: this is cool.

seijihg (approved):

exporter/applications/views/common.py
```diff
@@ -430,11 +431,28 @@ def get_success_url(self):
...
+        "Error creating appeal document",
+        "Unexpected error creating appeal document",
+    )
+    def post_appeal_document(self, request, appeal_pk, data):
+        return post_appeal_document(request, appeal_pk, data)
+
     def form_valid(self, form):
-        self.post_appeal(
+        cleaned_data = form.cleaned_data.copy()
+        documents = cleaned_data.pop("documents")
```
currycoder, line 444: Is there any limit to the number of documents/size of documents an exporter can upload?

unit_tests/exporter/applications/views/test_appeal.py
```diff
@@ -118,10 +147,59 @@ def test_post_appeal(authorized_client, appeal_url, application_url, mock_post_a
...
+        data={
+            "grounds_for_appeal": "These are my grounds for appeal",
+            "documents": [
+                SimpleUploadedFile("file 1", b"File 1 contents"),
+                SimpleUploadedFile("file 2", b"File 2 contents"),
+            ],
+        },
+    )
+
+    assertContains(response, "Unexpected error creating appeal document")
```
currycoder, line 205: Should we assert a status code too?

exporter/applications/views/common.py
```diff
@@ -430,11 +431,28 @@ def get_success_url(self):
...
+
+        appeal, _ = self.post_appeal(
```
currycoder, line 446: Should we check the response status here before proceeding?

unit_tests/exporter/applications/views/test_appeal.py
kevincarrogan: `assertContains` already checks for 200.

exporter/applications/views/common.py
kevincarrogan: Nope. No limit has been given so far.

exporter/applications/views/common.py
kevincarrogan: `self.post_appeal` uses `expect_status` so already checks the status code.

currycoder: TIL, thank you!

exporter/applications/views/common.py
currycoder: OK so just to talk it through (and I'm aware this might not be a regression), but do we have anything stopping an exporter uploading a 2GB txt file and saturating a frontend/backend process?  I guess that a timeout would kick in at some point.

kevincarrogan: Yes. There are limits to the size of a file actually, I meant there wasn't a limit to the number of files that they can upload here.

I can't fully remember where the limit is set but there is one.

currycoder (approved):

PR #1464, opened 2023-08-21
Set a href value for queue menu link

kevincarrogan:
### Aim

This is to make sure that the queue link is accessible via a keyboard and a blank href stops keyboard input.

[LTD-4109](https://uktrade.atlassian.net/browse/LTD-4109)


Tllew (approved):
Looks good to me

saruniitr (approved):

PR #1468, opened 2023-08-21
Ltd 4126 appeal details

kevincarrogan:
### Aim

Add appeal details to exporter application page.

This also pulls in some changes so that we can more easily test our interactions with s3.

[LTD-4126](https://uktrade.atlassian.net/browse/LTD-4126)

[LTD-4126]: https://uktrade.atlassian.net/browse/LTD-4126?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


exporter/applications/views/documents.py
```diff
@@ -198,6 +200,23 @@ def get(self, request, case_pk, document_pk):
...
+    )
+    def get_appeal_document(self, request, appeal_pk, document_pk):
+        return get_appeal_document(request, appeal_pk, document_pk)
+
+    def get(self, request, case_pk, appeal_pk, document_pk):
+        document, _ = self.get_appeal_document(request, appeal_pk, document_pk)
+        if document["safe"]:
+            return download_document_from_s3(document["s3_key"], document["name"])
+        else:
+            return error_page(request, strings.applications.AttachDocumentPage.DOWNLOAD_GENERIC_ERROR)
```
currycoder, line 217: Presumably we don't want to be more specific here so that we don't leak virus scanning details?

exporter/templates/applications/tabs/appeal-details.html
```diff
@@ -0,0 +1,24 @@
...
+        <div class="govuk-body">
+            {{ application.appeal.grounds_for_appeal|linebreaks }}
+        </div>
+
+        {% if application.appeal.documents %}
+            <h3 class="govuk-heading-m">Documents</h2>
+            <ul class="app-appeal-details-files-list">
+                {% for document in application.appeal.documents %}
+                    <li class="app-appeal-details-files-list__item">
+                        {% if document.safe %}
```
currycoder, line 14: Minor thought for the future...  Maybe it could be better to have a `document.is_downloadable` property generated in the API rather than checking specifically for `safe==True` in lots of different places.

currycoder (approved):

exporter/applications/views/documents.py
kevincarrogan: From an end user perspective you probably won't ever get here because we don't display an actual link if the file has a virus so I think a generic error if in fact they do somehow work out the link is probably sufficient.

seijihg (approved):

PR #1470, opened 2023-08-22
LTD-4115: Add option to start appeal if an application is refused

saruniitr:
### Aim

Exporters can appeal a refusal decision within 28 days.
Provide an option in exporter dashboard to start the appeal process.

[LTD-4115](https://uktrade.atlassian.net/browse/LTD-4115)

### DIFF
--- a/exporter/applications/rules.py
+++ b/exporter/applications/rules.py
@@ -7,16 +7,38 @@ from django.conf import settings
 
 @rules.predicate
 def is_appeal_feature_flag_set(request, application):
-    if not settings.FEATURE_FLAG_APPEALS:
+    return settings.FEATURE_FLAG_APPEALS
+
+
+@rules.predicate
+def is_application_finalised(request, application):
+    if not application:
         return False
 
-    if application and (application.licence or not application.appeal_deadline):
+    return application.status == "finalised"
+
+
+@rules.predicate
+def is_application_refused(request, application):
+    if not application:
         return False
 
-    appeal_deadline_date_str = application.appeal_deadline.split("T")[0]
-    appeal_deadline = datetime.strptime(appeal_deadline_date_str, "%Y-%m-%d")
-
-    return datetime.today().date() <= appeal_deadline.date()
+    return not application.licence
 
 
-rules.add_rule("can_user_appeal_case", is_appeal_feature_flag_set)
+@rules.predicate
+def appeal_within_deadline(request, application):
+    if not application:
+        return False
+
+    if not application.appeal_deadline:
+        return False
+
+    appeal_deadline = datetime.fromisoformat(application.appeal_deadline)
+    return appeal_deadline.date() >= datetime.today().date()
+
+
+rules.add_rule(
+    "can_user_appeal_case",
+    is_appeal_feature_flag_set & is_application_finalised & is_application_refused & appeal_within_deadline,
+)
--- a/exporter/applications/rules.py
+++ b/exporter/applications/rules.py
@@ -1,11 +1,22 @@
 import rules
 
+from datetime import datetime
+
 from django.conf import settings
 
 
 @rules.predicate
-def is_appeal_feature_flag_set(request):
-    return settings.FEATURE_FLAG_APPEALS
+def is_appeal_feature_flag_set(request, application):
+    if not settings.FEATURE_FLAG_APPEALS:
+        return False
+
+    if application and (application.licence or not application.appeal_deadline):
+        return False
+
+    appeal_deadline_date_str = application.appeal_deadline.split("T")[0]
+    appeal_deadline = datetime.strptime(appeal_deadline_date_str, "%Y-%m-%d")
+
+    return datetime.today().date() <= appeal_deadline.date()
 
 
 rules.add_rule("can_user_appeal_case", is_appeal_feature_flag_set)
--- a/exporter/templates/applications/application.html
+++ b/exporter/templates/applications/application.html
@@ -39,7 +39,7 @@
 						</a>
 					{% endif %}
 					{# Using licence_duration to check that there is a licence to surrender #}
-					{% if application.licence.status.key != "surrendered" %}
+					{% if application.licence and application.licence.status.key != "surrendered" %}
 						<a href="{% url 'applications:surrender' application.id %}" id="button-surrender-application" class="govuk-button govuk-button--secondary">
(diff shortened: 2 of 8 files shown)


exporter/templates/applications/application.html
```diff
@@ -39,11 +39,16 @@ <h1 class="govuk-heading-l">
 						</a>
 					{% endif %}
 					{# Using licence_duration to check that there is a licence to surrender #}
-					{% if application.licence.status.key != "surrendered" %}
+					{% if application.licence and application.licence.status.key != "surrendered" %}
```
saruniitr, line 42: Currently we are showing "surrender licence" option even for refused applications which is not a valid option.

exporter/templates/applications/application.html
```diff
@@ -39,11 +39,16 @@ <h1 class="govuk-heading-l">
...
 						<a href="{% url 'applications:surrender' application.id %}" id="button-surrender-application" class="govuk-button govuk-button--secondary">
 							{% lcs 'applications.ApplicationSummaryPage.Buttons.SURRENDER_APPLICATION_BUTTON' %}
 						</a>
 					{% endif %}
+					{% if not application.licence and application.appeal_deadline %}
+						<a href="{% url 'applications:appeal' application.id %}" id="button-appeal-refusal" class="govuk-button govuk-button--secondary">
+							Appeal refusal decision
+						</a>
+					{% endif %}
```
kevincarrogan, line 51: I've setup a Django rules rule here for this so we can use that instead of writing the logic in the template directly.

https://github.com/uktrade/lite-frontend/blob/dev/exporter/applications/rules.py

exporter/templates/applications/application.html
```diff
@@ -44,7 +46,8 @@ <h1 class="govuk-heading-l">
 							{% lcs 'applications.ApplicationSummaryPage.Buttons.SURRENDER_APPLICATION_BUTTON' %}
 						</a>
 					{% endif %}
-					{% if not application.licence and application.appeal_deadline %}
+					{% test_rule 'can_user_appeal_case' request as can_user_appeal_case %}
+					{% if can_user_appeal_case and not application.licence and application.appeal_deadline %}
```
kevincarrogan, line 50: Sorry. I didn’t make this clear in my previous review. 

The logic here should be transferred into the rule itself. 

So `can_user_appeal_case` should check that there isn’t a lindens and there’s a deadline (in fact it should check whether todays date is within the deadline).

saruniitr: ok, that makes sense, will update

exporter/applications/rules.py
```diff
@@ -1,11 +1,22 @@
...
 
+from datetime import datetime
+
 from django.conf import settings
 
 
 @rules.predicate
-def is_appeal_feature_flag_set(request):
-    return settings.FEATURE_FLAG_APPEALS
+def is_appeal_feature_flag_set(request, application):
```
kevincarrogan, line 9: This function (in fact it's a predicate in `rules` terms) is called `is_appeal_feature_flag_set` but now has additional logic that isn't just feature flags.

Having looked at how `rules` works I think this should probably be split out into separate predicates and then combined later in the `add_rule` statement, so something like:

```
rules.add_rule("can_user_appeal_case", is_appeal_feature_flag_set & application_has_licence & appeal_within_deadline)
```
```

File: exporter/applications/rules.py [hunk 5]
```diff
@@ -1,11 +1,22 @@
...
+    if not settings.FEATURE_FLAG_APPEALS:
+        return False
+
+    if application and (application.licence or not application.appeal_deadline):
+        return False
+
+    appeal_deadline_date_str = application.appeal_deadline.split("T")[0]
+    appeal_deadline = datetime.strptime(appeal_deadline_date_str, "%Y-%m-%d")
```
kevincarrogan, line 17: If we get this in isoformat can't we read this in as a datetime more easily than doing our own string manipulation and then convert it into a date?

unit_tests/exporter/applications/views/test_application_detail.py
```diff
@@ -19,3 +19,27 @@ def test_edit_button(authorized_client, data_standard_case, mock_application_get
...
+def test_appeal_refusal_decision_button(
+    authorized_client, settings, data_standard_case, mock_refused_application_get, feature_appeals_status
+):
+    settings.FEATURE_FLAG_APPEALS = feature_appeals_status
+    pk = data_standard_case["case"]["id"]
+
+    application_url = reverse("applications:application", kwargs={"pk": pk})
+    response = authorized_client.get(application_url)
+    soup = BeautifulSoup(response.content, "html.parser")
+    assert not (not soup.find(id="button-appeal-refusal")) is feature_appeals_status
```
kevincarrogan, line 34: This double negative is hurting my head. Is there a clearer way of expressing this?

saruniitr: this is done to have the expected value as a parameter.
Double negative to convert a non-boolean to a boolean.

exporter/applications/rules.py
saruniitr: As we only need date it is split like this.

unit_tests/exporter/applications/views/test_application_detail.py
kevincarrogan: `bool` exists for that purpose or we can just use `==` and it will be coerced to a bool anyway I think.

exporter/applications/rules.py
kevincarrogan: But can't we do something like `datetime.fromisoformat(application.appeal_deadline).date()` instead of us doing our own string splitting?

unit_tests/exporter/applications/views/test_application_detail.py
saruniitr: No, if element exists then `==` is not resulting into a bool.
One option is to use `if-else` which I am trying to avoid.

kevincarrogan: The alternative is:

```
@pytest.mark.parametrize(
    "feature_appeals_status, should_button_display",
    [
        (True, False),
        (False, True),
    ],
)
```

kevincarrogan: Ok. Why does `bool(soup.find(id="button-appeal-refusal"))` not work then?

kevincarrogan (approved):

Tllew (approved):

PR #1473, opened 2023-08-23
 LTD-4132-remove-inform-on-finialization-screen 

depsiatwal:
https://uktrade.atlassian.net/browse/LTD-4129

This is to hide the inform document from the finalisation screen.  Can't be feature flagged since we are looking to migrate documents away from this screen long term.

### DIFF
--- a/unit_tests/caseworker/cases/views/test_generate_document.py
+++ b/unit_tests/caseworker/cases/views/test_generate_document.py
@@ -170,13 +170,17 @@ def test_finalise_document_create(
     assert response.url == redirect_url
 
 
-
-
 @pytest.mark.parametrize(
     "document_data, expected",
     (
-        ({'documents': {'refuse': {'value': 'Refuse'}, 'inform': {'value': 'Inform'}}}, {'refuse': {'value': 'Refuse'}}),
-        ({'documents': {'refuse': {'value': 'Refuse'}, 'nla': {'value': 'nla'}}}, {'refuse': {'value': 'Refuse'}, 'nla': {'value': 'nla'}}),
+        (
+            {"documents": {"refuse": {"value": "Refuse"}, "inform": {"value": "Inform"}}},
+            {"refuse": {"value": "Refuse"}},
+        ),
+        (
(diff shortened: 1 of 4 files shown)


unit_tests/caseworker/cases/views/test_generate_document.py
```diff
@@ -168,3 +168,37 @@ def test_finalise_document_create(
...
+            {"documents": {"refuse": {"value": "Refuse"}, "inform": {"value": "Inform"}}},
+            {"refuse": {"value": "Refuse"}},
+        ),
+        (
+            {"documents": {"refuse": {"value": "Refuse"}, "nla": {"value": "nla"}}},
+            {"refuse": {"value": "Refuse"}, "nla": {"value": "nla"}},
+        ),
+    ),
+)
+def test_finalise_documents(
```
currycoder, line 186: Do we have a test anywhere that proves that finalisation can proceed without the inform letter being created?  It would be good to double check this

depsiatwal: I don't think such journey exists on FE.
Since most of the logic is done by BE which now has a series of tests to ensure it can work with inform switched on and a refusal. Actually just added a specific test to sure finalisation is successful in this scenario.

currycoder: OK but what about the scenario where the feature flag is off?  Where's the test that proves that finalisation can still go ahead even without an inform letter?

depsiatwal: This PR is just about hiding the inform letter in finalisation screen. I guess if we need more coverage etc we can raise another jira.

Tllew (approved):
Looks good to me

unit_tests/caseworker/cases/views/test_generate_document.py
currycoder: OK, I've added https://uktrade.atlassian.net/browse/LTD-4142

PR #1475, opened 2023-08-24
Ltd-4139-label-change-on-edit-screen

depsiatwal:
This Jira was about a label change but during the flow and testing discovered that the inform letter template selection is breaking. 

This is due to the decision being changed from refuse to inform hence including this in this PR.


Tllew (approved):
Looks good to me

currycoder (approved):

PR #1480, opened 2023-08-25
 LTD-4146-wireup-recreate-inform-letter 

depsiatwal:
wire up the  recrate inform letter when a document has been crated and send to an exporter we have an option to recrate the letter. 

https://uktrade.atlassian.net/browse/LTD-4146


seijihg (approved):

currycoder:
From the ticket;
> The end effect should be that only the newly recreated inform letter exists on the case.

I see that when I recreate the inform letter on my local copy, the old inform letter remains (and is still visible to the exporter).  I think we need some clarity from UCD on whether the old inform letter should be removed when "recreate" is used - my gut feeling is that it probably should.

currycoder (commented):
Just to be clear on my comment... What I see locally is...

Create inform letter - send it - recreate inform letter;
- The original inform letter is still present and visible to exporter in "Documents" tab
- The newly recreated inform letter is present and NOT visible to exporter in "Decision Documents" table
- After sending the new document, both inform letters are present and visible to the exporter in the "Documents" tab

currycoder (approved):

stuaxo (approved):

PR #1481, opened 2023-08-25
Ltd 4121 case details appeal details

kevincarrogan:
### Aim

Add appeals detail to case details page.

[LTD-4121](https://uktrade.atlassian.net/browse/LTD-4121)

[LTD-4121]: https://uktrade.atlassian.net/browse/LTD-4121?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/caseworker/cases/views/main.py
+++ b/caseworker/cases/views/main.py
@@ -238,6 +238,7 @@ class CaseDetail(CaseTabsMixin, CaseView):
             Slices.END_USE_DETAILS,
             Slices.SUPPORTING_DOCUMENTS,
             Slices.FREEDOM_OF_INFORMATION,
+            conditional(self.case.data["appeal"], Slices.APPEAL_DETAILS),
         ]
         self.additional_context = self.get_advice_additional_context()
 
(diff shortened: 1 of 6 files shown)


caseworker/cases/views/main.py
```diff
@@ -238,6 +238,7 @@ def get_standard_application(self):
             Slices.END_USE_DETAILS,
             Slices.SUPPORTING_DOCUMENTS,
             Slices.FREEDOM_OF_INFORMATION,
+            conditional(self.case.data["appeal"], Slices.APPEAL_DETAILS),
```
currycoder, line 241: Just double checking; this won't KeyError from feature flags being off etc?

kevincarrogan: Nope. The API doesn't have a feature flag so will have this key regardless.

In fact the feature flag is only whether an exporter can make the appeal or not so the rest of the code doesn't have any knowledge of a feature switch for this functionality.

saruniitr (approved):

PR #1501, opened 2023-09-08
Ltd 4170 sub status UI

kevincarrogan:
### Aim

Add a view to allow sub-statuses to be edited.

At the moment there is no link to this view, I'm thinking that we should add this link once we add the sub-status onto the product details page.

[LTD-4170](https://uktrade.atlassian.net/browse/LTD-4170)

[LTD-4170]: https://uktrade.atlassian.net/browse/LTD-4170?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/caseworker/cases/services.py
+++ b/caseworker/cases/services.py
@@ -439,3 +439,19 @@ def post_compliance_person_present(request, case_id, json):
 
         return {**json, "errors": translated_errors}, data.status_code
     return data.json(), data.status_code
+
+
+def get_case_sub_statuses(request, case_id):
+    response = client.get(request, f"/applications/{case_id}/sub-statuses/")
+    response.raise_for_status()
+    return response.json()
+
+
+def put_case_sub_status(request, case_id, data):
+    response = client.put(
+        request,
+        f"/applications/{case_id}/sub-status/",
+        data=data,
+    )
+    response.raise_for_status()
+    return response.json(), response.status_code
--- a/caseworker/cases/views/main.py
+++ b/caseworker/cases/views/main.py
@@ -5,6 +5,7 @@ from http import HTTPStatus
 
 from django.conf import settings
 from django.contrib import messages
+from django.contrib.messages.views import SuccessMessageMixin
 from django.http import Http404
 from django.shortcuts import redirect
 from django.urls import reverse, reverse_lazy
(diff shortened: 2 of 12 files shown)


caseworker/cases/services.py
```diff
@@ -439,3 +439,19 @@ def post_compliance_person_present(request, case_id, json):
 
         return {**json, "errors": translated_errors}, data.status_code
     return data.json(), data.status_code
+
+
+def get_case_sub_statuses(request, case_id):
+    response = client.get(request, f"/applications/{case_id}/sub-statuses/")
```
depsiatwal, line 445: Does this end point exist is it part of another PR ?

caseworker/cases/services.py
```diff
@@ -439,3 +439,19 @@ def post_compliance_person_present(request, case_id, json):
...
+    response.raise_for_status()
+    return response.json()
+
+
+def put_case_sub_status(request, case_id, data):
+    response = client.put(
+        request,
+        f"/applications/{case_id}/sub-status/",
```
depsiatwal, line 453: same as above

caseworker/templates/case/form.html
```diff
@@ -0,0 +1,11 @@
+{% extends 'layouts/case.html' %}
```
depsiatwal, line 1: do we need another form or can we just use core form ?

depsiatwal (commented):
lgtm just some minor comments more for my info

caseworker/cases/services.py
kevincarrogan: It exists. It's on the API side so can't be a part of this PR.

https://github.com/uktrade/lite-api/pull/1578

It's already merged in.

caseworker/templates/case/form.html
kevincarrogan: Core form doesn't have the case navigation header. This uses an include to make sure that we aren't duplicating the form display logic though.

caseworker/cases/views/main.py
```diff
@@ -392,6 +401,75 @@ def get_success_url(self):
...
+    def form_valid(self, form):
+        self.put_case_sub_status(
+            self.request,
+            self.case.id,
+            form.cleaned_data,
+        )
+        return super().form_valid(form)
+
+    def get_success_url(self):
+        return reverse(
```
currycoder, line 463: Is a success message injected in some way?  I think I always forget this..

currycoder (approved):

caseworker/cases/views/main.py
kevincarrogan: Not automatically. I did forget that in this case, I will add it in.

depsiatwal (approved):
lgtm

PR #1505, opened 2023-09-13
LTD-4173-show-sub-status

seijihg:
https://uktrade.atlassian.net/browse/LTD-4173


depsiatwal (commented):
probably worth adding a test using BS to show status/sub status is being displayed in the template

Tllew (commented):
Looks good to me, will agree with Gurdeep on beautiful soup test if possible

depsiatwal (approved):
lgtm

PR #1509, opened 2023-09-14
 LTD-4196-add-sub-status-link 

depsiatwal:
https://uktrade.atlassian.net/browse/LTD-4196

### DIFF
--- a/caseworker/templates/case/slices/summary.html
+++ b/caseworker/templates/case/slices/summary.html
@@ -117,8 +117,7 @@
 		</dd>
 			{% if case_sub_statuses_options %}
 				<dd class="app-case__summary-list__actions">
-					<a id="link-case-sub-status-change" class="govuk-link govuk-link--no-visited-state" href="{% url 'cases:change_sub_status' queue.id case.id  %}">
-						Change
+					<a id="link-case-sub-status-change" class="govuk-link govuk-link--no-visited-state" href="{% url 'cases:change_sub_status' queue.id case.id  %}">Change</a>
 				</dd>
 			{% endif %}
 	</div>
--- a/caseworker/templates/case/slices/summary.html
+++ b/caseworker/templates/case/slices/summary.html
@@ -115,12 +115,12 @@
 				<span class="govuk-hint govuk-!-margin-0">No sub-status set</span>
 			{% endif %}
 		</dd>
-		{% if case.data.sub_status %}
-			<dd class="app-case__summary-list__actions">
-				<a id="link-case-sub-status-change" class="govuk-link govuk-link--no-visited-state" href="{% url 'cases:change_sub_status' queue.id case.id  %}">
-					Change
-			</dd>
-		{% endif %}
+			{% if case_sub_statuses_options %}
+				<dd class="app-case__summary-list__actions">
+					<a id="link-case-sub-status-change" class="govuk-link govuk-link--no-visited-state" href="{% url 'cases:change_sub_status' queue.id case.id  %}">
+						Change
+				</dd>
+			{% endif %}
 	</div>
 	<div class="app-case__summary-list__row">
 		<dt class="app-case__summary-list__key">
--- a/caseworker/templates/case/slices/summary.html
+++ b/caseworker/templates/case/slices/summary.html
@@ -115,6 +115,12 @@
 				<span class="govuk-hint govuk-!-margin-0">No sub-status set</span>
 			{% endif %}
 		</dd>
+		{% if case.data.sub_status %}
+			<dd class="app-case__summary-list__actions">
+				<a id="link-case-sub-status-change" class="govuk-link govuk-link--no-visited-state" href="{% url 'cases:change_sub_status' queue.id case.id  %}">
+					Change
+			</dd>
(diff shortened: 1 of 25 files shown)


caseworker/templates/case/slices/summary.html
```diff
@@ -115,6 +115,12 @@
 				<span class="govuk-hint govuk-!-margin-0">No sub-status set</span>
 			{% endif %}
 		</dd>
+		{% if case.data.sub_status %}
+			<dd class="app-case__summary-list__actions">
+				<a id="link-case-sub-status-change" class="govuk-link govuk-link--no-visited-state" href="{% url 'cases:change_sub_status' queue.id case.id  %}">
+					Change
+			</dd>
+		{% endif %}
```
kevincarrogan, line 123: This if statement is probably too restrictive.

It is possible that a sub-status hasn't been set but they would like to be able to set one, this is only allowing them to change it if one is already set.

Tllew (commented):
Agreed with kevin's comment

caseworker/templates/case/slices/summary.html
depsiatwal: so do we get rid of the condition completely ?
i'e allow even when not set ?
on jira is says
"This link should only be available if there are available sub-statuses to change "  - Any ideas what would determine that ?

kevincarrogan: We should definitely allow it when not set.

> "This link should only be available if there are available sub-statuses to change " - Any ideas what would determine that ?

Yes, we know what sub-statuses are available given the parent status. As long as there are sub-statuses available that can be set then we should add this link.

This also reminds me that there are other rules that I think mean one has to be assigned to the case to change things like status, so we should also replicate that.

Tllew (approved):

unit_tests/caseworker/cases/views/test_case_details.py
```diff
@@ -244,3 +244,34 @@ def test_case_details_sub_status(
...
+):
+    data_standard_case["case"]["data"]["sub_status"] = value
+    case_url = reverse("cases:case", kwargs={"queue_pk": data_queue["id"], "pk": data_standard_case["case"]["id"]})
+    response = authorized_client.get(case_url)
+
+    html = BeautifulSoup(response.content, "html.parser")
+    dt = html.find("dt", string=re.compile("Sub-status"))
+    assert dt
+    dd = dt.find_next().find_next()
+    assert dd.get_text().replace("\n", "").replace("\t", "") == expected
```
kevincarrogan, line 277: This test hasn't been updated to work with the new update to show the "change" link depending on what is being sent back from the API.

depsiatwal: the 2nd test checks the next item DT and checks it has the text 'change' when we have a return value from api. 
the first test has no dt item so .find_next() just returns the dt displaying the sub status value. I guess it could check explictyl for the </a> link

kevincarrogan: But you’ve changed the logic away from choosing to display the link based on whether there is a sub-status or not on the case but this test is still using that. 

I would expect the test to change the returned value coming back from the new call you’ve added.

depsiatwal: updated

unit_tests/caseworker/cases/views/test_case_details.py
```diff
@@ -244,3 +245,38 @@ def test_case_details_sub_status(
...
+    "value, expected",
+    (
+        ({}, 0),
+        (
+            {
+                "id": "status-1",
+                "name": "Status 1",
+            },
+            1,
+        ),
```
kevincarrogan, line 260: The return type of `/applications/{case_id}/sub-statuses/` is a list not a dictionary.

unit_tests/caseworker/conftest.py
```diff
@@ -2542,3 +2542,16 @@ def mock_cases_search(requests_mock, data_cases_search, queue_pk):
...
+def mock_get_case_sub_statuses(data_standard_case, requests_mock):
+    case_id = data_standard_case["case"]["id"]
+    url = client._build_absolute_uri(f"/applications/{case_id}/sub-statuses/")
+    requests_mock.get(
+        url=url,
+        json=[
+            {"id": "status-1", "name": "Status 1"},
+            {"id": "status-2", "name": "Status 2"},
+        ],
+    )
```
kevincarrogan, line 2557: This fixture already exists elsewhere in `unit_tests/caseworker/cases/views/test_change_sub_status.py`. We should extract it and only have one.

unit_tests/caseworker/cases/views/test_case_details.py
depsiatwal: ahh good point i've removed as it's required in multiple places now.

unit_tests/caseworker/conftest.py
depsiatwal: ahh good point i've removed as it's required in multiple places now.

kevincarrogan (approved):

PR #1522, opened 2023-09-19
Ltd 4221 sub status permissions

kevincarrogan:
### Aim

Add rules for case sub-status link and view.

This will only allow a case sub-status to be changed if a user is allocated and if there are available sub-statuses. This wraps both of these rules in Django Rules permissions.

[LTD-4221](https://uktrade.atlassian.net/browse/LTD-4221)


currycoder (approved):

depsiatwal (approved):
lgtm

PR #1776, opened 2024-02-07
Added hcsat, posts and basic tests

Tllew:
Adding HCSAT form to the successpage

[LTD-4661](https://uktrade.atlassian.net/browse/LTD-4661)

[LTD-4661]: https://uktrade.atlassian.net/browse/LTD-4661?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


hnryjmes:
I have added screenshots of my local copy to the ticket

Tllew:
@hnryjmes 

> I didn't expect the stars to be yellow, but if they have to be yellow that's fine. would gov blue work?
- Yeah yellow doesn't look great but it's what we were given as an example
> some of the spacing was not as expected
- all looks good although the stars on the second page look a little odd, they're being removed anyway
> I would think that the questions (implemented currently as form labels) should be bigger and bolder, or maybe styled as headings like it appears in the design
- Good catch
> you can submit without filling in any of the fields, is that expected?
- Expected
> when I tried to submit the form with all data filled in, I got a bad request error "Error sending feedback" but I'm not sure why
- This is strange I'll take another look.

saruniitr (approved):

core/assets/javascripts/star-rating.js
```diff
@@ -0,0 +1,13 @@
+const initStarRating = () => {
```
currycoder, line 1: This feedback is too late for this PR, but for future reference we have a component pattern for new javascript e.g.; https://github.com/uktrade/lite-frontend/blob/dev/caseworker/assets/javascripts/list-expander.js

PR #1791, opened 2024-02-15
LTD-4682: Add UI test to ensure Case cannot be finalised if there is an open query

saruniitr:
## Change description

We recently made a change to ensure Caseworker cannot finalise a case if there is a query that is not resolved.
If this is the case we display a warning message that a query needs resolving and 'Finalise case' button is hidden.

Add UI test to ensure we display the warning message and 'Finalise case' button is not visible.

[LTD-4682](https://uktrade.atlassian.net/browse/LTD-4682)

[LTD-4682]: https://uktrade.atlassian.net/browse/LTD-4682?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ui_tests/caseworker/conftest.py
```diff
@@ -474,6 +490,23 @@ def select_template_by_name(driver, template_name):  # noqa
...
+    ecju_queries_pages = EcjuQueriesPages(driver)
+    ecju_queries_pages.enter_question_text(query)
+    context.ecju_question = query
+
+
+@then(parsers.parse('I see "{query}" as the query under open queries'))
+def query_in_open_list(driver, query):
+    assert query in EcjuQueriesPages(driver).get_open_queries_text()
+
+
```
hnryjmes, line 509: I assume this is deliberately the same as the ones in `./ui_tests/caseworker/step_defs/test_ecju_queries.py` but were moved here? do we need two sets of them?

hnryjmes (approved):
looks good, one minor comment

ui_tests/caseworker/conftest.py
saruniitr: there is only one now because it is moved to this file.
Because we are reusing the same steps in two different feature files it needs to be in the common area for all features which is `conftest.py`

depsiatwal (approved):
lgtm

PR #1800, opened 2024-02-20
moved lite_forms/generators to core/views

Tllew:
### Aim


seijihg (approved):

saruniitr (approved):

core/views.py
kevincarrogan, line 1: I feel like just moving this file is sort of missing the point of why we want to get rid of `lite_forms`.

For example, this just moves `confirm_form` wholesale when in fact this should just be replaced or removed as it's kind of bad.

The same for `error_page` it's not really consistent with how we write views and should probably be changed as a result.

Tllew: So I thought about this for a while, and came to the conclusion that in the case of error_page it's a re_render and didn't really fit the new view/warrant a separate url, we have a set of handleSTATUSCODE urls, but the case for error_page in particular is to show an error page with a set message

I just wanted to move it to a more generic location as a first step.

kevincarrogan: I guess it's the fact though that a bunch of other `lite_forms` code got dragged across with `error_page` as well which made me feel like we weren't really solving any problems, we're just moving stuff so it doesn't say `lite_forms` in the directory name.

Tllew: this can wait to be fair, the rest is more important and removing the unused lite_form is pretty useful

PR #1809, opened 2024-02-21
Ltd 4708 parallelise e2e

kevincarrogan:
### Aim

Parallelise end-to-end tests to speed up the our overall test runs.

[LTD-4708](https://uktrade.atlassian.net/browse/LTD-4708)

[LTD-4708]: https://uktrade.atlassian.net/browse/LTD-4708?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.circleci/config.yml
```diff
@@ -254,8 +270,17 @@ jobs:
...
     resource_class: large
+    parallelism: 10
+    environment:
+      PIPENV_YES: 1
     steps:
+      - browser-tools/install-chrome
+      - browser-tools/install-chromedriver
+      - run:
+            name: "Install dependencies"
+            command: pip3 install pipenv
```
currycoder, line 283: It seems like this string of steps is repeated, could they instead be put inside `prepare_e2e_tests` or some other single setup step?

Pipfile
```diff
@@ -32,13 +32,14 @@ python-dotenv = "*"
...
 django-requests-debug-toolbar = "*"
-moto = {version = "*", extras = ["s3"]}
+moto = {version = "~=4.2", extras = ["s3"]}
 requirements-detector = "==1.2.2"
 pylint = "==2.17.7"
 pylint-celery = "==0.3"
 pylint-django = "==2.5.3"
 pylint-flask = "==0.6"
 pylint-plugin-utils = "==0.7"
+pytest-circleci-parallelized = "*"
```
currycoder, line 42: Should this be more locked down?

currycoder (approved):
Looks great, just some minor questions

Pipfile
kevincarrogan: For the dev ones I tend not to mind so much, in fact I think it's a bit mad that some of them are very precisely locked down to a single specific version.

.circleci/config.yml
kevincarrogan: Yes. I can do that.

It wasn't possible for a long time whilst I was developing and I hadn't noticed I'd removed what was blocking me from doing it.

PR #1845, opened 2024-03-13
LTD-4822-enable-registration-django-forms 

depsiatwal:
This PR revitalises an old attempt to get registration converted to Django forms. 

This is an incremental update to get this functioning and add the labels which were missing. 
The functionally is behind a FF so don't impact


exporter/core/views.py
```diff
@@ -211,6 +211,8 @@ def form_valid(self, form):
         # Hold in session
         self.request.session["first_name"] = form.cleaned_data["first_name"]
         self.request.session["last_name"] = form.cleaned_data["last_name"]
+        self.request.session["email"] = profile["email"]
```
depsiatwal, line 214: This is a potential change which isn't hidden behind a Feature flag. 

I did thing about this but figured it's the first thing we do in the triage form with old flow switched on. 
So I think it's safe as long as we QA/Product review that part of the flow

exporter/core/organisation/views.py
```diff
@@ -40,10 +40,9 @@ class Registration(
 
     def get_form_kwargs(self, step=None):
         kwargs = super().get_form_kwargs(step)
-        if step == RegistrationSteps.ADDRESS_DETAILS:
-            kwargs["is_uk_based"] = self.is_uk_based
-        if step == RegistrationSteps.REGISTRATION_DETAILS:
+        if step in (RegistrationSteps.ADDRESS_DETAILS, RegistrationSteps.REGISTRATION_DETAILS):
             kwargs["is_individual"] = self.is_individual
+            kwargs["is_uk_based"] = self.is_uk_based
```
Tllew, line 45: has this logic changed?

It looks to be doing something different now?

Tllew (commented):
looks good just one question about the change in get_form_kwargs

exporter/core/organisation/views.py
depsiatwal: Yeah so both this forms require these values now, 

They both hide/show fields or change the title based on these values.

Tllew (approved):

exporter/core/organisation/forms.py
```diff
@@ -15,11 +18,26 @@ class RegistrationTypeForm(BaseForm):
     class Layout:
         TITLE = "Commercial organisation or private individual"
 
+    class RegistrationTypeChoices(models.TextChoices):
+        COMMERCIAL = "commercial", "Commercial organisation"
+        INDIVIDUAL = "individual", "Private individual"
```
currycoder, line 23: Minor, but can we make the tuples a bit more explicit here by using brackets?  I did a double take on this.

exporter/core/organisation/forms.py
```diff
@@ -15,11 +18,26 @@ class RegistrationTypeForm(BaseForm):
...
+            ),
+        ]
+        self.fields["type"].choices = type_choices
+
     type = forms.ChoiceField(
-        choices=(
-            ("commercial", "Commercial organisation"),
-            ("individual", "Private individual"),
-        ),
+        choices=(),  # updated in init
```
currycoder, line 40: Why do we need to do this work in `__init__()`?  There doesn't seem to be anything dynamic going on here so can't see why the choices can't be defined here..

exporter/core/organisation/forms.py
```diff
@@ -52,22 +70,56 @@ def get_layout_fields(self):
...
+    EORI_LABEL = "European Union registration and identification number (EORI)"
+    SIC_CODE_LABEL = "SIC Code"
+    REGISTRATION_LABEL = "Company registration number (CRN)"
+
     class Layout:
         TITLE = "Register a private individual"
 
-    def __init__(self, is_individual, *args, **kwargs):
+    def __set_all_fields_to_default__(self):
+        # This is required to rest the fields to a known state since the wizzard maintains state
```
currycoder, line 83: I think this can be avoided.  I'm not entirely sure why we are using `self.declared_fields` in this file but that is essentially a class-level variable which is why you are needing to reset state like this between Form instances.  If we instead use `self.fields` we shouldn't need to do the resetting; so the defaults can be set once in the field definitions and overridden as you are doing in the `__init__()` method.  Would need to use `self.fields` all throughout the `__init__()` method, though.

exporter/core/views.py
currycoder: So right now, this bit gets done somewhere else?  For my understanding, where is that?

exporter/core/organisation/forms.py
depsiatwal: Ahh I just tried this and I think I remember vaguely why it had been done using self.declared_fields
to use self.fields I need to call base class
super().__init__(*args, **kwargs) which calls get_layout_fields and that depends on class level variables.

I've manage to work out the correct ordering now and put in a comment so it obvious why it's done in this order.

currycoder: Having looked at `get_layout_fields()`;
```
    def get_layout_fields(self):
        if self.is_individual:
            return ("name", "eori_number", "vat_number")
        else:
            return ("name", "eori_number", "sic_number", "vat_number", "registration_number")
```

It seems that so long as you make the super call after `self.is_individual` is set in `__init__()` it should be ok?

This business of using/resetting class level variables (which should really only be used in django internals) seems like a bit of a mess and I'm struggling to understand why we *have* to do it this way.

exporter/core/organisation/forms.py
depsiatwal: good point moved

depsiatwal:
This is the link to where we set the session on RegisterAnOrganisationTriage
On the old flow once name is set it's the first thing it does as the page loads. 
https://github.com/uktrade/lite-frontend/blob/5e48131fee44c16aabd70a4866b0fa5112b09014/exporter/core/views.py#L92

exporter/core/organisation/forms.py
depsiatwal: that's correct in terms of the calls. 
sorry not sure was you suggesting further changes here ?

Basically this is an attempt to reuse the form for 4 different scenarios.

currycoder: Looks like you've dropped the use of `self.declared_fields` - that's what I was requesting 👍

exporter/core/organisation/forms.py
```diff
@@ -79,30 +82,79 @@ def __init__(self, is_individual, *args, **kwargs):
...
+            if not is_uk_based:
+                # commercial non-uk
+                self.fields["eori_number"].label = self.EORI_LABEL + " (optional)"
+                self.fields["eori_number"].required = False
+                self.fields["sic_number"].label = self.SIC_CODE_LABEL + " (optional)"
+                self.fields["sic_number"].required = False
+                self.fields["vat_number"].label = self.VAT_LABEL + " (optional)"
+                self.fields["vat_number"].required = False
+                self.fields["registration_number"].label = self.REGISTRATION_LABEL + " (optional)"
+                self.fields["registration_number"].required = False
```
currycoder, line 156: Realise we've already discussed this bit a fair amount, but the thing you pointed out about using the form in 4 different places is interesting as I think that's where the complexity of all these embedded if statements come from.

Would creating a `BaseRegisterDetailsForm` and 4 other form classes that inherited from it help?  I feel like we would then have 4 declarative forms rather than one big one that tries to account for all scenarios.

saruniitr (approved):
looks good to me.

PR #1847, opened 2024-03-15
Match API Hawk response when receiving streaming response from API

kevincarrogan:
The API uses the content disposition in case where we are sending streaming responses so we need to match that on the frontend.

[LTD-4668](https://uktrade.atlassian.net/browse/LTD-4668)

[LTD-4668]: https://uktrade.atlassian.net/browse/LTD-4668?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core/client.py
+++ b/core/client.py
@@ -65,14 +65,19 @@ def _seen_nonce(access_key_id, nonce, timestamp):
     return seen_cache_key
 
 
-def verify_hawk_response(response, sender):
+def verify_hawk_response(response, sender, stream=False):
     if "server-authorization" not in response.headers:
         sentry_sdk.set_context("response", {"content": response.content})
         raise RuntimeError("Missing server_authorization header. Probable API HAWK auth failure")
 
+    if stream:
+        content = response.headers["Content-Disposition"]
+    else:
+        content = response.content
+
     sender.accept_response(
         response.headers["server-authorization"],
-        content=response.content,
+        content=content,
         content_type=response.headers["Content-Type"],
     )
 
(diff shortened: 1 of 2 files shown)


unit_tests/core/test_api_client.py
```diff
@@ -20,3 +38,122 @@ def test_zipkin_headers(settings):
...
+    )
+    response.headers["content-type"] = "application/octet-stream"
+    response.headers["content-disposition"] = 'attachment; filename="filename.jpg"'
+    response.status_code = 200
+    response._content = b"test"
+
+    assert verify_hawk_response(response, sender, stream=True) is None
+
+
+def test_perform_request(settings, rf, client, requests_mock):
```
hnryjmes, line 117: is the `settings` variable here just an object created when this individual test is run?

core/client.py
```diff
@@ -65,14 +65,19 @@ def _seen_nonce(access_key_id, nonce, timestamp):
...
 
 
-def verify_hawk_response(response, sender):
+def verify_hawk_response(response, sender, stream=False):
     if "server-authorization" not in response.headers:
         sentry_sdk.set_context("response", {"content": response.content})
         raise RuntimeError("Missing server_authorization header. Probable API HAWK auth failure")
 
+    if stream:
+        content = response.headers["Content-Disposition"]
```
hnryjmes, line 74: I've read this file in api and now I understand what this is doing
https://github.com/uktrade/lite-api/blob/dev/api/conf/middleware.py

hnryjmes (approved):

unit_tests/core/test_api_client.py
kevincarrogan: It's a fixture https://pytest-django.readthedocs.io/en/latest/configuring_django.html#overriding-individual-settings.

currycoder (approved):

PR #1853, opened 2024-03-21
Ltd 4829 new reg country dropdown

Tllew:
New contract with data-module="autocomplete-select"

[LTD-4829](https://uktrade.atlassian.net/browse/LTD-4829)

### DIFF
--- a/exporter/core/organisation/forms.py
+++ b/exporter/core/organisation/forms.py
@@ -272,6 +272,7 @@ class RegisterAddressDetailsOverseasForm(RegisterAddressDetailsBaseForm):
             "required": "Enter a country",
         },
     )
+    # country = AutocompleteInput(name="country", title="country", options=get_countries(request, True)),
 
     def get_layout_fields(self):
         return ("name", "address", "phone_number", "website", "country")
--- a/exporter/core/organisation/forms.py
+++ b/exporter/core/organisation/forms.py
@@ -272,7 +272,6 @@ class RegisterAddressDetailsOverseasForm(RegisterAddressDetailsBaseForm):
             "required": "Enter a country",
         },
     )
-    # country = AutocompleteInput(name="country", title="country", options=get_countries(request, True)),
 
     def get_layout_fields(self):
         return ("name", "address", "phone_number", "website", "country")
--- a/exporter/organisation/sites/forms.py
+++ b/exporter/organisation/sites/forms.py
@@ -137,7 +137,10 @@ class NewSiteInternationalAddressForm(SiteFormMixin, forms.Form):
     )
     website = forms.CharField(label="Website", required=False)
     country = forms.ChoiceField(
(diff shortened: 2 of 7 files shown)


exporter/organisation/sites/forms.py
```diff
@@ -137,7 +137,10 @@ class NewSiteInternationalAddressForm(SiteFormMixin, forms.Form):
     )
     website = forms.CharField(label="Website", required=False)
     country = forms.ChoiceField(
```
depsiatwal, line 139: This is the wrong form we want to add it to the new registration flow 
https://github.com/uktrade/lite-frontend/blob/dev/exporter/core/organisation/forms.py#L269

Also would need to make sure the selected value is sent to the BE as it's currently hardcoded.

kevincarrogan (approved):

exporter/core/organisation/forms.py
```diff
@@ -273,6 +266,25 @@ def __init__(self, is_individual, *args, **kwargs):
         },
     )
 
+    country = forms.ChoiceField(
+        choices=[],
+        widget=forms.widgets.Select(attrs={"data-module": "autocomplete-select"}),
+        error_messages={
+            "required": "Enter a country",
+        },
+    )
```
kevincarrogan, line 275: The `CharField` should be deleted instead of having two declarations for the same field.

Tllew: Oops, Thought I had deleted it, thanks

depsiatwal:
Also Code needs to be update to pick up the value of the new field 
https://github.com/uktrade/lite-frontend/blob/dev/exporter/core/organisation/payloads.py#L20
Basically remove above and update tests

depsiatwal (approved):
LGTM

unit_tests/exporter/core/organisation/test_forms.py
```diff
@@ -199,47 +209,36 @@ def test_register_details_form_field_validation(data, valid, error, form_class):
...
-        "phone_number": "12334",
-        "website": "notreal.com",
-    }
-
-    form = forms.RegisterAddressDetailsUKForm(is_individual=False, data=data)
-    assert not form.is_valid()
-    assert form.errors == {
-        "phone_number": ["Invalid telephone number"],
-        "website": ["Enter a valid URL"],
-    }
```
kevincarrogan, line 240: Why are we getting rid of this test? It seems useful to still have.

kevincarrogan (approved):

unit_tests/exporter/core/organisation/test_forms.py
Tllew: test_register_**non_uk**_address_details_form

calls RegisterAddressDetailsUKForm

it's a duplicate of the above that wasn't removed when parameterisation was added or updated to be non uk

PR #1859, opened 2024-03-25
Fix ultimate end user missing from exporter application summary

hnryjmes:
### Aim

This fixes an issue where ultimate end users were not showing in the exporter application summary ("Check your answers" page). 

For some applications where an exporter has added incorporated goods, in the good on application, it is possible that `is_good_incorporated` can be `False` but `is_onward_incorporated` can be `True`.

Previously the way that ultimate end users were toggled on and off for this page was just by checking the value of `is_good_incorporated`.

We would like to check for the presence of `application["ultimate_end_users"]` but as this can be unreliable (as we don't delete this data if an application is edited) it is best to check both of `is_good_incorporated` and `is_onward_incorporated`. If either of these booleans are true then we consider that the application has incorporated goods. 

So, if the application has ultimate end users and has incorporated goods then we would like to show the ultimate end users section in the exporter summary.

[LTD-4853](https://uktrade.atlassian.net/browse/LTD-4853)

[LTD-4853]: https://uktrade.atlassian.net/browse/LTD-4853?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/exporter/applications/helpers/check_your_answers.py
+++ b/exporter/applications/helpers/check_your_answers.py
@@ -128,7 +128,7 @@ def _convert_standard_application(application, editable=False, is_summary=False)
         product_location = {"Product location and journey": _get_product_location_and_journey(application)}
         converted = {**product_location, **converted}
 
-    if has_ultimate_end_users(application):
+    if has_ultimate_end_users(application) and has_incorporated_goods(application):
         ultimate_end_users = [convert_party(item, application, editable) for item in application["ultimate_end_users"]]
         converted[strings.ULTIMATE_END_USERS] = ultimate_end_users
 
@@ -712,7 +712,7 @@ def has_ultimate_end_users(application):
 
 def has_incorporated_goods(application):
     for good in application["goods"]:
-        if good["is_good_incorporated"]:
+        if good.get("is_good_incorporated") or good.get("is_onward_incorporated"):
             return True
 
     return False
--- a/exporter/applications/helpers/check_your_answers.py
+++ b/exporter/applications/helpers/check_your_answers.py
@@ -128,7 +128,7 @@ def _convert_standard_application(application, editable=False, is_summary=False)
         product_location = {"Product location and journey": _get_product_location_and_journey(application)}
         converted = {**product_location, **converted}
 
-    if has_ultimate_end_users(application) and has_incorporated_goods(application):
+    if has_ultimate_end_users(application) and has_incorporated_goods_on_application(application):
         ultimate_end_users = [convert_party(item, application, editable) for item in application["ultimate_end_users"]]
         converted[strings.ULTIMATE_END_USERS] = ultimate_end_users
 
@@ -710,9 +710,9 @@ def has_ultimate_end_users(application):
     return bool(application["ultimate_end_users"])
 
 
-def has_incorporated_goods(application):
-    for good in application["goods"]:
-        if good.get("is_good_incorporated") or good.get("is_onward_incorporated"):
+def has_incorporated_goods_on_application(application):
+    for goa in application["goods"]:
+        if goa.get("is_good_incorporated") or goa.get("is_onward_incorporated"):
             return True
 
     return False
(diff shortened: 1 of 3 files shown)


exporter/applications/helpers/check_your_answers.py
```diff
@@ -706,6 +706,10 @@ def is_application_export_type_permanent(application):
     return False if not application.get("export_type") else (application.get("export_type").get("key") == PERMANENT)
 
 
+def has_ultimate_end_users(application):
+    return bool(application["ultimate_end_users"])
+
+
```
saruniitr, line 712: Do we understand why `is_good_incorporated` can be `False` when `is_onward_incorporated` is `True`?

In some places `is_good_incorporated` is given the same value as `is_onward_incorporated`,
eg https://github.com/uktrade/lite-frontend/blob/dev/exporter/applications/views/goods/platform/views/edit.py#L409

So is it possible this is missing in some places?

Also just using `application["ultimate_end_users"]` does not help because we have an existing bug where initially user answered 'Yes' to onward exported question so we ask for Ultimate end user details but later they edited and answered 'No' to that question. In this case ultimate end is still present (it is not decided whether we need to delete that automatically or not iirc).

In the above scenario we will still show ultimate end user which is not consistent.

hnryjmes:
@kevincarrogan 

> Could we get a test for this please?

I've added a BeautifulSoup test as I think testing the view is more like how we normally write tests

an alternative is to add a test here which is where tests have been written for the `check_your_answers` module: https://github.com/uktrade/lite-frontend/blob/dev/unit_tests/exporter/applications/helpers/test_check_your_answers.py

exporter/applications/helpers/check_your_answers.py
hnryjmes: if what you've said is correct it might not be sufficient to check `application["ultimate_end_users"]` and I can update the predicate to check multiple things before we show ultimate end users in the summary view. I can also add a comment to explain to make this more clear in future to anyone working on this bit of code.

I'm not sure exactly what to check yet but will investigate a bit more based on what you've written there.

saruniitr: https://uktrade.atlassian.net/browse/LTD-4782 has more details

exporter/applications/helpers/check_your_answers.py
```diff
@@ -706,9 +706,13 @@ def is_application_export_type_permanent(application):
...
 def has_incorporated_goods(application):
     for good in application["goods"]:
-        if good["is_good_incorporated"]:
+        if good.get("is_good_incorporated") or good.get("is_onward_incorporated"):
```
hnryjmes, line 715: This is the main change. We consider an application to have incorporated goods if either of these booleans are True.

exporter/applications/helpers/check_your_answers.py
hnryjmes: this has now been updated. in pseudocode the updated logic I'm using is something like
```
if has_ultimate_end_users and (is_good_incorporated or is_onward_incorporated)
```

I believe this covers the situations we've talked about

depsiatwal:
Do we need to also update the API where this check is enforced ?
https://github.com/uktrade/lite-api/blob/dev/api/applications/creators.py#L156

unit_tests/exporter/applications/views/conftest.py
```diff
@@ -106,6 +106,151 @@ def good_on_application(data_standard_case):
     }
 
 
+@pytest.fixture
+def data_goa_good_incorporated(data_standard_case):
+    return {
+        "id": str(uuid.uuid4()),
+        "created_at": "2024-03-26T15:22:35.714893Z",
+        "updated_at": "2024-03-26T15:22:35.714893Z",
+        "good": {
```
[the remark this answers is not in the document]
hnryjmes: I don't know how I would use this here, creating a factory object in lite-api doesn't seem like it will help me to mock the get_application response json in lite-frontend?

exporter/applications/helpers/check_your_answers.py
hnryjmes: updated

unit_tests/exporter/applications/views/conftest.py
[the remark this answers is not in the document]
saruniitr: ah sorry forgot this is FE

PR #1873, opened 2024-04-05
[F680] Queue view case type filter

currycoder:
### Aim

Adds a filter for "Case type" to the queue view - to allow caseworkers to view F680s or SIELs.


kevincarrogan (approved):

depsiatwal (approved):
lgtm

PR #1875, opened 2024-04-08
[F680] F680 OGD advice

currycoder:
### Aim

Ensure OGDs can offer advice/move case forward for F680 applications.


depsiatwal (approved):
lgtm

PR #1881, opened 2024-04-11
Setup changes

markj0hnst0n:
Made some amendments to caseworker.env and readme to improve inital setup flow.  Tried to amend exporter.env similarly but ran into 'pre-commit detect-private-key' errors so only included these files.


example.caseworker.env
```diff
@@ -27,22 +26,26 @@ TEST_TYPE_HEADLESS=False
...
 # AWS
 AWS_S3_ENDPOINT_URL=http://s3:9000
 AWS_ACCESS_KEY_ID=minio_username
+AWS_REGION=eu-west-2
 AWS_SECRET_ACCESS_KEY=minio_password
 AWS_STORAGE_BUCKET_NAME=uploads
 
+LITE_API_HAWK_KEY=<<FROM_VAULT>>
+LITE_E2E_HAWK_KEY=<<FROM_VAULT>>
+LITE_EXPORTER_HAWK_KEY=LITE_EXPORTER_HAWK_KEY
```
hnryjmes, line 42: I think this is probably not needed, I had it in mine by mistake but I've just removed it

it only gets read in `exporter.py` https://github.com/uktrade/lite-frontend/blob/dev/conf/exporter.py#L83

markj0hnst0n: I've just removed that and pushed up again

hnryjmes (approved):

PR #1883, opened 2024-04-12
LTD 4841 add request for people with accessibility needs to contact lite support to help with user research

markj0hnst0n:
Added users with accessibility needs callout with email

### DIFF
--- a/exporter/templates/core/hub.html
+++ b/exporter/templates/core/hub.html
@@ -260,11 +260,9 @@
 			</div>
 		</div>
 	{% endif %}
-	<div class="govuk-main-wrapper--l">
-		<div class="govuk-grid-row">
-			<div class="govuk-grid-column-two-thirds govuk-body">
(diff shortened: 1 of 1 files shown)


exporter/templates/core/hub.html
```diff
@@ -260,4 +260,11 @@ <h2 class="govuk-!-margin-top-0">
 			</div>
 		</div>
 	{% endif %}
+	<div class="govuk-main-wrapper--l">
```
hnryjmes, line 263: I think you probably don't need a main wrapper here as there is already one set in base.html
https://github.com/uktrade/lite-frontend/blob/dev/exporter/templates/layouts/base.html#L117-L119

if you want to set vertical spacing there might be a different way to do it. I would remove this layer of div and just have the `<div class="govuk-grid-row"></div>` element on its own

markj0hnst0n: No worries.  I suspected that might be the case so prob should have investigated that further at base.  I've removed  that div and added in govuk-!-margin-top-5 class to improve spacing on page

hnryjmes (approved):
looks good, others might have different opinions on exactly what div structure / styles to use but I'm looking at it locally and what you've done works for me.

markj0hnst0n:
tests failing so will re-submit

PR #1886, opened 2024-04-12
LTD-4847 type on bill of lading question page

markj0hnst0n:
re-submission of previous pull request


markj0hnst0n:
tests failing to need to re-create

PR #1889, opened 2024-04-12
Filter advice by team not user.team

kevincarrogan:
### Aim

Use team field directly from the advice objects as opposed to looking up through the user.

This pairs with API changes that always store the team at a point in time instead of changing if the user changes team.

[LTD-4899](https://uktrade.atlassian.net/browse/LTD-4899)

[LTD-4899]: https://uktrade.atlassian.net/browse/LTD-4899?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ui_tests/caseworker/features/give_advice/mod_approve_advice.feature
+++ b/ui_tests/caseworker/features/give_advice/mod_approve_advice.feature
@@ -78,7 +78,7 @@ Feature: I want to record my user advice and any comments and conditions relatin
     When I click the application previously created
     And I assign myself to the case
     And I click the recommendations and decision tab
-    And I expand the details for "MOD-ECJU has approved with licence conditions"
+    And I expand the details for "MOD-CapProt has approved with licence conditions"
     Then I see "reason for approving" as the reasons for approving
     And I see "licence condition" as the licence condition
     And I see "instruction for exporter" as the instructions for the exporter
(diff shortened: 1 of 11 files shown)


ui_tests/caseworker/features/give_advice/mod_approve_advice.feature
```diff
@@ -78,7 +78,7 @@ Feature: I want to record my user advice and any comments and conditions relatin
     When I click the application previously created
     And I assign myself to the case
     And I click the recommendations and decision tab
-    And I expand the details for "MOD-ECJU has approved with licence conditions"
+    And I expand the details for "MOD-CapProt has approved with licence conditions"
```
hnryjmes, line 81: has this needed to change as a result of the other changes in this PR to filter by `advice["team"]`? or for some other reason?

hnryjmes (approved):

ui_tests/caseworker/features/give_advice/mod_approve_advice.feature
kevincarrogan: Yes. This test was written relying on the fact that previous advice given would follow the users team but now we are statically linking the team then this needed to change.

depsiatwal (approved):
lgtm

PR #1892, opened 2024-04-15
LTD-4923-rename url path for denials upload

markj0hnst0n:
Amended urls.py and tests.  Is this an obscure enough change of the URL?


kevincarrogan (approved):

hnryjmes (commented):
looks good, I would also add a TODO comment something like
```
# TODO: rename back to "denials/upload/" and "denials-upload" when we are ready to release this to users
```

hnryjmes (approved):

PR #1893, opened 2024-04-15
Update build_caseworker script to build desnz js file

hnryjmes:
### Aim

The desnz js file was not getting built and this was causing errors in prod e.g. [sentry link removed]

This change updates the `build_caseworker` script which is what generates the built js files, which should fix the issue once this makes its way to prod environment. We can test on UAT first by running through a DESNZ caseworker journey to verify this fixes the issue.

[LTD-4925](https://uktrade.atlassian.net/browse/LTD-4925)


hnryjmes (commented):
I also updated the `watch_caseworker` script as that had a reference to `beis` also

saruniitr (approved):

kevincarrogan (approved):

PR #1906, opened 2024-04-23
LTD-4948 Add link to Denial records to caseworker menu bar

hnryjmes:
### Aim

This renames the denials upload url back to what we had before as it is no longer a hidden url but something we want to release to users. The Denial records link is also added back to the menu bar so that users can find their way to the Denials upload page.

[LTD-4948](https://uktrade.atlassian.net/browse/LTD-4948)

[LTD-4948]: https://uktrade.atlassian.net/browse/LTD-4948?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


markj0hnst0n (approved):
Looks grand

depsiatwal (approved):
lgtm

PR #1907, opened 2024-04-24
LTD-4949 Update denials upload example csv

hnryjmes:
### Aim

This adds the `entity_type` column to the example csv as we want users to be able to set this directly in their data upload. Also, the columns specific to Denial have been moved to the left and columns specific to DenialEntity have been moved to the right - this is what we did for our demo to them so we should reflect that in our example csv file.

[LTD-4949](https://uktrade.atlassian.net/browse/LTD-4949)

[LTD-4949]: https://uktrade.atlassian.net/browse/LTD-4949?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/caseworker/external_data/example.csv
+++ b/caseworker/external_data/example.csv
@@ -1,2 +1,2 @@
-reference,regime_reg_ref,name,address,notifying_government,country,item_list_codes,item_description,consignee_name,end_use,reason_for_refusal,spire_entity_id
-DN2000/0000,AB-CD-EF-000,Organisation Name,"1000 Street Name, City Name",Country Name,Country Name,0A00100,Medium Size Widget,Example Name,Used in industry,Risk of outcome,123
+reference,regime_reg_ref,notifying_government,item_list_codes,item_description,end_use,reason_for_refusal,name,address,country,consignee_name,spire_entity_id,end_user_flag,consignee_flag,other_role
+DN2000/0000,AB-CD-EF-000,Country Name,0A00100,Medium Size Widget,Used in industry,Risk of outcome,Organisation Name,"1000 Street Name, City Name",Country Name,Example Name,123,TRUE,FALSE,Other Role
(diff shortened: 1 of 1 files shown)


markj0hnst0n (approved):
looks grand

caseworker/external_data/example.csv
```diff
@@ -1,2 +1,2 @@
-reference,regime_reg_ref,name,address,notifying_government,country,item_list_codes,item_description,consignee_name,end_use,reason_for_refusal,spire_entity_id
-DN2000/0000,AB-CD-EF-000,Organisation Name,"1000 Street Name, City Name",Country Name,Country Name,0A00100,Medium Size Widget,Example Name,Used in industry,Risk of outcome,123
+reference,regime_reg_ref,notifying_government,item_list_codes,item_description,end_use,reason_for_refusal,name,address,country,spire_entity_id,end_user_flag,consignee_flag,other_role
+DN2000/0000,AB-CD-EF-000,Country Name,0A00100,Medium Size Widget,Used in industry,Risk of outcome,Organisation Name,"1000 Street Name, City Name",Country Name,123,TRUE,FALSE,Other Role
```
depsiatwal, line 2: If we merge this 
https://uktrade.atlassian.net/browse/LTD-4936
Then wouldn't this be enitity_type instead ?

depsiatwal (commented):
see comment

caseworker/external_data/example.csv
hnryjmes: this is the example csv that we show to users. these users are uploading denials data from spire into lite, unless I'm mistaken. so they would have data with these columns set and this data should be handled by our system which is what LTD-4936 does

depsiatwal: yeah I get that but why does a person uploading a CSV need to set end_user_flag,consignee_flag,other_role

is it optional ? and then on a related note should we not also include entity_type in this sample ?

hnryjmes: they don't need to set it. having the columns here is more of a UX thing - it signals clearly that we can accept these columns in the csv. but the idea was not to add these to the required headers.

I'm not sure what we would do with an `entity_type` column. what data would you propose that users put in there? and how would this be handled? at the moment we only support setting entity type based on these flags

depsiatwal: I see now slight misunderstanding I was thinking we can move away from the data thing and sift to populating it directly. But I see in the CSV update it depends on these fields.

depsiatwal (approved):
nice :-)

PR #1929, opened 2024-05-01
Ltd 4951 change denial hyperlinks from reference to regime ref 

markj0hnst0n:
### Aim

Swap first column from reference to regime_ref in denial matches screen after denials search is conducted on case.  Make sure hyperlink for denial detail is on regime_ref in the first column

[LTD-4951](https://uktrade.atlassian.net/browse/LTD-4951)

[LTD-4951]: https://uktrade.atlassian.net/browse/LTD-4951?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


depsiatwal (approved):
lgtm

hnryjmes (approved):
I tested this locally and it works fine. I realised that the column order is changed and that this was discussed on the ticket.

caseworker/templates/case/denial-for-case.html
```diff
@@ -65,13 +65,14 @@ <h2 class="govuk-heading-l govuk-!-margin-top-6 govuk-!-margin-bottom-3">
                                 <td class="govuk-table__cell govuk-table__cell--checkbox">
                                     <div>
                                         <input class="govuk-checkboxes__input" type="checkbox" name="objects" value="{{ denial.id }}" id="{{ denial.id }}">
-                                        <label class="govuk-label govuk-checkboxes__label" for="{{ denial.id }}">{{ denial.reference|sentence_case }}</label>
+                                        <label class="govuk-label govuk-checkboxes__label" for="{{ denial.id }}">{{ denial.regime_reg_ref|sentence_case }}</label>
```
hnryjmes, line 68: you could update this slightly, the `|sentence_case` tag is meant to run on the `reference` not the `regime_reg_ref` so to be consistent this should be moved to stay with `reference`

markj0hnst0n: This was the label for the checkbox so wasn't sure what impact that would have as it's only used in by screen readers.  I've changed it to follow reference to it's cell? @hnryjmes

hnryjmes: yeah that looks good now

PR #1931, opened 2024-05-02
Ltd 4941 remove data field and use entity type 

markj0hnst0n:
### Aim

Front end changes for using entity type field

[LTD-4941](https://uktrade.atlassian.net/browse/LTD-4941)

### DIFF
--- a/caseworker/templates/case/denial-for-case.html
+++ b/caseworker/templates/case/denial-for-case.html
@@ -78,7 +78,17 @@
                                 <td class="govuk-table__cell">{{ denial.item_list_codes }}</td>
                                 <td class="govuk-table__cell">{{ denial.item_description }}</td>
                                 <td class="govuk-table__cell">{{ denial.end_use }}</td>
-                                <td class="govuk-table__cell">{{ denial.entity_type }}</td>
+                                <td class="govuk-table__cell">
+                                    {% if denial.entity_type == "consignee" %}
+                                        Consignee
+                                    {% elif denial.entity_type == "end_user" %}
+                                        End User
+                                    {% elif denial.entity_type == "third_party" %}
+                                        Third Party
+                                    {% else %}
+                                        {{ denial.entity_type }}
+                                    {% endif %}
+                                </td>
                             </tr>
                         {% endfor %}
 


caseworker/templates/case/denial-for-case.html
```diff
@@ -78,7 +78,17 @@ <h2 class="govuk-heading-l govuk-!-margin-top-6 govuk-!-margin-bottom-3">
                                 <td class="govuk-table__cell">{{ denial.item_list_codes }}</td>
                                 <td class="govuk-table__cell">{{ denial.item_description }}</td>
                                 <td class="govuk-table__cell">{{ denial.end_use }}</td>
-                                <td class="govuk-table__cell">{{ denial.entity_type }}</td>
+                                <td class="govuk-table__cell">
+                                    {% if denial.entity_type == "consignee" %}
```
depsiatwal, line 82: as mentioned in Arun comment you can display the key from choices in the UI.

markj0hnst0n: changes made

caseworker/templates/case/denial-for-case.html
```diff
@@ -79,15 +79,11 @@ <h2 class="govuk-heading-l govuk-!-margin-top-6 govuk-!-margin-bottom-3">
...
-                                    {% if denial.entity_type == "consignee" %}
-                                        Consignee
-                                    {% elif denial.entity_type == "end_user" %}
-                                        End User
-                                    {% elif denial.entity_type == "third_party" %}
-                                        Third Party
-                                    {% else %}
-                                        {{ denial.entity_type }}
-                                    {% endif %}
+                                    {% for key, value in denial.entity_type.items %}
```
depsiatwal, line 82: I think it's enough to just use denial.entity_type[1].value as template are value safe

markj0hnst0n: I tried something like `denial.entity_type.values[1]` `denial.entity_type[1].values` (which doesn't feel right looking at the dict output) and a few other things along those lines but the template language didn't like it

markj0hnst0n: changed now to use customer method

core/builtins/custom_tags.py
```diff
@@ -526,6 +526,14 @@ def values(dictionary):
     return dictionary.values()
 
 
+@register.filter
+def entity_type_value(dict):
+    if not dict:
```
depsiatwal, line 531: nicely done BTW :-) 

it may  also worth checking that it's a actual dict instance

caseworker/templates/external_data/denial-detail.html
```diff
@@ -41,7 +41,17 @@ <h1 class="govuk-heading-l">Denial details record</h1>
             </tr>
             <tr class="govuk-table__row app-table__row">
                 <th scope="row" class="govuk-table__header">Entity type</th>
-                <td class="govuk-table__cell">{{ denial.entity_type }}</td>
+                <td class="govuk-table__cell">
+                    {% if denial.entity_type == "consignee" %}
```
depsiatwal, line 45: also can we use the same function here ?

caseworker/templates/case/denial-for-case.html
```diff
@@ -79,7 +79,8 @@ <h2 class="govuk-heading-l govuk-!-margin-top-6 govuk-!-margin-bottom-3">
                                 <td class="govuk-table__cell">{{ denial.item_list_codes }}</td>
                                 <td class="govuk-table__cell">{{ denial.item_description }}</td>
                                 <td class="govuk-table__cell">{{ denial.end_use }}</td>
-                                <td class="govuk-table__cell">{{ denial.entity_type }}</td>
+                                <td class="govuk-table__cell">{{ denial.entity_type|entity_type_value }}
```
saruniitr, line 82: I don't think this filter is necessary. You should just be able to say,
`denial.entity_type.value`

caseworker/templates/external_data/denial-detail.html
saruniitr: same here

saruniitr (changes requested):
I think the filter is not required

core/builtins/custom_tags.py
```diff
@@ -526,6 +526,15 @@ def values(dictionary):
     return dictionary.values()
 
 
+@register.filter
+def entity_type_value(db_value):
+    if not dict:
+        return db_value
+    if isinstance(db_value, dict):
+        values = db_value.values()
+        return list(values)[1]
```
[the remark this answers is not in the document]
markj0hnst0n: I don't think anyone would believe me (i prob wouldn't believe myself but denial.entity_type.value wasn't working any other time I tried it locally but it is working now.  I think I'm having some kind of docker issue.

saruniitr (approved):

unit_tests/caseworker/cases/test_views.py
```diff
@@ -309,6 +309,10 @@ def test_search_denials(authorized_client, data_standard_case, requests_mock, qu
         "end_use",
         "entity_type",
     ]
+
+    table_body_values = table_body_values[:-1]
+    table_body_values.append("consignee")
+
```
kevincarrogan, line 315: This seems odd to me. We're extracting the data from the table and then manipulating it to make the test pass?

depsiatwal (approved):
lgtm

markj0hnst0n (commented):
looks good

PR #1935, opened 2024-05-03
LTD-4993-add-dn-highlighting-party-details 

depsiatwal:
https://uktrade.atlassian.net/browse/LTD-4993 

Adds Party Denial Details on the screen together with ability to highlight. Highlight is depended on API change which has a PR but it won’t break without that change.


markj0hnst0n (approved):
Looks great.  Big improvement

kevincarrogan (approved):

PR #1937, opened 2024-05-08
Ltd 4994 add tau permission to denials csv upload link

markj0hnst0n:
### Aim

Add permissions to the Denial Upload option in the dropdown menu on case index page to ensure that only TAU team or Admin users can see the link as per product search permission.  Tested front end to make sure link is only showing if the rule check has completed.

[LTD-4994](https://uktrade.atlassian.net/browse/LTD-4994)

[LTD-4994]: https://uktrade.atlassian.net/browse/LTD-4994?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/caseworker/core/constants.py
+++ b/caseworker/core/constants.py
@@ -72,6 +72,8 @@ class Role(Enum):
     TAU_SENIOR_MANAGER = "TAU Senior Manager"
     SUPER_USER = "Super User"
 
+    tau_roles = ["TAU Manager", "TAU Officer", "TAU Senior Manager", "Super User"]
+
 
 class FlagLevels:
     CASES = "cases"
--- a/caseworker/core/constants.py
+++ b/caseworker/core/constants.py
@@ -66,6 +66,13 @@ class Permission(Enum):
     REMOVE_HEAD_OF_LICENSING_UNIT_FLAGS = "REMOVE_HEAD_OF_LICENSING_UNIT_FLAGS"
 
 
+class Role(Enum):
+    TAU_MANAGER = "TAU Manager"
+    TAU_OFFICER = "TAU Officer"
+    TAU_SENIOR_MANAGER = "TAU Senior Manager"
+    SUPER_USER = "Super User"
+
+
 class FlagLevels:
     CASES = "cases"
     GOODS = "goods"
--- a/caseworker/core/context_processors.py
+++ b/caseworker/core/context_processors.py
@@ -2,8 +2,8 @@ from django.conf import settings
 from django.urls import reverse_lazy
 
 
-from caseworker.core.constants import Permission
-from caseworker.core.services import get_user_permissions, get_menu_notifications, get_new_mention_count
+from caseworker.core.constants import Permission, Role
+from caseworker.core.services import get_user_permissions, get_menu_notifications, get_new_mention_count, get_user_role
 from lite_content.lite_internal_frontend import strings, open_general_licences
 from lite_content.lite_internal_frontend.flags import FlagsList
 from lite_content.lite_internal_frontend.organisations import OrganisationsPage
@@ -49,6 +49,7 @@ def lite_menu(request):
     has_notifications = False
     if "lite_api_user_id" in request.session:
         permissions = get_user_permissions(request)
+        role_name = get_user_role(request)["name"]
         notifications = get_menu_notifications(request)
         notification_data = notifications["notifications"]
         has_notifications = notifications["has_notifications"]
@@ -95,7 +96,13 @@ def lite_menu(request):
                 or Permission.MANAGE_ALL_ROUTING_RULES.value in permissions,
                 {"title": "Routing rules", "url": reverse_lazy("routing_rules:list"), "icon": "menu/routing-rules"},
             ),
-            {"title": "Denial records", "url": reverse_lazy("external_data:denials-upload"), "icon": "menu/cases"},
+            conditional(
+                Role.TAU_SENIOR_MANAGER.value in role_name
+                or Role.TAU_MANAGER.value in role_name
+                or Role.TAU_OFFICER.value in role_name
+                or Role.SUPER_USER.value in role_name,
+                {"title": "Denial records", "url": reverse_lazy("external_data:denials-upload"), "icon": "menu/cases"},
+            ),
         ]
     else:
         pages = []
--- a/caseworker/core/context_processors.py
+++ b/caseworker/core/context_processors.py
@@ -49,7 +49,7 @@ def lite_menu(request):
     has_notifications = False
     if "lite_api_user_id" in request.session:
         permissions = get_user_permissions(request)
-        role_name = get_user_role(request)["name"]
+        role = get_user_role(request).values()
         notifications = get_menu_notifications(request)
         notification_data = notifications["notifications"]
         has_notifications = notifications["has_notifications"]
@@ -97,10 +97,10 @@ def lite_menu(request):
                 {"title": "Routing rules", "url": reverse_lazy("routing_rules:list"), "icon": "menu/routing-rules"},
             ),
             conditional(
-                Role.TAU_SENIOR_MANAGER.value in role_name
-                or Role.TAU_MANAGER.value in role_name
-                or Role.TAU_OFFICER.value in role_name
-                or Role.SUPER_USER.value in role_name,
+                Role.TAU_SENIOR_MANAGER.value in role
+                or Role.TAU_MANAGER.value in role
+                or Role.TAU_OFFICER.value in role
+                or Role.SUPER_USER.value in role,
                 {"title": "Denial records", "url": reverse_lazy("external_data:denials-upload"), "icon": "menu/cases"},
             ),
         ]
--- a/caseworker/core/context_processors.py
+++ b/caseworker/core/context_processors.py
@@ -55,12 +55,6 @@ def lite_menu(request):
     if "lite_api_user_id" in request.session:
         permissions = get_user_permissions(request)
         role_name = get_user_role_name(request)
-        tau_roles = [
-            Role.TAU_SENIOR_MANAGER.value,
-            Role.TAU_MANAGER.value,
-            Role.TAU_OFFICER.value,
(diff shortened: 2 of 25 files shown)


caseworker/core/rules.py
```diff
@@ -80,4 +80,4 @@ def case_is_nlr(request, case):
 rules.add_rule("can_user_generate_document", is_user_allocated)
 rules.add_rule("can_user_add_contact", is_user_allocated)
 rules.add_rule("can_user_change_sub_status", is_user_allocated & has_available_sub_statuses)
-rules.add_rule("can_user_search_products", is_user_in_admin_team | is_user_in_tau_team)  # noqa
+rules.add_rule("can_user_search_products_or_upload_denials", is_user_in_admin_team | is_user_in_tau_team)  # noqa
```
kevincarrogan, line 83: This would be better as two separate rules instead of renaming one that already exists.

caseworker/templates/layouts/base.html
```diff
@@ -128,13 +128,28 @@
...
+						{% svg item.icon %}
+					</div>
+					{{ item.title }}
+					{% if item.notifications %}<span class="app-menu__item-notification">{{ item.notifications }}</span>{% endif %}
+				</a>
+				{% else %}
+				<span></span>
+				{% endif %}
+			{% else %}
+			<a class="app-menu__item app-menu__item--icon-subtitle app-menu__item--disabled" href="#">
```
kevincarrogan, line 145: Why are we presenting a disabled item? We should probably just not show it at all.

caseworker/templates/layouts/base.html
```diff
@@ -128,13 +128,28 @@
 
 	<div id="lite-menu" class="app-menu app-hidden">
 		{% for item in LITE_MENU %}
-			<a class="app-menu__item app-menu__item--icon-subtitle {% if item.url == CURRENT_PATH %}app-menu__item--selected{% endif %}" href="{{ item.url }}">
+			{% if item.title == 'Denial records' %}
```
kevincarrogan, line 131: Instead of checking the name of the item and doing all of this logic in the template could it be better to solve this in the place that the `LITE_MENU` is constructed itself?

caseworker/core/rules.py
depsiatwal: ahh that was probably me a suggestion but as the rule makes it quite explict I agree 
rules.add_rule("can_user_search_products", is_user_in_admin_team | is_user_in_tau_team)
add another

caseworker/templates/layouts/base.html
markj0hnst0n: Moved this logic into the context processor where the LITE_MENU is constructed and added constants to check the role string.  I felt this was neater than adding a number of conditional checks for the already existing permissions variable.

caseworker/core/context_processors.py
```diff
@@ -95,7 +96,13 @@ def lite_menu(request):
                 or Permission.MANAGE_ALL_ROUTING_RULES.value in permissions,
                 {"title": "Routing rules", "url": reverse_lazy("routing_rules:list"), "icon": "menu/routing-rules"},
             ),
-            {"title": "Denial records", "url": reverse_lazy("external_data:denials-upload"), "icon": "menu/cases"},
+            conditional(
+                Role.TAU_SENIOR_MANAGER.value in role
```
kevincarrogan, line 100: This could be written the other way round i.e. `role in [manager, senior_manager]`

caseworker/core/context_processors.py
```diff
@@ -49,6 +49,7 @@ def lite_menu(request):
     has_notifications = False
     if "lite_api_user_id" in request.session:
         permissions = get_user_permissions(request)
+        role = get_user_role(request).values()
```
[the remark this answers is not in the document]
markj0hnst0n: made changes to address this

caseworker/core/context_processors.py
markj0hnst0n: made changes to address this

caseworker/core/context_processors.py
```diff
@@ -49,6 +54,13 @@ def lite_menu(request):
     has_notifications = False
     if "lite_api_user_id" in request.session:
         permissions = get_user_permissions(request)
+        role_name = get_user_role_name(request)
+        tau_roles = [
```
depsiatwal, line 58: we could just add something to the class to give us a list of values.

depsiatwal (commented):
I don't see any tests to check the menu is visible or not depending on role.

markj0hnst0n:
tests added for valid and invalid roles

caseworker/core/constants.py
```diff
@@ -66,6 +66,15 @@ class Permission(Enum):
     REMOVE_HEAD_OF_LICENSING_UNIT_FLAGS = "REMOVE_HEAD_OF_LICENSING_UNIT_FLAGS"
 
 
+class Role(Enum):
+    TAU_MANAGER = "TAU Manager"
+    TAU_OFFICER = "TAU Officer"
+    TAU_SENIOR_MANAGER = "TAU Senior Manager"
+    SUPER_USER = "Super User"
+
+    tau_roles = ["TAU Manager", "TAU Officer", "TAU Senior Manager", "Super User"]
```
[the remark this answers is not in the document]
markj0hnst0n: Amended

depsiatwal (approved):
LGTM

unit_tests/caseworker/core/test_context_processors.py
```diff
@@ -1,25 +1,46 @@
...
+    mocker.patch("caseworker.core.context_processors.get_queue")
+
+    request = mocker.MagicMock()
+    request.session = {"lite_api_user_id": "user_id", "default_queue": "00000000-0000-0000-0000-000000000001"}
+
+    result = lite_menu(request)
+
+    assert result["LITE_MENU"][-1]["title"] == "Denial records"
+    assert result["LITE_MENU"][-1]["url"] == reverse_lazy("external_data:denials-upload")
+    assert result["LITE_MENU"][-1]["icon"] == "menu/cases"
```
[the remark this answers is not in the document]
markj0hnst0n: @saruniitr Test now doesn't break when more items are added

unit_tests/caseworker/core/test_context_processors.py
```diff
@@ -1,25 +1,45 @@
...
+    mocker.patch("caseworker.core.context_processors.get_user_permissions")
+    mocker.patch("caseworker.core.context_processors.get_user_role_name", return_value=invalid_user_role)
+    mocker.patch("caseworker.core.context_processors.get_queue")
+
+    request = mocker.MagicMock()
+    request.session = {"lite_api_user_id": "user_id", "default_queue": "00000000-0000-0000-0000-000000000001"}
+
+    result = lite_menu(request)
+    assert "Denial records" not in [x["title"] for x in result["LITE_MENU"]]
+    assert reverse_lazy("external_data:denials-upload") not in [x["url"] for x in result["LITE_MENU"]]
```
saruniitr, line 45: this is better but instead of separate tests for valid and invalid users, I think you can combine both in a single test with one other parameter. please consider,

```
@pytest.mark.parametrize(
    "role,status",    # rename variables as appropriate
    ("TAU Manager", True),
    ("TAU User", True),
    ("LU User", False),
    ("Admin", True),
)
def test_lite_menu_user_can_access_denial_records(role, status):
 ...
 ...

 result = lite_menu(request)
menu_options = [item['title'] for item in item['LITE_MENU']
 assert ("Denial records" in menu_options) is status
```

saruniitr (commented):
one last suggestion otherwise it is good

unit_tests/caseworker/core/test_context_processors.py
```diff
@@ -1,25 +1,40 @@
...
+    request = mocker.MagicMock()
+    request.session = {"lite_api_user_id": "user_id", "default_queue": "00000000-0000-0000-0000-000000000001"}
+
+    result_for_side_effect_1 = lite_menu(request)
+
+    for item in result_for_side_effect_1["LITE_MENU"]:
+        if item["title"] == "Denial records":
+            assert item["url"] == reverse_lazy("external_data:denials-upload")
+
+    result_for_side_effect_2 = lite_menu(request)
```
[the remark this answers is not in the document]
markj0hnst0n: I was just refactoring this part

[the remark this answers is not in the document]
markj0hnst0n: Still wanted to keep the reference to the Role constant however, so I do still have 2 parameters

unit_tests/caseworker/core/test_context_processors.py
```diff
@@ -1,25 +1,36 @@
...
-    url = reverse(
-        "cases:attach_documents",
-        kwargs={"queue_pk": queue, "pk": data_standard_case["case"]["id"]},
-    )
-    resp = authorized_client.get(url)
-    assert resp.context.get("is_all_cases_queue") == is_all_cases_queue
+roles_and_status = []
+for role in Role.tau_roles.value:
+    role = (role, True)
+    roles_and_status.append(role)
```
saruniitr, line 11: I would just type out the valid roles also but easy either way

unit_tests/caseworker/core/test_context_processors.py
```diff
@@ -1,25 +1,36 @@
...
+    mocker.patch("caseworker.core.context_processors.get_user_role_name", return_value=role)
+    mocker.patch("caseworker.core.context_processors.get_queue")
+
+    request = mocker.MagicMock()
+    request.session = {"lite_api_user_id": "user_id", "default_queue": "00000000-0000-0000-0000-000000000001"}
+
+    result = lite_menu(request)
+    menu_options = [item["title"] for item in result["LITE_MENU"]]
+
+    assert ("Denial records" in menu_options) == status
```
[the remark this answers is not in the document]
markj0hnst0n: I've added a check like this.  Thanks for getting back!

PR #1945, opened 2024-05-09
Add a longer grace period for bringing up containers in E2E tests

saruniitr:
## Change description

When running E2E tests the current timeout to wait for containers to come up is 5 min.
We have couple of data migrations recently related to Denials and adding new ones related to Report summaries which are pushing it over 5min limit. These migrations are already optimised as possible but we may exceeding the timeout slightly so increase the timeout to allow for these changes.


kevincarrogan (approved):

depsiatwal (approved):
lgtm

PR #1946, opened 2024-05-09
Ltd 5000 add country highlighting

markj0hnst0n:
### Aim

Highlight the country on denials search page

Corresponding API pr here: https://github.com/uktrade/lite-api/pull/1982

[LTD-5000](https://uktrade.atlassian.net/browse/LTD-5000)

[LTD-5000]: https://uktrade.atlassian.net/browse/LTD-5000?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/caseworker/cases/views/main.py
+++ b/caseworker/cases/views/main.py
@@ -698,6 +698,9 @@ class Denials(LoginRequiredMixin, TemplateView):
         if search:
             response = search_denials(request=self.request, search=search, filter=filter).json()
             results = response["results"]
+            for result in results:
+                if result["country"]:
+                    result["country"] = f"<mark>{result['country']}</mark>"
             total_pages = response["total_pages"]
 
         return super().get_context_data(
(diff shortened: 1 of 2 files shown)


caseworker/cases/views/main.py
```diff
@@ -698,6 +698,9 @@ def get_context_data(self, **kwargs):
         if search:
             response = search_denials(request=self.request, search=search, filter=filter).json()
             results = response["results"]
+            for result in results:
+                if result["country"]:
+                    result["country"] = f"<mark>{result['country']}</mark>"
```
depsiatwal, line 703: As marking is happening on BE it makes sense to perhaps contain this logic there. 

I would check the 'filter' and if it matches the country returned then add the markup

markj0hnst0n: I think it's _technically_ possible to highlight filter fields in the backend but not very convenient with `django_elasticsearch_dsl_drf` it requires manually re-writing the search queries and I'm not even sure how to get under the hood to do that.

From what I understand highlighting is used for highlighting search terms only usually not filter terms.

depsiatwal: yeah I was thinking more from a consistent stand point. Marking is then done all in one place. 
any how I think we should still definitely check the 'filter' query matches the country field.

depsiatwal: yeah I think you would need to write something on the get or filter if it's too much of a pain 
also happy for it to go here for first iteration.

markj0hnst0n:
closed in favour of back end change

PR #1958, opened 2024-05-15
resolved optional registration number

Tllew:
### Aim

### DIFF
--- a/exporter/core/organisation/forms.py
+++ b/exporter/core/organisation/forms.py
@@ -137,7 +137,7 @@ class RegisterDetailsBaseForm(BaseForm):
         super().__init__(*args, **kwargs)
 
     def clean_registration_number(self):
-        if self.fields["registration_number"].required:
+        if self.cleaned_data["registration_number"]:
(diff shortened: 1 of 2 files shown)


depsiatwal (commented):
if the tests don't fail because of this change I think it's worth adding one to check 
code cov will fail anyway

exporter/core/organisation/forms.py
```diff
@@ -137,10 +137,11 @@ def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
 
     def clean_registration_number(self):
-        response, status_code = validate_registration_number(self.request, self.cleaned_data)
-        if status_code != 200:
-            self.add_error("registration_number", response["errors"]["registration_number"])
-            return
+        if self.fields["registration_number"].required:
```
depsiatwal, line 140: The filed may not be required (optional) but could have a value. 
and if it has a value we would want to check for duplicates.

depsiatwal (approved):
lgtm

currycoder (approved):

PR #1975, opened 2024-05-22
LTD-5012-complex-denial-searching 

depsiatwal:
https://uktrade.atlassian.net/browse/LTD-5012)

This fundamentally changes how we search denials it allows more complex queries which can be customised by the user. 

To validate the default results UAT vs Test Env will use the same data and compare old vs new search/ This will allow us to validate this is sensible to roll out. 

Also dependant on API PR

https://github.com/uktrade/lite-api/pull/2002


caseworker/cases/views/denials.py
```diff
@@ -50,8 +50,8 @@ def get_search_filter(self):
         }
 
         for party in self.parties_to_search:
-            search_filter.append(f'name:"{party["name"]}"')
-            search_filter.append(f'address:"{party["address"]}"')
+            search_filter.append(f'name:({party["name"]})')
+            search_filter.append(f'address:({party["address"]})')
```
kevincarrogan, line 54: What’s going on here with the brackets? It might be good to comment what this is doing.

markj0hnst0n (approved):

kevincarrogan (approved):

PR #1978, opened 2024-05-23
LTD-5004-e2e-create-org

seijihg:
https://uktrade.atlassian.net/browse/LTD-5004


ci.exporter.env
```diff
@@ -45,10 +45,11 @@ SsF52iSXTw==
 TOKEN_SESSION_KEY=exporter_token_session_key
 
 MOCK_SSO_ACTIVATE_ENDPOINTS=True
-MOCK_SSO_USER_EMAIL=lite-ops@digital.trade.gov.uk
```
seijihg, line 48: I have deleted that so we can log in as a new user and give us more flexibility. Otherwise it will keeps logging in the above user by default.

EXPORTER_TEST_SSO_EMAIL was added so the fixture can pick it up and log in this user when is needed

ui_tests/exporter/step_defs/test_register_an_organisation.py
```diff
@@ -53,13 +58,88 @@ def register_individual(driver):
...
 
 
-@when("I sign in as user without an organisation registered")  # noqa
-def go_to_exporter_when(driver, exporter_url, context):  # noqa
+fake = Faker("en_GB")
+
+
+@given("I register but I don't belong to an organisation")
+def new_log_in(context):
+    response = create_govuk_sso_user()
```
kevincarrogan, line 66: I feel a little like the name of this function `create_govuk_sso_user` is misleading because I don't think it does that.

ui_tests/exporter/step_defs/test_register_an_organisation.py
```diff
@@ -53,13 +58,88 @@ def register_individual(driver):
...
+
+
+@then(parsers.parse('I pick an organisation "{organisation}"'))
+def pick_org(driver, organisation):
+    if "register-an-organisation" in driver.current_url:
+        no = utils.get_element_index_by_text(Shared(driver).get_radio_buttons_elements(), organisation)
+        Shared(driver).click_on_radio_buttons(no)
+        functions.click_submit(driver)
+    else:
+        assert False
```
kevincarrogan, line 135: It's probably better to raise a useful message here instead of just `assert False` as this is going to be hard to debug if/when it fails.

ui_tests/exporter/step_defs/test_register_an_organisation.py
```diff
@@ -53,13 +58,88 @@ def register_individual(driver):
...
+def new_log_in(context):
+    response = create_govuk_sso_user()
+    context.newly_registered_email = response["email"]
+    context.newly_registered_password = response["password"]
+
+
+@given("I am not signed into LITE but signed into GREAT SSO")
+def not_logged_into_LITE(exporter_url, driver, context):
+    driver.get(exporter_url.rstrip("/") + "/auth/logout")
+    if "accounts/logout" in driver.current_url:
```
kevincarrogan, line 74: I'm not really sure what this is doing here. We go to the page and then _something_ happens where it might go somewhere else?

ui_tests/exporter/features/register_an_organisation.feature
```diff
@@ -4,10 +4,22 @@ Feature: I want to register an organisation
...
+    When I sign in as a new user without an organisation registered
+    Then I pick an organisation "Commercial organisation"
+    And I choose the option "In the United Kingdom"
+    When I enter my information to register a commercial organisation
+    Then the organisation is registered successfully
+    And I logout
+
+  @register_individual_organisation
+  Scenario: Register an individual organisation that is based in UK
+    Given I am not signed into LITE but signed into GREAT SSO
```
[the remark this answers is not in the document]
seijihg: thanks removed!

ui_tests/exporter/step_defs/test_register_an_organisation.py
seijihg: its creating user but it was here already and being used by some places so I will leave it as it is.

ui_tests/exporter/step_defs/test_register_an_organisation.py
seijihg: done thanks

ui_tests/exporter/step_defs/test_register_an_organisation.py
seijihg: probably it was left here when I was testing things making sure i was logged out first if I was logged in.

ui_tests/exporter/step_defs/test_register_an_organisation.py
```diff
@@ -53,25 +58,84 @@ def register_individual(driver):
...
+
+
+@then(parsers.parse('I pick an organisation "{organisation}"'))
+def pick_org(driver, organisation):
+    if "register-an-organisation" in driver.current_url:
+        no = utils.get_element_index_by_text(Shared(driver).get_radio_buttons_elements(), organisation)
+        Shared(driver).click_on_radio_buttons(no)
+        functions.click_submit(driver)
+    else:
+        assert False, "The current URL does not contain 'register-an-organisation'."
```
kevincarrogan, line 131: I think https://docs.pytest.org/en/7.1.x/reference/reference.html#pytest-fail is probably the right thing to use here.

markj0hnst0n (approved):

ci.exporter.env
saruniitr: Does this not break if we are using mock sso?
Is it verified that it is still working as expected?

ui_tests/exporter/conftest.py
```diff
@@ -125,7 +125,11 @@ def go_to_exporter(driver, register_organisation, sso_sign_in, exporter_url, con
     if "signin" in driver.current_url and not settings.MOCK_SSO_ACTIVATE_ENDPOINTS:
         GovukSigninPage(driver).sign_in(exporter_info["email"], exporter_info["password"])
 
-    if settings.MOCK_SSO_ACTIVATE_ENDPOINTS and not getattr(settings, "MOCK_SSO_USER_EMAIL", None):
+    if (
+        "authorize" in driver.current_url
+        and settings.MOCK_SSO_ACTIVATE_ENDPOINTS
+        and not getattr(settings, "MOCK_SSO_USER_EMAIL", None)
+    ):
```
saruniitr, line 132: same as above?
Also relying presence of certain strings is not good. If the url changes then this fails.

ui_tests/exporter/features/register_an_organisation.feature
```diff
@@ -4,10 +4,22 @@ Feature: I want to register an organisation
...
+  @register_commercial_organisation
   Scenario: Register a commercial organisation that is based in UK
-    Given I am not signed into LITE but signed into GREAT SSO
-    When I sign in as user without an organisation registered
-    And I enter my information from steps 1-4 to register a commercial organisation
+    Given I make sure that I am not logged in
+    When I sign in as a new user without an organisation registered
+    Then I pick an organisation "Commercial organisation"
+    And I choose the option "In the United Kingdom"
+    When I enter my information to register a commercial organisation
```
saruniitr, line 13: This single step is hiding a lot of actions. I think each should perform individual action and this is very open and doesn't convey what is happening.
Eg It is not clear what information user has to enter for commercial organisation (unless you look at the step definition).

By reading the scenario it should be clear what information is entered and what actions are performed. I think this needs breaking down into individual steps.

And I enter organisation name as ""
And I enter EORI number as ""
...

ui_tests/exporter/features/register_an_organisation.feature
```diff
@@ -4,10 +4,22 @@ Feature: I want to register an organisation
...
 
-  @skip @current
+  @register_commercial_organisation
   Scenario: Register a commercial organisation that is based in UK
-    Given I am not signed into LITE but signed into GREAT SSO
-    When I sign in as user without an organisation registered
-    And I enter my information from steps 1-4 to register a commercial organisation
+    Given I make sure that I am not logged in
+    When I sign in as a new user without an organisation registered
+    Then I pick an organisation "Commercial organisation"
```
saruniitr, line 11: just like the next step please change it "I choose/select the option"

ui_tests/exporter/features/register_an_organisation.feature
```diff
@@ -4,10 +4,22 @@ Feature: I want to register an organisation
...
+    Then the organisation is registered successfully
+    And I logout
+
+  @register_individual_organisation
+  Scenario: Register an individual organisation that is based in UK
+    Given I make sure that I am not logged in
+    When I sign in as a new user without an organisation registered
+    Then I pick an organisation "Private individual"
+    And I choose the option "In the United Kingdom"
+    When I enter my information to register an individual organisation
```
saruniitr, line 23: Same here and the same steps can be reused here.

ui_tests/exporter/step_defs/test_register_an_organisation.py
```diff
@@ -53,25 +60,84 @@ def register_individual(driver):
...
+    register.enter_random_company_name()
+    register.enter_random_eori_number(get_eori_number)
+    register.enter_random_sic_number()
+    register.enter_random_vat_number()
+    register.enter_random_registration_number()
+    functions.click_submit(driver)
+
+    register.enter_random_site()
+    functions.click_submit(driver)
+    functions.click_submit(driver)
```
saruniitr, line 94: why are we submitting, entering some data and submitting twice again?

There is also phone number that we enter in the form.

ui_tests/exporter/step_defs/test_register_an_organisation.py
```diff
@@ -53,25 +60,84 @@ def register_individual(driver):
...
+@when("I sign in as a new user without an organisation registered")  # noqa
+def go_to_exporter_when(driver, exporter_url):  # noqa
+    driver.get(exporter_url)
+    StartPage(driver).try_click_sign_in_button()
+
+    if "signin" in driver.current_url and not settings.MOCK_SSO_ACTIVATE_ENDPOINTS:
+        GovukSigninPage(driver).sign_in(fake.email(), fake.password())
+
+    if settings.MOCK_SSO_ACTIVATE_ENDPOINTS and not getattr(settings, "MOCK_SSO_USER_EMAIL", None):
+        MockSigninPage(driver).sign_in(fake.email())
```
saruniitr, line 119: Is this the result of deleting `MOCK_SSO_USER_EMAIL` above?
Not sure what is the reason to remove but it is just making things more complicated.

ui_tests/exporter/step_defs/test_register_an_organisation.py
```diff
@@ -53,25 +60,84 @@ def register_individual(driver):
...
+
+        driver.find_element(by=By.ID, value="id_first_name").send_keys(fake.first_name())
+        driver.find_element(by=By.ID, value="id_last_name").send_keys(fake.last_name())
+        driver.find_element(by=By.CSS_SELECTOR, value="[type='submit']").click()
+
+
+@then(parsers.parse('I pick an organisation "{organisation}"'))
+def pick_org(driver, organisation):
+    if "register-an-organisation" in driver.current_url:
```
saruniitr, line 128: this pattern is used in multiple places and not the correct approach.
When this step is executed then we are expected to be on the correct page, if not test fails as it should be so no need to check again.

ui_tests/exporter/step_defs/test_register_an_organisation.py
```diff
@@ -53,25 +60,84 @@ def register_individual(driver):
...
+        no = utils.get_element_index_by_text(Shared(driver).get_radio_buttons_elements(), organisation)
+        Shared(driver).click_on_radio_buttons(no)
+        functions.click_submit(driver)
+    else:
+        pytest.fail("The current URL does not contain 'register-an-organisation'.", pytrace=False)
+
+
+@then(parsers.parse('I choose the option "{option}"'))
+def pick_option(driver, option):
+    no = utils.get_element_index_by_text(Shared(driver).get_radio_buttons_elements(), option)
```
saruniitr, line 138: this is not the efficient way to select element as it loops through all elements.
I think we should always go with one of the following variants
```
driver.find_element(by=By.ID, value)
driver.find_element(by=By.CLASS_NAME, value)
driver.find_element(by=By.CSS_SELECTOR, value)
driver.find_element(by=By.XPATH, value)
```

ui_tests/exporter/step_defs/test_register_an_organisation.py
```diff
@@ -53,25 +60,84 @@ def register_individual(driver):
...
+    Shared(driver).click_on_radio_buttons(no)
+    functions.click_submit(driver)
```
saruniitr, line 140: if we are also submitting then the step should reflect that,

eg "I choose the option "option" and submit

saruniitr (changes requested):
I think it requires changes and may go through few iterations.
I can take a look once it is ready again.

ci.exporter.env
seijihg: yes its working as expected

ui_tests/exporter/conftest.py
seijihg: I can base it on something else, but yea because we are not logging in automatically now it will use exporter_info["email"] to log in.

ui_tests/exporter/step_defs/test_register_an_organisation.py
seijihg: the reason it was removed because to create a new organisation you need to log in as a new user.

ui_tests/exporter/step_defs/test_register_an_organisation.py
seijihg: we confirm data and then we in checks data is correct page where we click submit, just reused the same function to click button thats all

ui_tests/exporter/features/register_an_organisation.feature
seijihg: I am not sure if it does matter much, since we are testing if flow of registering is working correctly so things are just random inputs.

saruniitr: Random input is fine, it just need to be broken down into individual action instead of one step doing all the actions.
We can say,
"And I enter EORI number"

and in the definition you can enter random EORI number following the format it is expecting.
Similarly for other steps.

ui_tests/exporter/step_defs/test_register_an_organisation.py
saruniitr: ok it looks like it is submitting multiple times to skip the confirmation screen?
I think this needs to reflect in the test eg with a step saying that I see summary screen with the details validated etc.
This again goes back to doing multiple different actions in a single step.

ui_tests/exporter/step_defs/test_register_an_organisation.py
```diff
@@ -20,58 +29,81 @@ def new_log_in(context):
     context.newly_registered_password = response["password"]
 
 
-@when("I enter my information from steps 1-4 to register a commercial organisation")
-def register_commercial(driver, get_eori_number):
+@given("I make sure that I am not logged in")
+def not_logged_into_LITE(exporter_url, driver, context):
+    driver.get(exporter_url.rstrip("/") + "/auth/logout")
```
saruniitr, line 34: I think the previous comment is still valid here, what is the purpose of this step?
What is the reason to ensure you are not logged in? when the test start you are. not logged in anyway so we can just proceed to login. Please remove this step.

ui_tests/exporter/step_defs/test_register_an_organisation.py
```diff
@@ -20,58 +29,81 @@ def new_log_in(context):
...
-def go_to_exporter_when(driver, exporter_url, context):  # noqa
+@when("I sign in as a new user without an organisation registered")  # noqa
+def go_to_exporter_when(driver, exporter_url):  # noqa
     driver.get(exporter_url)
     StartPage(driver).try_click_sign_in_button()
 
-    if "signin" in driver.current_url:
-        GovukSigninPage(driver).sign_in(context.newly_registered_email, context.newly_registered_password)
+    if settings.MOCK_SSO_ACTIVATE_ENDPOINTS and not getattr(settings, "MOCK_SSO_USER_EMAIL", None):
+        MockSigninPage(driver).sign_in(fake.email())
```
saruniitr, line 85: This is the confusing bit and I am not following the need to check this.

`MOCK_SSO_USER_EMAIL` is already deleted above so what is the need to check here?
Also in this step it should not matter whether mock sso endpoints are active or not, this step should just try to sign in. Please remove this condition.

ui_tests/exporter/step_defs/test_register_an_organisation.py
```diff
@@ -20,58 +29,81 @@ def new_log_in(context):
...
-    response = create_govuk_sso_user()
-    context.newly_registered_email = response["email"]
-    context.newly_registered_password = response["password"]
+@then(parsers.parse('I choose an organisation "{organisation}" and go to next page'))
+def pick_org(driver, organisation):
+    if organisation == "Commercial organisation":
+        driver.find_element(By.ID, "id_REGISTRATION_TYPE-type_1").click()
+
+    if organisation == "Private individual":
+        driver.find_element(By.ID, "id_REGISTRATION_TYPE-type_2").click()
```
saruniitr, line 98: this pattern is not encouraged in the tests where step behaves differently based on a condition.
Usually we create separate steps for each option. So in this case you have two options,
- create separate steps for commerical and private individual
- if you want single step then without using condition find the element using the given parameter and select that option.

ui_tests/exporter/features/register_an_organisation.feature
```diff
@@ -4,10 +4,30 @@ Feature: I want to register an organisation
...
+    Given I make sure that I am not logged in
+    When I sign in as a new user without an organisation registered
+    Then I choose an organisation "Commercial organisation" and go to next page
+    And I select In the United Kingdom and go to next page
+    When I enter random name
+    Then I input random EORI number
+    And I input random SIC number
+    And I input random VAT number
+    And I enter random registration number and go to next page
+    Then I enter random site details and finish submitting
```
saruniitr, line 18: I understand we are entering random details but we don't have to specify that in the step, these can be,
```
    When I enter company name
    And I enter company EORI number
    And I enter company SIC number
    And I enter company VAT number
    And I enter company registration number and continue
    And I enter random site details and finish submitting
```

`Then` only at the last step as that defines the outcome of all above steps

ui_tests/exporter/conftest.py
```diff
@@ -122,10 +123,19 @@ def go_to_exporter(driver, register_organisation, sso_sign_in, exporter_url, con
...
+
     if "signin" in driver.current_url and not settings.MOCK_SSO_ACTIVATE_ENDPOINTS:
         GovukSigninPage(driver).sign_in(exporter_info["email"], exporter_info["password"])
 
-    if settings.MOCK_SSO_ACTIVATE_ENDPOINTS and not getattr(settings, "MOCK_SSO_USER_EMAIL", None):
+    if (
+        mock_sso_login_screen
+        and settings.MOCK_SSO_ACTIVATE_ENDPOINTS
+        and not getattr(settings, "MOCK_SSO_USER_EMAIL", None)
+    ):
```
saruniitr, line 138: As mentioned below in other comment, I am not following the need to check for these here.
`MOCK_SSO_USER_EMAIL` is deleted so what is the need to check again?

ci.exporter.env
saruniitr: I think deleting this is causing you more issues and various conditions have been added to overcome this.
I am not seeing the need to delete this.

saruniitr (commented):
I think it is moving in the right direction, few more changes requested and this is good.

ci.exporter.env
seijihg: no, without deleting this I cant log in as a new user to test the flow because when the image is being built it will use MOCK_SSO_USER_EMAIL to authenticate automatically on the API side. I cant overwrite this with a command etc. I spent two looooong time on this trying to solve this issue

ui_tests/exporter/step_defs/test_register_an_organisation.py
seijihg: I have added it because at the beginning I was logged in for some reason maybe because of parallel thingy since its using the same container to run chrome.

ui_tests/exporter/conftest.py
seijihg: probably there is no need :P just left it there since its the pattern used previously :D

kevincarrogan (approved):

saruniitr (approved):

PR #1982, opened 2024-05-28
ltd-5027-denial-search-validation-message 

depsiatwal:
Add error message to form

### DIFF
--- a/caseworker/cases/forms/denial_forms.py
+++ b/caseworker/cases/forms/denial_forms.py
@@ -10,11 +10,13 @@ from crispy_forms_gds.layout import (
 from django import forms
 from django.template.loader import render_to_string
 
+from core.forms.layouts import RenderTemplate
+
 
 class DenialSearchForm(forms.Form):
 
     search_string = forms.CharField(
-        widget=forms.Textarea(attrs={"rows": "2"}),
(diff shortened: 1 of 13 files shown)


caseworker/cases/forms/denial_forms.py
```diff
@@ -14,22 +14,44 @@
...
 
     search_string = forms.CharField(
-        widget=forms.Textarea(attrs={"rows": "2"}),
+        widget=forms.Textarea(attrs={"rows": "1"}),
         label="",
         required=False,
     )
 
+    def query_error(self):
+        return HTML(
```
kevincarrogan, line 23: I feel this would be better as a template that gets rendered via `render_to_string`.

caseworker/assets/styles/components/_denial-search.scss
kevincarrogan, line 1: Is this a copy and paste from product search? Could this be extracted out to something more common?

caseworker/cases/forms/denial_forms.py
depsiatwal: added new layout

kevincarrogan (approved):

markj0hnst0n (approved):

PR #1991, opened 2024-05-31
Allow ODT file uploads in exporter

hnryjmes:
### Aim

This updates the list of acceptable mime types in lite-frontend which is what is used to determine if an uploaded file is acceptable or not. This has been tested locally by uploading an odt document, and on the demo environment where I was able to upload an odt document as an exporter and then submit this application and see that the document can be downloaded from the caseworker side.

[LTD-4928](https://uktrade.atlassian.net/browse/LTD-4928)

[LTD-4928]: https://uktrade.atlassian.net/browse/LTD-4928?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/conf/base.py
+++ b/conf/base.py
@@ -166,10 +166,11 @@ FILE_UPLOAD_HANDLERS = env.list("FILE_UPLOAD_HANDLERS", default=["core.file_hand
 ACCEPTED_FILE_UPLOAD_MIME_TYPES = env.list(
     "ACCEPTED_FILE_UPLOAD_MIME_TYPES",
     default=(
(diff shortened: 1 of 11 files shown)


kevincarrogan (commented):
The fact this small change required us updating 10 files suggests to me that we might want to centralise this in some way so we're only updating this in one place moving forward.

saruniitr (approved):

kevincarrogan (approved):

PR #1992, opened 2024-05-31
Ltd 5054 expose elasticsearch score in denial search results

markj0hnst0n:
### Aim

Expose 'score' in the denial results table and add feature flag to turn it on and off

Corresponding API PR [here](https://github.com/uktrade/lite-frontend/pull/1992)

[LTD-5054](https://uktrade.atlassian.net/browse/LTD-5054)

[LTD-5054]: https://uktrade.atlassian.net/browse/LTD-5054?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


unit_tests/caseworker/cases/test_denials_view.py
```diff
@@ -200,6 +208,9 @@ def test_search_denials(
         json={"count": "26", "total_pages": "2", "results": denials_data * 26},
     )
 
+    # Disable search score feature flag to make sure default behaviour is used
+
+    settings.FEATURE_FLAG_SEARCH_SCORE = False
```
kevincarrogan, line 213: Instead of changing this directly you should use the `settings` fixture https://pytest-django.readthedocs.io/en/latest/helpers.html#settings here so that the value gets reset correctly at the end of the test.

unit_tests/caseworker/cases/test_denials_view.py
```diff
@@ -186,6 +187,13 @@ def test_search_denials_session_search_string_matchs(
     )
 
 
+def test_search_score_feature_flag(authorized_client, data_standard_case, url):
```
kevincarrogan, line 190: We should set the settings explicitly here (by using the `settings` fixture) and make sure that when we toggle the setting we get the right value.

unit_tests/caseworker/cases/test_denials_view.py
markj0hnst0n: Refactored to use fixture.  Feedback most welcome

unit_tests/caseworker/cases/test_denials_view.py
markj0hnst0n: Refactored to use fixture.   Feedback most welcome

unit_tests/caseworker/cases/test_denials_view.py
```diff
@@ -60,6 +61,18 @@ def mock_denials_search(requests_mock, denials_search_results):
     return requests_mock.get(url=url, json=denials_search_results)
 
 
+@pytest.fixture
+def search_score_flag_on(settings):
+    settings.FEATURE_FLAG_SEARCH_SCORE = True
+    return settings.FEATURE_FLAG_SEARCH_SCORE
```
saruniitr, line 67: no need to return this.

unit_tests/caseworker/cases/test_denials_view.py
```diff
@@ -186,8 +199,22 @@ def test_search_denials_session_search_string_matchs(
     )
 
 
+def test_search_score_feature_flag(authorized_client, data_standard_case, url, search_score_flag_on):
+    end_user_id = data_standard_case["case"]["data"]["end_user"]["id"]
+    settings.FEATURE_FLAG_SEARCH_SCORE = search_score_flag_on
```
saruniitr, line 204: As it is set in fixture, this is not required either (in other places below).

conf/caseworker.py
```diff
@@ -120,6 +120,7 @@
 }
 
 FEATURE_FLAG_PRODUCT_SEARCH = env.bool("FEATURE_FLAG_PRODUCT_SEARCH", False)
+FEATURE_FLAG_SEARCH_SCORE = env.bool("FEATURE_FLAG_SEARCH_SCORE", False)
```
saruniitr, line 123: Could this be renamed to specify that it is Denials search score?

unit_tests/caseworker/cases/test_denials_view.py
```diff
@@ -60,6 +61,18 @@ def mock_denials_search(requests_mock, denials_search_results):
...
 
 
+@pytest.fixture
+def search_score_flag_on(settings):
+    settings.FEATURE_FLAG_SEARCH_SCORE = True
+    return settings.FEATURE_FLAG_SEARCH_SCORE
+
+
+@pytest.fixture
+def search_score_flag_off(settings):
```
depsiatwal, line 71: I think you need a test that accepts parameter value with feature ON/OFF and checks the HTML content is expected

unit_tests/caseworker/cases/test_denials_view.py
markj0hnst0n: removed a few instances of this type of code

unit_tests/caseworker/cases/test_denials_view.py
markj0hnst0n: Added single job tests for these now @depsiatwal

unit_tests/caseworker/cases/test_denials_view.py
```diff
@@ -186,8 +197,31 @@ def test_search_denials_session_search_string_matchs(
...
 
+def test_search_score_feature_flag_on(authorized_client, data_standard_case, url, denials_search_score_flag_on):
+    end_user_id = data_standard_case["case"]["data"]["end_user"]["id"]
+    response = authorized_client.get(f"{url}?end_user={end_user_id}")
+    soup = BeautifulSoup(response.content, "html.parser")
+    search_header = soup.find("th", string="Search score")
+    assert search_header
+
+
+def test_search_score_feature_flag_off(authorized_client, data_standard_case, url, denials_search_score_flag_off):
```
depsiatwal, line 208: These two are effectively doing the same thing you could 
@pytest.mark.parametrize to set the feature flag and use boolean for the asset

depsiatwal: Here's an example 
https://github.com/uktrade/lite-frontend/blob/a2e184e292c5a93947774f6068feca4de21a8e05/unit_tests/exporter/apply_for_a_licence/forms/test_triage_questions.py#L14

saruniitr (approved):

unit_tests/caseworker/cases/test_denials_view.py
```diff
@@ -60,6 +61,16 @@ def mock_denials_search(requests_mock, denials_search_results):
     return requests_mock.get(url=url, json=denials_search_results)
 
 
+@pytest.fixture
+def denials_search_score_flag_on(settings):
```
depsiatwal, line 65: i'm not sure we need the off since we should just leave in as default 

Then you could just use thing fixture denials_search_score_flag_on when you need it on in a test or just set it directly

markj0hnst0n: I sort of thought it was redundant to use the `denials_search_score_flag_off ` test as it's done in `test_search_denials` I've removed the it now

caseworker/cases/views/denials.py
```diff
@@ -94,11 +95,14 @@ def get_context_data(self, **kwargs):
             search_results, _ = search_denials(request=self.request, search=search, filter=filter)
             total_pages = search_results.get("total_pages", 0)
 
-        return super().get_context_data(
+        context = super().get_context_data(
```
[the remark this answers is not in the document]
markj0hnst0n: It was a hangover from a different way of adding in the `search_score_feature_flag` but it felt more readable so I left it in

unit_tests/caseworker/cases/test_denials_view.py
markj0hnst0n: as below, removed flag_off

caseworker/cases/views/denials.py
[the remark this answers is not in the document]
hnryjmes: there are lots of ways to write this that end up being equivalent. 

if I had to choose, I might follow this example from the documentation which shows 3 steps: (1) use `get_context_data(**kwargs)` and store it in a variable `context`, (2) modify the context however you want, (3) return the variable `context`

https://docs.djangoproject.com/en/5.0/topics/class-based-views/generic-display/#adding-extra-context

hnryjmes (approved):

PR #1995, opened 2024-06-03
LTD-5050: Allow editing of quantity, value after adding a product

saruniitr:
## Change description

Allow editing of quantity, value for existing products on the application.
Exporter can edit these when application is in draft state or as a minor edit when it is already submitted.

[LTD-5050](https://uktrade.atlassian.net/browse/LTD-5050)

[LTD-5050]: https://uktrade.atlassian.net/browse/LTD-5050?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


hnryjmes (approved):

exporter/applications/rules.py
```diff
@@ -40,6 +40,16 @@ def is_application_appealed(request, application):
...
 
 
+@rules.predicate
+def is_application_in_draft(request, application):
+    return application and application.status == "draft"
+
+
+@rules.predicate
+def is_application_in_major_edit(request, application):
+    return application and application.status == "applicant_editing"
```
kevincarrogan, line 50: Do we have an enum or constants for these statuses somewhere?

saruniitr: No, they are in API. I will update.

saruniitr: constants defined

kevincarrogan (approved):

PR #1996, opened 2024-06-03
LTD-5039-edit-application

seijihg:
https://uktrade.atlassian.net/browse/LTD-5039

It was a bit hacky to add things to lite-forms as it was, so I rewrote this to Django Form.


saruniitr (approved):
Nice one !
This is mainly content change so there is no need to convert to django forms but you made those changes as well.

hnryjmes (approved):
nice work!

PR #1997, opened 2024-06-04
Update website field validation

hnryjmes:
### Aim

This updates the website validation for exporter organisation registration to allow website URLs that don't have a URL scheme prefix e.g. "https://". This is because user research has shown that exporters will often type a website address that looks like "www.example.com" and this would cause the Django URL validator to throw a ValidationError as it is expecting something like "https://www.example.com" which is a full URL including the scheme prefix. 

Rather than change the form data that is entered, the value stored in LITE is kept the same, but the validator function has an extra step to check if the entered website string is a valid URL with the "https://" prefix added. If so, then the value is considered valid and no error is raised.

[LTD-2285](https://uktrade.atlassian.net/browse/LTD-2285)

[LTD-2285]: https://uktrade.atlassian.net/browse/LTD-2285?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/unit_tests/exporter/core/organisation/test_validators.py
+++ b/unit_tests/exporter/core/organisation/test_validators.py
@@ -2,7 +2,7 @@ import pytest
 
 from django.core.exceptions import ValidationError
 
-from exporter.core.organisation.validators import validate_phone, validate_registration
+from exporter.core.organisation.validators import validate_phone, validate_registration, validate_website
 
 
 @pytest.mark.parametrize(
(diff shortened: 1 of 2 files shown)


seijihg (approved):

unit_tests/exporter/core/organisation/test_validators.py
```diff
@@ -53,3 +53,14 @@ def test_validate_registration_number_valid_numbers(registration_number):
 def test_validate_registration_number_invalid_numbers(registration_number):
     with pytest.raises(ValidationError):
         validate_registration(registration_number)
+
+
+@pytest.mark.parametrize(("website"), ["https://www.example.com", "www.example.com", "example.com"])
```
saruniitr, line 58: Can you add empty string also as it is optional?

hnryjmes: I think validators are only run if the field is nonempty, so if the execution gets to this point then the assumption is that value cannot be an empty string. there is also L41 `if value:` that would catch an empty string.

saruniitr: ok if this is not called if the value is empty then its ok.
Correct L41 fails if it is empty in which case the return value should also be empty which is reason to include that as a test case.

hnryjmes: ok, I've added the empty string test case now

markj0hnst0n (approved):

PR #2005, opened 2024-06-06
Add https to incomplete URLs in org registration

hnryjmes:
### Aim

Previously the exporter organisation registration form was changed to allow incomplete URLs such as `example.com` to pass a custom validator used for the `website` field. However this still causes a problem when the data reaches the backend as the model and db uses `URLField` and so an error just appears in the backend instead.

With this change the custom validator is no longer used and instead a `clean_website` method does something similar for the `RegisterAddressDetailsBaseForm`, including that when a URL does not have a scheme prefix such as `https://`, the `clean_website` method prepends this to the form data and then this is checked using the standard Django URL validator, raising a `ValidationError` if it fails that validator. 

This is probably better than changing the backend and database to treat the website field as plain text. HTTPS has been the default for a few years now and it is unlikely to cause problems if we add `https://` to all incomplete URLs submitted through the exporter org forms.

[LTD-5068](https://uktrade.atlassian.net/browse/LTD-5068)

[LTD-5068]: https://uktrade.atlassian.net/browse/LTD-5068?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/exporter/core/organisation/forms.py
+++ b/exporter/core/organisation/forms.py
@@ -226,12 +226,15 @@ class RegisterAddressDetailsBaseForm(BaseForm):
             try:
                 validator(website)
             except ValidationError:
-                if not website.startswith("http://") or website.startswith("https://"):
-                    website = "https://" + website
-                    try:
-                        validator(website)
-                    except ValidationError:
-                        raise ValidationError("Enter a valid URL.")
+                website = "https://" + website
+                try:
+                    validator(website)
+                except ValidationError:
+                    raise ValidationError("Enter a valid URL.")
+                else:
+                    return website
+            else:
+                return website
 
         return website
 
--- a/exporter/core/organisation/forms.py
+++ b/exporter/core/organisation/forms.py
@@ -1,5 +1,7 @@
 from django import forms
 from django.db import models
+from django.core.validators import URLValidator
+from django.core.exceptions import ValidationError
 
 from crispy_forms_gds.layout import HTML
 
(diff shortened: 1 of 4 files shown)


exporter/core/organisation/forms.py
```diff
@@ -214,7 +215,28 @@ class RegisterAddressDetailsBaseForm(BaseForm):
         validators=[validate_phone],
     )
 
-    website = forms.CharField(label="Website", required=False, validators=[validate_website])
+    website = forms.CharField(label="Website", required=False)
```
depsiatwal, line 218: tors=[URLValidator],
can't we just add URLValidator when we define the field ?

hnryjmes: no, because then we would be where we were originally - an error would be shown if the user typed in an incomplete url like "www.mywebsite.com".

to handle this it's better to use the `clean_` method I think

depsiatwal: I mean is it possible to make a custom validator and then include in the validations. 

similar to validate_sic_number rather then clean method

hnryjmes: it would be possible, but as the point of this change is the modify the value, you would need to have a validator that modified the value. the normal tests for a validator are just that the value is returned unchanged. it's also not good practice to do that kind of change in a validator, but you can do it in a clean method

depsiatwal (approved):
lgtm

markj0hnst0n (approved):

PR #2012, opened 2024-06-06
Ltd 5013 allow pagination to post as forms

kevincarrogan:
### Aim

Allow pagination to work with forms that have to send their payload via a post.

This in cases where the payload itself shouldn't be shared in the URL, as it would with a `get`, so this provides the means to do this via the current `pagination` template tag.

[LTD-5013](https://uktrade.atlassian.net/browse/LTD-5013)

[LTD-5013]: https://uktrade.atlassian.net/browse/LTD-5013?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/caseworker/cases/views/denials.py
+++ b/caseworker/cases/views/denials.py
@@ -86,53 +86,29 @@ class Denials(LoginRequiredMixin, FormView):
     def get_form_kwargs(self):
         kwargs = super().get_form_kwargs()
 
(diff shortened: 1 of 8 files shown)


depsiatwal (approved):
nice :-)

markj0hnst0n (approved):

PR #2013, opened 2024-06-06
[LTD-5055] Amend application by copy

currycoder:
### Aim

This change allows selected exporters to use a new "amendment by copy" flow.

To do this, a case must be in the "Re-opened for changes" status.  When the exporter edits the application using the "major edit" choice a new amendment application will be created which is in draft status for the exporter to adjust and submit.  The original application will still exist in "Superseded by amendment" status.

The caseworker is able to see that an application is an amendment of an existing one through a static banner on the case detail view.  Similarly, the superseded application has a banner that refers to the amendment.

Most of the context/details exist in the companion API PR: https://github.com/uktrade/lite-api/pull/2020

[LTD-5055](https://uktrade.atlassian.net/browse/LTD-5055)

[LTD-5055]: https://uktrade.atlassian.net/browse/LTD-5055?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


conf/exporter.py
```diff
@@ -90,6 +90,7 @@
 SPIRE_URL = "https://www.spire.trade.gov.uk/spire/fox/espire/LOGIN/login"
 
 FEATURE_FLAG_FIREARMS_ENABLED = env.bool("FEATURE_FLAG_FIREARMS_ENABLED", False)
+FEATURE_AMENDMENT_BY_COPY_EXPORTER_IDS = env.list("FEATURE_AMENDMENT_BY_COPY_EXPORTER_IDS", default=[])
```
hnryjmes, line 93: we've not done feature flagging like this before where it's been specific to individual exporters/organisations. could you give me some extra context for why it's needed for this work? I like the idea, just curious why a basic feature flag would not work here.

caseworker/templates/layouts/case.html
```diff
@@ -143,6 +143,32 @@ <h1 id="heading-reference-code" class="app-case-header__heading">
...
 			</div>
 		{% endif %}
+        {% if case.superseded_by %}
+			<div id="superseded-warning" class="app-case-warning-banner-important">
+				<div class="app-case-warning-banner__warning">
+					<span class="app-case-warning-banner__icon" aria-hidden="true">!</span>
+					<span class="app-case-warning-banner__text">
+                        This case has been amended by the exporter and is now superseded.<br/>
+                        {% if case.superseded_by.reference_code %}
+                        The amendment case is <a class="govuk-link govuk-link--inverse" href="{% url 'cases:case' "00000000-0000-0000-0000-000000000001" case.superseded_by.id %}">{{case.superseded_by.reference_code}}</a>.
```
hnryjmes, line 153: you could consider using the constant defined here and injecting it into the template, but what you've done is also fine.

https://github.com/uktrade/lite-frontend/blob/dev/caseworker/core/constants.py#L6

unit_tests/exporter/applications/views/test_edit_application.py
```diff
@@ -38,3 +40,45 @@ def test_edit_application_form_valid_major_edit(requests_mock, authorized_client
...
+    url = reverse("applications:edit_type", kwargs={"pk": application_id})
+
+    amendment_url = f"/applications/{application_id}/amendment/"
+    synthetic_amendment_id = str(uuid.uuid4())
+    requests_mock.post(amendment_url, json={"error": "some whoops"}, status_code=500)
+    response = authorized_client.post(url, {"edit_type": "major"})
+    response_body = BeautifulSoup(response.content, "html.parser")
+
+    assert response.status_code == 200
+    assert "An error occurred" in response_body.text
```
hnryjmes, line 84: is it possible to make a more specific assertion? I'm not sure if the full message from the api would even make it to the frontend, but wondered if it would show any additional text or just `An error occurred`

hnryjmes (approved):
looks good, just minor style comments

conf/exporter.py
currycoder: Yes so it was decided that to help us get this out quicker and be more confident about this, we would make it so that we could turn this on selectively and test it with a small group.  It's not quite determined who we will test it with yet.  But this will allow us to merge this in now and have fairly fine control over testing it.

saruniitr (approved):

unit_tests/exporter/applications/views/test_edit_application.py
currycoder: Yeah TBH it's a pretty generic error page

PR #2017, opened 2024-06-07
LTD-5011 Private individual registration should display CRN number on 'check your answers', not UK VAT number

markj0hnst0n:
### Aim

Changed VAT number to CRN on confirm details screen when registering a private individual

[LTD-5011](https://uktrade.atlassian.net/browse/LTD-5011)

[LTD-5011]: https://uktrade.atlassian.net/browse/LTD-5011?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


exporter/templates/core/registration/includes/registration-details-individual.html
```diff
@@ -28,10 +28,10 @@ <h1 class="govuk-heading-m">
     </div>
     <div class="govuk-summary-list__row" id="org-sic_number">
```
kevincarrogan, line 29: I think the id for this row is probably wrong here too.

markj0hnst0n: It seems to match the pattern in `registration-details-corporation.html` for the CRN number so I followed that.  Is it not correct?

kevincarrogan (approved):

exporter/templates/core/registration/includes/registration-details-individual.html
```diff
@@ -28,10 +28,10 @@ <h1 class="govuk-heading-m">
...
         <dt class="govuk-summary-list__key">
-            UK VAT number
+            CRN number
         </dt>
         <dd class="govuk-summary-list__value">
-            {{ registration_data.vat_number}}
+            {{ registration_data.registration_number}}
```
currycoder, line 34: This looks good, but it would be nice if there was a unit test proving that it worked.  Here's a test that's sort of similar to what I'm thinking about; https://github.com/uktrade/lite-frontend/blob/dev/unit_tests/exporter/applications/views/test_application_summary.py#L96

markj0hnst0n: @currycoder  I think this is already achieved with `test_registration_individual_end_to_end_uk_based` and`test_registration_individual_end_to_end_non_uk_based` [here](https://github.com/uktrade/lite-frontend/blob/dev/unit_tests/exporter/core/organisation/test_organisation_views.py).  I can write a more html specific one if you prefer but these do test the value of that variable, I think.

markj0hnst0n: No worries

markj0hnst0n: changes made @currycoder

currycoder: Nice one!

currycoder (approved):

PR #2021, opened 2024-06-11
LTD-5066-5065-content

seijihg:
https://uktrade.atlassian.net/browse/LTD-5066
https://uktrade.atlassian.net/browse/LTD-5065

### DIFF
--- a/exporter/templates/applications/applications.html
+++ b/exporter/templates/applications/applications.html
@@ -1,4 +1,5 @@
 {% extends 'layouts/base.html' %}
+{% load url_filters %}
 
 {% block back_link %}
 	{% include 'includes/breadcrumbs.html' with title='hub.Tiles.APPLICATIONS' %}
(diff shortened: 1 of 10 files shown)


markj0hnst0n (approved):
The ticket mentions that this change was dependent on LTD-5065 but happy to approve given that 5065's completion is not a blocker for this one.

hnryjmes (approved):

exporter/templates/applications/applications.html
```diff
@@ -87,20 +115,52 @@ <h1 class="govuk-heading-l">
...
+					The most common statuses you'll see are:
+				</div>
+				<ul>
+					<li>submitted - application pending initial checks</li>
+					<li>initial checks - application details checked</li>
+					<li>application editing - if you make changes</li>
+					<li>under review - your products assessed and other details reviewed</li>
+					<li>OGD advice - other government departments (OGD) review your application</li>
+					<li>under final review - final checks</li>
+					<li>finalised - view the decision made on your application</li>
```
hnryjmes, line 156: here I would use en dashes instead of hyphens
https://en.wikipedia.org/wiki/Dash#En_dash

unit_tests/exporter/applications/views/test_applications_list.py
```diff
@@ -0,0 +1,50 @@
...
+            ],
+        },
+    )
+
+
+def test_get_applications(authorized_client, mock_get_applications):
+    url = reverse("applications:applications")
+    response = authorized_client.get(url)
+    assert response.status_code == 200
+    assertTemplateUsed(response, "applications/applications.html")
```
hnryjmes, line 50: I think you could have more testing of the sorting functionality than just a template test

hnryjmes (dismissed):
please add another test for the sorting functionality in addition to your view test that checks the template

unit_tests/exporter/applications/tags/test_url_filters.py
```diff
@@ -0,0 +1,21 @@
...
+        ("http://example.com/applications/", "param1=value1", False),  # Testing with no query params
+        (
+            "http://example.com/applications/?param1=value1&param1=value2",
+            "param1=value2",
+            True,
+        ),  # Multiple same-name params
+    ],
+)
+def test_has_param(url, param, expected_result):
+    assert has_param(url, param) == expected_result
```
hnryjmes, line 21: I don't see what this test is for, it looks like you're only working with test data that you've defined and not testing any other code

hnryjmes: oh, never mind I see it's testing the function `has_param` that you wrote. I had assumed it was some built-in function but I was wrong

hnryjmes (approved):

PR #2023, opened 2024-06-11
Update SIEL declaration page

hnryjmes:
### Aim

The SIEL declaration page has been updated to use a new form and view. The previous declaration page used lite_forms and was difficult to modify so the opportunity was taken to rewrite the declaration page as a Django form using crispy forms.

There are tests for each bit of text on the declaration as this was signed off by DBT legal and data protection (see ticket for details) so it is important that all of this text is checked for accuracy.

[LTD-906](https://uktrade.atlassian.net/browse/LTD-906)


exporter/applications/forms/declaration.py
```diff
@@ -0,0 +1,99 @@
...
+
+class ApplicationDeclarationForm(BaseForm):
+    class Layout:
+        TITLE = "Submit your application"
+        BACK_LINK_TEXT = "Back to check your answers"
+        HEADING_FOI_DISCLOSURE = "Freedom of Information disclosure"
+        HEADING_DECLARATION = "Declaration"
+        SUBMIT_BUTTON_TEXT = "Accept and submit"
+
+    # TODO: we should rename agreed_to_foi to be less confusing. The old
```
depsiatwal, line 21: not sure this is the correct place for all these comments

exporter/applications/views/declaration.py
```diff
@@ -0,0 +1,52 @@
...
+        return context
+
+    @expect_status(HTTPStatus.OK, "Error submitting application", "Unexpected error submitting application")
+    def submit_application(self, request, application_pk, data):
+        return submit_application(request, application_pk, json=data)
+
+    def form_valid(self, form):
+        data = form.cleaned_data
+        # TODO: we should update lite-api so we can remove these extra fields
+        data.update({"submit_declaration": "True", "agreed_to_declaration_text": "I AGREE"})
```
saruniitr, line 50: is this really `"True"` or just boolean `True`?

exporter/applications/forms/declaration.py
```diff
@@ -0,0 +1,99 @@
...
+        We want the foi_reason field to be required if agreed_to_foi is No, but
+        using the default required=True makes it error if this question is never
+        shown. This is a difficulty with using ConditionalRadios. To achieve the
+        desired behaviour we can raise a ValidationError if the No answer is given
+        and foi_reason is left blank which displays a field error to the user.
+        """
+        cleaned_data = super().clean()
+        agreed_to_foi = cleaned_data.get("agreed_to_foi")
+        foi_reason = cleaned_data.get("foi_reason")
+        if agreed_to_foi == "True" and not foi_reason:
```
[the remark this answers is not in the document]
hnryjmes: I did have `clean_foi_reason` initially, but found it wasn't getting run, I think the `clean_` methods aren't run if the form sees no input. 

also passing a dict to `ValidationError()` with the key as the form field means that the error is shown on that field, which I verified myself locally

exporter/applications/forms/declaration.py
hnryjmes: fair point. this is sort of deliberately verbose and over the top; I wrote a lot because I wanted to start a conversation about this in the PR review comments and draw people's attention to it.

I mainly want to know if others think it's alright to keep using this variable `agreed_to_foi` as it is just because it means not making api changes, or if because the question text has changed, it's worth changing how this is handled in the api as well.

https://github.com/uktrade/lite-api/blob/dev/api/applications/creators.py#L201-L205

another option would be just to get rid of the above checks in lite-api completely. especially the `I AGREE` one as it's no longer relevant with this new declaration form.
https://github.com/uktrade/lite-api/blob/dev/api/applications/creators.py#L207-L211

but this might be getting beyond the scope of the Assisted Digital workstream.

exporter/applications/views/declaration.py
hnryjmes: you're right, this can be boolean `True`, I must have got confused with something else I was doing that needed `"True"` the string. I updated it to use `True` as this is more obvious.

actually the api doesn't really care, it can be anything truthy:
https://github.com/uktrade/lite-api/blob/dev/api/applications/views/applications.py#L347

but also see my other comment on this PR, I was thinking that maybe some of these checks in api are unnecessary. originally `submit_declaration` comes from the hidden field defined in the old lite forms declaration form here: https://github.com/uktrade/lite-frontend/blob/dev/exporter/applications/forms/common.py#L69
but it's purpose is unclear to me

exporter/applications/forms/declaration.py
```diff
@@ -0,0 +1,99 @@
...
+                "Failure to do so is an offence."
+            ),
+            HTML.p(
+                "Information provided in this application may be passed to international organisations or other governments in accordance with commitments "
+                "entered into by His Majesty's Government."
+            ),
+            HTML.p(
+                "If ECJU staff have completed this application on your behalf, they based it on the information you provided. You as the exporter are "
+                "responsible for the accuracy of the information."
+            ),
```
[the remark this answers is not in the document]
hnryjmes: this has been updated, there are now two template includes with all of the same content.

exporter/applications/views/declaration.py
```diff
@@ -0,0 +1,52 @@
...
+    form_class = ApplicationDeclarationForm
+    template_name = "core/form.html"
+
+    def dispatch(self, request, **kwargs):
+        try:
+            self.application = get_application(request, kwargs["pk"])
+        except HTTPError:
+            raise Http404()
+
+        return super().dispatch(request, **kwargs)
```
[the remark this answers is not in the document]
hnryjmes: yes, this is me copying a pattern that looked safe and not really thinking about it. have updated so we just use set pk from kwargs and use that, we don't really need the application object for any other reason.

exporter/applications/views/declaration.py
```diff
@@ -0,0 +1,52 @@
...
+
+from core.auth.views import LoginRequiredMixin
+from core.decorators import expect_status
+from exporter.applications.forms.declaration import ApplicationDeclarationForm
+from exporter.applications.services import get_application, submit_application
+
+
+class ApplicationDeclarationView(LoginRequiredMixin, FormView):
+    form_class = ApplicationDeclarationForm
+    template_name = "core/form.html"
```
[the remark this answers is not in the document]
hnryjmes: I'm still unsure about this, for now I've rewritten to use 2 partial templates that we load with `render_to_string` but I could be persuaded to rewrite to use a separate template. I wanted to try to make use of `BaseForm` as much as possible as this is quite a simple form layout
https://github.com/uktrade/lite-frontend/blob/dev/core/common/forms.py#L17

exporter/applications/forms/declaration.py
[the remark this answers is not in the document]
kevincarrogan: We also have a component https://github.com/uktrade/lite-frontend/blob/900e7afbc996fdcbdd5794dea77412477e1f4cc6/caseworker/cases/forms/denial_forms.py#L45 for this if necessary.

exporter/applications/forms/declaration.py
```diff
@@ -0,0 +1,86 @@
...
+        We want the foi_reason field to be required if agreed_to_foi is No, but
+        using the default required=True makes it error if this question is never
+        shown. This is a difficulty with using ConditionalRadios. To achieve the
+        desired behaviour we can raise a ValidationError if the No answer is given
+        and foi_reason is left blank which displays a field error to the user.
+        """
+        cleaned_data = super().clean()
+        agreed_to_foi = cleaned_data.get("agreed_to_foi")
+        foi_reason = cleaned_data.get("foi_reason")
+        if agreed_to_foi == "True" and not foi_reason:
```
saruniitr, line 62: I think if you use `coerce_str_to_bool` then you can compare with `True` rather than string `'True'`,
eg https://github.com/uktrade/lite-frontend/blob/dev/caseworker/advice/forms.py#L293

saruniitr (changes requested):
one last comment

exporter/applications/forms/declaration.py
hnryjmes: you're right, this is a good change, have updated to use `forms.TypedChoiceField` so I can use `coerce=`, this is used a lot in exporter already so makes sense to continue that pattern

hnryjmes: also changed to use `self.add_error` for the same reason, makes it more consistent

exporter/applications/forms/declaration.py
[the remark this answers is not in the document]
hnryjmes: have updated to use `self.add_error()`

exporter/applications/forms/declaration.py
hnryjmes: this has now been updated to be a shorter comment, as I think it was too long to begin with

exporter/applications/forms/declaration.py
```diff
@@ -0,0 +1,80 @@
...
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+    def get_layout_fields(self):
+        return [
+            HTML(
+                render_to_string(
+                    "applications/forms/declaration/declaration-heading-question.html",
+                    context={"HEADING_FOI_DISCLOSURE": self.Layout.HEADING_FOI_DISCLOSURE},
```
kevincarrogan, line 70: What's the benefit of passing this into the template? Isn't this template just used for this one purpose and could just be hardcoded in the template itself?

hnryjmes: I think not much benefit, this is probably left over from when I had my tests written a different way, but now the tests just define string literals and don't make use of these variables anyway, I can remove these and make it simpler

exporter/applications/forms/declaration.py
```diff
@@ -0,0 +1,76 @@
...
+        """
+        cleaned_data = super().clean()
+        agreed_to_foi = cleaned_data.get("agreed_to_foi")
+        foi_reason = cleaned_data.get("foi_reason")
+        if agreed_to_foi and not foi_reason:
+            self.add_error("foi_reason", "Explain why the disclosure of information would be harmful to your interests")
+        return cleaned_data
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
```
saruniitr, line 61: this can be removed now

saruniitr (approved):
one minor comment

seijihg (approved):

PR #2025, opened 2024-06-12
Update is_onward_exported question text

hnryjmes:
### Aim

The onward exported question has been changed to ask about ultimate end-users instead, as this is more correct. There are sometimes cases where a good is onward exported to an ultimate end-user in the same country, and the existing question wording means exporters cannot answer properly if that is their situation.

Because the question is still asking about goods being onward exported just in a different way, it seems fine to keep the `is_onward_exported` variable the same and not change that in other places. 

[LTD-5053](https://uktrade.atlassian.net/browse/LTD-5053)

[LTD-5053]: https://uktrade.atlassian.net/browse/LTD-5053?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/caseworker/tau/templates/tau/firearms_accessory_table.html
+++ b/caseworker/tau/templates/tau/firearms_accessory_table.html
@@ -91,7 +91,7 @@
           </td>
         </tr>
         <tr class="govuk-table__row">
-          <th class="govuk-table__header">Will the product be onward exported to any additional countries?</th>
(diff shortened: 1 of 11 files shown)


depsiatwal (approved):
LGTM

saruniitr (approved):

PR #2035, opened 2024-06-18
LTD-5049: Add option to archive restore a product

saruniitr:
## Change description

To help exporters better manage their product list add option to archive products that are already submitted on an application. Verified products also can be archived.
Once archived they will be hidden from the product list but the exporter can always restore it if required.

Exporter can archive/restore products any number of times and each action is recorded.
Product details page also displays archive history.

These actions are managed using a single view but separate forms are created for archive/restore as the content is different.

[LTD-5049](https://uktrade.atlassian.net/browse/LTD-5049)

[LTD-5049]: https://uktrade.atlassian.net/browse/LTD-5049?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


exporter/applications/rules.py
```diff
@@ -63,3 +64,37 @@ def is_application_in_major_edit(request, application):
 )
 
 rules.add_rule("can_edit_quantity_value", is_application_in_draft | is_application_in_major_edit)  # noqa
+
+
+# Rules for Goods
+# TODO: Move to rules.py under goods/
```
saruniitr, line 70: there is an issue with the way goods app is registered and not recognising rules.
Before merging I will try to fix that and move under goods/

core/goods/apps.py
```diff
@@ -0,0 +1,6 @@
+from django.apps import AppConfig
+
+
+class CoreGoodsConfig(AppConfig):
+    name = "core.goods"
+    label = "core_goods"
```
saruniitr, line 6: defining a unique label as it is clashing with `goods` app.
Without this the rules are not registered correctly

core/summaries/summaries.py
```diff
@@ -775,3 +777,39 @@ def get_summary_type_for_good_on_application(good_on_application):
     item_category = item_category["key"]
 
     return get_summary_type_from_item_category(item_category)
+
+
+def get_summary_url_for_good(good):
```
saruniitr, line 782: When archiving/restoring we ask for user confirmation.
If they cancel they need to be redirected to product details page, this helper is useful to get the url based on product type.

exporter/goods/common/views.py
```diff
@@ -0,0 +1,51 @@
...
+
+from core.auth.views import LoginRequiredMixin
+from core.decorators import expect_status
+from core.summaries.summaries import get_summary_url_for_good
+from exporter.goods.forms import GoodArchiveForm, GoodRestoreForm
+from exporter.goods.services import get_good, edit_good
+
+
+class GoodArchiveRestoreView(LoginRequiredMixin, FormView):
+    template_name = "core/form.html"
```
saruniitr, line 13: this view manages both archive/restore but the forms are separate because they have different content.
View is kept the same because the action is same in both cases and doesn't require any user input.

exporter/goods/common/base.py
```diff
@@ -48,4 +49,19 @@ def get_context_data(self, **kwargs):
...
+            user = f"{item['user']['first_name']} {item['user']['last_name']}"
+            if item["is_archived"]:
+                archive_history.append(
+                    f"Archived by {user} at {date_obj.strftime('%H:%M')} on {date_obj.strftime('%d %B %Y')}"
+                )
+            else:
+                archive_history.append(
+                    f"Restored by {user} at {date_obj.strftime('%H:%M')} on {date_obj.strftime('%d %B %Y')}"
+                )
+
```
hnryjmes, line 64: you could consider moving this out of `get_context_data()` and into a separate function

markj0hnst0n (approved):

hnryjmes (approved):

exporter/goods/common/views.py
```diff
@@ -0,0 +1,51 @@
...
+        return kwargs
+
+    def is_archiving(self):
+        return self.kwargs["action"] == "archive"
+
+    def get_form(self):
+        if self.is_archiving():
+            return GoodArchiveForm(**self.get_form_kwargs())  # noqa
+        else:
+            return GoodRestoreForm(**self.get_form_kwargs())  # noqa
```
currycoder, line 32: Would suggest that this could be a bit cleaner with a base class/two inherited concrete View classes

currycoder: Kind of a minor though, but in general I do think we should get away from peppering ifs and in to leaning on OOP a bit more

exporter/goods/rules.py
```diff
@@ -0,0 +1,33 @@
...
+def is_verified_product(request, product):
+    return product and product["status"]["key"] == GoodStatus.VERIFIED
+
+
+@rules.predicate
+def is_archived_product(request, product):
+    return product and product["is_archived"] is True
+
+
+rules.add_rule("can_delete_product", is_draft_product)
```
currycoder, line 26: Nice!

exporter/goods/services.py
```diff
@@ -14,6 +14,13 @@ def get_goods(
     data = client.get(request, "/goods/" + convert_parameters_to_query_params(locals()))
     return data.json()
 
+def get_archived_goods(
+    request, page: int = 1, name=None, part_number=None, control_list_entry=None, for_application=None
+):
+    response = client.get(request, "/goods/archived-goods/" + convert_parameters_to_query_params(locals()))
```
currycoder, line 20: Can we get away from `locals()` here?

currycoder (commented):
Could I suggest a follow-up ticket for E2E tests too?

exporter/goods/common/views.py
saruniitr: Yes I switched back and forth about splitting/combining views and combined mainly because there is no user input.
In both cases they click a button and takes you to a confirmation page with a different content.

I will split them for future proofing.

saruniitr:
> Could I suggest a follow-up ticket for E2E tests too?

Yes, created a ticket already for E2E tests.

PR #2036, opened 2024-06-18
Update party website forms to use same help text and validation

hnryjmes:
### Aim

Adds the same help text and validation to the end user website form and party website form (this means it will update for adding an ultimate end-user, third party, consignee) as was already done for the exporter org registration form. 

[LTD-5124](https://uktrade.atlassian.net/browse/LTD-5124)

[LTD-5124]: https://uktrade.atlassian.net/browse/LTD-5124?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/exporter/applications/forms/parties.py
+++ b/exporter/applications/forms/parties.py
@@ -14,15 +14,7 @@ from exporter.core.services import get_countries
 from lite_content.lite_exporter_frontend import strings
 from lite_content.lite_exporter_frontend.applications import PartyForm, PartyTypeForm
 from lite_forms.common import country_question
-from lite_forms.components import (
-    BackLink,
-    RadioButtons,
-    Form,
-    Option,
-    TextArea,
-    TextInput,
-    FormGroup,
-)
+from lite_forms.components import BackLink, RadioButtons, Form, Option, TextArea, TextInput, FormGroup, Label
 from lite_forms.generators import confirm_form
 
 
@@ -66,7 +58,7 @@ def party_name_form(title, button):
 def party_website_form(title, button):
     return Form(
         title=title,
-        questions=[TextInput("website")],
+        questions=[Label("Use the format https://www.example.com", classes=["govuk-hint"]), TextInput("website")],
         default_button_name=button,
     )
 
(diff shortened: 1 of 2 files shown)


exporter/applications/forms/parties.py
```diff
@@ -235,7 +228,28 @@ class Layout:
...
+            try:
+                validator(website)
+            except ValidationError:
+                website = "https://" + website
+                try:
+                    validator(website)
+                except ValidationError:
+                    raise ValidationError("Enter a valid URL.")
+                else:
+                    return website
```
saruniitr, line 248: nested try-catch not a good idea. As we are showing the expected format using help text, I think it is better to validate just once and fail. Once user sees the error then they will correct as per expected format.

By prepending `https://` in the case where they just entered `www.example.com` then we are going against the expected format.

hnryjmes: I want to push back on this slightly. Part of this ticket is to be consistent in how we handle URLs entered by the user. So this code is copied from the check we now have in exporter for registering an organisation: https://github.com/uktrade/lite-frontend/blob/dev/exporter/core/organisation/forms.py#L220-L239

I even thought about extracting that logic out into a reusable function that I could then use here, to keep things DRY, but when I tried that it did not save many lines of code so I didn't do that in the end.

A lot of the point of this Assisted Digital work is that we have to not assume the user will always know what we intend them to do. Hence this "belt and braces" approach where we don't assume that just because the help text is there the user won't still enter a website like `www.example.com`.

hnryjmes: To only validate once you need to write a custom validator, and then if you use that, you can't have the validator change the URL input, because validators aren't meant to have side effects. So the try-catch approach to both validate (using the built-in URL validator) and then possibly modify the value is what I ended up with.

saruniitr (approved):

markj0hnst0n (approved):

exporter/applications/forms/parties.py
kevincarrogan: If you want this to be consistent you could just have a function somewhere that does this without it being a validator.

PR #2038, opened 2024-06-19
Update unused import check to use ruff

hnryjmes:
### Aim

Previously we have explored using `ruff` as a replacement for all our pre-commit linting and formatting needs. This is because it is extremely fast, so it's possible to eventually have our CI checks and local checks be exactly the same. In this end scenario, the CI check would literally just run `pre-commit` so each check would be identical to what is checked locally. 

However, instead of migrating all of our existing rules over to use `ruff` in one go, we now think that we want to add individual rules and rule sets slowly to make sure there are no big code changes associated with linting and formatting that become hard to manage. This is also the recommended approach in the docs here: https://docs.astral.sh/ruff/linter/#rule-selection (normally you would add rule sets not individual rules; e.g. to add all the pyflakes rules you would add `"F"` to `pyproject.toml`)

With that in mind, this PR changes the existing unused import check which used `autoflake` to use `ruff`. To do this we simply add rule `F401` to the `pyproject.toml` and make sure that `ruff` is run with the `--fix` argument at the top of our `pre-commit` config file.

Eventually we can find time to move all of our existing linting and formatting rules over to `ruff` but for the time being it is ok to use multiple tools (`ruff` and `black` and related tools) in parallel given there is a negligible time cost for the additional `ruff` hook.


kevincarrogan (approved):

hnryjmes (commented):
note to self: don't merge this until Kevin has had a chance to test something

pyproject.toml
```diff
@@ -21,3 +21,6 @@ exclude = '''
   | dist
 )/
 '''
+
+[tool.ruff.lint]
+select = ["F401"]
```
currycoder, line 26: Is this the mechanism that restricts this to just tackling unused imports right now?

currycoder (approved):

pyproject.toml
hnryjmes: yes, you can add rules by updating `select` or remove rules by updating `ignore`. there are some nice examples of what you might end up with later on if you scroll down on this page: https://docs.astral.sh/ruff/linter/#rule-selection

PR #2040, opened 2024-06-19
 LTD-4966-add-dbt-healthcheck 

depsiatwal:
https://uktrade.atlassian.net/browse/LTD-4966

This is using the new platform format health check for use on dbt platform. 

Normally the UI checks if your services are also up i.e DB. 

But since we have a separate API the direction i'm proposing and to avoid confusion when we get alerts

https://platform.readme.trade.gov.uk/migrations/conducting-a-migration/02b-codebase-preparation/#logging

- Caseworker - webserver only health check
- FrontEnd -  webserver only health check
- API - webserver and database health. check 
- API bac end web services health check REDIS/Celery etc using recommend BE services check for DBT platform using The dbt-copilot-python package

https://github.com/uktrade/copilot-python?tab=readme-ov-file#celery-health-check


markj0hnst0n (approved):

caseworker/urls.py
```diff
@@ -2,12 +2,14 @@
...
 
 import caseworker.core.views
+from core.health_check.views import pingdom
 
 from core.accessibility.views import CaseworkerAccessibilityStatementView
 
 
 urlpatterns = [
     path("healthcheck/", include("health_check.urls")),
+    path("health_check/pingdom/", pingdom, name="healthcheck-pingdom"),
```
currycoder, line 12: Why do we need a whole new healthcheck?  Don't we already have one that's wired up with pingdom in the URL/view above?

depsiatwal: i've overridden as recommend here 
https://pypi.org/project/django-health-check/
 however looks like it still needs to be included in the urls. 

is that what you had in mind ?

core/templates/healthcheck/pingdom.xml
```diff
@@ -0,0 +1,13 @@
...
+<?xml version="1.0" encoding="UTF-8"?>
+<pingdom_http_custom_check>
+    <status>
+    {% for plugin in errored_plugins %}
+        {{ plugin.identifier }}: {{ plugin.pretty_status }}
+    {% empty %}
+        OK
+    {% endfor %}
+    </status>
+    <response_time>{{ total_response_time }}</response_time>
```
currycoder, line 11: Is there a document somewhere that explains the format the consumer of this is looking for?

depsiatwal: Yeah DBT reffer to this 

https://documentation.solarwinds.com/en/success_center/pingdom/content/topics/http-custom-check.htm

depsiatwal: same pattern used on other platforms that have migrated

hnryjmes (approved):

PR #2045, opened 2024-06-21
LTD-4963 Update logging to use dbt copilot python helper

markj0hnst0n:
### Aim

Updated logging to output simple to local env, ECS format to Gov Paas and ASIM format to DBT Platform

[LTD-4963](https://uktrade.atlassian.net/browse/LTD-4963)

[LTD-4963]: https://uktrade.atlassian.net/browse/LTD-4963?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


depsiatwal (approved):
lgtm

Pipfile
```diff
@@ -70,7 +70,6 @@ django-redis = "~=4.12.1"
 notifications-python-client = "~=6.4.1"
 python-magic = "~=0.4.27"
 django-log-formatter-ecs = "==0.0.5"
-importlib-metadata = "==4.6.3"  # Dependency for some packages. Might not be needed for Python3.8+
```
hnryjmes, line 73: nice

conf/base.py
```diff
@@ -202,20 +203,25 @@
...
-    "handlers": {
-        "stdout": {"class": "logging.StreamHandler", "formatter": "simple"},
-        "ecs": {"class": "logging.StreamHandler", "formatter": "ecs_formatter"},
-        "asim": {"class": "logging.StreamHandler", "formatter": "asim_formatter"},
-    },
-    "root": {"handlers": ["stdout", "ecs", "asim"], "level": env.str("LOG_LEVEL", "info").upper()},
 }
+ENVIRONMENT = env.str("ENVIRONMENT", "")
+
+if ENVIRONMENT == "local":
```
hnryjmes, line 209: out of interest, what happens if the logging dict isn't set correctly? suppose there's a situation where `ENVIRONMENT=local` is not set in the local env files, and so this line evaluates to false - how bad is that?

hnryjmes (approved):

conf/base.py
markj0hnst0n: @hnryjmes The `else` case is used to output ASIM format for logs as that was the default behaviour I was going for overall.  Same pattern is used in the other apps.

PR #2046, opened 2024-06-24
Fix FOI answer shown in case details tab

hnryjmes:
### Aim

Updates the FOI question and answer shown in the case details tab in caseworker now that the SIEL declaration has been updated in exporter.

[LTD-906](https://uktrade.atlassian.net/browse/LTD-906)

[LTD-906]: https://uktrade.atlassian.net/browse/LTD-906?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core/builtins/custom_tags.py
+++ b/core/builtins/custom_tags.py
@@ -227,7 +227,10 @@ def default_na(value):
 @register.filter()
 def get_agreed_to_foi_value(boolean):
     """
-    Returns 'No' if the boolean is equal to True, else 'Yes'
+    Returns 'No' if the boolean is equal to some variant of True, else 'Yes'.
+    This is needed as long as the agreed_to_foi model field has yet to be
+    updated in lite-api. For more context see the agreed_to_foi question in
+    exporter: exporter/applications/forms/declaration.py
     """
     if boolean is True or boolean == "true" or boolean == "True" or boolean == "yes" or boolean == "Yes":
         return "No"
--- a/core/builtins/custom_tags.py
+++ b/core/builtins/custom_tags.py
@@ -225,14 +225,14 @@ def default_na(value):
 
 
 @register.filter()
-def get_agreed_to_foi_value(boolean):
+def get_agreed_to_foi_display_value(value):
     """
-    Returns 'No' if the boolean is True, else 'Yes'.
+    Returns 'No' if the boolean value is True, else 'Yes'.
     This is needed as long as the agreed_to_foi model field has yet to be
     updated in lite-api. For more context see the agreed_to_foi question in
     exporter: exporter/applications/forms/declaration.py
     """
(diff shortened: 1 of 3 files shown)


markj0hnst0n (approved):

core/builtins/custom_tags.py
```diff
@@ -224,6 +224,20 @@ def default_na(value):
         return mark_safe(f'<span class="govuk-hint govuk-!-margin-0">N/A</span>')  # nosec
 
 
+@register.filter()
+def get_agreed_to_foi_value(boolean):
+    """
```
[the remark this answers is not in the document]
hnryjmes: you're right, have updated this and the test. now that `agreed_to_foi` is coerced to a boolean the tag does not have to do very much.

core/builtins/custom_tags.py
```diff
@@ -224,6 +224,17 @@ def default_na(value):
...
+    Returns 'No' if the boolean is True, else 'Yes'.
+    This is needed as long as the agreed_to_foi model field has yet to be
+    updated in lite-api. For more context see the agreed_to_foi question in
+    exporter: exporter/applications/forms/declaration.py
+    """
+    return "No" if boolean is True else "Yes"
```
saruniitr, line 235: this looks inverted to me,

if `agreed_to_foi` is `True` then we should be returning `Yes` isn't it?
`return "Yes" if agreed_to_foi else "No"` ??

Instead of variable name `boolean` it can be `value` which sounds better.

hnryjmes: There is some more explanation here: https://github.com/uktrade/lite-frontend/blob/dev/exporter/applications/forms/declaration.py#L20-L26

I don't think that the boolean was well named to begin with, I think it should have been called something like `agreed_that_foi_harmful` where `True` would then make sense as the answer that leads to the text box showing to let the exporter add the explanation. But that was the old question.

Now, the question in the declaration has actually changed, and the new question is `Do you agree to make the application details publicly available?` where answering "No" means the text box should be shown. This is how we get the weird choices tuple of `choices=((False, "Yes"), (True, "No"))` which you see here: https://github.com/uktrade/lite-frontend/blob/dev/exporter/applications/forms/declaration.py#L28

Instead of updating the api to change the sense of the boolean (the business meaning of True / False for that field `agreed_to_foi`) I/we chose to keep the boolean the same as I didn't want to have to migrate lots of old cases which I think we would have had to do if the model was changed. 

So right now it's a bit confusing but I think I have it the right way around. If there's more time we can update the api and have a newly named field that tracks the new question.

saruniitr (approved):

PR #2047, opened 2024-06-24
LTD-4960-application-settings 

depsiatwal:
https://github.com/uktrade/lite-frontend/tree/LTD-4960-application-settings

### DIFF
--- a/conf/base.py
+++ b/conf/base.py
@@ -6,6 +6,7 @@ from sentry_sdk.integrations.django import DjangoIntegration
 from django_log_formatter_ecs import ECSFormatter
 from django_log_formatter_asim import ASIMFormatter
 from dbt_copilot_python.utility import is_copilot
+from dbt_copilot_python.network import setup_allowed_hosts
 
 from django.urls import reverse_lazy
 
@@ -29,6 +30,13 @@ SECRET_KEY = env.str("DJANGO_SECRET_KEY")
 DEBUG = env.bool("DEBUG", False)
 
 ALLOWED_HOSTS = env.list("ALLOWED_HOSTS", default=[])
+
+ENVIRONMENT = env.str("ENVIRONMENT", "")
+VCAP_SERVICES = env.json("VCAP_SERVICES", {})
+
+
+IS_ENV_DBT_PLATFORM = is_copilot()
+IS_ENV_GOV_PAAS = bool(VCAP_SERVICES)
 
 # django-allow-cidr
 ALLOWED_CIDR_NETS = ["10.0.0.0/8"]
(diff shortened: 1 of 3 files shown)


markj0hnst0n (approved):

conf/base.py
```diff
@@ -205,19 +206,24 @@
...
     LOGGING.update({"handlers": {"ecs": {"class": "logging.StreamHandler", "formatter": "ecs_formatter"}}})
     LOGGING.update({"root": {"handlers": ["ecs"]}})
 
+elif VCAP_SERVICES:
+    REDIS_URL = VCAP_SERVICES["redis"][0]["credentials"]["uri"]
 else:
+    REDIS_URL = env.str("REDIS_URL", "")
     LOGGING.update({"formatters": {"asim_formatter": {"()": ASIMFormatter}}})
     LOGGING.update({"handlers": {"asim": {"class": "logging.StreamHandler", "formatter": "asim_formatter"}}})
     LOGGING.update({"root": {"handlers": ["asim"]}})
```
currycoder, line 229: I was concerned this was becoming a mess in a previous review and I think the problem has been realised with these changes.

The middle elif block is now a combination of settings for gov paas/DBT platform which is wrong.

There's now 3 blocks for 4 target environments so I don't get that.

I really like `IS_DBT_PLATFORM`, I think that's the route we should go down to make this cleaner e.g.

```
IS_ENV_LOCAL = ENVIRONMENT == "local"
IS_ENV_DBT_PLATFORM = is_copilot()
IS_ENV_GOV_PAAS = bool(VCAP_SERVICES)

# Bail out with an exception if the environment is not as we expect
assert [IS_ENV_LOCAL, IS_ENV_DBT_PLATFORM, IS_ENV_GOV_PAAS].count(True) == 1, "local/dbt platform/gov paas environment could not be reliably identified"

if IS_ENV_LOCAL:
    # set up local stuff
    ...

if IS_ENV_DBT_PLATFORM:
    # set up DBT platform stuff
    ...

if IS_ENV_GOV_PAAS:
    # set up gov paas stuff
    ...
```

conf/base.py
```diff
@@ -205,19 +206,24 @@
     "disable_existing_loggers": False,
 }
 ENVIRONMENT = env.str("ENVIRONMENT", "")
+IS_DBT_PLATFORM = is_copilot()
 
 if ENVIRONMENT == "local":
```
hnryjmes, line 211: similar to Brendan's comment, I think we only need to think about 3 cases - the local case, the DBT platform case, and the VCAP_SERVICES case. that's what it looks like on the lite-api PR as well. either we can achieve that with 3 if-statements as Brendan suggests, or an if-elif-else. probably we don't need if-elif-elif-else

conf/base.py
depsiatwal: yeah agree it's getting a bit messy moved to above pattern hopefully it's a little cleaner 

also remember GOV PASS will be gone soon so it will be just two envs

conf/base.py
depsiatwal: new changes hopefully cleaner will move the others to same pattern if this ones ok

conf/base.py
depsiatwal: as I was thinking earlier I think we have 4 env 
GOV_PASS
DBT_PLATFORM 
Local and ciricle CI 

Hence the above now fails for ciricle CI  having said that I think these two can be identical hence this can be our fallback case.

currycoder (approved):

Tllew (approved):
Looks good to me

PR #2052, opened 2024-06-27
LTD-5071-forward-slash-breaking-search 

depsiatwal:
https://uktrade.atlassian.net/browse/LTD-5071

This is a scenario where / break query_string ES searching 
replacing with double // appears to fix the issue and allows us to match as expected. 

it appears to be a real waky way to fix it but ATM it works. Other option may be to put the hack into API instead.


seijihg (approved):

depsiatwal:
This can be closed as we deal with this on the API now

PR #2053, opened 2024-06-28
LTD: Task list is only for applications that can be edited

saruniitr:
### Aim


hnryjmes (commented):
the fact that this change has caused a lot of existing test failures for the application task list is probably a good thing, it makes me think this was a good change to make

kevincarrogan (approved):

PR #2056, opened 2024-07-01
Add error for failed api request

kevincarrogan:
### Aim

Add error checking for cases where the API received a bad request back when editing a product.


saruniitr (approved):

currycoder (approved):

currycoder (commented):
Agree with the changes, but I do think we should probably have tests that exercise the paths of these decorators.

PR #2057, opened 2024-07-01
LTD-5072 remove safe from top form and newline from address

Tllew:
quick fix for un needed safe in the template and remove newline from address

[LTD-5072](https://uktrade.atlassian.net/browse/LTD-5072)

[LTD-5072]: https://uktrade.atlassian.net/browse/LTD-5072?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


caseworker/cases/views/denials.py
```diff
@@ -53,8 +53,9 @@ def get_initial_search_params(self):
         country_list = set()
 
         for party in self.parties_to_search:
+            address = party["address"].replace("\n", "")
```
depsiatwal, line 56: might wanna add a quick test for this function 
probably a gap here anyway

seijihg (approved):

unit_tests/caseworker/conftest.py
```diff
@@ -2769,3 +2769,578 @@ def mock_roles(requests_mock):
         url=url,
         json={"roles": [{"id": SUPER_USER_ROLE_ID, "name": "Super User"}]},
     )
+
+
+@pytest.fixture
+def data_newline_enduser_address(
```
depsiatwal, line 2775: quite a lot to put conftest.py i'm conscious this is polluting the file. 
it's quite long to just add new line to end user.

saruniitr, line 2775: All of this is essentially `data_standard_case` fixture. Is it not an option to take that as an argument and update relevant fields in `end_user` and use this fixture in the test?

Tllew: done

Tllew: Yeah wasn't sure if it looked neater to just import a new case, wasn't sure the exact fields it needed, I've modified data_standard_case instead

PR #2058, opened 2024-07-02
LTD-5153: Disable Good archive/restore temporarily

saruniitr:
## Change description

Some of the recent changes breaks this feature.
This will be re-enabled after reworking all the changes (API and FE).

The options to archive/restore are linked to rules so we are just disabling the rules now.

### DIFF
--- a/exporter/goods/rules.py
+++ b/exporter/goods/rules.py
@@ -27,7 +27,7 @@ rules.add_rule("can_delete_product", is_draft_product)
 
 rules.add_rule(
     "can_archive_product",
-    (is_submitted_product | is_verified_product) & ~is_archived_product,  # noqa
+    lambda _: False,  # noqa
 )
 
-rules.add_rule("can_restore_product", is_archived_product)  # noqa
+rules.add_rule("can_restore_product", lambda _: False)  # noqa
--- a/unit_tests/exporter/goods/views/test_archive_restore.py
+++ b/unit_tests/exporter/goods/views/test_archive_restore.py
@@ -138,7 +138,7 @@ def test_submitted_product_details_context(
     assert response.status_code == 200
     soup = BeautifulSoup(response.content, "html.parser")
 
-    assert "Archive product" in soup.find("a", {"id": "archive-good"}).text
+    assert soup.find("a", {"id": "archive-good"}) is None
     assert soup.find("a", {"id": "restore-good"}) is None
     assert soup.find("a", {"id": "delete-good"}) is None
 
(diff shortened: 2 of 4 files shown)


unit_tests/exporter/goods/views/test_archive_restore.py
```diff
@@ -207,7 +207,7 @@ def test_restore_product_asks_for_confirmation(
     assert soup.find("a", {"id": "cancel-id-cancel"})["href"] == firearm_product_details_url
 
 
-def test_archived_product_details_context(
+def archived_product_details_context(
```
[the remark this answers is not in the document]
saruniitr: Yes but as the changes are merged I wanted to have tests that check that these buttons are indeed not available hence updated the tests to have some coverage.
Also skipping them might rise coverage errors.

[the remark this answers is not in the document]
saruniitr: If the API change is merged then this PR is not necessary or can just be updated to use the new endpoint.

exporter/goods/rules.py
```diff
@@ -27,7 +27,7 @@ def is_archived_product(request, product):
 
 rules.add_rule(
     "can_archive_product",
-    (is_submitted_product | is_verified_product) & ~is_archived_product,  # noqa
+    lambda _: False,  # noqa
```
seijihg, line 30: Could not we just disabled the button instead of changing the rule to always return False here?

I guess does not matter at the end of day.

seijihg (approved):

exporter/goods/rules.py
saruniitr: If these rules are not there that could be an option (disable button).
If we just disable buttons now then we get coverage errors.

exporter/goods/common/views.py
```diff
@@ -41,9 +41,8 @@ class GoodArchiveView(GoodArchiveRestoreBaseView):
     form_class = GoodArchiveForm
 
     def form_valid(self, form):
-        data = {"is_archived": True}
 
-        edit_good(self.request, self.good_id, data)
+        archive_restore_good(self.request, self.good_id, True)
```
currycoder, line 45: Could I suggest making the last argument a kwarg for readability; e.g.
`archive_restore_good(self.request, self.good_id, is_archived=True)`

exporter/goods/services.py
```diff
@@ -351,3 +351,10 @@ def post_technology(request, json):
 def edit_technology(request, pk, json):
     response = client.put(request, f"/goods/{pk}", json)
     return response.json(), response.status_code
+
+
+def archive_restore_good(request, pk, status):
```
currycoder, line 356: Sort of in tandem with the above but `is_archived` I feel is much more readable and explicit than `status`

currycoder (approved):
A couple of minor suggestions inline for readability.  Otherwise this looks good.  I assume the plan is to do our usual release sequence and ensure that the dependent API endpoint is up first before deploying frontend?

saruniitr:
> A couple of minor suggestions inline for readability. Otherwise this looks good. I assume the plan is to do our usual release sequence and ensure that the dependent API endpoint is up first before deploying frontend?

Yes that's correct. API change is already merged so both will go out along with other tickets as usual.

seijihg (approved):

PR #2060, opened 2024-07-03
Initiating major edits

kevincarrogan:
### Aim

### DIFF
--- a/core/constants.py
+++ b/core/constants.py
@@ -50,8 +50,7 @@ class CaseStatusEnum:
 
     @classmethod
     def all(cls):
-        is_all_upper = re.compile(r"^[A-Z_]+$")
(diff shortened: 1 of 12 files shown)


core/constants.py
```diff
@@ -48,22 +48,9 @@ class CaseStatusEnum:
...
-        return status in [
-            cls.CLOSED,
-            cls.DEREGISTERED,
-            cls.FINALISED,
-            cls.REGISTERED,
-            cls.REVOKED,
-            cls.SURRENDERED,
-            cls.WITHDRAWN,
-        ]
-
```
hnryjmes, line 62: glad to see this be removed (which I assume is because whether a status is terminal is already defined in the api)

hnryjmes (approved):
approving because I think the e2e test failures are to be expected. we can test this on support env to gain more confidence but that shouldn't stop us from merging this now.

PR #2063, opened 2024-07-05
LTD-5136: Add confirmation page when major edit is selected

saruniitr:
## Change description

Major edit creates a new application by copying the current application. This is enabled for specific exporters and in such cases we ask for their confirmation before creating amendment. This PR adds this new confirmation page for these exporters.

All other exporters see existing confirmation page which follows existing way of doing major edits.

Add tests that check for each category and ensure they redirect to correct success urls in each case.

[LTD-5136](https://uktrade.atlassian.net/browse/LTD-5136)

[LTD-5136]: https://uktrade.atlassian.net/browse/LTD-5136?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/exporter/applications/forms/common.py
+++ b/exporter/applications/forms/common.py
@@ -82,7 +82,7 @@ class ApplicationMajorEditConfirmationForm(BaseForm):
     def get_layout_fields(self):
         return [
             HTML.p(
-                "Progress on the case will stop while you are making changes. The application will appear in your draft until you re-submit."
+                "Progress on the case will stop while you are making changes. The application will appear in your drafts until you re-submit."
             ),
             HTML.p(
                 "Re-submitting the application with changes means it will take longer to process and will have a new ECJU case reference."
--- a/exporter/applications/forms/common.py
+++ b/exporter/applications/forms/common.py
@@ -1,10 +1,11 @@
 from django import forms
 
 from crispy_forms_gds.helper import FormHelper
-from crispy_forms_gds.layout import Submit
+from crispy_forms_gds.layout import HTML, Submit
 
 from django.urls import reverse_lazy
 
+from core.common.forms import BaseForm
 from exporter.applications.forms.edit import told_by_an_official_form, reference_name_form
 from exporter.core.constants import STANDARD
 from lite_content.lite_exporter_frontend import strings
@@ -77,3 +78,35 @@ class EditApplicationForm(forms.Form):
         super().__init__(*args, **kwargs)
         self.helper = FormHelper()
         self.helper.add_input(Submit("submit", "Continue"))
+
+
+class ApplicationMajorEditConfirmationForm(BaseForm):
+    class Layout:
+        TITLE = "Are you sure you want to open your application for editing?"
+
+    def __init__(self, *args, application_reference, cancel_url, **kwargs):
+        self.application_reference = application_reference
+        self.cancel_url = cancel_url
+        super().__init__(*args, **kwargs)
+
+    def get_layout_fields(self):
+        return [
+            HTML.p(
+                "Progress on the case will stop while you are making changes. The application will appear in your draft until you re-submit."
+            ),
+            HTML.p(
+                "Re-submitting the application with changes means it will take longer to process and will have a new ECJU case reference."
+            ),
+            HTML.p(f"Your own application reference '{self.application_reference}' will remain the same."),
+        ]
+
+    def get_layout_actions(self):
+        layout_actions = super().get_layout_actions()
+
+        layout_actions.append(
+            HTML(
+                f'<a class="govuk-button govuk-button--secondary" href="{self.cancel_url}" id="cancel-id-cancel">Cancel</a>'
+            ),
+        )
+
+        return layout_actions
--- a/exporter/applications/rules.py
+++ b/exporter/applications/rules.py
@@ -2,6 +2,7 @@ import dateparser
 import rules
 
 from datetime import datetime
+from django.conf import settings
 
 from exporter.applications.services import get_status_properties
 from exporter.applications.constants import ApplicationStatus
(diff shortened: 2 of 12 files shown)


exporter/applications/rules.py
```diff
@@ -75,3 +81,8 @@ def can_invoke_major_editable(request, application):
     "can_invoke_major_editable",
     can_invoke_major_editable,
 )
+
+rules.add_rule(
+    "is_major_edit_whitelisted_organisation",
```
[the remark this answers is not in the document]
saruniitr: ok, looking to update it as `can_make_amendment_by_copy` or `can_amend_by_copy`

exporter/applications/views/common.py
```diff
@@ -117,47 +135,79 @@ def get_success_url(self):
...
 
-        context.update(
-            {
-                "application_id": self.application_id,
-                "data": self.data,
-            }
-        )
-        return context
+    def handle_major_edit(self):
+        set_application_status(self.request, self.application_id, APPLICANT_EDITING)
```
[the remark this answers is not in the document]
saruniitr: This view is already covered under the rule `can_invoke_major_editable` so edit button is not visible if application is not in the right status.

exporter/applications/forms/common.py
```diff
@@ -67,3 +68,35 @@ def __init__(self, *args, **kwargs):
...
+
+    def __init__(self, *args, application_reference, cancel_url, **kwargs):
+        self.application_reference = application_reference
+        self.cancel_url = cancel_url
+        super().__init__(*args, **kwargs)
+
+    def get_layout_fields(self):
+        return [
+            HTML.p(
+                "Progress on the case will stop while you are making changes. The application will appear in your draft until you re-submit."
```
currycoder, line 85: in your draft**s**

exporter/applications/views/common.py
```diff
@@ -117,47 +135,84 @@ def get_success_url(self):
...
     def handle_major_edit(self):
-        organisation = get_organisation(self.request, self.request.session["organisation"])
-        if organisation["id"] in settings.FEATURE_AMENDMENT_BY_COPY_EXPORTER_IDS:
-            data, _ = self.create_amendment()
-            return redirect(reverse_lazy("applications:task_list", kwargs={"pk": data["id"]}))
-        else:
-            set_application_status(self.request, self.application_id, APPLICANT_EDITING)
-            return redirect(reverse_lazy("applications:task_list", kwargs={"pk": self.application_id}))
+        if not can_amend_by_copy(self.request, self.application):
+            raise ValueError
```
currycoder, line 206: How likely is it that a user can get here?  Presumably, we won't show them the button that leads to this flow if they do not have permission and this is just a backup check?

currycoder: Is there some way to push this up higher so that either GET or POST to this URL will raise an appropriate permissions error?

currycoder (approved):
Really like this PR - splitting out the feature flagged if statement to a separate permission rule/django view is great.

Just an optional couple of suggestions inline.

exporter/applications/views/common.py
saruniitr: Yes, it is really a backup check.

The GET link is in a template so this is highest I could add. In the template the link is covered with two rules, one is that it can get into major edit state and the other is whitelisted organisation so only valid users come here.

kevincarrogan: I don't think we need to make any further changes here but I do think we should start thinking a bit more about having checks on the views that are a lot higher up and not just relying on whether a link exists or not.

I don't really fully understand why this is the highest this could be added. To me this whole view should return absolutely nothing if the permission isn't granted so I'm not sure why we have to only check this at the point of someone actually submitting.

Really I want us to think about the opposite idea that the view is the canonical checker of permissions i.e. even if you mangle a URL you can't get to it unless you're supposed to and whether a link is displayed or not is the "backup".

exporter/applications/forms/common.py
```diff
@@ -67,3 +68,35 @@ def __init__(self, *args, **kwargs):
...
+
+    def get_layout_fields(self):
+        return [
+            HTML.p(
+                "Progress on the case will stop while you are making changes. The application will appear in your drafts until you re-submit."
+            ),
+            HTML.p(
+                "Re-submitting the application with changes means it will take longer to process and will have a new ECJU case reference."
+            ),
+            HTML.p(f"Your own application reference '{self.application_reference}' will remain the same."),
```
hnryjmes, line 90: in the design in the ticket it has commas surrounding the app reference
```
Your own application reference, [app ref], will remain the same.
```

however I actually prefer the way you've got it written as I think it looks better. maybe we can keep it the way you have it but I wanted to comment this for others to think about.

exporter/applications/views/common.py
kevincarrogan: To add to this, I'd also like it that possibly by default all views aren't accessible unless we specifically state the allowed permissions for any given view.

exporter/applications/forms/common.py
saruniitr: I will check with UCD and change if they prefer the way it is in the ticket.

hnryjmes (approved):
I've tested out this branch locally now, with and without the exporter org id being in the whitelist, and I'm happy that I understand the changes

exporter/applications/views/common.py
saruniitr: @kevincarrogan promoted check to be primary check

exporter/applications/forms/common.py
saruniitr: ok they are fine with quotes as well.

PR #2068, opened 2024-07-15
LTD-5130: Add BDD scenario for amending an application by copy

saruniitr:
## Change description

The test first creates a draft application and submits it.
It then initiates a major edit which creates a new application by copying the existing application.
Note that major edit is initiated from the `Submitted` status itself. Ideally would've been good to do it after assessments but that involves switching to caseofficer and pushing the case forward. As the main purpose of this test that we can trigger major edit and ensure we create a new one successfully this is done from the submitted status itself. 

Once we create amended application it checks that the previous application is superseded by this new application and that it cannot be edited.

The new application is ensured to be in draft state and ready for amendments.
Finally it also ensures products and party details match with the original application.

**Note:**
Opening an application for editing by copying an existing application is allowed for configurable list of Exporters. This behaviour is not possible to imitate in E2E test as it is a runtime check and the Exporter Id changes for each iteration.

As an alternative we added this test assuming it is available for all exporters and skipping the test for now.

Once this feature is opened for all then we will enable this test.

[LTD-5130]: https://uktrade.atlassian.net/browse/LTD-5130?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/exporter/templates/applications/applications.html
+++ b/exporter/templates/applications/applications.html
@@ -44,10 +44,10 @@
 			</a>
 			<a href="{% url 'applications:applications' %}?sort_by=submitted_at&selected_filter=finalised_applications"
 				class="lite-tabs__tab {% if request.build_absolute_uri|has_param:"selected_filter=finalised_applications" %}lite-tabs__tab--selected{% endif %}"
-				id="applications-tab-draft">
+				id="applications-tab-finalised">
 					Finalised
 			</a>
-			<a href="{% url 'applications:applications' %}?sort_by=-created_at&selected_filter=draft_tab" class="lite-tabs__tab" id="applications-tab-draft">
+			<a href="{% url 'applications:applications' %}?sort_by=-created_at&selected_filter=draft_tab" class="lite-tabs__tab" id="applications-tab-drafts">
 				{% lcs "applications.ApplicationsList.DRAFTS_TAB" %}
 			</a>
 		</div>
--- a/ui_tests/exporter/features/amendments/siel_amendments.feature
+++ b/ui_tests/exporter/features/amendments/siel_amendments.feature
@@ -3,7 +3,7 @@ Feature: I want to submit SIEL applications and be able to make major amendments
   As a logged in exporter
   I want to submit SIEL applications and be able to make major amendments and resubmit
 
-  @siel_amend
+  @skip
   Scenario: Exporter making major amendments to an already submitted application
     Given I signin and go to exporter homepage and choose Test Org
     And I create a standard draft application with "Amend by copy" as reference
--- a/ui_tests/exporter/features/amendments/siel_amendments.feature
+++ b/ui_tests/exporter/features/amendments/siel_amendments.feature
@@ -4,6 +4,7 @@ Feature: I want to submit SIEL applications and be able to make major amendments
   I want to submit SIEL applications and be able to make major amendments and resubmit
 
   @skip
+  # Enable this scenario when this flow is enabled by default for all Exporters
   Scenario: Exporter making major amendments to an already submitted application
     Given I signin and go to exporter homepage and choose Test Org
     And I create a standard draft application with "Amend by copy" as reference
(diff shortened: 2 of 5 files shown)


exporter/templates/applications/applications.html
```diff
@@ -44,10 +44,10 @@ <h1 class="govuk-heading-l">
 			</a>
 			<a href="{% url 'applications:applications' %}?sort_by=submitted_at&selected_filter=finalised_applications"
 				class="lite-tabs__tab {% if request.build_absolute_uri|has_param:"selected_filter=finalised_applications" %}lite-tabs__tab--selected{% endif %}"
-				id="applications-tab-draft">
+				id="applications-tab-finalised">
```
currycoder, line 47: Maybe not one for right now, but this block looks like it's begging to be factored out in to a separate include of some kind.

ui_tests/exporter/features/amendments/siel_amendments.feature
```diff
@@ -0,0 +1,56 @@
...
+    When I go to task list of the amended draft application
+    And I click on "Tell us about the products" section
+    Then I see products with below details as json:
+      ["Sporting shotgun", "Sodium chloride", "Ammunition", "Magnetic sensor", "Imaging device"]
+    When I go to task list of the amended draft application
+    And I click on "Consignee" section
+    Then I see Consignee with details "Consignee", "1234 Consignee address, France"
+    When I go to task list of the amended draft application
+    And I click on "End user" section
+    Then I see End-user with details "End user", "1234 End-user address Australia"
```
currycoder, line 56: Could this be further extended with a meaningful edit, submission of the new case and verification that it was submitted successfully?  I feel like right now this stops a little short of covering the whole journey.

ui_tests/exporter/features/amendments/siel_amendments.feature
```diff
@@ -0,0 +1,56 @@
+@all
+Feature: I want to submit SIEL applications and be able to make major amendments and resubmit
+  As a logged in exporter
+  I want to submit SIEL applications and be able to make major amendments and resubmit
+
+  @skip
```
currycoder, line 6: Could it be possible to add a comment saying we should switch this on when we make the new flow the default?

exporter/templates/applications/applications.html
saruniitr: Yes definitely. There is a ticket to add a separate tab for archived applications so we can change it as part of that.

ui_tests/exporter/features/amendments/siel_amendments.feature
saruniitr: yeah, makes sense to add few more steps.

ui_tests/exporter/step_defs/amendments/test_siel_amendments.py
```diff
@@ -0,0 +1,279 @@
...
+@when("I continue to submit application")
+def continue_submitting_application(driver):
+    functions.click_submit(driver)
+
+
+@then("I record application reference code")
+def record_application_reference_code(driver, context):
+    message = driver.find_element(by=By.ID, value="application-processing-message-value").text
+    index = message.index("GBSIEL")
+    context.reference_code = message[index:]
```
hnryjmes, line 135: If the application processing message changes then this might not work. One way you could change this if you wanted would be to use the 21 characters after the index, as reference codes I believe are always of length 21. Or at least it will be for this UI test scenario. I think something like this would work:
```
context.reference_code = message[index:index+21]
```

hnryjmes (approved):
Looks good, one minor comment but not a blocker

ui_tests/exporter/step_defs/amendments/test_siel_amendments.py
```diff
@@ -132,7 +132,7 @@ def continue_submitting_application(driver):
 def record_application_reference_code(driver, context):
     message = driver.find_element(by=By.ID, value="application-processing-message-value").text
     index = message.index("GBSIEL")
-    context.reference_code = message[index:]
+    context.reference_code = message[index : index + 21]
```
currycoder, line 135: I do think there must be a better way to extract the reference code e.g. by regex.  But will leave this with you.

currycoder (approved):

PR #2069, opened 2024-07-16
LTD-5069-product-location-journey

seijihg:
https://uktrade.atlassian.net/browse/LTD-5069

Since I did not want refactor the whole wizard journey and this approach is not affecting it.

The aim is to create a separate views to handle the individual edit where is needed. Where things could be reused I did not rewrite the lite-forms since it is not the aim of this PR

TODO:

-refactor to make it a bit shorter
-unitest
-e2e test (in a seperate PR/Jira Ticket)


hnryjmes:
closing this PR as we will revisit this work in more detail at a later date.

PR #2079, opened 2024-07-25
[LTD-5225] Prevent caseworker allocation for cases which are not operable by caseworkers

currycoder:
### Aim

Ensure that cases which caseworkers are unable to modify cannot be allocated.

This is a companion PR to https://github.com/uktrade/lite-api/pull/2106 - it ensures that when a case is not caseworker operable, the caseworker allocation button is not visible.

**NOTE:** Before merge, the commit forcing the branch to use in the E2E tests must be dropped.

[LTD-5225](https://uktrade.atlassian.net/browse/LTD-5225)

[LTD-5225]: https://uktrade.atlassian.net/browse/LTD-5225?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


caseworker/cases/services.py
```diff
@@ -25,11 +25,6 @@ def get_case_basic_details(request, pk):
     return response.json()
 
 
-def patch_case(request, pk, json):
```
currycoder, line 28: This appears to be completely unused, so we can remove it.  The endpoint itself is removed in the companion API PR.

.circleci/config.yml
```diff
@@ -133,7 +133,7 @@ commands:
           command: |
             git clone git@github.com:uktrade/lite-api.git
             cd lite-api
-            git checkout $(python ../which_branch.py <<parameters.api_branch>>)
+            git checkout LTD-5225-prevent-caseworker-operations-superseded-case
```
Tllew, line 136: are we not able to set the api_branch as a parameter here to avoid this?

Tllew (approved):
Looks good, the initial change to conf.py without the changes to tests confused me a little, thought it was a formatting change, made a comment but I imagine there's a good reason to not just set the parameter so I'll approve.

saruniitr (approved):
looks good to me.
Tried these changes locally on a Case which is superseded by edits and the links to allocate case advice as well as the button on the banner are missing.

PR #2080, opened 2024-07-25
LTD-5245 - LU Senior Manager Permission to change licence status

markj0hnst0n:
### Aim

Create a new permission we can use in both templates and views to say the user is LU Senior Manager.

The idea for the next stage is to follow the pattern [here](https://github.com/uktrade/lite-frontend/blob/1e590bd54a65c30032781eee9a315b96e7786716/exporter/applications/views/common.py#L480) to check the user is eligible to make changes to licence status.  This is to be done in a ChangeLicence FormView view  with `get_context` of `dispatch` as shown in the link.

[LTD-5245](https://uktrade.atlassian.net/browse/LTD-5245)

[LTD-5245]: https://uktrade.atlassian.net/browse/LTD-5245?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


caseworker/core/constants.py
```diff
@@ -25,6 +25,7 @@
 
 # Role IDs
 SUPER_USER_ROLE_ID = "00000000-0000-0000-0000-000000000002"
+LICENSING_UNIT_SENIOR_MANAGER_ROLE_ID = "3ae08e0c-47b3-47ba-965f-48318129c147"
```
depsiatwal, line 28: Was susceptible using ID this appears to match not ideal but this is what we've used so far. Ideally we should have a static data store with this info.

caseworker/core/rules.py
```diff
@@ -82,3 +88,4 @@ def case_is_nlr(request, case):
 rules.add_rule("can_user_change_sub_status", is_user_allocated & has_available_sub_statuses)
 rules.add_rule("can_user_search_products", is_user_in_admin_team | is_user_in_tau_team)  # noqa
 rules.add_rule("can_user_rerun_routing_rules", rules.always_deny)
+rules.add_rule("can_user_change_licence_status", is_user_licencing_unit_senior_manager)
```
saruniitr, line 91: How is this rule going to be used?

caseworker/core/constants.py
saruniitr, line 28: Do we know if this id is consistent across all envs?

markj0hnst0n: Yes I have checked this id across all envs in cf to make sure

caseworker/core/rules.py
markj0hnst0n: The intention is to use this rule in the `ListOpenAndStandardLicences` view to allow only LU Senior Managers to see the change licence status link.  Then in turn have a view that only LU Senior Managers are allowed to see where they can change the licence status.

caseworker/core/constants.py
depsiatwal: Having checked since we have done role checks else where and see that the ID is consistent happy to approve.

depsiatwal (approved):
approved

caseworker/core/constants.py
markj0hnst0n: The pattern is pretty common across the api too but I do agree with the initial point.  We probably shouldn't have db ids in the code at all really.

caseworker/core/rules.py
markj0hnst0n: The idea is to follow the pattern [here](https://github.com/uktrade/lite-frontend/blob/1e590bd54a65c30032781eee9a315b96e7786716/exporter/applications/views/common.py#L480) checking the rules in a FormView to put protections in place so that the correct status is present in Case and Licence before the licence status can be suspended or revoked

caseworker/core/constants.py
saruniitr: There is no specific permission defined for this, may be that is a better alternative?
Eg there is a permission defined for changing licence duration so something similar in this case? then we can check if that permission exists.

kevincarrogan (commented):
How does this relate to https://github.com/uktrade/lite-frontend/pull/2082?

It looks like we're adding a lot of rules to essentially check the same thing, what's the difference in all of these rules?

markj0hnst0n:
In this PR the rule checks if the user has the correct role to change licence status so they are all distinct checks.  Is it a problem with the naming @kevincarrogan or is there maybe an issue I'm not seeing?

kevincarrogan:
> In this PR the rule checks if the user has the correct role to change licence status so they are all distinct checks. Is it a problem with the naming @kevincarrogan or is there maybe an issue I'm not seeing?

I've written comments around this on the other PR. I think my overall problem is that I don't know how this is going to be used so I'm finding it hard to know why we need to have some many distinct rules.

caseworker/core/constants.py
markj0hnst0n: This is something I'd brought up at the splke stage @saruniitr and certainly something that is possible.  I decided against it initially as It would just be adding a permission on to users with this role anyway so it may just be adding another permission layer that does the same job.  Happy to add if you have a different perspective though!

I think there would be value in this if there were only particular LU Senior Managers that need to do this action but that's not been the intention so I've just kept it to that role as that's the approach that meets the requirements defined by UCD.

markj0hnst0n:
I probably should have given more background in the description of this PR.   I've updated the description and made comments on the other PR mentioned.

depsiatwal:
As agreed the template change are being handled in another jira so we don't have change links switched. 
so the plan is 
1) Permissions Changes first 
2) Create views that utilise these permission and allows status change 
3) When we ready to test with uses add the link 
https://uktrade.atlassian.net/browse/LTD-5244

PR #2081, opened 2024-07-25
LTD-5135: Applications archived tab

saruniitr:
## Change description

Add a new tab on the exporter dashboard to group archived applications. Currently this will only show the list of applications that are superseded by amendments but the idea is to group similar ones under this tab (eg withdrawn etc)

Used this opportunity to refactor this a bit
 - Drafts tab was using a separate template which has lot of duplication from the main template so updated view such that same template can be used for all tabs
 - Sort options form was directly rendered in the template, moved to a separate form
 - Application state tabs urls, query params were defined in the template itself - changing the filter, sort options was fiddly and can easily go wrong. Moved this to the view so they can be managed better.
 - Some of the template code moved into other include templates
 - Fix failing ui tests because of id changes

Add unit tests for all tabs.

API changes are here https://github.com/uktrade/lite-api/pull/2107

[LTD-5135](https://uktrade.atlassian.net/browse/LTD-5135)

[LTD-5135]: https://uktrade.atlassian.net/browse/LTD-5135?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


exporter/templates/applications/applications.html
```diff
@@ -20,72 +21,42 @@ <h1 class="govuk-heading-l">
             </h1>
      	</div>
 		<div class="lite-app-bar__controls">
-			{% if organisation.type.key == 'hmrc' %}
-				<a href="{% url 'hmrc:raise_a_query' %}" class="govuk-button" draggable="false" role="button">
```
saruniitr, line 24: We don't support these at the moment hence removed this.
We may do things differently when we add support for this later.

exporter/templates/applications/applications.html
```diff
@@ -20,72 +21,42 @@ <h1 class="govuk-heading-l">
...
-			</a>
-			<a href="{% url 'applications:applications' %}?sort_by=-created_at&selected_filter=draft_tab" class="lite-tabs__tab" id="applications-tab-drafts">
-				{% lcs "applications.ApplicationsList.DRAFTS_TAB" %}
-			</a>
+			{% for tab in tabs %}
+				<a id="{{ tab.filter }}" href="{{ tab.url }}"
+					class="lite-tabs__tab {% if selected_filter == tab.filter %}lite-tabs__tab--selected{% endif %}">
+					{{ tab.name }}
+				</a>
+			{% endfor %}
```
saruniitr, line 40: Moved this to view mainly to avoid fiddling with url, query params in the template, they can be managed better within the view.

exporter/templates/applications/applications.html
```diff
@@ -20,72 +21,42 @@ <h1 class="govuk-heading-l">
...
-				Sort by
-			</label>
-			<select class="govuk-select" id="sort_by" name="sort_by">
-            <option value="submitted_at" {% if request.GET.sort_by == 'submitted_at' %}selected{% endif %}>Date submitted</option>
-						<option value="-updated_at" {% if request.GET.sort_by == '-updated_at' %}selected{% endif %}>Date updated</option>
-			</select>
-		</div>
+	{% if sort_options %}
+		{% crispy form %}
+	{% endif %}
```
saruniitr, line 46: Refactored this to a separate form so can be managed better.

exporter/applications/forms/common.py
```diff
@@ -100,3 +100,31 @@ def get_layout_actions(self):
         )
 
         return layout_actions
+
+
+class ApplicationsListSortForm(BaseForm):
+    class Layout:
+        TITLE = ""
```
kevincarrogan, line 107: Why no title?

exporter/templates/applications/applications.html
```diff
@@ -20,80 +21,50 @@ <h1 class="govuk-heading-l">
...
-			{% else %}
-				<a href="." class="govuk-button govuk-button--secondary" draggable="false" role="button">
-					{% lcs "applications.ApplicationsList.REFRESH_BUTTON" %}
-				</a>
-				<a href="{% url 'apply_for_a_licence:start' %}" class="govuk-button" draggable="false" role="button">
-					{% lcs "applications.ApplicationsList.APPLY_FOR_A_LICENCE_BUTTON" %}
-				</a>
-			{% endif %}
+			<a href="." class="govuk-button govuk-button--secondary" draggable="false" role="button">
+				{% lcs "applications.ApplicationsList.REFRESH_BUTTON" %}
```
kevincarrogan, line 25: Are we not deprecating the usage of these `lcs` things?

exporter/templates/applications/applications.html
```diff
@@ -20,80 +21,50 @@ <h1 class="govuk-heading-l">
...
+					<th class="govuk-table__header" scope="col">{% lcs "applications.ApplicationsSummaryPage.STATUS" %}</th>
 				</tr>
 			</thead>
 			<tbody class="govuk-table__body">
 				{% for application in applications.results %}
 					<tr class="govuk-table__row">
 						<td class="govuk-table__cell">
 							<a class="govuk-link govuk-link--no-visited-state app-icon-label"
-								{% if application.status.key == 'applicant_editing' %}
+								{% if application.status.key == "draft" or application.status.key == "applicant_editing" %}
```
kevincarrogan, line 67: Can we check to make sure we're not using this as a crude permission and we can't get to this URL even if the case status is one of these statuses?

kevincarrogan: Is this also possibly a place where we could introduce a django-rules rule?

exporter/templates/applications/applications.html
saruniitr: we are, I just did n't change in all places, will remove.

exporter/templates/applications/applications.html
saruniitr: Yes this is already protected in other places.
If you try to get to task list with other status then user will be redirected to `applications:application` url.

I thought of adding a rule but as this is not a specific user action I did not continue.

kevincarrogan: Ok. I will leave this one up to you.

exporter/applications/views/common.py
```diff
@@ -83,33 +85,71 @@ def get_application_task_list_url(self, pk):
...
+            **context,
             "applications": applications,
             "organisation": organisation,
-            "params": params,
+            "tabs": self.get_tabs(),
+            "selected_filter": params["selected_filter"],
             "page": params.pop("page"),
-            "params_str": convert_dict_to_query_params(params),
             "is_user_multiple_organisations": is_user_multiple_organisations,
+            "sort_options": params["selected_filter"] != "draft_applications",
```
hnryjmes, line 150: maybe this could be `show_sort_options`, then it is more action-oriented? as this variable does not store the sort options it's just a boolean

exporter/applications/views/common.py
```diff
@@ -83,33 +85,71 @@ def get_application_task_list_url(self, pk):
...
+            },
+        ]
+        for tab in tabs:
+            sort_by = tab.get("sort_by", "-submitted_at")
+            query_params = {"selected_filter": tab["filter"], "sort_by": sort_by}
+            tab["url"] = f"/applications/?{urlencode(query_params, doseq=True)}"
+
+        return tabs
+
+    def get_context_data(self, **kwargs):
```
hnryjmes, line 129: I like this general change to override `get_context_data()` instead of the `get()` method, this is a better pattern to use in my view

unit_tests/exporter/applications/views/test_applications_list.py
```diff
@@ -1,81 +1,201 @@
...
+            **base_application_data(),
+        },
+        {
+            "status": {"id": "00000000-0000-0000-0000-000000000002", "key": "under_review", "value": "Under review"},
+            "reference_code": "GBSIEL/2024/0000002/P",
+            **base_application_data(),
+        },
+        {
+            "status": {"id": "00000000-0000-0000-0000-000000000001", "key": "ogd_advice", "value": "OGD Advice"},
+            "reference_code": "GBSIEL/2024/0000001/P",
```
hnryjmes, line 58: I think I'd slightly prefer if these weren't real existing case references in our prod database, but I can't really explain why. also I see that before, the case references in the test data looked like `GBSIEL/2000/0000001/P`. maybe that's better? happy to go with what you think is best.

saruniitr: I think its fine. By changing it to 2000 we are not really masking anything.
The `id` values also are dummy ones.

exporter/templates/applications/applications.html
saruniitr: @kevincarrogan 
Tried converting this to a rule which is the same as this one https://github.com/uktrade/lite-frontend/blob/dev/exporter/applications/rules.py#L78

But the rule expects application as instance of `Application()` https://github.com/uktrade/lite-frontend/blob/dev/exporter/applications/services.py#L56
whereas in this case we get a paginated list of applications.

It may be possible to convert each of those entries in the list but may need more changes to the template so keeping it as it is now.

hnryjmes (approved):
approving as code changes look good to me and my questions have been answered

exporter/applications/views/common.py
```diff
@@ -83,33 +85,71 @@ def get_application_task_list_url(self, pk):
...
+            "sort_by": self.request.GET.get("sort_by", "-submitted_at"),
+        }
+
+    def get_form_kwargs(self):
+        kwargs = super().get_form_kwargs()
+        kwargs["action"] = reverse("applications:applications")
+        return kwargs
+
+    def get_tabs(self):
+        tabs = [
```
depsiatwal, line 103: Is it worth having this as a constant of tis form or even as a separate class just to keep views lighter and easier to read ?

exporter/assets/javascripts/sort_by.js
```diff
@@ -1,5 +1,5 @@
 const initSortBy = () => {
-  const sortBy = document.getElementById("sort_by");
+  const sortBy = document.getElementById("id_sort_by");
```
depsiatwal, line 2: do we know if this is used else where ?
and covered all areas where a change is needed

saruniitr: yes, have updated UI tests accordingly

depsiatwal (approved):
lgtm

PR #2082, opened 2024-07-26
LTD 5246 add license status change rules

markj0hnst0n:
### Aim

Add rules to check whether a case in in the correct state to allow a licence status to be changed and to check whether licence status can be changed so that this maybe be checked in views/forms to only those licences and cases that we want will be impacted.

Original ticket mentions 'a permission' but thought it better to check each object separately and then the rules can be checked in the views as needed.

[LTD-5246](https://uktrade.atlassian.net/browse/LTD-5246)

### DIFF
--- a/caseworker/core/rules.py
+++ b/caseworker/core/rules.py
@@ -92,3 +92,4 @@ rules.add_rule("can_user_add_contact", is_user_allocated)
 rules.add_rule("can_user_change_sub_status", is_user_allocated & has_available_sub_statuses)
 rules.add_rule("can_user_search_products", is_user_in_admin_team | is_user_in_tau_team)  # noqa
 rules.add_rule("can_user_rerun_routing_rules", rules.always_deny)
+rules.add_rule("can_licence_status_be_changed", is_case_finalised & is_licence_status_able_to_be_changed)
--- a/caseworker/core/rules.py
+++ b/caseworker/core/rules.py
@@ -71,7 +71,7 @@ def case_is_nlr(request, case):
 
 
 @rules.predicate
-def is_case_finalised(request, case):
+def is_case_finalised_or_withdrawn(request, case):
     return case["status"] in ["finalised", "withdrawn"]
 
 
@@ -92,4 +92,5 @@ rules.add_rule("can_user_add_contact", is_user_allocated)
 rules.add_rule("can_user_change_sub_status", is_user_allocated & has_available_sub_statuses)
 rules.add_rule("can_user_search_products", is_user_in_admin_team | is_user_in_tau_team)  # noqa
 rules.add_rule("can_user_rerun_routing_rules", rules.always_deny)
-rules.add_rule("can_licence_status_be_changed", is_case_finalised & is_licence_status_able_to_be_changed)
+rules.add_rule("can_licence_status_be_changed_on_case", is_case_finalised_or_withdrawn)
+rules.add_rule("can_licence_status_be_changed", is_licence_status_able_to_be_changed)
--- a/caseworker/core/rules.py
+++ b/caseworker/core/rules.py
@@ -70,6 +70,16 @@ def case_is_nlr(request, case):
     return has_flag(case, GOODS_NOT_LISTED_ID)
 
 
+@rules.predicate
+def is_case_finalised(request, case):
+    return case["status"] in ["finalised", "withdrawn"]
+
+
+@rules.predicate
+def is_licence_status_able_to_be_changed(request, licence):
+    return licence["status"] in ["issued", "reinstated", "suspended"]
+
+
 rules.add_rule("can_user_change_case", is_user_allocated)
 rules.add_rule("can_user_move_case_forward", is_user_allocated)
 rules.add_rule("can_user_review_and_countersign", is_user_allocated)
--- a/unit_tests/caseworker/core/test_rules.py
+++ b/unit_tests/caseworker/core/test_rules.py
@@ -67,7 +67,7 @@ def get_mock_request(client):
 )
 def test_is_user_assigned(data, mock_gov_user, get_mock_request, expected_result):
     assigned_users = {"assigned_users": data}
-    assert caseworker_rules.is_user_assigned(get_mock_request(mock_gov_user["user"]), assigned_users) == expected_result
+    assert caseworker_rules.is_user_assigned(get_mock_request(mock_gov_user["user"]), assigned_users) is expected_result
 
 
 def test_is_user_assigned_request_missing_attribute():
@@ -97,7 +97,7 @@ def test_is_user_case_officer_none(get_mock_request):
     ),
 )
 def test_is_user_case_officer(data, mock_gov_user, get_mock_request, expected_result):
-    assert caseworker_rules.is_user_case_officer(get_mock_request(mock_gov_user["user"]), data) == expected_result
+    assert caseworker_rules.is_user_case_officer(get_mock_request(mock_gov_user["user"]), data) is expected_result
 
 
 def test_is_user_case_officer_request_missing_attribute():
@@ -167,7 +167,7 @@ def test_user_assignment_based_rules(data, mock_gov_user, get_mock_request, expe
         "can_user_add_contact",
         "can_user_change_sub_status",
     ):
-        assert rules.test_rule(rule_name, get_mock_request(mock_gov_user["user"]), data) == expected_result
+        assert rules.test_rule(rule_name, get_mock_request(mock_gov_user["user"]), data) is expected_result
 
 
 @pytest.mark.parametrize(
@@ -270,7 +270,7 @@ def test_can_user_search_products(mock_gov_user, get_mock_request, mock_gov_user
 
     request = get_mock_request(user)
 
-    assert rules.test_rule("can_user_search_products", request) == expected
+    assert rules.test_rule("can_user_search_products", request) is expected
 
 
 @pytest.mark.parametrize(
@@ -297,7 +297,7 @@ def test_can_assigned_user_assess_products(mock_gov_user, get_mock_request, mock
 
     request = get_mock_request(user)
 
-    assert rules.test_rule("can_user_assess_products", request, case) == expected
+    assert rules.test_rule("can_user_assess_products", request, case) is expected
 
 
 @pytest.mark.parametrize(
@@ -324,7 +324,7 @@ def test_can_unassigned_user_assess_products(mock_gov_user, get_mock_request, mo
 
     request = get_mock_request(user)
 
-    assert rules.test_rule("can_user_assess_products", request, case) == expected
+    assert rules.test_rule("can_user_assess_products", request, case) is expected
 
 
 @pytest.mark.parametrize(
@@ -391,7 +391,7 @@ def test_can_user_review_and_combine_based_on_allocation(mock_gov_user, get_mock
     user = mock_gov_user["user"]
     request = get_mock_request(user)
 
-    assert rules.test_rule("can_user_review_and_combine", request, case) == expected_result
+    assert rules.test_rule("can_user_review_and_combine", request, case) is expected_result
 
 
 @pytest.mark.parametrize(
@@ -427,7 +427,7 @@ def test_can_user_review_and_combine_based_on_advice(mock_gov_user, get_mock_req
     user = mock_gov_user["user"]
     request = get_mock_request(user)
 
-    assert rules.test_rule("can_user_review_and_combine", request, case) == expected
+    assert rules.test_rule("can_user_review_and_combine", request, case) is expected
 
 
 def test_can_user_rerun_routing_rules(get_mock_request):
@@ -435,3 +435,45 @@ def test_can_user_rerun_routing_rules(get_mock_request):
     user = {}
     request = get_mock_request(user)
     assert not rules.test_rule("can_user_rerun_routing_rules", request, case)
+
+
+@pytest.mark.parametrize(
+    ("case_status", "expected"),
+    (
+        ("finalised", True),
+        ("withdrawn", True),
+        ("under_review", False),
+        ("revoked", False),
+        ("open", False),
+        ("draft", False),
+    ),
+)
+def test_can_licence_status_be_changed_on_case(mock_gov_user, get_mock_request, case_status, expected):
+    case = {
+        "status": case_status,
+    }
+    user = mock_gov_user["user"]
+    request = get_mock_request(user)
+
+    assert rules.test_rule("can_licence_status_be_changed_on_case", request, case) is expected
+
+
+@pytest.mark.parametrize(
+    ("licence_status", "expected"),
+    (
+        ("issued", True),
+        ("reinstated", True),
+        ("suspended", True),
+        ("expired", False),
+        ("exhausted", False),
+        ("cancelled", False),
+    ),
+)
+def test_can_licence_status_be_changed(mock_gov_user, get_mock_request, licence_status, expected):
+    licence = {
+        "status": licence_status,
+    }
+    user = mock_gov_user["user"]
+    request = get_mock_request(user)
+
+    assert rules.test_rule("can_licence_status_be_changed", request, licence) is expected


caseworker/core/rules.py
```diff
@@ -70,6 +70,16 @@ def case_is_nlr(request, case):
     return has_flag(case, GOODS_NOT_LISTED_ID)
```
saruniitr, line 72: I appreciate that these are in preparation for upcoming changes but it helps if explain about how these are going to be used later.

caseworker/core/rules.py
```diff
@@ -70,6 +70,16 @@ def case_is_nlr(request, case):
...
 
 
+@rules.predicate
+def is_case_finalised_or_withdrawn(request, case):
+    return case["status"] in ["finalised", "withdrawn"]
```
saruniitr, line 75: I think there are constants defined for these, if not please add any missing ones.

This is used for the rule `can_licence_status_be_changed_on_case` - why is 'withdrawn' included here?
Is this for the cases where licence is issued and exporter withdrawn?
If yes in that case that would be surrender rather than withdrawn.
My understanding of withdrawn is exporter makes this decision before we reach licence state.

caseworker/core/rules.py
markj0hnst0n: The idea is to follow the pattern [here](https://github.com/uktrade/lite-frontend/blob/1e590bd54a65c30032781eee9a315b96e7786716/exporter/applications/views/common.py#L480) to check the case and licence are eligible for change.  This is to be done in a ChangeLicence FormView view `get_context`

caseworker/core/rules.py
```diff
@@ -70,6 +70,16 @@ def case_is_nlr(request, case):
...
+
+
+@rules.predicate
+def is_licence_status_able_to_be_changed(request, licence):
+    return licence["status"] in ["issued", "reinstated", "suspended"]
```
depsiatwal, line 80: as per comment above add to some constants as a list of editable states

caseworker/core/rules.py
```diff
@@ -70,6 +70,16 @@ def case_is_nlr(request, case):
     return has_flag(case, GOODS_NOT_LISTED_ID)
 
 
+@rules.predicate
+def is_case_finalised_or_withdrawn(request, case):
+    return case["status"] in ["finalised", "withdrawn"]
+
+
+@rules.predicate
+def is_licence_status_able_to_be_changed(request, licence):
```
depsiatwal, line 79: may be better name is 
is_license_status_editable ?

caseworker/core/rules.py
depsiatwal, line 75: why "withdrawn" it's just finalised I believe

markj0hnst0n: I think withdrawn has been included incorrectly here so I've removed it.  I think it was in the criteria at some point but it's been removed.

markj0hnst0n: I don't know how that got in there tbh but I've removed

markj0hnst0n: Constant for finalised added

caseworker/core/rules.py
markj0hnst0n: Added in some constants for these

caseworker/core/rules.py
```diff
@@ -82,3 +93,5 @@ def case_is_nlr(request, case):
 rules.add_rule("can_user_change_sub_status", is_user_allocated & has_available_sub_statuses)
 rules.add_rule("can_user_search_products", is_user_in_admin_team | is_user_in_tau_team)  # noqa
 rules.add_rule("can_user_rerun_routing_rules", rules.always_deny)
+rules.add_rule("can_licence_status_be_changed_on_case", is_case_finalised)
+rules.add_rule("can_licence_status_be_changed", is_license_status_editable)
```
kevincarrogan, line 97: Why are there two different rules with practically the same name? What's the different usage requirements here?

markj0hnst0n: The first rule checks the case to see if it is in a finalised state as that's the only one allowed for changing licence status.  The second rule checks the licence to see if it is in one of the 3 acceptable states (issued, reinstated and suspended) for changing licence status.

kevincarrogan: Ok. I guess I want to know then how you are intending to use these rules? I think want I don't get is why we are creating so many distinct rules and how they actually pertain to actions that someone might take.

markj0hnst0n: The idea for the next stage is to follow the pattern [here](https://github.com/uktrade/lite-frontend/blob/1e590bd54a65c30032781eee9a315b96e7786716/exporter/applications/views/common.py#L480) to check the case and licence are eligible for change.  This is to be done in a ChangeLicence FormView view  with `get_context` of `dispatch` as shown in the link.  Same with this [this](https://github.com/uktrade/lite-frontend/pull/2080) PR.

I've updated the description of the pull request with this info too.  Probably superfluous at this stage but this should have been done at the beginning.

kevincarrogan: Ok. So we have three new rules, what are the three (or more) actions that you are intending to use these permissions for?

markj0hnst0n: The idea was that the 3 rules would be checked in the view and there would be one action off the back of the rules being checked.  That being that the Licence status can be changed.  They way this conversation is going is making me think I should be making one rule that checks those 3 things and gives a result?

kevincarrogan: If there is a single action then there only needs to be a single rule. That rule could be made up of multiple predicates but I think there should be a one-to-one correlation with a rule and an action.

If you have a view that is doing a single action, which it sounds like you are, then I would agree it should be a single rule.

markj0hnst0n: I've rolled the permissions into one rule which takes a dict which will be populated in the view with the active licence status and the case status.

markj0hnst0n: I've also put the rule into the view so that it's easier to see the direction I'm going with it hopefully

caseworker/cases/helpers/case.py
```diff
@@ -190,6 +197,10 @@ def get_context(self):
         context = super().get_context_data()
         default_tab = "quick-summary"
         current_tab = default_tab if self.kwargs["tab"] == "default" else self.kwargs["tab"]
+        licence_dict = {
+            "case": {"status": self.case.data["status"]["key"]},
+            "licence": {"status": self.case.data["licence"]["status"]["key"]},
```
depsiatwal, line 202: I don't think this is going to work as you will need to check each license 
to see if it's editable when we display the change link.

caseworker/templates/case/tabs/licences.html
```diff
@@ -27,6 +27,11 @@
 					<td class="govuk-table__cell govuk-table__cell--tight">
 						{{ licence.status }}
 					</td>
+					{% if can_change_licence_status %}
```
depsiatwal, line 30: So in the html template I would expect something like 

{% test_rule 'can_licence_status_be_changed' request case %}

{% set my_dict = { 'case_status': case.status,'license_status': license.status, } %}

of course you could even register a tag to create you the dict if were going to use this multiple times.

caseworker/cases/helpers/case.py
depsiatwal: I've put an example below to show my thoughts

unit_tests/caseworker/core/test_rules.py
```diff
@@ -427,11 +434,65 @@ def test_can_user_review_and_combine_based_on_advice(mock_gov_user, get_mock_req
...
+
+
+@pytest.mark.parametrize(
+    ("user_role_id", "licence_status", "case_status", "expected"),
+    (
+        (LICENSING_UNIT_SENIOR_MANAGER_ROLE_ID, LicenceStatusEnum.ISSUED, CaseStatusEnum.FINALISED, True),
+        (LICENSING_UNIT_SENIOR_MANAGER_ROLE_ID, LicenceStatusEnum.ISSUED, CaseStatusEnum.FINALISED, True),
+        (LICENSING_UNIT_SENIOR_MANAGER_ROLE_ID, LicenceStatusEnum.REINSTATED, CaseStatusEnum.FINALISED, True),
+        (LICENSING_UNIT_SENIOR_MANAGER_ROLE_ID, LicenceStatusEnum.SUSPENDED, CaseStatusEnum.FINALISED, True),
+        ("00c341d1-d83e-4a12-b103-c3fb575a5962", LicenceStatusEnum.ISSUED, CaseStatusEnum.FINALISED, False),
```
depsiatwal, line 475: why an ID here and not constant ?

caseworker/core/constants.py
```diff
@@ -25,6 +25,7 @@
 
 # Role IDs
 SUPER_USER_ROLE_ID = "00000000-0000-0000-0000-000000000002"
+LICENSING_UNIT_SENIOR_MANAGER_ROLE_ID = "3ae08e0c-47b3-47ba-965f-48318129c147"
```
markj0hnst0n, line 29: The LU Senior manager role matches exactly the criteria for these changes as only LU Senior managers may change licence status.  Using any existing permission may not match this criteria exactly.

caseworker/core/rules.py
```diff
@@ -70,6 +75,23 @@ def case_is_nlr(request, case):
...
+@rules.predicate
+def is_license_and_case_in_editable_state(user, licence):
+    is_case_editable = licence["case_status"] == CaseStatusEnum.FINALISED.capitalize()
+    is_licence_editable = licence["status"] in [
+        LicenceStatusEnum.ISSUED,
+        LicenceStatusEnum.REINSTATED,
+        LicenceStatusEnum.SUSPENDED,
+    ]
+    return is_case_editable and is_licence_editable
+
```
markj0hnst0n, line 93: API now provides case on the licence in the correct context so that the case id can be checked to see fi it is finalised.

caseworker/core/rules.py
```diff
@@ -70,6 +75,23 @@ def case_is_nlr(request, case):
...
 
+@rules.predicate
+def is_user_licencing_unit_senior_manager(user):
+    user_role_id = user["role"]["id"]
+    return user_role_id is LICENSING_UNIT_SENIOR_MANAGER_ROLE_ID
+
+
+@rules.predicate
+def is_license_and_case_in_editable_state(user, licence):
+    is_case_editable = licence["case_status"] == CaseStatusEnum.FINALISED.capitalize()
```
depsiatwal, line 86: This doesn't feel correct shouldn't need to capitalize() probably not return the correct value from API 
see my comment regarding API side

caseworker/templates/case/tabs/licences.html
```diff
@@ -6,10 +7,17 @@
 				<th class="govuk-table__header" scope="col">{% lcs 'cases.CasePage.LicencesTab.PRODUCTS_COLUMN' %}</th>
 				<th class="govuk-table__header" scope="col">{% lcs 'cases.CasePage.LicencesTab.USAGE_COLUMN' %}</th>
 				<th class="govuk-table__header" scope="col">{% lcs 'cases.CasePage.LicencesTab.STATUS_COLUMN' %}</th>
+				{% for licence in case.licences %}
+					{% test_rule 'can_licence_status_be_changed' request.lite_user licence as can_licence_status_be_changed %}
+					{% if can_licence_status_be_changed and is_case_finalised %}
```
depsiatwal, line 12: don't all these test rules need to be update not to is_license_and_case_in_editable_state ?
Also We don't use constants anymore so we can use Action as a string directly. 
Also if we find any license that is editable we can break out this loop or we could just pass into context from the view which determines if action should be displayed.

caseworker/core/rules.py
markj0hnst0n: removed

caseworker/templates/case/tabs/licences.html
markj0hnst0n: The rules are in the correct stage to work correctly in template now.  The rule `can_licence_status_be_changed` checks the predicate `is_license_and_case_in_editable_state`.   There was an issue with the is_case_finalised variable left over from a previous iteration but it has been removed.

Used "Actions" string directly now.  Should I amend and delete the other constants?

We are iterating through the licences in template and assigning the link.  I wouldn't expect a huge number of licences on each template so this feels appropriate.  The `get_context` view is already pretty chunky and while that isn't necessarily a reason not to loop through the licences there and assign them a boolean to show the change link, I feel like the template language accommodates this well so wen't with that.

caseworker/templates/case/tabs/licences.html
```diff
@@ -6,10 +7,17 @@
 				<th class="govuk-table__header" scope="col">{% lcs 'cases.CasePage.LicencesTab.PRODUCTS_COLUMN' %}</th>
 				<th class="govuk-table__header" scope="col">{% lcs 'cases.CasePage.LicencesTab.USAGE_COLUMN' %}</th>
 				<th class="govuk-table__header" scope="col">{% lcs 'cases.CasePage.LicencesTab.STATUS_COLUMN' %}</th>
+				{% for licence in case.licences %}
+					{% test_rule 'can_licence_status_be_changed' request.lite_user licence as can_licence_status_be_changed %}
+					{% if can_licence_status_be_changed %}
+						<th class="govuk-table__header" scope="col">Actions</th>
```
depsiatwal, line 13: Lets break out if we get here

depsiatwal (commented):
Can we also add a test for licence page and check using BS ?
The change link is displayed as expected with the correct url

caseworker/templates/case/tabs/licences.html
```diff
@@ -27,6 +35,12 @@
 					<td class="govuk-table__cell govuk-table__cell--tight">
 						{{ licence.status }}
 					</td>
+					{% test_rule 'can_licence_status_be_changed' request.lite_user licence as can_licence_status_be_changed %}
+					{% if can_licence_status_be_changed %}
+						<td class="govuk-table__cell govuk-table__cell--tight">
```
depsiatwal, line 40: also might need an id here so we can test using e2e tests

markj0hnst0n: template and view changes removed

caseworker/templates/case/tabs/licences.html
markj0hnst0n: removed view and template changes

unit_tests/caseworker/core/test_rules.py
markj0hnst0n: That is just there to fail on that UUID as it's different to LICENSING_UNIT_SENIOR_MANAGER_ROLE_ID.  I didn't want to add more Role Id constants without a purpose other than testing.

caseworker/cases/helpers/case.py
```diff
@@ -172,6 +175,14 @@ def get_open_ecju_queries_with_forms(self, open_ecju_queries):
             open_ecju_queries_with_forms.append((open_query, CloseQueryForm(prefix=str(open_query["id"]))))
         return open_ecju_queries_with_forms
 
+    def get_licence_dict(self):
```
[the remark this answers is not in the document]
markj0hnst0n: removed

caseworker/core/rules.py
```diff
@@ -70,6 +75,25 @@ def case_is_nlr(request, case):
...
+@rules.predicate
+def is_user_licencing_unit_senior_manager(user):
+    user_role_id = user["role"]["id"]
+
+    return user_role_id is LICENSING_UNIT_SENIOR_MANAGER_ROLE_ID
+
+
+@rules.predicate
+def is_license_and_case_in_editable_state(user, licence):
+    is_case_editable = licence["case_status"] == CaseStatusEnum.FINALISED
```
saruniitr, line 87: Can this be renamed to `is_case_finalised` or similar? `editable` is misleading.

This is also where I was thinking to check for sub_status as well but licence won't exist if it is refusal.
So whereever this rule is used we need to make sure `licence` is valid otherwise it will break.

unit_tests/caseworker/core/test_rules.py
saruniitr: you can just use `str(uuid.uuid4())` to get a random uuid

unit_tests/caseworker/core/test_rules.py
```diff
@@ -427,11 +434,42 @@ def test_can_user_review_and_combine_based_on_advice(mock_gov_user, get_mock_req
...
+        (LICENSING_UNIT_SENIOR_MANAGER_ROLE_ID, LicenceStatusEnum.SUSPENDED, CaseStatusEnum.SUSPENDED, False),
+    ),
+)
+def test_can_licence_status_be_changed(
+    mock_gov_user, get_mock_request, user_role_id, licence_status, case_status, expected
+):
+    licence = {
+        "status": licence_status,
+        "case_status": case_status,
+    }
```
saruniitr, line 468: this is odd - licence and case need to be fixtures matching the response you usually get from the API.

saruniitr (commented):
last set of comments

caseworker/core/rules.py
depsiatwal: agree perhaps is_case_finalised_licence_editable

unit_tests/caseworker/core/test_rules.py
depsiatwal: can add to data_standard_case where there's license details so all tests can consume the new payload

caseworker/core/rules.py
markj0hnst0n: updated variable name

unit_tests/caseworker/core/test_rules.py
markj0hnst0n: made that change

unit_tests/caseworker/core/test_rules.py
markj0hnst0n: changes made

depsiatwal (approved):
LGTM

markj0hnst0n:
As agreed with squad lead the template change are being handled in another jira so we don't have change links switched. 
so the plan is 

1) Permissions Changes first 
2) Create views that utilise these permission and allows status change 
3) When we ready to test with uses add the link 

https://uktrade.atlassian.net/browse/LTD-5244

saruniitr (approved):

PR #2089, opened 2024-08-05
LTD-5242-add-change-status-view-2 

depsiatwal:
https://uktrade.atlassian.net/browse/LTD-5242

This is the new view where a user selects a single  License state to be changed. 

The view will not be accessible until we switch it on via a template URL. 
It builds on the previous permissions where a user must be LU_Senior_manager and the case/License must be in specific states. 

This is the initial form that gives a choice presented to the user to select the new license state. The next step will be the confirmation screen where there will be a post action to call the API save endpoint. 

Also require a slight change to end point 
https://github.com/uktrade/lite-api/pull/2120/
Which pulls in case_status


unit_tests/caseworker/cases/views/test_change_licence_status.py
```diff
@@ -0,0 +1,216 @@
...
+def test_get_change_licence_status_case_non_lu_manager_404(
+    authorized_client,
+    mock_queue,
+    mock_case,
+    mock_get_licence_details,
+    change_licence_status_url,
+    mock_gov_user,
+):
+
+    mock_gov_user["user"]["role"]["id"] = "12345"
```
markj0hnst0n, line 214: Maybe useful to use UUID with `str(uuid.uuid4()`

markj0hnst0n (approved):
looks grand.  Follows pattern as outlined in POC building on rules usage.  Minor comments but happy to approve

core/constants.py
```diff
@@ -119,9 +119,24 @@ class SecurityClassifiedApprovalsType:
...
-    REINSTATED = "Reinstated"
-    SUSPENDED = "Suspended"
+    ISSUED = "issued"
+    CANCELLED = "cancelled"
+    EXPIRED = "expired"
+    EXHAUSTED = "exhausted"
+    REINSTATED = "reinstated"
+    SUSPENDED = "suspended"
+    REVOKED = "revoked"
+
```
markj0hnst0n, line 129: The licence object provided by the api call has the licence status capitalised so these changes would break the rules code.  You can either change these back to capitalised or change the rules code to capitalise the licence status.

depsiatwal: Yeah I thought about this. They should definitely be lower case as that's the DB entry. 
I think we will end up calling the permission check to deal with both as short term. Long term bother serialiser should fetch they key.

markj0hnst0n: I agree with this principle but the db entry is Capialised by the serializer on the api side so it will never reach us in the lower case state.  The rule would not be able to deal with the actual data from the api correctly.

depsiatwal: Yeah that's wrong the serializer that capitalises should be using key/value pairs. 
Key which in options are usual lower case and the value is for UI. 

Hence why i'm reverting the enums to be the correct way round. Then we would either need to correct the 'bad' serializer which may have a large impact or change the permission check to deal with both for now.

PR #2090, opened 2024-08-06
LTD-5072 Added replacement for \r

Tllew:
### Aim

[LTD-](https://uktrade.atlassian.net/browse/LTD-5072)

### DIFF
--- a/unit_tests/caseworker/cases/test_denials_view.py
+++ b/unit_tests/caseworker/cases/test_denials_view.py
@@ -289,7 +289,13 @@ def test_search_denials_query_string_error(authorized_client, requests_mock, que
 
 @pytest.fixture
 def mock_newline_case(requests_mock, data_standard_case):
-    data_standard_case["case"]["data"]["end_user"]["address"] = "42 \nWallaby \nWay"
+    data_standard_case["case"]["data"]["end_user"][
+        "address"
+    ] = """42 \n\r\n\r
+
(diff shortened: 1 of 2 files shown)


unit_tests/caseworker/cases/test_denials_view.py
```diff
@@ -289,7 +289,13 @@ def test_search_denials_query_string_error(authorized_client, requests_mock, que
...
 @pytest.fixture
 def mock_newline_case(requests_mock, data_standard_case):
-    data_standard_case["case"]["data"]["end_user"]["address"] = "42 \nWallaby \nWay"
+    data_standard_case["case"]["data"]["end_user"][
+        "address"
+    ] = """42 \n\r\n\r
+
+Wallaby\r\n\r Way\n
+
+"""
```
hnryjmes, line 298: I think we should have test data that uses either `\n` or `\r\n` as this corresponds to the Unix-like and Windows-like examples here: https://en.wikipedia.org/wiki/Newline#Representation

Tllew: the test data I got from the replication in the ticket was specifically \r\n\r and I copied the exact newline character from the database just to prove that multiple newlines are all stripped out.

hnryjmes: that seems reasonable to me

hnryjmes (approved):

PR #2092, opened 2024-08-06
LTD-5083: Use exporter specific endpoint to update quantity/value of existing goods on application

saruniitr:
## Change description

A new endpoint is added that allows only updating of quantity/value of goods that are already added to an application.

Previously we were using a common url using which other fields can also be updated.

API PR: https://github.com/uktrade/lite-api/pull/2121

[LTD-5083](https://uktrade.atlassian.net/browse/LTD-5083)

[LTD-5083]: https://uktrade.atlassian.net/browse/LTD-5083?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/unit_tests/exporter/applications/views/goods/common/test_edit_existing_goods_on_application.py
+++ b/unit_tests/exporter/applications/views/goods/common/test_edit_existing_goods_on_application.py
@@ -22,11 +22,18 @@ def edit_quantity_value_url(application, good_on_application):
     return url
 
 
+@pytest.fixture
+def mock_good_on_application_edit_quantity_value(requests_mock, data_standard_case, good_on_application):
+    application_id = data_standard_case["case"]["id"]
+    url = f"/applications/{application_id}/good-on-application/{good_on_application['id']}/quantity-value/"
+    return requests_mock.patch(url, json={})
+
+
 def test_edit_quantity_value_existing_good(
     authorized_client,
     edit_quantity_value_url,
     application_products_url,
-    mock_good_on_application_put,
+    mock_good_on_application_edit_quantity_value,
 ):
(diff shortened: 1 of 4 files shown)


unit_tests/exporter/applications/views/goods/common/test_edit_existing_goods_on_application.py
```diff
@@ -22,11 +22,18 @@ def edit_quantity_value_url(application, good_on_application):
...
+    url = f"/applications/{application_id}/good-on-application/{good_on_application['id']}/quantity-value/"
+    return requests_mock.patch(url, json={})
+
+
 def test_edit_quantity_value_existing_good(
     authorized_client,
     edit_quantity_value_url,
     application_products_url,
-    mock_good_on_application_put,
+    mock_good_on_application_edit_quantity_value,
```
depsiatwal, line 36: so nothing needs to be added here as we using a different API endpoint ?

saruniitr: This is a new fixture corresponding the new endpoint.
`mock_good_on_application_put` is the old one which uses a common endpoint

saruniitr: nothing else required as the form (hence payload) is the same

depsiatwal (approved):
Approving assume API will be merged before this is used

Tllew (approved):
Looks good

PR #2091, opened 2024-08-06
LTD-5243-save-licence-confirmation-save 

depsiatwal:
https://uktrade.atlassian.net/browse/LTD-5242
https://uktrade.atlassian.net/browse/LTD-5243

This is the new view where a user selects a single License state to be changed.

The view will not be accessible until we switch it on via a template URL.
It builds on the previous permissions where a user must be LU_Senior_manager and the case/License must be in specific states.

View ChangeLicenceStatus
This is the initial form that gives a choice presented to the user to select the new license state. The next step will be the confirmation screen where there will be a post action to call the API save endpoint.

Also require a slight change to end point
https://github.com/uktrade/lite-api/pull/2120
Which pulls in case_status

View ChangeLicenceStatusConfirmation

This view is responsible for submitting the update_licence_details to api once user has confirmed.

PR #2093, opened 2024-08-07
LTD 5244 add change licence link

markj0hnst0n:
### Aim


unit_tests/caseworker/cases/views/test_case_details.py
```diff
@@ -289,6 +289,97 @@ def test_case_details_sub_status_change_displayed(
...
+                "status": "issued",
+                "case_status": "finalised",
+                "reference_code": "12345AB",
+            },
+            str(uuid.uuid4()),
+            False,
+        ),
+    ],
+)
+def test_licence_details_actions_column_and_licence_status_change_link_display(
```
depsiatwal, line 357: think it's worth adding a couple of tests where we have multiple license since that will be reality 
i.e One expired, another Issued 
etc etc

markj0hnst0n: changes made

caseworker/cases/helpers/case.py
```diff
@@ -190,6 +191,11 @@ def get_context(self):
         context = super().get_context_data()
         default_tab = "quick-summary"
         current_tab = default_tab if self.kwargs["tab"] == "default" else self.kwargs["tab"]
+        show_actions_column = False
+        for licence in self.case.licences:
+            if rules.test_rule("can_licence_status_be_changed", self.request, licence):
+                show_actions_column = True
+                break
```
markj0hnst0n, line 199: This is in view so that the template can be told definitively to display a column if any licence in the case can be displayed.  Done here to remove logic from template and also `break` not available in template so could iterate further through object unnecessarily.

caseworker/core/rules.py
```diff
@@ -86,7 +86,7 @@ def is_user_licencing_unit_senior_manager(request):
 @rules.predicate
 def is_case_finalised_and_licence_editable(request, licence):
     is_case_finalised = licence.get("case_status") == CaseStatusEnum.FINALISED
-    is_licence_editable = licence.get("status") in [
+    is_licence_editable = licence.get("status").lower() in [
         LicenceStatusEnum.ISSUED,
```
markj0hnst0n, line 90: Licence status received from the api is capitalised so this is normalised so that it matches db representation provided by `LicenceStatusEnum`

caseworker/templates/case/tabs/licences.html
```diff
@@ -27,6 +31,12 @@
 					<td class="govuk-table__cell govuk-table__cell--tight">
 						{{ licence.status }}
 					</td>
+					{% test_rule 'can_licence_status_be_changed' request licence as can_licence_status_be_changed %}
+					{% if can_licence_status_be_changed %}
+						<td class="govuk-table__cell govuk-table__cell--tight">
+							<a id="{{ licence.id }}" href="{% url 'cases:change_licence_status' queue.id case.id licence.id %}" class="govuk-link">Change status</a>
```
currycoder, line 37: Does the `change_licence_status` view itself also have these sorts of permission checks?  How about the API endpoint?

markj0hnst0n: They do check the role.  Gurdeep made these changes and they are already merged into dev.   PRs [here](https://github.com/uktrade/lite-frontend/pull/2091/) and [here](https://github.com/uktrade/lite-api/pull/2109/)

depsiatwal: Yeah they all have identical checks 
The view uses this permission directly and the API endpoint also has these checks 
https://github.com/uktrade/lite-api/blob/dev/api/licences/views/main.py#L108

currycoder (approved):

depsiatwal (approved):
also looks good to me
only merge this after sign-off and testing in Dev2

PR #2096, opened 2024-08-08
LTD-5161: Update breadcrumbs on product details page if it is an archived product

saruniitr:
## Change description

Once a product is archived the path to view its details is,

Home -> Products list -> Archived products -> Product detail.

Update breadcrumbs links to reflect the same.

[LTD-5161](https://uktrade.atlassian.net/browse/LTD-5161)

[LTD-5161]: https://uktrade.atlassian.net/browse/LTD-5161?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/exporter/goods/views.py
+++ b/exporter/goods/views.py
@@ -160,20 +160,6 @@ class Goods(LoginRequiredMixin, TemplateView):
 
 class ArchivedGoods(LoginRequiredMixin, TemplateView):
 
-    def get_breadcrumbs(self):
-        breadcrumbs = [
-            {
-                "title": "Account home",
-                "url": reverse("core:home"),
-            },
-            {
-                "title": "Product list",
-                "url": reverse("goods:goods"),
-            },
-        ]
-
-        return breadcrumbs
-
     def get(self, request, **kwargs):
         name = request.GET.get("name", "").strip()
         part_number = request.GET.get("part_number", "").strip()
--- a/exporter/goods/views.py
+++ b/exporter/goods/views.py
@@ -37,6 +37,7 @@ from exporter.core.helpers import (
     has_valid_rfd_certificate,
     str_to_bool,
 )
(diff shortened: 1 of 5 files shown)


seijihg (approved):

exporter/goods/views.py
```diff
@@ -158,6 +158,21 @@ def get(self, request, **kwargs):
 
 
 class ArchivedGoods(LoginRequiredMixin, TemplateView):
+
+    def get_breadcrumbs(self):
```
depsiatwal, line 162: any reason why this is in two places ?

unit_tests/exporter/goods/views/test_archive_restore.py
```diff
@@ -230,6 +233,21 @@ def test_archived_product_details_context(
         "Archived by Exporter User1 at 12:46 on 10 April 2024",
     ]
 
+    assert response.context["breadcrumbs"] == [
```
depsiatwal, line 236: do we need a test to show it's not displayed when we don't have HTTP_REFERER=reverse("goods:archived_goods")

exporter/goods/views.py
saruniitr: these are two separate views

unit_tests/exporter/goods/views/test_archive_restore.py
saruniitr: There is only one way to get to this view (unless you save the url) so that may not be required.
I will see if it can added quickly.

exporter/goods/views.py
depsiatwal: but same breadcrumb dict ?
possible to add to a common place in Core ?

depsiatwal (approved):
LGTM

PR #2099, opened 2024-08-09
LTD 5262 filter by licence status on case queue screen

markj0hnst0n:
### Aim

Caseworkers require the ability to filter cases by licence status in case they need to make any changes to licences of a specific status due to a change in circumstances.

This matches current designs given by UCD for where this filter should show.

Corresponding API PR [here](https://github.com/uktrade/lite-api/pull/2127/files#diff-c8faa4a1b6f07603b1967f0b629f858fee1796613876327d3fcbc3bcb9de827d)

[LTD-5262](https://uktrade.atlassian.net/browse/LTD-5262)

[LTD-5262]: https://uktrade.atlassian.net/browse/LTD-5262?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


caseworker/queues/views/forms.py
```diff
@@ -130,6 +131,8 @@ def __init__(self, queue, filters_data, all_flags, all_cles, all_regimes, countr
             (choice["id"], choice["full_name"]) for choice in filters_data["gov_users"]
         ]
 
+        licence_status_choices = self.get_field_choices(LicenceStatusEnum.licence_choices_dict, "licence_statuses")
```
depsiatwal, line 134: do we need the work licence_choices_dict on the basis that it's Licence Enum anyway

markj0hnst0n: Probably not as it adds a little bit of unnecessary complexity.  I've removed this now and used a `choices` tuple which follows existing patterns.

core/constants.py
```diff
@@ -140,3 +142,15 @@ class LicenceStatusEnum:
...
+
+    choices = [
+        (ISSUED, "Issued"),
+        (REINSTATED, "Reinstated"),
+        (REVOKED, "Revoked"),
+        (SURRENDERED, "Surrendered"),
+        (SUSPENDED, "Suspended"),
+        (EXHAUSTED, "Exhasted"),
+        (EXPIRED, "Expired"),
+        (DRAFT, "Draft"),
```
saruniitr, line 154: I think this is not a valid option in the list as you cannot change from draft to any of the other statuses directly.

core/constants.py
```diff
@@ -140,3 +142,15 @@ class LicenceStatusEnum:
...
         "Reinstated",
     )
+
+    choices = [
+        (ISSUED, "Issued"),
+        (REINSTATED, "Reinstated"),
+        (REVOKED, "Revoked"),
+        (SURRENDERED, "Surrendered"),
+        (SUSPENDED, "Suspended"),
+        (EXHAUSTED, "Exhasted"),
```
saruniitr, line 152: There is a typo

core/constants.py
markj0hnst0n: Removed draft option

core/constants.py
markj0hnst0n: amended

saruniitr (approved):

PR #2105, opened 2024-08-12
[LTD-5299] Prevent caseworkers setting superseded status

currycoder:
### Aim

TODO: Rebase on dev after LTD-5225 merges.

[LTD-5299](https://uktrade.atlassian.net/browse/LTD-5299)


saruniitr (commented):
Changes looks good to me.
It should be possible to test that we don't see this option in drop down, are there any issues with the test?
We have relevant tests in this file https://github.com/uktrade/lite-frontend/blob/dev/unit_tests/caseworker/cases/views/test_change_status.py

caseworker/core/services.py
```diff
@@ -84,6 +84,8 @@ def get_permissible_statuses(request, case):
     case_type = case["case_type"]["type"]["key"]
     case_type_applicable_statuses = []
 
+    # TODO: Make this list of dis-allowed caseworker-settable statuses driven
+    # by the API
     if case_type == CaseType.APPLICATION.value:
```
depsiatwal, line 89: happy with this for now as mentioned and I think it's been on the cards for a while but some kind of static service to obtain these from API.

depsiatwal (approved):

PR #2112, opened 2024-08-16
LTD-5284-Fix-show-correct-cle

seijihg:
https://uktrade.atlassian.net/browse/LTD-5284

Exporter fix for the licence template is on BE

This shows only where it matters in `ApplicationDetail` page where it should show only the one where Licence is required and have been assessed by TAU. 

In the `CheckYourAnswers` and `GoodsDetailSummaryCheckYourAnswers` it will stay how it was before so showing the ones added by Users.

### DIFF
--- a/exporter/applications/helpers/check_your_answers.py
+++ b/exporter/applications/helpers/check_your_answers.py
@@ -32,21 +32,20 @@ from core.builtins.custom_tags import (
 )
 from exporter.core.helpers import convert_to_link, convert_control_list_entries
 from exporter.goods.helpers import requires_serial_numbers
+from exporter.applications.constants import ApplicationStatus
 
 from lite_content.lite_exporter_frontend import applications
 from lite_content.lite_exporter_frontend.strings import Parties
 from lite_forms.helpers import conditional
 
 
-def convert_application_to_check_your_answers(application, editable=False, summary=False, is_application_detail=False):
+def convert_application_to_check_your_answers(application, editable=False, summary=False):
     """
     Returns a correctly formatted check your answers page for the supplied application
     """
     sub_type = application.sub_type
     if sub_type == STANDARD:
-        return _convert_standard_application(
-            application, editable, is_summary=summary, is_application_detail=is_application_detail
-        )
+        return _convert_standard_application(application, editable, is_summary=summary)
     elif sub_type == OPEN:
         return _convert_open_application(application, editable)
     elif sub_type == HMRC:
@@ -102,14 +101,14 @@ def _convert_gifting_clearance(application, editable=False):
     }
 
 
-def _convert_standard_application(application, editable=False, is_summary=False, is_application_detail=False):
+def _convert_standard_application(application, editable=False, is_summary=False):
     strings = applications.ApplicationSummaryPage
     pk = application["id"]
     url = reverse(f"applications:good_detail_summary", kwargs={"pk": pk})
     old_locations = bool(application["goods_locations"])
     converted = {
         convert_to_link(url, strings.GOODS): convert_goods_on_application(
-            application, application["goods"], is_summary=is_summary, is_application_detail=is_application_detail
+            application, application["goods"], is_summary=is_summary
         ),
         strings.END_USE_DETAILS: _get_end_use_details(application),
         strings.END_USER: convert_party(application["end_user"], application, editable),
@@ -240,9 +239,7 @@ def _convert_hmrc_query(application, editable=False):
     }
 
 
-def convert_goods_on_application(
-    application, goods_on_application, is_exhibition=False, is_summary=False, is_application_detail=False
-):
+def convert_goods_on_application(application, goods_on_application, is_exhibition=False, is_summary=False):
     converted = []
 
     def requires_actions(application, good_on_application):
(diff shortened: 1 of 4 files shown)


exporter/applications/views/common.py
```diff
@@ -302,12 +302,11 @@ def dispatch(self, request, *args, **kwargs):
 
     def get(self, request, **kwargs):
         status_props, _ = get_status_properties(request, self.application["status"]["key"])
-
         context = {
             "case_id": self.application_id,
             "application": self.application,
             "type": self.view_type,
-            "answers": convert_application_to_check_your_answers(self.application),
+            "answers": convert_application_to_check_your_answers(self.application, is_application_detail=True),
```
kevincarrogan, line 309: Is this called on rge “Check your answers” page here?

exporter/applications/helpers/check_your_answers.py
```diff
@@ -238,15 +240,20 @@ def _convert_hmrc_query(application, editable=False):
...
+):
     converted = []
 
     def requires_actions(application, good_on_application):
         return not is_summary and requires_serial_numbers(application, good_on_application)
 
     requires_actions_column = any(requires_actions(application, g) for g in goods_on_application)
     for good_on_application in goods_on_application:
-        control_list_entries = convert_control_list_entries(good_on_application["good"]["control_list_entries"])
+        if is_application_detail:
```
kevincarrogan, line 253: If I’m following this through right then this is the bit that runs on the “Check your answers” page? Is my understanding correct?

seijihg: All of them use the same method:
`convert_application_to_check_your_answers` Thats the reason they all lead to that code. So the thing I am changing here is just for the `ApplicationDetail` View which have:
`is_application_detail=True`

`convert_application_to_check_your_answers(self.application, is_application_detail=True)`

exporter/applications/views/common.py
seijihg: no its not, CheckYourAnswers is here below:

https://github.com/uktrade/lite-frontend/blob/cdf70e47ed41edc35651e0d2abf397164df35775/exporter/applications/views/common.py#L420

kevincarrogan: Which page is the application detail view?

seijihg: `/applications/38232a08-c394-4a36-9c11-e144e84d70ae/`
https://github.com/uktrade/lite-frontend/blob/cdf70e47ed41edc35651e0d2abf397164df35775/exporter/applications/urls.py#L535

The ones where you go to Account Home -> Applicaitons then pick one.

kevincarrogan (approved):

exporter/applications/helpers/check_your_answers.py
saruniitr: This fix is an improvement over the current code but we will still be displaying CLEs from `Good` on check your answers page. That is something we need to address but out of scope of this PR.

@seijihg 
I think you can achieve this in a simpler way (instead of adding `is_application_detail` argument).

If it is application detail page, that means application is submitted where as check your answers page means it is in draft state. So you can check if the status is not draft (then it is application detail) and if it is draft (then it is check your answers page).

With that change I think this is good to go.

unit_tests/exporter/applications/helpers/test_check_your_answers.py
```diff
@@ -222,3 +222,18 @@ def test_convert_standard_application_has_security_approvals(application, settin
     test_application = Application(application)
     actual = check_your_answers._convert_standard_application(test_application)
     assert "Do you have a security approval?" in actual.keys()
+
+
+def test_goods_on_application_when_application_detail(application, data_good_on_application):
```
saruniitr, line 227: you may need to parameterize this test for 'draft' and some other status other than draft and check CLEs or separate tests if it cannot be parameterized.

exporter/applications/helpers/check_your_answers.py
seijihg: I have updated the code! thanks

unit_tests/exporter/applications/helpers/test_check_your_answers.py
```diff
@@ -222,3 +226,30 @@ def test_convert_standard_application_has_security_approvals(application, settin
...
+        (
+            {"key": ApplicationStatus.DRAFT, "value": "Draft"},
+            False,
+            "<span data-definition-title='ML4' data-definition-text='Rifle for research and development...'>ML4</span>, <span data-definition-title='ML5' data-definition-text='Smart ammunition...'>ML5</span>",
+        ),
+    ),
+)
+def test_goods_on_application_when_application_detail(
+    application, data_good_on_application, status, is_good_controlled, cle_results
+):
```
saruniitr, line 248: This is not the correct approach because we are only calling the function where the changes are done. It doen't confirm whether the correct CLEs are displayed on the final template.
We should just `get` the application detail page, check your answers page and ensure CLEs are correctly displayed.

Also currently only one product is being checked but this application has two products so it is incomplete.

seijihg: Updated. I did not write a test for summary page since it exists already.

unit_tests/exporter/applications/helpers/test_check_your_answers.py
```diff
@@ -222,3 +238,25 @@ def test_convert_standard_application_has_security_approvals(application, settin
     test_application = Application(application)
     actual = check_your_answers._convert_standard_application(test_application)
     assert "Do you have a security approval?" in actual.keys()
+
+
+def test_application_detail_shows_correct_cles(authorized_client, mock_application_get, application):
+    application["status"] = {"key": ApplicationStatus.SUBMITTED, "value": "Submitted"}
```
saruniitr, line 244: this line is not necessary as `mock_application_get` already gives a submitted application.

unit_tests/exporter/applications/helpers/test_check_your_answers.py
```diff
@@ -222,3 +238,25 @@ def test_convert_standard_application_has_security_approvals(application, settin
...
+    url = reverse("applications:application", kwargs={"pk": application["id"]})
+    response = authorized_client.get(url)
+    soup = BeautifulSoup(response.content, "html.parser")
+
+    products_table = soup.find_all("table", attrs={"class": "govuk-table"})[0]
+    products_data = []
```
saruniitr, line 250: As it is just saving CLEs, may be name it accordingly?

saruniitr (approved):

PR #2119, opened 2024-08-19
LTD-5340-add-draft-licence-filter 

depsiatwal:
https://uktrade.atlassian.net/browse/LTD-5340

Draft was missing from the Licence filter so added this and a test to check they're all there

### DIFF
--- a/core/constants.py
+++ b/core/constants.py
@@ -152,4 +152,5 @@ class LicenceStatusEnum:
         (EXHAUSTED, "Exhausted"),
         (EXPIRED, "Expired"),
         (CANCELLED, "Cancelled"),
+        (DRAFT, "Draft"),
     ]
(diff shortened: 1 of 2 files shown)


currycoder (approved):

core/constants.py
```diff
@@ -152,4 +152,5 @@ class LicenceStatusEnum:
         (EXHAUSTED, "Exhausted"),
         (EXPIRED, "Expired"),
         (CANCELLED, "Cancelled"),
+        (DRAFT, "Draft"),
```
saruniitr, line 155: What is the need to search for `draft` licences?

depsiatwal: Raised as a bug 
gives the user the freedom to allow to search for a draft Licence during Licence status filtering.

depsiatwal: Just to clarify this is only used for searching and allows the user to search for Draft statuses no impact on ability to change the status.

saruniitr (approved):

PR #2120, opened 2024-08-20
LTD-5332 added validator for good name

Tllew:
### Aim

[LTD-5332](https://uktrade.atlassian.net/browse/LTD-5332)

[LTD-5332]: https://uktrade.atlassian.net/browse/LTD-5332?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/exporter/goods/validators.py
+++ b/exporter/goods/validators.py
@@ -4,6 +4,7 @@ import re
 
 def validate_name(value):
     if value:
-        match_regex = re.sub(r"[^a-zA-Z0-9 .,\\-\\)\\/'+:=\\?\\!\"%&\\*;\\<\\>]", "", value)
-        if len(match_regex) < len(value):
-            raise ValidationError("""Invalid character, allowed characters: A-Z a-z 0-9 -()/'+:=?!"_ ."%&*;<>""")
+        match_regex = re.compile(r"^[a-zA-Z0-9 .,\-\)\(\/'+:=\?\!\"%&\*;\<\>]+$")
+        is_value_valid = bool(match_regex.match(value))
+        if not is_value_valid:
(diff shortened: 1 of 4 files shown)


exporter/goods/validators.py
```diff
@@ -0,0 +1,9 @@
+from django.core.exceptions import ValidationError
+import re
+
+
+def validate_name(value):
+    if value:
+        match_regex = re.sub(r"[^a-zA-Z0-9 .,\\-\\)\\/'+:=\\?\\!\"%&\\*;\\<\\>]", "", value)
+        if len(match_regex) < len(value):
```
currycoder, line 8: Maybe this could be a bit more readable?

```
match_regex = re.compile(r"...")
is_value_valid = bool(match_regex.match(value))
if not is_value_valid:
    raise ValidationError(...)
```

unit_tests/exporter/goods/test_validators.py
```diff
@@ -0,0 +1,22 @@
...
+    assert validate_name(name) is None
+
+
+@pytest.mark.parametrize(
+    "name",
+    ("\r\n"),
+)
+def test_validate_goods_name_invalid(name):
+    with pytest.raises(ValidationError):
+        validate_name(name)
```
currycoder, line 22: I think I'd like to see a bunch more tests.  e.g. one with a string with every single valid special character, ones with known invalid characters (that kicked off this work in the first place) etc etc

Right now, I don't think the regex works for `"good-name"` when I think it should do..

depsiatwal (approved):
approved

PR #2123, opened 2024-08-21
LTD-5253-licence-change-e2e-tests 

depsiatwal:
Added e2e test to prove flow for changing licence works from starting application until licence is suspended.

The only user role which can complete the licence change is LU Senior manager and as such we had to log out the standard test user and login a user with that role.

We added the ability to logout of the mock_sso system on casewoker which previously was not present.

We have removed code which automatically logs a user in id the mock_sso_email is present.  This means that all mock_sso users must now login explicitly and so all caseworker e2e tests needed to be updated to reflect this change.

[LTD-5253](https://uktrade.atlassian.net/browse/LTD-5253)

[LTD-5253]: https://uktrade.atlassian.net/browse/LTD-5253?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


core/mock_sso/views.py
```diff
@@ -97,12 +98,6 @@ def get(self, request, *args, **kwargs):
         if session_user_email:
             return self.redirect_to_redirect_uri(request, session_user_email)
 
-        # If we have an explicit email set in settings then we don't need to
-        # show the mock sso login prompt so we can just redirect back.
-        mock_sso_user_email = getattr(settings, "MOCK_SSO_USER_EMAIL", None)
-        if mock_sso_user_email:
-            return self.redirect_to_redirect_uri(request, mock_sso_user_email)
-
         return super().get(request, *args, **kwargs)
```
markj0hnst0n, line 101: Removed this so that mock user needs to login explicitly each time rather than login details taken from env var.  This gives us the ability to login and log out during e2e tests and so makes e2e test closer to real user behaviour.

caseworker/mock_sso/urls.py
```diff
@@ -9,4 +9,5 @@
     path("o/authorize/", views.Authorize.as_view(), name="authorize"),
     path("o/token/", views.Token.as_view(), name="token"),
     path("api/v1/user/me/", views.APIUserMe.as_view(), name="api_user_me"),
+    path("auth/logout/", views.Logout.as_view(), name="logout"),
 ]
```
markj0hnst0n, line 13: Logout added to enable e2e tests to logout user.

ui_tests/caseworker/conftest.py
```diff
@@ -490,11 +524,6 @@ def should_see_case_status(driver, status):  # noqa
     assert CasePage(driver).get_status() == status
 
 
-@when("I go to users")  # noqa
-def go_to_users(driver, sso_sign_in, internal_url):  # noqa
-    driver.get(internal_url.rstrip("/") + "/users/")
-
-
 @given("an Exhibition Clearance is created")  # noqa
```
markj0hnst0n, line 527: Duplicate code removed

unit_tests/caseworker/mock_sso/test_views.py
markj0hnst0n, line 1: the tests for the settings email have been removed as that code has been removed and does not need to be tested

unit_tests/exporter/mock_sso/test_views.py
markj0hnst0n, line 1: again the tests for the settings email have been removed as that code has been removed and does not need to be tested

caseworker/templates/case/tabs/licences.html
```diff
@@ -28,13 +28,13 @@
 						{{ good.usage }} of {{ good.quantity }}<br>
 						{% endfor %}
 					</td>
-					<td class="govuk-table__cell govuk-table__cell--tight">
+					<td id="licence-status-id" class="govuk-table__cell govuk-table__cell--tight">
 						{{ licence.status }}
```
markj0hnst0n, line 32: id added to easily find in e2e tests

hnryjmes:
@markj0hnst0n there are a lot of commits here. I'm not sure if it's best to leave it as is, or squash them down into a few bigger changes that are sensibly grouped together.

docker-compose.yml
```diff
@@ -48,7 +48,7 @@ services:
       - "host.docker.internal:host-gateway"
 
   chrome:
-    image: "seleniarm/node-chromium:latest"
+    image: selenium/node-chromium:latest
```
hnryjmes, line 51: what was the reason for this change?

markj0hnst0n: I actually changed it because I thought it was a spelling mistake but on investigation it's just another image which seems to do the same job.  I can change it back.

makefile
```diff
@@ -100,7 +100,7 @@ stop-exporter:
 
 caseworker-e2e-selenium-test:
 	@echo "*** Requires starting the caseworker stack, which can be started running: 'make start-caseworker' ***"
-	$(docker-e2e-caseworker) exec caseworker bash -c '$(wait-for-caseworker)' && PIPENV_DOTENV_LOCATION=caseworker.env pipenv run pytest --circleci-parallelize --headless --chrome-binary-location=/usr/bin/google-chrome -vv --gherkin-terminal-reporter --junitxml=test_results/output.xml ./ui_tests/caseworker
+	$(docker-e2e-caseworker) exec caseworker bash -c '$(wait-for-caseworker)' && docker exec -it api pipenv run ./manage.py seedinternalusers && PIPENV_DOTENV_LOCATION=caseworker.env pipenv run pytest --circleci-parallelize --headless --chrome-binary-location=/usr/bin/google-chrome -vv --gherkin-terminal-reporter --junitxml=test_results/output.xml ./ui_tests/caseworker
```
hnryjmes, line 103: maybe the `docker exec -it api pipenv run ./manage.py seedinternalusers` could be defined as a separate make step? then it can be referenced more nicely in this command. would this work?
```
seed-internal-users:
    docker exec -it api pipenv run ./manage.py seedinternalusers
```
then
```
$(seed-internal-users)
```
can be used instead

docker-compose.yml
hnryjmes: I mean I would do what's best for cross compatibility. there was a whole saga a while ago (April 2023?) about getting e2e tests working locally on our (then new) ARM based macs. however it's possible that the selenium image now supports ARM:

https://stackoverflow.com/questions/76857893/is-there-a-known-working-configuration-for-using-selenium-on-linux-arm64
https://www.selenium.dev/blog/2024/multi-arch-images-via-docker-selenium/

reading the above, it looks like from image `4.21.0` onwards (May 2024), ARM compatibility is added, so there is potentially no need for seleniarm anymore. although I see elsewhere in this PR you pin the version at `4.20.0` so maybe we should try and work out what the impacts of that are.

@Tllew @kevincarrogan FYI

ci.api.env
```diff
@@ -2,7 +2,7 @@ DEBUG=True
 DATABASE_URL=postgres://postgres:password@db:5432/lite-api
 PORT=8100
 ALLOWED_HOSTS=*
-INTERNAL_USERS='[{"email": "test-uat-user@digital.trade.gov.uk", "role": "Super User"}, {"email": "lite-ops@digital.trade.gov.uk", "role": "Super User"}, {"email": "lite-team-1@digital.trade.gov.uk", "role": "Super User"}, {"email": "lite-team-2@digital.trade.gov.uk", "role": "Super User"}, {"email": "lite-team-3@digital.trade.gov.uk", "role": "Super User"}]'
+INTERNAL_USERS='[{"email": "test-uat-user@digital.trade.gov.uk", "role": "Super User"}, {"email": "lite-ops@digital.trade.gov.uk", "role": "Super User"}, {"email": "lite-team-1@digital.trade.gov.uk", "role": "Super User"}, {"email": "lite-team-2@digital.trade.gov.uk", "role": "Super User"}, {"email": "lite-team-3@digital.trade.gov.uk", "role": "Super User"}, {"email": "luseniormanager@example.com", "role": "Licencing Unit Senior Manager", "first_name":"LU", "last_name":"Senior Manager"}]'
```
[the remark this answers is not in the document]
hnryjmes: also, very minor - internal user emails always end with `gov.uk` I think. what about `lu.seniormanager@example.gov.uk` ?

docker-compose.yml
markj0hnst0n: The selinium image for chrome was pinned to the `4.20.0` version as the latest version wouldn't run the tests locally for mine nor Gurdeep's mac so we had to move to an earlier version.

ui_tests/caseworker/features/case_notes.feature
```diff
@@ -6,20 +6,22 @@ Feature: I want to add an internal note to a case and view notes
 
   @add_case_note
   Scenario: Add a new valid case note
-    Given I sign in to SSO or am signed into SSO
+    Given I sign in as "test-uat-user@digital.trade.gov.uk" # /PS-IGNORE
```
saruniitr, line 9: I think it is better not to include email addresses in the steps because we have to repeat whereever it is used. Being the login step we will have to repeat in all scenarios. Like in this case which is not related to licence at all.

What is missing in the previous step which required this change?
One alternative is that we change the step to,

`Given I sign in as Test UAT user` which we can have a different step for different users and in the definition we can this email address.
`Give I sign in as LU Senior manager`

ui_tests/caseworker/conftest.py
```diff
@@ -76,6 +78,34 @@ def sign_into_sso(driver, sso_sign_in):  # noqa
...
+        MockSigninPage(driver).sign_in(email)
+
+
+@when(parsers.parse('I sign in as "{email}"'))
+def when_i_go_to_mock_sso_caseworker_sign_in(driver, internal_url, email):  # noqa
+    driver.get(internal_url)
+    mock_sso_login_screen = driver.find_element(By.XPATH, "//*[contains(text(), 'Mock SSO Login')]")
+
+    if mock_sso_login_screen and settings.MOCK_SSO_ACTIVATE_ENDPOINTS:
+        MockSigninPage(driver).sign_in(email)
```
kevincarrogan, line 96: Does there need to be two of these?

Can you possibly use two decorators wrapped around the same function?

ui_tests/caseworker/features/case_notes.feature
kevincarrogan: I agree with this. The email here doesn't show the intention of who this test is being run as.

ui_tests/caseworker/features/case_notes.feature
```diff
@@ -6,20 +6,22 @@ Feature: I want to add an internal note to a case and view notes
...
   Scenario: Case note cancel button
-    Given I sign in to SSO or am signed into SSO
+    Given I sign in as "test-uat-user@digital.trade.gov.uk" # /PS-IGNORE
     And I create standard application or standard application has been previously created
     When I go to application previously created
     And I click on the notes and timeline tab
     And I enter "Case note to cancel" as the case note
     And I click cancel button
     Then entered text is no longer in case note field
+    And I logout
```
kevincarrogan, line 27: There is a part of me that thinks we shouldn't necessarily log out at the end of every test and maybe it's just something that happens automatically.

One of the reasons why I'm thinking about this is because an earlier step might fail and this wouldn't get executed and could then make the rest of the tests fail.

core/mock_sso/views.py
```diff
@@ -149,3 +144,10 @@ def get(self, request, **kwargs):
             "access_profiles": [],
         }
         return JsonResponse(response_data)
+
+
+class Logout(RedirectView):
+    def get_redirect_url(self, *args, **kwargs):
+        redirect_url = self.request.build_absolute_uri("/")
+        self.request.session.flush()
```
kevincarrogan, line 152: This is a pretty quick and dirty thing to do but it is odd that we are flushing the session when retrieving the URL as opposed to in a more suitable place.

ui_tests/caseworker/conftest.py
markj0hnst0n: There does need to be 2 of these.  I'm not sure what's driving it.  Either the 'given'/'when' decorator or the parsing.  I expect the decorator but I have tidied the 'I logout' by wrapping it in 2 decorators.

markj0hnst0n: I've removed them in favour of more explicit naming of the login actions to refer to the user they are logging in.

ui_tests/caseworker/features/case_notes.feature
markj0hnst0n: I've added this type of explicit login action now

core/mock_sso/views.py
markj0hnst0n: I think it just looks quick and dirty as it's a mock logout but maybe I'm wrong.

ci.api.env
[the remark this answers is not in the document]
markj0hnst0n: Changed that email to a more realistic one

makefile
markj0hnst0n: I have done this in every conceivable way I can think of and it never seeds the db correctly.

tests_common/fixtures/core.py
```diff
@@ -8,20 +8,20 @@
 MAX_WORKERS = 3
 
 
-@fixture(scope="session")
+@fixture()
```
saruniitr, line 11: this is a new change in this revision, is this to address any issue?

saruniitr (approved):
With any outstanding comments from other reviewers addressed it looks good to me.

PR #2132, opened 2024-08-30
LTD-5380: LU countersigning restrictions

saruniitr:
## Change description

After LU gives their final recommendation some cases go through countersigning depending on the destinations/assessments. We have a requirement to make sure the same user who has given final recommendation (case officer) also cannot countersign the case.
In some cases two countersignatures are required so in such situations requirement is to ensure each of those countersignatures are done by a unique user.

To achieve this a new rule is added which determines whether current user is allowed to countersign or not.
If allowed then the action buttons are displayed otherwise a warning message is displayed.

Add unit tests to cover all scenarios.

[LTD-5380](https://uktrade.atlassian.net/browse/LTD-5380)

[LTD-5380]: https://uktrade.atlassian.net/browse/LTD-5380?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


caseworker/advice/templates/advice/view-advice.html
```diff
@@ -36,14 +36,20 @@ <h2 class="govuk-error-summary__title" id="error-summary-title">
...
+            {% if buttons.review_and_countersign and can_user_be_allowed_to_lu_countersign %}
                 <a id="review-and-countersign-decision-button" draggable="false" class="govuk-button govuk-button--primary" href="{% url 'cases:countersign_decision_review' queue_pk case.id %}">Review and countersign</a>
-            {% else %}
-                <a id="review-and-countersign-button" draggable="false" class="govuk-button govuk-button--primary" href="{% url 'cases:countersign_review' queue_pk case.id %}">Review and countersign</a>
+            {% elif not can_user_be_allowed_to_lu_countersign %}
+                {% include "advice/countersign_not_allowed_warning.html" %}
             {% endif %}
+        {% elif countersign and not is_lu_countersigning and can_user_review_and_countersign%}
+            <!-- FCDO countersigning -->
+            <a id="review-and-countersign-button" draggable="false" class="govuk-button govuk-button--primary" href="{% url 'cases:countersign_review' queue_pk case.id %}">Review and countersign</a>
```
saruniitr, line 52: Re-arranged to separate LU and FCDO countersigning branches.

unit_tests/caseworker/advice/views/test_countersign.py
```diff
@@ -303,3 +509,142 @@ def test_lu_countersign_get_shows_previous_countersignature(
...
+    assert countersign_button is None
+
+    message = soup.find("div", {"id": "lu-user-not-allowed-to-countersign"})
+    assert message
+    assert "You cannot countersign this case because you were the officer that assessed it." in message.text
+    return True
+
+
+@patch("caseworker.advice.views.get_gov_user")
+@patch("caseworker.core.rules.get_logged_in_caseworker")
```
saruniitr, line 527: this also need to be mocked as the user is actually set in request middleware hence mocking the function where we retrieve this user.

caseworker/advice/services.py
```diff
@@ -593,7 +609,13 @@ def get_advice_tab_context(case, caseworker, queue_id):
...
         if queue_alias in (LU_LICENSING_MANAGER_QUEUE, LU_SR_LICENSING_MANAGER_QUEUE):
             advice_to_countersign = get_advice_to_countersign(case.advice, caseworker)
-            countersigned_by = get_countersigners_decision_advice(case, caseworker)
+
+            order = FIRST_COUNTERSIGN
+            if queue_alias == LU_SR_LICENSING_MANAGER_QUEUE:
+                order = SECOND_COUNTERSIGN
+
+            # Get the countersigners of the current queue
+            countersigned_by = get_countersigners_decision_advice(case, caseworker, order)
```
saruniitr, line 618: This change is to fix a hole in the current code.
We are first checking if the user has already countersigned or not so that we present countersign/edit countersign options and we are looking for the countersigned advice on both queues (as they are for the same team). However if licencing manager countersigned and he also switches to senior licencing manager queue then we see that he has already countersigned and present edit options. So potentially user can edit countersignature from a different queue.
User should not be able to perform any action in this case hence check the advice given for that queue only.

ui_tests/caseworker/features/give_advice/lu_countersign.feature
```diff
@@ -4,7 +4,7 @@ Feature: I want to record my user advice and any comments and conditions relatin
   I want to record my user advice and any comments and conditions relating to my recommendation
   So that other users can see my decision and know that I have finished assessing this case
 
-  @lu_countersign
+  @skip @lu_countersign
```
saruniitr, line 7: Skipping this test along with other tests because these changes breaks these tests.
This PR is already large so these are going to be fixed in the next PR where new E2E tests are going to be add to cover all cases.
Also changes required to fix these are a bit involved as we need to switch users when countersigning.
This is not risky because the plan is to get these changes user tested first before merging to `dev` and release.

caseworker/core/rules.py
```diff
@@ -71,6 +94,23 @@ def case_has_ogd_advice(request, case):
...
+
+    if not case_has_ogd_advice(request, case):
+        return False
+
+    if not filter_advice_by_teams(case["advice"], [LICENSING_UNIT_TEAM]):
+        return False
+
+    if not filter_advice_by_level(case["advice"], [AdviceLevel.FINAL]):
+        return False
+
```
currycoder, line 110: These stack of ifs feel a bit off to me.  Is there any way that we could change things around to just find some advice with the characteristics we are looking for a bit more explicitly?

e.g. something like

```
final_advice = filter_advice_by_level(case["advice"], [AdviceLevel.FINAL])
final_advice_from_lu = [advice for advice in final_advice if advice["team"] == LICENSING_UNIT_TEAM]
return len(final_advice_from_lu) > 0
```

currycoder (commented):
A few comments/questions inline.

If I were to pull this branch down locally to try this out, could you give me some hints on how to get a case in to a state where I could do that?

caseworker/advice/services.py
```diff
@@ -223,13 +230,22 @@ def get_countersigners(advice_to_countersign):
...
     """Get a set of user ids representing the users that have already
-    countersigned the advice on this case with accept/reject decision.
+    countersigned the advice of the given order on this case with accept/reject decision.
     """
     countersigned_by = set()
+    countersign_orders = [FIRST_COUNTERSIGN, SECOND_COUNTERSIGN]
+
+    if order:
+        countersign_orders = [order]
+
```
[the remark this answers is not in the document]
saruniitr: Correct, this is only for LU.

caseworker/core/rules.py
saruniitr: I was only planning to check for final level advice as only LU can give that but your suggest looks good to me, will update.

caseworker/core/rules.py
```diff
@@ -98,17 +138,52 @@ def is_case_finalised_and_licence_editable(request, licence):
...
+
+@rules.predicate
+def user_not_yet_countersigned(request, case):
+
+    caseworker = get_logged_in_caseworker(request)
+    if not caseworker:
+        return False
+
+    countersigners = get_countersigners_decision_advice(case, caseworker)
```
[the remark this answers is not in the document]
saruniitr: Yes it possible to go through multiple times. When things go back previous countersigned advice is marked as invalid (there is a field `valid` which will be set to `False`) and this function only looks through valid advice only so this case is covered.

This is the current behaviour already.

caseworker/core/rules.py
saruniitr: fixed

caseworker/core/rules.py
```diff
@@ -98,17 +134,52 @@ def is_case_finalised_and_licence_editable(request, licence):
...
+
+    return caseworker["id"] not in countersigners
```
hnryjmes, line 168: this is serialized data correct? so it is checking a string is not in a list of strings? i.e. it is not UUID type anymore?

hnryjmes (approved):
happy with the code changes and unit test coverage. I understand that we have skipped the e2e tests but this is to be worked on in the other PR so not a concern here.

caseworker/core/rules.py
saruniitr: Correct, `case` contains all serialized data from which we are extracting ids of users who countersigned, they are UUIDs in string form, not UUID type.

currycoder (approved):

caseworker/advice/services.py
```diff
@@ -223,13 +230,22 @@ def get_countersigners(advice_to_countersign):
...
 
 
-def get_countersigners_decision_advice(case, caseworker):
+def get_countersigners_decision_advice(case, caseworker, order=None):
     """Get a set of user ids representing the users that have already
-    countersigned the advice on this case with accept/reject decision.
+    countersigned the advice of the given order on this case with accept/reject decision.
     """
     countersigned_by = set()
+    countersign_orders = [FIRST_COUNTERSIGN, SECOND_COUNTERSIGN]
```
kevincarrogan, line 238: This is probably something to refactor later, but if someone chose to update the `CountersignOrder` enum this is going to get out of sync.

What ensures that someone would know to update this if they decided to update `CountersignOrder`?

unit_tests/caseworker/conftest.py
```diff
@@ -637,23 +637,52 @@ def mock_gov_tau_user(requests_mock, mock_notifications, mock_case_statuses, moc
...
 
+@pytest.fixture
+def mock_gov_lu_case_officer(requests_mock, LU_case_officer):
+    url = client._build_absolute_uri("/gov-users/")
+    data = {"user": LU_case_officer}
+
+    requests_mock.get(url=f"{url}me/", json=data)
+    requests_mock.get(url=re.compile(f"{url}{LU_case_officer['id']}/"), json=data)
+
+    yield data
```
kevincarrogan, line 656: We don't need to `yield` here as there's not clean up after, this could just be `return`.

unit_tests/caseworker/core/test_rules.py
```diff
@@ -483,3 +486,191 @@ def test_can_licence_status_be_changed(
...
+    case = {
+        "case_officer": LU_case_officer,
+        "advice": [
+            {
+                "level": AdviceLevel.FINAL,
+                "user": LU_case_officer,
+            },
+        ],
+    }
+    user = request.getfixturevalue(gov_user)
```
kevincarrogan, line 577: I've learnt something new today watch me use this way too much in the near future.

caseworker/core/rules.py
```diff
@@ -98,17 +134,52 @@ def is_case_finalised_and_licence_editable(request, licence):
...
+
     return case.status in get_caseworker_operable_case_statuses(request)
 
 
+@rules.predicate
+def user_is_not_final_adviser(request, case):
+
+    caseworker = get_logged_in_caseworker(request)
+    if not caseworker:
+        return False
```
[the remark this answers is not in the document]
saruniitr: yes very likely things fail much earlier as this is set in the middleware.

unit_tests/caseworker/conftest.py
saruniitr: ok, copied from another fixture

unit_tests/caseworker/core/test_rules.py
saruniitr: yes this is very handy to parameterize fixture with different data.

caseworker/advice/services.py
saruniitr: agree, this should've been an enum instead of individual constants.

unit_tests/caseworker/conftest.py
saruniitr: fixed

PR #2133, opened 2024-08-30
Update exporter app to use new CLEs list view

hnryjmes:
This is dependent on this API change here: https://github.com/uktrade/lite-api/pull/2160/

This changes the exporter frontend to use the new exporter-specific CLEs endpoint.

[LTD-5381](https://uktrade.atlassian.net/browse/LTD-5381)

[LTD-5381]: https://uktrade.atlassian.net/browse/LTD-5381?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/exporter/core/services.py
+++ b/exporter/core/services.py
@@ -198,19 +198,23 @@ def put_organisation_user(request, user_pk, json):
     return data.json(), data.status_code
 
 
-def get_control_list_entries(request, convert_to_options=False):
-    response = client.get(request, "/exporter/static/control-list-entries/")
-
+def get_control_list_entries(request, convert_to_options=False, converted_control_list_entries_cache=[]):  # noqa
     if convert_to_options:
-        return [
-            Option(
-                key=control_list_entry["rating"],
-                value=control_list_entry["rating"],
-                description=control_list_entry["text"],
-            )
-            for control_list_entry in response.json()
-        ]
+        if converted_control_list_entries_cache:
+            return converted_control_list_entries_cache
+        else:
+            response = client.get(request, "/exporter/static/control-list-entries/")
+            for control_list_entry in response.json():
+                converted_control_list_entries_cache.append(
+                    Option(
+                        key=control_list_entry["rating"],
+                        value=control_list_entry["rating"],
+                        description=control_list_entry["text"],
+                    )
+                )
+            return converted_control_list_entries_cache
 
+    response = client.get(request, "/exporter/static/control-list-entries/")
     return response.json()
 
 
--- a/exporter/core/services.py
+++ b/exporter/core/services.py
@@ -198,26 +198,20 @@ def put_organisation_user(request, user_pk, json):
     return data.json(), data.status_code
 
 
-def get_control_list_entries(request, convert_to_options=False, converted_control_list_entries_cache=[]):  # noqa
+def get_control_list_entries(request, convert_to_options=False):
+    response = client.get(request, "/static/control-list-entries/exporter-list/")
+
     if convert_to_options:
-        if converted_control_list_entries_cache:
-            return converted_control_list_entries_cache
-        else:
-            data = client.get(request, "/static/control-list-entries/?flatten=True")
-
-        for control_list_entry in data.json().get("control_list_entries", []):
-            converted_control_list_entries_cache.append(
-                Option(
-                    key=control_list_entry["rating"],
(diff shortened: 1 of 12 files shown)


exporter/core/services.py
```diff
@@ -198,26 +198,20 @@ def put_organisation_user(request, user_pk, json):
     return data.json(), data.status_code
 
 
-def get_control_list_entries(request, convert_to_options=False, converted_control_list_entries_cache=[]):  # noqa
```
currycoder, line 201: That mutable default `converted_control_list_entries_cache` was doing some very heavy lifting that we might want to have a small discussion about...

It would make it so that the control list entries API endpoint was only called once for the running exporter django instance.  Subsequent calls would use the values cached in that variable instead.  e.g. the first time an exporter went to a page which needed CLEs, the API endpoint would be called.  Subsequent times, the cached value would be used instead.  Up until the running exporter container was recycled (or until we did a new release).

Maybe the performance impact/load increase will be negligible but it could be good to think that through...  Could it be possible to measure how long this endpoint takes to return on a real environment?

currycoder: BTW, I'm in no way saying that doing this with a mutable default is the right way to do it - just thinking through the performance concerns of getting rid of that (odd) caching mechanism.

hnryjmes: I did notice this, but I thought I saw that `converted_control_list_entries_cache` wasn't used anywhere. If you search for `get_control_list_entries(` in lite-frontend you can see the references, but there aren't any function calls which pass a `converted_control_list_entries_cache` list as an argument. If it's never called with that list of previously cached results, then it always does an API call, so it's like that caching mechanism doesn't exist. From that I concluded that it's just a legacy bit of code that no longer is actively used.

currycoder: Well that's the fun thing about this way of caching...  It works because **the argument is never sepcified**.

In general, mutable values should **never** be set as defaults for keyword arguments.  That's because the variable exists at the file level.  e.g. it's like setting a variable completely outside of the function.
```
def add_to_cache(cache=[]):
    cache.append("foo")
    print(cache)

add_to_cache()
> ["foo"]
add_to_cache()
> ["foo", "foo"]
add_to_cache()
> ["foo", "foo", "foo"]
```

So this caching hack that we have is depending on this very detail about mutable default values.  With this in mind, the points I was making in the parent comment;

> It would make it so that the control list entries API endpoint was only called once for the running exporter django instance. Subsequent calls would use the values cached in that variable instead. e.g. the first time an exporter went to a page which needed CLEs, the API endpoint would be called. Subsequent times, the cached value would be used instead. Up until the running exporter container was recycled (or until we did a new release).

> Maybe the performance impact/load increase will be negligible but it could be good to think that through... Could it be possible to measure how long this endpoint takes to return on a real environment?

hnryjmes: thanks for this and very well explained.

is there a way we can add the caching mechanism again but do it in a more Djangoish way? can you think of anything that might help us here? dw if not, I can also look into this.

otherwise I suppose the fall back is to literally just add in that mutable default that was there before.

hnryjmes: what about `from django.core.cache import cache` then `cache.get()` and `cache.set()`? https://docs.djangoproject.com/en/4.2/topics/cache/#basic-usage

currycoder: Yes we could use django's caching mechanism; https://docs.djangoproject.com/en/5.1/topics/cache/#basic-usage

I believe both the exporter and caseworker apps are setup to use redis as a cache.  If it feels like it's blowing out the PR scope though, maybe it would make sense to get this out with the mutable default first and have this as a follow-up ticket.

hnryjmes: I've reverted this function back to the original which uses the mutable default variable caching. I'm interested in rewriting it at some point though so made a ticket to do this: https://uktrade.atlassian.net/browse/LTD-5397 which maybe we can get to "at the end" if that comes

hnryjmes:
@currycoder have updated this PR and plan to look at rewriting the caching in a separate ticket, as detailed in the comment, so ready for re-review when you can thanks

kevincarrogan (approved):

currycoder (approved):

PR #2134, opened 2024-09-02
Dev-migration-DBT-Platform

depsiatwal:


conf/caseworker.py
```diff
@@ -130,3 +133,24 @@
 
 LITE_FEEDBACK_EMAIL = env.str("LITE_FEEDBACK_EMAIL", "")
 CONFIG_ADMIN_USERS_LIST = env.list("CONFIG_ADMIN_USERS_LIST", default=[])
+
+# Logging formatting
+
+LOGGING = {
```
depsiatwal, line 139: any reason why this is here when we have it in the base which is backward compatible. 
https://github.com/uktrade/lite-frontend/blob/dev-migration/conf/base.py#L218

conf/exporter.py
```diff
@@ -123,3 +126,24 @@
 ]
 
 SURVEY_URL = env.str("SURVEY_URL", "")
+
+# Logging formatting
+
+LOGGING = {
```
depsiatwal, line 132: any reason why this is here when we have it in the base which is backward compatible. 
https://github.com/uktrade/lite-frontend/blob/dev-migration/conf/base.py#L218

depsiatwal:
Testing on Dev2 and this looks good

depsiatwal:
Testing on Dev2 and this looks good

markj0hnst0n:
Would it be an idea to deploy this somewhere so we can see it's backwards compatible okay?

depsiatwal:
> Would it be an idea to deploy this somewhere so we can see it's backwards compatible okay?

I did test these branches on dev2 all was ok, probably need to retest with most recent changes.

conf/base.py
```diff
@@ -200,9 +202,9 @@
 # AWS
 
 AWS_S3_ENDPOINT_URL = env.str("AWS_S3_ENDPOINT_URL", None)
-AWS_ACCESS_KEY_ID = env.str("AWS_ACCESS_KEY_ID")
-AWS_SECRET_ACCESS_KEY = env.str("AWS_SECRET_ACCESS_KEY")
-AWS_REGION = env.str("AWS_REGION")
+AWS_ACCESS_KEY_ID = env.str("AWS_ACCESS_KEY_ID", None)
+AWS_SECRET_ACCESS_KEY = env.str("AWS_SECRET_ACCESS_KEY", None)
+AWS_REGION = env.str("AWS_REGION", "eu-west-2")
```
kevincarrogan, line 207: What's the reason this needs to change because of DBT Platform?

PR #2137, opened 2024-09-03
 LTD-5382-fix-hawk-dbt-platform

depsiatwal:
When migration to DBT Platform we had issue with CloudFront rejecting the request 
when we send some content when we doing a GET request. 

Then additional we required the hawk parameters to be empty else the hawkheader is missing hence verifyhawk function fails.


core/client.py
```diff
@@ -83,17 +83,22 @@ def verify_hawk_response(response, sender, stream=False):
 
 
 def perform_request(method, request, appended_address, data=None, stream=False):
-    data = data or {}
```
depsiatwal, line 86: required to removed this as an empty dict is being sent on GET requests on DBT platform

core/client.py
```diff
@@ -83,17 +83,22 @@ def verify_hawk_response(response, sender, stream=False):
...
     session = request.requests_session  # provided by RequestsSessionMiddleware
     url = _build_absolute_uri(appended_address.replace(" ", "%20"))
 
+    if not data:
+        # If we not sending any data then the hawk header should be defined with no content and no header
+        content_type = ""
```
depsiatwal, line 91: we require an empty content and content_type when we not sending any data as per hawk spec.

depsiatwal:
This has been smoke tested on dev2 and all appears to be working fine.

markj0hnst0n (approved):
Did a bit of v basic testing on dev 2.  Looks grand.

kevincarrogan (approved):

PR #2143, opened 2024-09-06
Ltd 5391 disable major edit button after click

kevincarrogan:
### Aim

Disable the major edit submission button after clicking to block multiple amendments from being generated.

[LTD-5391](https://uktrade.atlassian.net/browse/LTD-5391)

[LTD-5391]: https://uktrade.atlassian.net/browse/LTD-5391?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/unit_tests/exporter/applications/forms/test_amendments.py
+++ b/unit_tests/exporter/applications/forms/test_amendments.py
@@ -1,12 +1,4 @@
-import pytest
-import re
-
 from bs4 import BeautifulSoup
-
-from django.template import (
-    RequestContext,
-    Template,
-)
 
 from exporter.applications.forms.common import ApplicationMajorEditConfirmationForm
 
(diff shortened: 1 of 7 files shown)


unit_tests/exporter/applications/forms/test_amendments.py
```diff
@@ -9,3 +11,15 @@ def test_application_major_edit_confirm_form():
...
+def test_disabling_button_data_module(render_form):
+    form = ApplicationMajorEditConfirmationForm(
+        data={},
+        application_reference="your reference",
+        cancel_url="",
+    )
+    rendered = render_form(form)
+    soup = BeautifulSoup(rendered, "html.parser")
+    submit_button = soup.find("input", {"id": "submit-id-submit"})
+    assert submit_button.attrs["data-module"] == "disabling-button"
```
saruniitr, line 25: Is this the same as checking whether button is disabled?

kevincarrogan: This is checking that the right data module is set for the JS to get applied that disables the button. 

This can’t check that the button is disabled as it doesn’t actually run the JS.

saruniitr: ok understand. If JS is running then it will pick this up and set the attribute to `False`

saruniitr (approved):

PR #2147, opened 2024-09-07
LTD-5394: E2E tests for LU Countersigning restrictions

saruniitr:
## Change description

Re-enable skipped LU countersigning tests and add new ones for countersigning restrictions
The new restrictions ensure user cannot countersign their own cases.
 - Case officer who gave final recommendation should not be able to countersign as either a licensing manager or a senior licensing manager
 - Licensing manager should not be able to countersign as a Senior licensing manager

All of these steps (final recommendation, two-tier countersignatures) are to be done by unique users.

Dedicated users are created with specific roles as we need to logout and login during the test.

[LTD-5394](https://uktrade.atlassian.net/browse/LTD-5394)

[LTD-5394]: https://uktrade.atlassian.net/browse/LTD-5394?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ui_tests/caseworker/conftest.py
```diff
@@ -78,22 +78,42 @@ def sign_into_sso(driver, sso_sign_in):  # noqa
...
+
+
+@when("I sign in as Licensing Unit Manager")
+def mock_sso_lu_manager_user_caseworker_sign_in(driver, internal_url):  # noqa
+    mock_sso_signin_with_email(driver, internal_url, "lumanager@digital.trade.gov.uk")
+
+
+@when("I sign in as Licensing Unit Senior Manager")
+def mock_sso_lu_senior_manager_user_caseworker_sign_in(driver, internal_url):  # noqa
+    mock_sso_signin_with_email(driver, internal_url, "luseniormanager@digital.trade.gov.uk")
```
saruniitr, line 116: Couldn't think of any better way to improve this further.
We need some of these steps as `given` and `when` and specifying under the same definition is throwing error.
Tried using `step` to cover both but that is also giving error.

ui_tests/caseworker/features/give_advice/lu_countersign.feature
```diff
@@ -4,16 +4,14 @@ Feature: I want to record my user advice and any comments and conditions relatin
   I want to record my user advice and any comments and conditions relating to my recommendation
   So that other users can see my decision and know that I have finished assessing this case
 
-  @skip @lu_countersign
+  @lu_countersign
   Scenario: LU countersign
-    Given I sign in as Test UAT user
+    Given I sign in as Licensing Unit Officer
```
saruniitr, line 9: Switched all countersign related tests to use correct roles instead of test user.

ui_tests/caseworker/conftest.py
kevincarrogan: You can use multiple `@when` and `@given` decorators on the same function e.g.

```
@given("I sign in as Licensing Unit Officer")
def mock_sso_lu_officer_signin_given(driver, internal_url):
    mock_sso_signin_with_email(driver, internal_url, "lucaseofficer@digital.trade.gov.uk")


@when("I sign in as Licensing Unit Officer")
def mock_sso_lu_officer_signin(driver, internal_url):
    mock_sso_signin_with_email(driver, internal_url, "lucaseofficer@digital.trade.gov.uk")
```

Could be:

```
@given("I sign in as Licensing Unit Officer")
@when("I sign in as Licensing Unit Officer")
def mock_sso_lu_officer_signin(driver, internal_url):
    mock_sso_signin_with_email(driver, internal_url, "lucaseofficer@digital.trade.gov.uk")
```

ui_tests/caseworker/conftest.py
```diff
@@ -404,6 +424,20 @@ def system_queue_shown_in_dropdown(driver, queue_name):  # noqa
...
+def switch_to_queue(driver, queue_name):  # noqa
+    driver.find_element(by=By.ID, value="link-queue").click()
+    WebDriverWait(driver, 30).until(expected_conditions.presence_of_element_located((By.ID, "filter-queues")))
+    driver.find_element(by=By.ID, value="filter-queues").send_keys(queue_name)
+    elements = [
+        item
+        for item in driver.find_elements(by=By.CLASS_NAME, value="app-menu__item--subtitle")
+        if item.text.split("\n")[0] == queue_name
+    ]
+    assert len(elements) == 1
```
kevincarrogan, line 437: I think we've suggested in the past giving an assertion error message for things like this so we can understand the error better.

kevincarrogan (approved):

ui_tests/caseworker/conftest.py
saruniitr: yeah this is what I mean by specifying under the same definition but was throwing error, will check again.

saruniitr: This gives fixture not found error

Tllew (approved):
Looks good to me

PR #2150, opened 2024-09-12
LTD-5361-caseworker-add-user 

depsiatwal:
https://uktrade.atlassian.net/browse/LTD-5361

This adds a new ability for a caseworker to add an administor to an organisation. 

The Org must be active and the user has appropriate permissions. A new specific caseworker endpoint has been created for this view. 

https://github.com/uktrade/lite-api/pull/2172


caseworker/organisations/members/users/views.py
```diff
@@ -0,0 +1,67 @@
...
+        return self.success_url
+
+    @expect_status(
+        HTTPStatus.CREATED,
+        "Error adding user to organisation",
+        "Unexpected error adding user to organisation",
+    )
+    def post_add_user(self, data):
+        # This is currently limited to Administrator
+        data["role"] = ExporterRoles.administrator.id
```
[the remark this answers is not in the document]
saruniitr: If we want to keep in the FE then I suggest setting this in `form_valid()` and this function just deal with posting the data without any modifications.

caseworker/organisations/members/users/views.py
```diff
@@ -0,0 +1,67 @@
...
+        except HTTPError:
+            raise Http404()
+
+        if not rules.test_rule("can_manage_organisation", self.request, self.organisation):
+            raise Http404()
+        return super().dispatch(*args, **kwargs)
+
+    def get_form_kwargs(self):
+        form_kwargs = super().get_form_kwargs()
+        users = get_organisation_members(self.request, self.organisation_id).get("results")
```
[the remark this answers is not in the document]
depsiatwal: Very well spotted I refactored during tests and change the get method to the caseworker one and forgot to test it. 
fixed now

caseworker/core/rules.py
```diff
@@ -192,3 +204,4 @@ def user_not_yet_countersigned(request, case):
 rules.add_rule(
     "can_licence_status_be_changed", is_user_licencing_unit_senior_manager & is_case_finalised_and_licence_editable
 )
+rules.add_rule("can_manage_organisation", is_user_manage_organisations_role & is_organisation_active)
```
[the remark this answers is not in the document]
depsiatwal: yeah that's a better name updated

caseworker/core/rules.py
```diff
@@ -168,6 +169,17 @@ def user_not_yet_countersigned(request, case):
     return caseworker["id"] not in countersigners
 
 
+@rules.predicate
+def is_user_manage_organisations_role(request, organisation):
+    user = getattr(request, "lite_user", None)
+    return Permission.MANAGE_ORGANISATIONS.value in user["role"]["permissions"]
```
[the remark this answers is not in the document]
depsiatwal: I can see us reusing this form for exporter at some point so would rather leave it to frontend. 
i've shifted to form_valid now.

unit_tests/caseworker/organisations/users/test_user_views.py
```diff
@@ -0,0 +1,238 @@
...
+    requests_mock,
+    url,
+    organisation_pk,
+):
+    """Test to check add exporter admin errors when we the organisation isn't approved"""
+
+    requests_mock.get(
+        client._build_absolute_uri(f"/organisations/{organisation_pk}/"),
+        json={"status": {"key": "in-review"}},
+    )
```
saruniitr, line 232: Do we need a similar test for rejected organisations also? or parameterize this.

unit_tests/caseworker/organisations/users/test_user_views.py
```diff
@@ -0,0 +1,238 @@
...
+    url,
+    organisation_pk,
+):
+    """Test to check add exporter admin errors when we fail to get a organisation"""
+
+    requests_mock.get(
+        client._build_absolute_uri(f"/organisations/{organisation_pk}/"),
+        json={},
+        status_code=500,
+    )
```
saruniitr, line 200: this along with other similar code below can also be fixtures which makes it even cleaner.

caseworker/organisations/members/users/views.py
```diff
@@ -0,0 +1,67 @@
...
+        organisation_users = set(user["email"] for user in users)
+        form_kwargs["cancel_url"] = self.get_success_url()
+        form_kwargs["organisation_users"] = organisation_users
+        form_kwargs["sites"] = get_organisation_sites(self.request, self.organisation_id)
+        return form_kwargs
+
+    def form_valid(self, form):
+        data = form.cleaned_data
+        # This is currently limited to Administrator
+        data["role"] = ExporterRoles.administrator.id
```
saruniitr, line 48: sorry didn't think about this earlier.
In similar cases I think we usually create a function eg `get_payload()` which returns the payload data, please see if it can be refactored that way.

saruniitr (approved):
minor improvements to consider but it looks good to me.

unit_tests/caseworker/organisations/users/test_user_views.py
depsiatwal: > Do we need a similar test for rejected organisations also? or parameterize this.

Added as parameters but the mock can't be a fixture but i've changed the rest so it's cleaner

unit_tests/caseworker/organisations/users/test_user_views.py
depsiatwal: done

caseworker/organisations/members/users/views.py
depsiatwal: yeah I did consider payload however the payload data is designed for wizards and not compatible with a single view.

markj0hnst0n (approved):
Looks good

PR #2153, opened 2024-09-13
Rewrite CLEs caching to use Django cache

hnryjmes:
### Aim

This rewrites the `get_control_list_entries` service functions in both caseworker and exporter to not use a mutable default variable cache and to use the Django cache instead.

[LTD-5397](https://uktrade.atlassian.net/browse/LTD-5397)

[LTD-5397]: https://uktrade.atlassian.net/browse/LTD-5397?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/caseworker/core/services.py
+++ b/caseworker/core/services.py
@@ -136,41 +136,43 @@ def get_user_role_name(request):
 # Control List Entries
 def get_control_list_entries(request, convert_to_options=False, include_parent=False):
     if convert_to_options:
-        if cache.get("caseworker_converted_control_list_entries_cache"):
-            return cache.get("caseworker_converted_control_list_entries_cache")
-        response = client.get(request, "/static/control-list-entries/")
-        response.raise_for_status()
-        caseworker_converted_control_list_entries_cache = [
+        if cache.get("caseworker_control_list_entries_cache"):
+            caseworker_control_list_entries = cache.get("caseworker_control_list_entries_cache")
+        else:
+            response = client.get(request, "/static/control-list-entries/")
+            response.raise_for_status()
+            caseworker_control_list_entries = response.json().get("control_list_entries")
+            cache.set("caseworker_control_list_entries_cache", caseworker_control_list_entries)
+
+        converted_control_list_entries = [
             Option(
                 key=control_list_entry["rating"],
                 value=control_list_entry["rating"],
                 description=control_list_entry["text"],
             )
-            for control_list_entry in response.json().get("control_list_entries")
+            for control_list_entry in caseworker_control_list_entries
         ]
-        cache.set("caseworker_converted_control_list_entries_cache", caseworker_converted_control_list_entries_cache)
-        return caseworker_converted_control_list_entries_cache
+        return converted_control_list_entries
 
     if include_parent:
-        if cache.get("caseworker_control_list_entries_cache__include_parent"):
-            return cache.get("caseworker_control_list_entries_cache__include_parent")
+        if cache.get("caseworker_control_list_entries_including_parent_cache"):
+            return cache.get("caseworker_control_list_entries_including_parent_cache")
         response = client.get(request, "/static/control-list-entries/?include_parent=True")
         response.raise_for_status()
-        caseworker_control_list_entries_cache__include_parent = response.json().get("control_list_entries")
+        control_list_entries_including_parent = response.json().get("control_list_entries")
         cache.set(
-            "caseworker_control_list_entries_cache__include_parent",
-            caseworker_control_list_entries_cache__include_parent,
+            "caseworker_control_list_entries_including_parent_cache",
+            control_list_entries_including_parent,
         )
-        return caseworker_control_list_entries_cache__include_parent
+        return control_list_entries_including_parent
 
-    if cache.get("caseworker_control_list_entries_cache__group"):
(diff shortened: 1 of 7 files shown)


unit_tests/conftest.py
```diff
@@ -96,10 +95,6 @@ def data_control_list_entries():
 
 @pytest.fixture
 def mock_control_list_entries(requests_mock, data_control_list_entries):
-    # We must clear app-level CLE cache used by core.services.get_control_list_entrie
-    # so that tests remain isolated from eachother
-    # TODO: Remove this when we have a better way of caching CLEs
-    CLC_ENTRIES_CACHE.clear()
```
hnryjmes, line 102: this turned out to be important

caseworker/core/services.py
```diff
@@ -131,40 +133,46 @@ def get_user_role_name(request):
     return user["user"]["role"]["name"]
 
 
-CLC_ENTRIES_CACHE = []
-
-
 # Control List Entries
-def get_control_list_entries(  # noqa
```
currycoder, line 138: I've simplified this a fair bit over in; https://github.com/uktrade/lite-frontend/pull/2162

Turns out that the callers didn't really need there to be options returned, parents included, etc.  I also binned the caching (temporarily because I know you are doing this); but it should be fine as actually the only part of the app that was using the caching was the TAU assessment pages (previously caching was only used if `convert_to_options=True`).

I'm sort of wondering if things might go more simply if the caching work is applied on top of my changes here.

caseworker/core/services.py
```diff
@@ -131,40 +133,46 @@ def get_user_role_name(request):
...
-    request, convert_to_options=False, include_parent=False, clc_entries_cache=CLC_ENTRIES_CACHE  # noqa
-):  # noqa
-    """
-    Preliminary caching mechanism, requires service restart to repopulate control list entries
-    """
+def get_control_list_entries(request, convert_to_options=False, include_parent=False):
     if convert_to_options:
-        if clc_entries_cache:
-            return clc_entries_cache
+        if cache.get("caseworker_control_list_entries_cache"):
```
currycoder, line 139: Sprinkling `if cache_value: ...  else: ...; cache_it()` throughout feels like it could be improved upon.

Previously I've used libraries that provide decorators to make things neater e.g.

```
@cached(180)
def my_func(a, b="foo"):
    return ...
```

In this pattern the `@cached()` decorator does a lot of what you're doing in the guts here; try to look the value up in the cache, otherwise generate the value -> cache it (with keys based on funcname/args) -> return it.

I found that django-cacheops is well supported and provides exactly what I'm talking about; https://github.com/Suor/django-cacheops?tab=readme-ov-file#simple-time-invalidated-cache

Could be a neat pattern to establish. What do you think?

caseworker/core/services.py
```diff
@@ -131,40 +133,46 @@ def get_user_role_name(request):
...
+            caseworker_control_list_entries = cache.get("caseworker_control_list_entries_cache")
         else:
```
saruniitr, line 141: It is good to use standard components for this but just wondering is this change functionally required for this work or it is an improvement?

I think the way this function is organised need improvement. It has three distinct `if` blocks which are totally independent and managed by separate default arguments.
Is it better to split these as separate functions given that they have no dependencies?

exporter/core/services.py
```diff
@@ -198,24 +199,27 @@ def put_organisation_user(request, user_pk, json):
...
-                )
-            return converted_control_list_entries_cache
+        converted_control_list_entries = [
+            Option(
+                key=control_list_entry["rating"],
+                value=control_list_entry["rating"],
+                description=control_list_entry["text"],
+            )
+            for control_list_entry in exporter_control_list_entries
+        ]
```
currycoder, line 219: I kind of feel like this function should have never been able to return a list of `Option` values because that is kind of coupling our API request (which should live here in services.py) to our lite-forms mechanism.

What would you feel about refactoring so that there's another helper function elsewhere which does the options converting?  That way we can keep this function to purely grabbing the API result.

caseworker/core/services.py
saruniitr, line 139: Could these keys be made as constants? they are long and if there is a typo we will see it as a miss and fetch.

caseworker/core/services.py
```diff
@@ -131,40 +133,46 @@ def get_user_role_name(request):
...
+            response = client.get(request, "/static/control-list-entries/")
+            response.raise_for_status()
+            caseworker_control_list_entries = response.json().get("control_list_entries")
+            cache.set("caseworker_control_list_entries_cache", caseworker_control_list_entries)
+
+        converted_control_list_entries = [
+            Option(
+                key=control_list_entry["rating"],
+                value=control_list_entry["rating"],
+                description=control_list_entry["text"],
```
saruniitr, line 151: This conversion to option is only for use in forms, could this be moved to that function itself as it is not related to fetching CLEs?

hnryjmes (commented):
I will need to make substantial changes to this based on https://github.com/uktrade/lite-frontend/pull/2162 (all good changes) but as this will take some time, moving the Jira ticket back to in progress now

hnryjmes:
closing as I'm now working on a different branch

PR #2154, opened 2024-09-14
LTD-5362 Handle multiple instances of conflicting final advice

markj0hnst0n:
### Aim

Support ticket was NLR but goods were not pulling through to NLR letter.  It was discovered there were 2 types of final advice on the application as goods were assessed as needing a licence then NLR.  System was using licence issue/reissue flow even though goods/app was nlr.  However, goods were only associated with NLR advice so put in check to make sure system will only use advice (any type) which is associated with a good when getting advice items for the application.  This will stop any cases like this in the future using the incorrect flow.

[LTD-5362](https://uktrade.atlassian.net/browse/LTD-5362)

[LTD-5362]: https://uktrade.atlassian.net/browse/LTD-5362?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


caseworker/cases/views/advice.py
```diff
@@ -156,14 +156,19 @@ def get(self, request, *args, **kwargs):
             all_nlr = False
         else:
             advice = filter_advice_by_level(case["advice"], "final")
-            items = [item["type"]["key"] for item in advice]
-            approve = any([item == "approve" or item == "proviso" for item in items])
-            all_nlr = not approve and "refuse" not in items
+
+            # Only advice with an associated good can be part of advice_items
```
kevincarrogan, line 160: I feel it would be good to have a comment that explains why we want to do this as opposed to just saying what it does.

unit_tests/caseworker/cases/test_views.py
```diff
@@ -476,3 +496,63 @@ def test_im_done_page_submit(
...
+    data_standard_case["case"]["advice"] = advice
+    url = reverse("cases:finalise", kwargs={"queue_pk": queue_pk, "pk": data_standard_case["case"]["id"]})
+
+    response = authorized_client.get(url)
+
+    assert response.status_code == 200
+
+    soup = BeautifulSoup(response.content, "html.parser")
+    approve_heading = soup.find("h1", class_="govuk-fieldset__heading")
+    assert approve_heading.text.strip() == "Approve"
```
kevincarrogan, line 558: This test doesn't really explain to me what's happening here.

I can see that we end up discarding something but I'm not really sure why that advice payload ends up doing that.

markj0hnst0n: It goes to the approve page to approve the licence part even if there's nlr goods.  I don't think it's discarded but it addresses the approve first.  It's just proving that it issues the licence even if there's nlr goods on the application.

caseworker/cases/views/advice.py
```diff
@@ -156,14 +156,22 @@ def get(self, request, *args, **kwargs):
             all_nlr = False
         else:
             advice = filter_advice_by_level(case["advice"], "final")
-            items = [item["type"]["key"] for item in advice]
-            approve = any([item == "approve" or item == "proviso" for item in items])
-            all_nlr = not approve and "refuse" not in items
+
+            # We get all advice on the application and then filter out the advice with
+            # no good associated with it as we can assume that the good has been removed
+            # from the application.
```
saruniitr, line 162: This still doesn't covey the message correctly.
Basically for NLR line items we also record `approve` decision but their `good_id` is set to `None`. So when choosing decisions we have to ignore these.

unit_tests/caseworker/cases/test_views.py
saruniitr, line 558: In this case we should also get an option to generate NLR letter correct? or is it after we submit on this view?

markj0hnst0n: I believe that is later in the process when the documents is produced.  The next step after finalise is approve licence where there are any goods that require a licence.  I've updated these tests so hopefully they make a bit more sense now.

caseworker/cases/views/advice.py
markj0hnst0n: I've hopefully changed it to reflect this a bit better

unit_tests/caseworker/cases/test_views.py
markj0hnst0n: I've made some further amendments trying to make it clearer

caseworker/cases/views/advice.py
markj0hnst0n: Comments has been amended

unit_tests/caseworker/cases/test_views.py
```diff
@@ -476,3 +496,109 @@ def test_im_done_page_submit(
...
+        {
+            "level": "final",
+            "note": "note",
+            "type": {"key": "no_licence_required"},
+            "good": {"id": "good_id"},
+            "text": "text",
+        }
+    ]
+    data_standard_case["case"]["advice"] = advice
+    url = reverse("cases:finalise", kwargs={"queue_pk": queue_pk, "pk": data_standard_case["case"]["id"]})
```
saruniitr, line 514: this can actually be moved to a fixture which returns a case with this specific advice.
eg https://github.com/uktrade/lite-frontend/blob/dev/unit_tests/caseworker/advice/views/test_countersign.py#L186

saruniitr (approved):
some more improvements to consider.

unit_tests/caseworker/cases/test_views.py
markj0hnst0n: Thanks. I've created fixtures for these in `conftest.py`

caseworker/cases/views/advice.py
```diff
@@ -179,7 +187,7 @@ def get(self, request, *args, **kwargs):
                 extra_data={
                     "any_nlr": any_nlr,
                     "case": case,
-                    "has_proviso": any([item == "proviso" for item in items]),
+                    "has_proviso": any([item == "proviso" for item in advice_items]),
```
Tllew, line 190: we're looping over advice_items a lot because of the list comps, is there any way we could loop over it once and build the variables instead?

caseworker/cases/views/advice.py
```diff
@@ -156,14 +156,22 @@ def get(self, request, *args, **kwargs):
...
             advice = filter_advice_by_level(case["advice"], "final")
-            items = [item["type"]["key"] for item in advice]
-            approve = any([item == "approve" or item == "proviso" for item in items])
-            all_nlr = not approve and "refuse" not in items
+
+            # For no licence required advice items we have recorded their decision as ‘approve’
+            # but their ‘good_id’ has been set to ‘None’ so it is best to filter out
+            # these advice items.
+
+            advice_items = []
```
Tllew, line 164:
```
approve = False
all_nlr = False
has_proviso = False
any_nlr = False

for advice_item in advice:
    ....
    if advice_item["good"]:
        item = advice_item["type"]["key"]
        advice_items.append(item)
    if item in ["approve", "item"]:
        approve = True
...

```

would something like this be more readable?

Tllew (commented):
looks good to me, comment just on some readability but happy to be pushed back

caseworker/cases/views/advice.py
```diff
@@ -156,14 +156,22 @@ def get(self, request, *args, **kwargs):
             all_nlr = False
         else:
             advice = filter_advice_by_level(case["advice"], "final")
-            items = [item["type"]["key"] for item in advice]
```
saruniitr, line 159: Actually changing this line as below is sufficient,
`items = [item["type"]["key"] for item in advice if item['good']]`

caseworker/cases/views/advice.py
markj0hnst0n: The fact that the first list comprehension filters the the useful advice from the not useful advice before the others use that to make their decisions seems to be more logical to me.

caseworker/cases/views/advice.py
```diff
@@ -155,15 +155,20 @@ def get(self, request, *args, **kwargs):
...
+
+            advice_items = [item["type"]["key"] for item in final_advice if item["good"]]
+            approve = any([item == "approve" or item == "proviso" for item in advice_items])
+            all_nlr = not approve and "refuse" not in advice_items
 
         case_id = case["id"]
 
         if approve:
-            any_nlr = any([item == "no_licence_required" for item in items])
+            any_nlr = any([item == "no_licence_required" for item in advice_items])
```
Tllew, line 171: can this be undefined if the case is approved and of type open?

Tllew (approved):

PR #2160, opened 2024-09-17
LTD-5430 added validation to party name

Tllew:
### Aim

[LTD-](https://uktrade.atlassian.net/browse/LTD-5430)


exporter/core/validators.py
```diff
@@ -65,6 +65,10 @@ class PartyAddressValidator(EdifactStringValidator):
     message = "Address must only include letters, numbers, and common special characters such as hyphens, brackets and apostrophes"
 
 
+class PartyNameValidator(EdifactStringValidator):
+    message = "Party name must only include letters, numbers, and common special characters such as hyphens, brackets and apostrophes"
```
depsiatwal, line 69: The message says no special chars but there's no reg expression for validation of this.

Tllew: inherited from parent

depsiatwal: ahh ok I didn't notice that. 
In which case I think we should add a few more tests for special chars even for this field.

Tllew: I added an _ one to show it's active, we discussed previously that if the parent was tested we probably don't want to bombard all of our tests with the same test cases

saruniitr (approved):

PR #2162, opened 2024-09-17
[LTD-5399] Ensure frontend is acting upon correct CLE listings

currycoder:
### Aim

The top level aim of this PR is to ensure that CLEs which should not be available during product assessment are hidden.

This change ensures that calls to the exporter/caseworker CLE endpoints appropriately include/exclude CLEs which are not `selectable_for_assessment`.

The only place where we really want to exclude such non-selectable CLEs is in the TAU product assessment views - so otherwise, the `include_non_selectable_for_assessment` is specified.

To make this change easier, the open general licences app and various views to add/update flags/flagging rules were removed.  The associated functionality is legacy so this is a good opportunity to get rid of it.

These changes depend on changes in lite-api in the companion PR; https://github.com/uktrade/lite-api/pull/2190

[LTD-5399](https://uktrade.atlassian.net/browse/LTD-5399)

[LTD-5399]: https://uktrade.atlassian.net/browse/LTD-5399?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


caseworker/tau/views.py
```diff
@@ -116,8 +116,8 @@ def goods(self):
 
     @cached_property
     def control_list_entries(self):
-        control_list_entries = get_control_list_entries(self.request, convert_to_options=True)
-        return [(item.value, item.key) for item in control_list_entries]
```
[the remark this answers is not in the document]
currycoder: Ha yep.  I assume that previously we were using options but over time we have departed but not really cleaned it up

caseworker/core/services.py
```diff
@@ -131,40 +131,14 @@ def get_user_role_name(request):
...
-    if include_parent:
-        response = client.get(request, "/static/control-list-entries/?include_parent=True")
-    else:
-        response = client.get(request, "/static/control-list-entries/?group=True")
+def get_control_list_entries(request, include_non_selectable_for_assessment=False):
+    url = "/caseworker/static/control-list-entries/"
+    if include_non_selectable_for_assessment:
+        url = f"{url}?include_non_selectable_for_assessment=True"
 
+    response = client.get(request, url)
```
hnryjmes, line 139: I guess we will have to try and update and merge the caching PR fairly soon after this PR so that we don't have a temporary period where there is no caching at all of CLEs

.circleci/config.yml
```diff
@@ -133,7 +133,7 @@ commands:
           command: |
             git clone git@github.com:uktrade/lite-api.git
             cd lite-api
-            git checkout $(python ../which_branch.py <<parameters.api_branch>>)
+            git checkout LTD-5399-CLE-api-calls-visibility
```
hnryjmes, line 136: making a note to myself that we need to change this back before merging

caseworker/tau/views.py
```diff
@@ -116,8 +116,8 @@ def goods(self):
...
+        control_list_entries = get_control_list_entries(self.request, include_non_selectable_for_assessment=False)
```
hnryjmes, line 119: I think technically this `include_non_selectable_for_assessment=False` is not needed because it is the default value, but probably it's nicer to be really explicit about this as it's quite an important detail

caseworker/tau/views.py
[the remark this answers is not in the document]
saruniitr: I think the options was more for `lite-forms`

saruniitr (approved):
this looks good to me, cleaning up a lot of unused code.

Tllew (approved):
Looks good to me!

hnryjmes (approved):
nice work

caseworker/core/services.py
currycoder: I've convinced myself that this should be OK now - the only place where we were really using the cached values was for TAU assessments which I think is a fraction of our overall traffic.  Previously I was concerned that we were using the cached values for e.g. something like the queue view page - but having dug in here we are not.
So my thinking is; this isn't going to be an awful performance degradation if we end up releasing this ahead of fixing up caching.

PR #2171, opened 2024-09-20
Exporter mock sso logout

kevincarrogan:
### Aim

Allow logout to work on exporter SSO by clearing session data.

### DIFF
--- a/core/mock_sso/views.py
+++ b/core/mock_sso/views.py
@@ -147,7 +147,10 @@ class APIUserMe(View):
 
 
 class Logout(RedirectView):
-    def get_redirect_url(self, *args, **kwargs):
+    def get_redirect_url(self):
         redirect_url = self.request.build_absolute_uri("/")
-        self.request.session.flush()
         return redirect_url
+
+    def get(self, request, *args, **kwargs):
+        self.request.session.flush()
+        return super().get(request, *args, **kwargs)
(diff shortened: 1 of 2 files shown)


core/mock_sso/views.py
```diff
@@ -147,7 +147,10 @@ def get(self, request, **kwargs):
 
 
 class Logout(RedirectView):
-    def get_redirect_url(self, *args, **kwargs):
+    def get_redirect_url(self):
```
depsiatwal, line 150: where we need fushing the session ?
isn't this what the real view does ?

core/mock_sso/views.py
```diff
@@ -147,7 +147,10 @@ def get(self, request, **kwargs):
...
         redirect_url = self.request.build_absolute_uri("/")
-        self.request.session.flush()
         return redirect_url
+
+    def get(self, request, *args, **kwargs):
+        self.request.session.flush()
```
depsiatwal, line 155: ahh I see it does a redirect to here ?

depsiatwal (approved):
nice

core/mock_sso/views.py
kevincarrogan: In my mind this method should only do what it's supposed to do without side-effects. In this case all it needs to do is return the URL that the `RedirectView` would redirect to.

core/mock_sso/views.py
kevincarrogan: Nothing "redirects" here as such.

This is just how views work in that they get routed to the method that corresponds to the HTTP method being used to retrieve the page.

In this case I'm flushing the session here and then letting the parent (which does redirects) do its own thing.

PR #2173, opened 2024-09-23
Rewrite CLEs caching to use django-cacheops cache

hnryjmes:
### Aim

This is a rewrite of https://github.com/uktrade/lite-frontend/pull/2153 based on the frontend changes made in https://github.com/uktrade/lite-frontend/pull/2162.

Instead of using the Django cache directly, the `django-cacheops` library is used instead, which provides useful ways to work with the cache in Django such as caching the result of a function just using a function decorator.

[LTD-5397](https://uktrade.atlassian.net/browse/LTD-5397)

[LTD-5397]: https://uktrade.atlassian.net/browse/LTD-5397?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


exporter/core/constants.py
```diff
@@ -15,6 +15,8 @@
 GIFTING = "gifting_clearance"
 F680 = "f680_clearance"
 
+CONTROL_LIST_ENTRIES_CACHE_TIMEOUT = 60 * 60 * 24  # 24 hours
```
Tllew, line 18: since this and the caseworker one are the case would it be sensible to move this to api/core/constants.py?

Tllew (approved):
minor comment but looks good to me

currycoder (commented):
This looks neat - nice work.  I would suggest popping this on an environment though and testing that things work as expected prior to merge.

hnryjmes:
@currycoder ok so I've been testing this out on `dev2`, I'm happy that the API calls for CLEs static data are being cached in both exporter app and caseworker app. it's a bit tricky to look at the cache directly but you can monitor what's happening with the API by watching for various calls and there aren't any to the CLEs endpoints as expected when I do the relevant actions. all CLEs functionality in the frontend is also working fine from what I can tell, I made a new application and went through to finalisation

this is giving me confidence that possibly we can look at extending the caching to all of the other static data used in the frontend, as prominently in the TAU assess screen you can see it go for other static data such as regimes. report summaries get requested when the dropdown is activated, that is a list of things that could also be cached. maybe one to think about if we do a dev firebreak.

conf/caseworker.py
```diff
@@ -114,6 +114,8 @@
     }
 }
 
+CACHEOPS_REDIS = REDIS_URL
```
saruniitr, line 117: Is this used by the framework, not used anywhere.

saruniitr (approved):
all looks good to me.

conf/caseworker.py
hnryjmes: yes, it fails without this line
https://github.com/Suor/django-cacheops/tree/master?tab=readme-ov-file#setup

PR #2177, opened 2024-09-24
LTD-5452-feedback-content-change

depsiatwal:
https://uktrade.atlassian.net/browse/LTD-5452

New feedback form where the exported get a link to the support-help desk link. 
Had to refactor and create an exporter specific feedback form. 

Also took this opportunity to use our standard base crisp_form.


exporter/core/feedback/forms.py
```diff
@@ -0,0 +1,16 @@
+from crispy_forms_gds.layout import HTML
+
+from core.feedback.forms import FeedbackForm
+
+
+class ExporterFeedbackForm(FeedbackForm):
```
saruniitr, line 6: `FeedbackForm` alone used separately? just wondering why it is split like this.

unit_tests/exporter/feedback/test_views.py
```diff
@@ -14,6 +15,16 @@ def test_feedback_get_success(url, authorized_client):
...
 
 
+def test_feedback_contains_support_link(authorized_client):
+    url = reverse("feedback")
+    response = authorized_client.get(url)
+    assert response.status_code == 200
+    soup = BeautifulSoup(response.content, "html.parser")
+    links = soup.find_all(class_="govuk-link")
+    support_link = [l["href"] for l in links if l["href"] == "/help-support/"]
+    assert support_link
```
saruniitr, line 25: Could you also add a test when this form is submitted?

hnryjmes (commented):
when trying to submit the form with some text in, even as a logged in user, I get this error: 
```
403 - [{'error': 'AuthError', 'message': 'Invalid token: service id is not the right data type'}]
```

possibly this is me doing something wrong, have you seen this error before?

core/feedback/forms.py
```diff
@@ -1,5 +1,19 @@
 from django import forms
+from crispy_forms_gds.layout import HTML
 
+from core.common.forms import BaseForm
 
-class FeedbackForm(forms.Form):
-    feedback = forms.CharField(label="Please type your feedback here: ", widget=forms.Textarea)
+
+class FeedbackForm(BaseForm):
```
hnryjmes, line 7: good use of BaseForm

core/feedback/forms.py
```diff
@@ -1,5 +1,19 @@
...
+    class Layout:
+        TITLE = "Leave Feedback"
```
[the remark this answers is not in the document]
depsiatwal: done

exporter/core/feedback/forms.py
depsiatwal: prior to this FeedbackForm was used by exporter/caseworker 

No we have a requirement to change some the content on exporter feedback hence why I created a specialised version for it rather then creating completely separate views/forms with lots of overlapping functionality

unit_tests/exporter/feedback/test_views.py
depsiatwal: This is the text below 
test_feedback_post_success

depsiatwal:
> when trying to submit the form with some text in, even as a logged in user, I get this error:
> 
> ```
> 403 - [{'error': 'AuthError', 'message': 'Invalid token: service id is not the right data type'}]
> ```
> 
> possibly this is me doing something wrong, have you seen this error before?

Could be quite possible you don't have email/ notify api not setup locally

markj0hnst0n (approved):
Setup notify locally and all seems to be working well.  I get that the support link is the main change here so test makes sense.

PR #2180, opened 2024-09-26
Ltd 5464-key-error-registration

depsiatwal:
https://uktrade.atlassian.net/browse/LTD-5464

This is to rectify an error that manifests itself when registering a new user. 
If the user takes too long to fill in their name then when we attempt to get the profile from the SSO service it silently fails. 
This is because the token used to make this call is very short lived but the session/token is still active hence we don't redirect using the loginmixin. 

By redirecting the user to the login screen  we force a token refresh and the user will either be directed to login if they don't have an active token or it gets refreshed and they return back to the register-name screen.


core/auth/utils.py
```diff
@@ -16,5 +20,13 @@ def get_client(request, **kwargs):
...
-    response = client.get(settings.AUTHBROKER_PROFILE_URL)  # .json()
+    try:
+        response = client.get(settings.AUTHBROKER_PROFILE_URL)
+        response.raise_for_status()
+    except HTTPError as ex:
+        logger.info(
+            "Error with SSO getting profile: %s ",
+            ex,
+        )
+        raise ex
```
kevincarrogan, line 31: This can be slightly simplified by doing:

```
response = client.get(settings.AUTHBROKER_PROFILE_URL)
try:
    response.raise_for_status()
except HTTPError:
    logger.info("Error with SSO", exc_info=True)
    raise
```

exporter/core/views.py
```diff
@@ -194,11 +195,17 @@ def get_success_url(self):
     def dispatch(self, request, *args, **kwargs):
         if self.request.session.get("first_name") and self.request.session.get("last_name"):
             return HttpResponseRedirect(resolve_url(settings.LOGIN_URL))
+        try:
+            self.profile = get_profile(self.request.authbroker_client)
+        except HTTPError:
```
kevincarrogan, line 200: Instead of checking for the `HTTPError` here could we actually check to see if the token that the authbroker_client has is still valid?

I believe most tokens have their own timeout associated with them and that means we could differentiate between a timeout on the token and some other kind of error coming from the auth provider.

kevincarrogan:
> This is because the token used to make this call is very short lived but the session/token is still active hence we don't redirect using the loginmixin.

Having looked at the ticket you suggest that this is because the refresh token is timing out, however that's a long-lived token but this comment implies that the token that's failing is a short-lived one (which isn't the refresh token).

What is the thing that's actually timing out?

core/auth/utils.py
depsiatwal: Done

exporter/core/views.py
depsiatwal: At the moment I don't want to change the behaviour too much. however this should never error as the middleware is response for token refresh. For now I just want to log and handle the scenario where getting a profile didn't work and leave the responsibility of refresh tokens with the middle ware.

depsiatwal:
This PR is about adding some extra logging 
- Handle and log when we fail to get profile which is an exception and shouldn't happen
- Log when we have a profile and they attempt to authenticate . the 401 is part of expected flow but for prod investigating it's useful to know so we can determine the users journey.

kevincarrogan (approved):

markj0hnst0n (approved):
Makes sense

PR #2181, opened 2024-09-27
LTD-5465 Amend exporter setting of consignee and add website length validation

markj0hnst0n:
### Aim

Amend set consignee to use similar flow to set end-user to enable front end validation of length of website address.

Both end-user and consignee website fields can cause errors in the api end if the address is too long as the max length of the website field is 200 chars in the model (default for URLfield).

More info on how this was uncovered in the ticket.

Adding this validator has been taken as an opportunity to change Consignee setting flow to use a form wizard and it also adds this website char length validation to end-user flow where these kind of errors have also been seen on prod.

[LTD-5465](https://uktrade.atlassian.net/browse/LTD-5465)

[LTD-5465]: https://uktrade.atlassian.net/browse/LTD-5465?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


exporter/applications/views/parties/consignees.py
```diff
@@ -32,26 +62,71 @@ def get(self, request, **kwargs):
...
+        context = super().get_context_data(**kwargs)
+        context["form_title"] = self.form_class.title
+        return context
 
+    def form_valid(self, form):
+        reuse_party = str_to_bool(form.cleaned_data.get("reuse_party"))
+        if reuse_party:
+            success_url = reverse("applications:consignees_copy", kwargs=self.kwargs)
+        else:
+            success_url = reverse("applications:set_consignee", kwargs=self.kwargs)
```
saruniitr, line 79: Can this be moved to get_success_url()?

saruniitr (commented):
few minor changes, looks good otherwise

exporter/applications/views/parties/consignees.py
```diff
@@ -32,26 +62,71 @@ def get(self, request, **kwargs):
...
+        kwargs = super().get_form_kwargs(step)
+
+        if step == SetPartyFormSteps.PARTY_ADDRESS:
+            kwargs["request"] = self.request
+
+        return kwargs
+
+    def get_context_data(self, form, **kwargs):
+        context = super().get_context_data(form, **kwargs)
+        return context
```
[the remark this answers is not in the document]
markj0hnst0n: removed

exporter/applications/views/parties/consignees.py
markj0hnst0n: Done.  Hopefully it's been done in a clean enough way.

saruniitr (approved):

exporter/applications/views/parties/end_users.py
```diff
@@ -146,8 +146,12 @@ def get_context_data(self, form, **kwargs):
         return context
 
     def get_form_kwargs(self, step=None):
-        kwargs = super().get_form_kwargs(step)
+        PartySubTypeSelectForm.title = "Select the type of end user"
+        PartyNameForm.Layout.TITLE = "End user name"
+        PartyWebsiteForm.Layout.TITLE = "End user website address (optional)"
+        PartyAddressForm.Layout.TITLE = "End user address"
```
kevincarrogan, line 152: I've seen us use this pattern before where we are manipulating the `TITLE` of the form and it does make me think that this is potentially breaking a bunch of stuff (because you're setting this at the class level).

It also feels a bit odd that this is in `get_form_kwargs` which I would see as something that doesn't cause side effects.

Why are we needing to dynamically change the title of these forms?

markj0hnst0n: We could use something generic for just updating 'party' but it felt better just to use the type of party and make it more relevant to the user.

When I just update the SetConsignee view then the SetEndUser view has the consignee personalisation in it.  I've been trying to work out why but I'm not sure what's happening so I set the form titles in both views to correct.

I thought it should just inherit the title from the form but maybe I've gone wrong somewhere

kevincarrogan: > When I just update the SetConsignee view then the SetEndUser view has the consignee personalisation in it.

I'm not really sure what you mean by this.

Do you have some code where you could describe what you mean here?

depsiatwal: Not 100% sure if it helps but we did some dynamic titles for registration using inheritance. 
https://github.com/uktrade/lite-frontend/blob/dev/exporter/core/organisation/forms.py

kevincarrogan: The only problem is that doesn't work properly either.

```
In [1]: from exporter.core.organisation.forms import RegisterDetailsIndividualUKForm, RegisterDetailsCommercialUKForm

In [2]: individual = RegisterDetailsIndividualUKForm()

In [3]: individual.Layout.TITLE
Out[3]: 'Enter organisation details'

In [4]: commercial = RegisterDetailsCommercialUKForm()

In [5]: commercial.Layout.TITLE
Out[5]: 'Register a commercial organisation'

In [6]: individual.Layout.TITLE
Out[6]: 'Register a commercial organisation'
```

kevincarrogan: I would highly suggest at no point should we manipulate `Layout.TITLE` anywhere in the form itself as it's going to cause a problem at some point.

markj0hnst0n: This is exactly the behaviour I was seeing.

markj0hnst0n: I've made changes and taken a different approach to setting the title dynamically by introducing the title as a parameter in the initialisation of the forms.

exporter/applications/forms/parties.py
```diff
@@ -221,6 +227,18 @@ class Layout:
         ],
     )
 
+    def __init__(self, title, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        if not title:
```
depsiatwal, line 232: why if not title ?

especially as there no default so user much deliberately setting it to None which is odd.

exporter/applications/forms/parties.py
```diff
@@ -230,7 +248,12 @@ class Layout:
         TITLE = "End user website address (optional)"
         TITLE_AS_LABEL_FOR = "website"
 
-    website = forms.CharField(required=False, label="", help_text="Use the format https://www.example.com")
+    website = forms.CharField(
+        required=False,
+        label="",
+        help_text="Use the format https://www.example.com",
+        validators=[MaxLengthValidator(200, f"Website address should be 200 characters or less")],
```
depsiatwal, line 255: just wondering why 200 ?

markj0hnst0n: It's the max length in the model field in the api.  Any more and the post request fails.

exporter/applications/forms/parties.py
markj0hnst0n: The idea was that if a title is not passed dynamically it will just fall back to using the `Layout.TITLE` which is the 'end user' wording

exporter/applications/forms/parties.py
depsiatwal: i'm not so sure we want to add validation for all max fields. 
I don't think it's likely a web address will be greater then 200

markj0hnst0n: It's actually happened a few times.  It was the root cause of the error message that precipitated this ticket.  I imagine people copy pasting long URLs.

exporter/applications/forms/parties.py
```diff
@@ -195,6 +198,9 @@ def __init__(self, *args, **kwargs):
             Submit("submit", "Continue"),
         )
 
+    def get_layout_fields(self):
+        return ("sub_type",)
+
```
kevincarrogan, line 203: Now I'm taking a second look at this form I'm noticing that it's doing a lot of things that `BaseForm` already does.

For example, this `get_layout_fields` is being overridden but not actually used because you're setting `self.helper.layout` in the `__init__`.

I think it's probably worth taking a look at what `BaseForm` does as I think there might be a bunch of redundant code going on here.

markj0hnst0n: The `BaseForm` forces the use of the `get_layout_fields` method so I have to keep it.  I could make a bespoke form but then I would have to import the `FormHelper` class so I think it's worth keeping for inheritance for that reason.

kevincarrogan: I don't think this method is actually really being used because it's purpose is to generate the `self.helper.layout` but you're completely resetting that in the `__init__`.

I think that you could just put the `ConditionalRadios` but in here instead and get rid of the entire `__init__`.

exporter/applications/views/parties/consignees.py
```diff
@@ -32,26 +61,64 @@ def get(self, request, **kwargs):
...
+        return SetConsigneePayloadBuilder().build(form_dict)
+
+    @expect_status(
+        HTTPStatus.CREATED,
+        "Error adding consignee to application",
+        "Unexpected error adding consignee to application",
+    )
+    def post_party_with_payload(self, pk, form_dict):
+        payload = self.get_payload(form_dict)
+        payload.update({"type": self.party_type})
```
kevincarrogan, line 111: Is `party_type` anything but `"consignee"`?

markj0hnst0n: In this case no but since it was using the more generic `post_party` to send it to the api and the payload being built uses `party_type` I kept the naming generic.  

`party_type` could also be `end_user` for that `post_party` service

exporter/applications/forms/parties.py
markj0hnst0n: Removed that and refactored a bit more

PR #2192, opened 2024-10-08
LTD-5463-reg-uk-based-content-changes 

depsiatwal:
Changes to content for registration
https://uktrade.atlassian.net/browse/LTD-5463

Also includes these 
https://uktrade.atlassian.net/browse/LTD-5502


core/common/forms.py
```diff
@@ -46,6 +46,9 @@ def __init__(self, *args, **kwargs):
             ),
         )
 
+    def get_title(self):
+        return self.Layout.TITLE
+
     def get_layout_fields(self):
```
markj0hnst0n, line 52: Is this change doing anything at the minute?

depsiatwal: yeah self.get_title() above it 
the idea being if we set it in initial it refreshes it's self.

markj0hnst0n: It's only returning `self.Layout.TITLE` at the moment though so I don't understand why it would be needed because it's not being overloaded with anything at the minute.  Possible it could have impact on anything that uses `BaseForm`.

depsiatwal: https://github.com/uktrade/lite-frontend/blob/bd8f297a0a23dd4d39722ccbe3178930a874d1a4/core/common/forms.py#L27

depsiatwal: so every time __init__ is called the title gets refreshed

kevincarrogan: When I suggested this it wasn't actually what I intended this to be used for.

The point I was trying to make was that if there is a form that wants to dynamically change the title instead of manipulating `self.Layout.TITLE` directly it could override `get_title` and do whatever it wants instead.

I've said it before and I'll say it again, we should not alter `self.Layout.TITLE` at runtime.

exporter/core/organisation/forms.py
```diff
@@ -239,14 +241,40 @@ def clean_website(self):
...
         super().__init__(*args, **kwargs)
 
+    def get_title(self):
+        return self.title
+
+    def set_address_help_text(self):
+        self.p1_address_help = HTML.details(**self.address_help_text)
+
+    def set_name_help_text(self):
+        self.base_fields["name"].help_text = self.name_help_text
```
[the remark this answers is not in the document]
depsiatwal: it just mean's one more layer of inheritance then selecting which form to use in the view. 

was trying to avoid 4 layers 

Baseform > RegisterAddressDetailsBaseForm > RegisterAddressDetailsUKForm > RegisterAddressDetailsUKIndividualForm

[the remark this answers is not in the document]
kevincarrogan: I agree that multiple levels of inheritance could be considered as a bad code smell, but it's not necessarily always something we shouldn't do.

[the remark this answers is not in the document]
depsiatwal: updated to use inheritance reduces some of the complexity I guess

kevincarrogan (approved):

markj0hnst0n:
Am I right in saying this pull request also incorporates content changes necessary for [https://uktrade.atlassian.net/browse/LTD-5502](https://uktrade.atlassian.net/browse/LTD-5502) ?  I was a bit confused because some of the changes weren't mentioned on the ticket in the PR description.

depsiatwal:
> Am I right in saying this pull request also incorporates content changes necessary for https://uktrade.atlassian.net/browse/LTD-5502 ? I was a bit confused because some of the changes weren't mentioned on the ticket in the PR description.

Yeah some how they got rolled in so I left them lol 
i've updated the description.

exporter/core/organisation/forms.py
```diff
@@ -97,11 +97,12 @@ class Layout:
     eori_number = forms.CharField(
         label=EORI_LABEL,
         help_text=(
-            "<a href='https://www.gov.uk/eori' class='govuk-link govuk-link--no-visited-state'"
-            "target='_blank'>Get an EORI number </a> if you don't have one."
+            """The first two letters are the country code, like GB or XI. This is followed by 12 or 15 numbers, like GB123456123456.
+            <a href='https://www.gov.uk/eori' class='govuk-link govuk-link--no-visited-state'"
```
kevincarrogan, line 101: I'm sort of surprised that this isn't getting escaped somewhere.

markj0hnst0n (approved):

exporter/core/organisation/forms.py
depsiatwal: ahh well spotted the under_blank target was being escaped so wasn't working test and fixed now.

PR #2197, opened 2024-10-10
[LTD-5504] CLE cache invalidation

currycoder:
### Aim

This change ensures that our CLE cache will refresh on frontend deployments by using `GIT_COMMIT` as a key to vary the cache on.  

This assumes that `GIT_COMMIT` will be available.  This appears to be the case for gov-PAAS environments.  For local environments this will not be the case (although we could potentially pass this through in docker-compose env vars).  We will also need to think about maintaining this for DBT platform environments when we make the switch.

In the worst case, where this env var is not available, we will have a lag of up to 24H for CLEs to refresh on the frontend.

Additionally, this PR ensures that the `python manage.py invalidate` command provided by django-cacheops can run and that things cached by cacheops are cached under redis DB 2.  This means if we needed to manually invalidate the cache on an environment, we would *only* invalidate cacheops records and not *all* redis records.

[LTD-5504](https://uktrade.atlassian.net/browse/LTD-5504)

[LTD-5504]: https://uktrade.atlassian.net/browse/LTD-5504?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Tllew (approved):
Looks good to me

kevincarrogan (approved):

currycoder:
Just want to triple check this on a real env before merging - so converting to draft in the meantime

conf/caseworker.py
```diff
@@ -114,7 +114,8 @@
     }
 }
 
-CACHEOPS_REDIS = REDIS_URL
+# Ensure cache on separate redis DB
+CACHEOPS_REDIS = f"{REDIS_URL}/2"
```
hnryjmes, line 118: glad to see this be set to something different. when I chose the value before I was wondering if it was wise to have it all connected to the same redis db, as the documentation shows you can vary this, so if this works it's better I think

PR #2201, opened 2024-10-14
LTD 5511 EORI, UK VAT number and SIC code validation changes

markj0hnst0n:
### Aim

Changes to validation and associated error messages for number fields during exporter registration

[LTD-5511](https://uktrade.atlassian.net/browse/LTD-5511)

[LTD-5511]: https://uktrade.atlassian.net/browse/LTD-5511?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/exporter/core/organisation/validators.py
+++ b/exporter/core/organisation/validators.py
@@ -54,22 +54,23 @@ def validate_phone(value):
     try:
         phone_number = phonenumbers.parse(value, "GB")
         if not phonenumbers.is_valid_number(phone_number):
-            errors.extend("Invalid telephone number")
+            raise ValidationError(errors)
     except phonenumbers.phonenumberutil.NumberParseException:
-        errors.extend("Invalid telephone number")
+        raise ValidationError(errors)
 
     if errors:
         raise ValidationError(errors)
 
 
 def validate_sic_number(value):
-    errors = []
 
+    validate_sic_number_functions = {
+        "SIC code can only include numbers": lambda v: not v.isdigit(),
+        "Enter a SIC code that is 5 numbers long, like 12345": lambda v: len(v) != Validation.SIC_LENGTH,
+    }
+    errors = []
     if value:
-        if not value.isdigit():
-            errors.append("SIC code can only include numbers")
-        if len(value) != Validation.SIC_LENGTH:
-            errors.append("Enter a SIC code that is 5 numbers long, like 12345")
+        errors.extend(error_message for error_message, func in validate_sic_number_functions.items() if func(value))
         if errors:
             raise ValidationError(errors)
 
--- a/exporter/core/organisation/validators.py
+++ b/exporter/core/organisation/validators.py
@@ -1,4 +1,5 @@
 import phonenumbers
+import re
 
 from django.core.exceptions import ValidationError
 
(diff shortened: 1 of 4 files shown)


exporter/core/organisation/validators.py
```diff
@@ -7,24 +7,39 @@
...
 
 def validate_eori(value):
+    errors = []
     if value:
-        eori = re.sub(r"[^A-Z0-9]", "", value)
-        if len(eori) > Validation.UK_EORI_MAX_LENGTH:
-            raise ValidationError("EORI numbers are 17 characters or less")
-        elif not re.match(Validation.UK_EORI_VALIDATION_REGEX, eori):
-            raise ValidationError("Invalid UK EORI number")
+        if not re.match(Validation.LETTERS_AND_NUMBERS_ONLY, value):
```
depsiatwal, line 29: This works but a lot of if statements now 
I thinking is it better to refactor to avoid this. 

may be a list of functions in a dict that we can

saruniitr: Yeah like a list of validators which we loop through and validate.
Is it possible that there can be multiple errors and we display them on the UI or one is displayed?

saruniitr: Also I think EORI number is GB followed by 12 or 15 digits so totally it is either 14 or 17 characters.

markj0hnst0n: Multiple errors can be displayed as I'm just using an error list to collect and display them on the user end.  I think the EORI validation is consistent with your comment @saruniitr also.  Min value is 14 chars and max is 17.

I've made changes so that the validators are a list of true/false functions for looping through.

exporter/core/organisation/validators.py
```diff
@@ -1,51 +1,78 @@
...
-        return stripped_vat
+        errors.extend(error_message for error_message, func in validate_vat_number_functions.items() if func(value))
+        if errors:
+            raise ValidationError(errors)
 
 
 def validate_eori(value):
+
+    validate_eori_number_functions = {
+        "EORI number can only include numbers and letters": lambda v: not re.match(
```
depsiatwal, line 32: nice 
however looks a bit messy perhaps shifting to use constants for the message and the RE expression will make it cleaner. 

Also if we don't require a dict perhaps tuple might be more relevant here or even a list of named tuples.

markj0hnst0n: Switched to using constants but I've kept the dict as originally asked for.  Feels more maintainable to me that way and not sure changing the data type would add anything.

exporter/core/organisation/validators.py
```diff
@@ -1,51 +1,79 @@
...
+        Validation.UK_EORI_MIN_LENGTH_ERROR_MESSAGE: lambda v: len(re.sub(Validation.STRIPPED_VALUE, "", v))
+        < Validation.UK_EORI_MIN_LENGTH,
+        Validation.UK_EORI_STARTING_LETTERS_ERROR_MESSAGE: lambda v: not (
+            re.sub(Validation.STRIPPED_VALUE, "", v).startswith("GB")
+            or re.sub(Validation.STRIPPED_VALUE, "", v).startswith("XI")
+        ),
+        Validation.UK_EORI_VALIDATION_ERROR_MESSAGE: lambda v: not re.match(
+            Validation.UK_EORI_VALIDATION_REGEX, re.sub(Validation.STRIPPED_VALUE, "", v)
+        ),
+    }
```
hnryjmes, line 48: I think this looks better now the sting literals aren't being used as keys

depsiatwal: I still think it's wrong as it's not really using the dict with keys but an error message as the key. 

May be have a key name that makes sense or shift to tuples

markj0hnst0n: Wrong is a strong word but I agree with making the changes as it reads better.  It's been amended and it looks much nicer now having implemented the list of tuples.

depsiatwal: lol sorry I mean't feel wrong to me

exporter/core/organisation/validators.py
```diff
@@ -1,51 +1,101 @@
...
 
 from .constants import Validation
 
 
 def validate_vat(value):
+
+    validate_vat_function_list = [
+        (
+            Validation.UK_VAT_LETTERS_AND_NUMBERS_ERROR_MESSAGE,
+            lambda v: not re.match(Validation.LETTERS_AND_NUMBERS_ONLY, v),
```
depsiatwal, line 14: looking good just one more thought would be more readable if the function return TRUE if it passed 
and the errors is appended if FALSE ?

saruniitr: looks good now.
Just another thought, not suggesting that you update it like that.
Now that we have a list of validators, does it work if we just feed that list in the form itself? (whereever `validate_vat` is called). These are all independent validators just grouped in the same function.

markj0hnst0n: The list of functions has no way of raising the ValidationError.  It's just a list of Boolean functions so it doesn't work in the form.  It doesn't make much difference whether they return true of false in the end.  I don't think there's a right or wrong way to do it in that sense.

depsiatwal (approved):
Happy with these changes

exporter/core/organisation/validators.py
kevincarrogan: I think we should be leaning on Django form validation itself more which I would say is _the_ right way to do it.

```
from django import forms


class MyForm(forms.Form):
    even_field = forms.IntegerField(validators=[validate_even])
```

You can pass in a list of validator functions to the field itself we don't need to make our own dict or list for this, we don't need to re-invent the wheel.

A lot of these look like checks for length or regular expression checks which are all already available validators from Django e.g. `MaxLengthValidator` and `RegexValidator`.

If the desire is to consolidate all of these validation rules into one place that could be used across LITE then we should be thinking of creating a Field e.g. VATField, that provides all of the validation, but again it should just lean on what Django gives us.

markj0hnst0n: Made changes which add this kind of thing now.

kevincarrogan (approved):

PR #2207, opened 2024-10-22
 LTD-5461-rc-changes 

depsiatwal:
https://uktrade.atlassian.net/browse/LTD-5461

This PR resolves RC number by adding 
validation rules during registration and by changing the labels.

### DIFF
--- a/caseworker/templates/organisations/organisation/details.html
+++ b/caseworker/templates/organisations/organisation/details.html
@@ -53,7 +53,7 @@
 		{% if organisation.registration_number %}
 			<div class="govuk-summary-list__row">
 				<dt class="govuk-summary-list__key">
-					{% lcs 'organisations.OrganisationPage.Details.SummaryList.COMPANY_REG_NUMBER' %}
+					Registered number
 				</dt>
 				<dd class="govuk-summary-list__value">
 					{{ organisation.registration_number }}
--- a/unit_tests/exporter/core/organisation/test_organisation_views.py
+++ b/unit_tests/exporter/core/organisation/test_organisation_views.py
@@ -135,7 +135,7 @@ def test_registration_address_details(
             "eori_number": "GB205672212000",
             "sic_number": "12345",
             "vat_number": "GB123456789",
-            "registration_number": "GB123456",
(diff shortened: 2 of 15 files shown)


caseworker/templates/organisations/organisation/details.html
```diff
@@ -53,7 +53,7 @@
 		{% if organisation.registration_number %}
 			<div class="govuk-summary-list__row">
 				<dt class="govuk-summary-list__key">
-					{% lcs 'organisations.OrganisationPage.Details.SummaryList.COMPANY_REG_NUMBER' %}
+					Registration number
 				</dt>
```
markj0hnst0n, line 57: In the comment on the ticket says this should read 'Registered number' in 'Internal changes' section of the comment.  Also thinking you could probably delete the constant now if it's not being used.

markj0hnst0n: It's still showing 'Registration number' at my end.  Maybe it's reverted?

depsiatwal: The Org details is defo showing company number for me. 
However the edit screen was still showing CRN so changed it

markj0hnst0n (approved):

unit_tests/exporter/core/organisation/test_organisation_views.py
```diff
@@ -276,7 +276,7 @@ def test_registration_individual_end_to_end_non_uk_based(
         "location": "abroad",
         "name": "joe",
         "eori_number": "GB205672212000",
-        "registration_number": "GB123456",
+        "registration_number": "12345678",
```
Tllew, line 279: This used to be a common case, are there any entries in the DB that this might effect?

Tllew (approved):
realised after leaving my comment that this doesn't have any model changes, happy to approve

is it worth adding any backend validation too?

unit_tests/exporter/core/organisation/test_organisation_views.py
depsiatwal: possibly but we don't want any new ones

PR #2212, opened 2024-10-23
LTD-5566 add address line validation

markj0hnst0n:
### Aim

Add validation for address lines to match HMRC 35 char limit

[LTD-5566](https://uktrade.atlassian.net/browse/LTD-5566)

[LTD-5566]: https://uktrade.atlassian.net/browse/LTD-5566?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/exporter/core/organisation/forms.py
+++ b/exporter/core/organisation/forms.py
@@ -12,7 +12,12 @@ from .validators import (
     validate_registration,
 )
 from exporter.core.organisation.services import validate_registration_number
-from .fields import *
+from .fields import (
+    AddressLineField,
+    VatField,
+    EoriField,
+    SicField,
+)
 
 
 class RegistrationConfirmation(BaseForm):
--- a/exporter/core/organisation/forms.py
+++ b/exporter/core/organisation/forms.py
@@ -15,6 +15,39 @@ from exporter.core.organisation.services import validate_registration_number
 from .constants import Validation
 
 
+class AddressLineField(forms.CharField):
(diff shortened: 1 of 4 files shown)


exporter/core/organisation/forms.py
```diff
@@ -15,6 +15,39 @@
 from .constants import Validation
 
 
+class AddressLineField(forms.CharField):
+    default_validators = [
+        MaxLengthValidator(Validation.ADDRESS_LINE_MAX_LENGTH, Validation.ADDRESS_LINE_MAX_LENGTH_ERROR_MESSAGE)
+    ]
+
+
+class VatField(forms.CharField):
```
depsiatwal, line 24: I know these are related to forms but probably doesn't below here 
Lets move to it's own file perhaps in core so it can be a place where we add more generic fields to to share. .

markj0hnst0n: Done.  Followed Django guidance to keep them in the same parent folder as `forms.py`.  If we start to use them across both the caseworker and exporter app (which I think is a good idea and there is scope for with org edit) the reusable stuff should prob be moved to core at that point and the old edit stuff removed.

depsiatwal (approved):
just some lint issue to resolve which i'll leave with you.

Tllew (approved):

PR #2217, opened 2024-10-28
Add flake8-bandit rules to ruff and run against all files

kevincarrogan:
### Aim


pyproject.toml
```diff
@@ -23,4 +23,8 @@ exclude = '''
 '''
 
 [tool.ruff.lint]
-select = ["F401"]
+select = ["F401", "S", "G", "LOG"]
```
hnryjmes, line 26: am I right to think that the logging rules `G` and `LOG` are newly added here? and we didn't have coverage of those rules before?

hnryjmes (commented):
just wondering about the file `unit_tests/caseworker/search/test_helpers.py` being removed - was this done by ruff?

pyproject.toml
kevincarrogan: Correct. They are newly added.

hnryjmes (approved):
checked the bandit rules against the ruff implementation of the same rules, using the docs, and everything seems legit

PR #2219, opened 2024-10-29
LTD-5573-make-makedown-safe-remove

depsiatwal:
https://uktrade.atlassian.net/browse/LTD-5573

First part here is to just remove the filter_tag


exporter/applications/components.py
```diff
@@ -13,4 +13,6 @@ def back_to_task_list(application_id):
 
 def footer_label(application_id):
     url = reverse("applications:task_list", kwargs={"pk": str(application_id)})
-    return Label(f'Or <a class="govuk-link" href="{url}">return to application overview</a>')
+    return Label(
+        f'Or <a class="govuk-link govuk-link--no-visited-state" rel="noreferrer noopener" target="_blank" href="{url}">return to application overview</a>'
```
currycoder, line 17: Just interested, but what is this change achieving?

lite_content/lite_exporter_frontend/applications.py
```diff
@@ -223,7 +223,7 @@ class ReferenceNameQuestion:
 class ExportLicenceQuestions:
     class HaveYouBeenInformedQuestion:
         TITLE = "Have you received a letter or email from Border Force or HMRC informing you to apply for a licence?"
-        DESCRIPTION = "You may know this as an 'end use control'."
+        DESCRIPTION = "Give the application a reference name so you can refer back to it when needed."
```
currycoder, line 226: This doesn't seem right

exporter/applications/components.py
depsiatwal: https://github.com/uktrade/lite-frontend/blob/c577a90bee6f8cb6e6e093ec3f7e3efc1d30782a/lite_forms/components.py#L175

Looks like it's used to add a link to the footer. 
Yes I could add it to the template and be done with. But I question how far do I go with this PR. For now the aim is to just ensure we can remove the markdown method and make it clearer where it was used and replace the constants along the way.

lite_forms/helpers.py
```diff
@@ -155,20 +154,6 @@ def heading_used_as_label(components):
     return single_input
 
 
-def convert_to_markdown(text):
```
currycoder, line 158: 🙌

lite_content/lite_exporter_frontend/applications.py
depsiatwal: not sure how that got changed looks like copied from above revered back.

saruniitr (approved):

PR #2224, opened 2024-10-31
Ltd 5599 connect csp reporting to sentry

markj0hnst0n:
### Aim

Connect CSP to Sentry via django-csp package.

Tested on dev2 environment and everything working as expected.

[LTD-5599](https://uktrade.atlassian.net/browse/LTD-5599)

[LTD-5599]: https://uktrade.atlassian.net/browse/LTD-5599?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Pipfile
```diff
@@ -65,7 +65,7 @@ rules="~=3.3.0"
 djangorestframework = "~=3.15.2"
 django-chunk-upload-handlers = "~=0.0.13"
 django-compressor = "~=2.4.1"
-django-csp = "~=3.7"
+django-csp = "~=3.8"
 django-environ = "~=0.9.0"
```
markj0hnst0n, line 69: Upgraded to enable use of CSP_REPORT_TO variable.  As this may be needed in the future.  Read docs and it suggests that the CSP_REPORT_URI will be deprecated at some point

conf/base.py
```diff
@@ -257,9 +257,12 @@
 CSP_DEFAULT_SRC = env.tuple("CSP_DEFAULT_SRC", default=("'self'",))
 CSP_STYLE_SRC = env.tuple("CSP_STYLE_SRC", default=("'self'",))
 CSP_SCRIPT_SRC = env.tuple("CSP_SCRIPT_SRC", default=("'self'",))
+CSP_IMG_SRC = env.tuple("CSP_IMG_SRC", default=("'self'",))
 CSP_FONT_SRC = env.tuple("CSP_FONT_SRC", default=("'self'",))
+CSP_FRAME_ANCESTORS = env.tuple("CSP_FRAME_ANCESTORS", default=("'none'",))
+CSP_INCLUDE_NONCE_IN = env.tuple("CSP_INCLUDE_NONCE_IN", default=("'script-src'",))
+
 CSP_REPORT_ONLY = env.bool("CSP_REPORT_ONLY", False)
```
markj0hnst0n, line 265: Added protection against embedding site within a malicious site (clickjacking) with `CSP_FRAME_ANCESTORS`.  `CSP_IMG_SRC` should prevent issue brought up in pen test [here](https://uktrade.atlassian.net/browse/LTD-5572)

Tllew (approved):

hnryjmes (commented):
I noticed that the usual setup for `django-csp` in the docs, like adding to `INSTALLED_APPS` (https://django-csp.readthedocs.io/en/latest/installation.html) and setting config (https://django-csp.readthedocs.io/en/latest/configuration.html) has not been done here. Is that because actually we're not using this library in the typical way? Is the CSP to Sentry integration separate of that setup?

markj0hnst0n:
@hnryjmes
I didn't amend the existing setup but as I understand it we're only using the middleware here so adding to installed apps wasn't needed.  The `CSP_` variables in `conf/base.py` allow us to set the policy and turn the reporting aspect on and off.  While it's less explicit from the CSP perspective it allows us to control the policy in more code based way, I think.

Sentry reporting is just handled by the `CSP_REPORT_URI` variable.  If you add in the Sentry project reporting URL there then the reports get sent and processed.

hnryjmes (commented):
@markj0hnst0n ok great. happy to approve once you've been able to test this out on a dev environment, as I think you're still waiting to do

markj0hnst0n:
@hnryjmes 
Also I think the docs you've linked to there are the latest version and we're on 3.8 here.  I don't think the latest version has been labeled as a stable release so I didn't go that far.

kevincarrogan (approved):

hnryjmes (approved):

PR #2225, opened 2024-10-31
LTD-5603-privilege-escalation-restrict-view-profile 

depsiatwal:
https://uktrade.atlassian.net/browse/LTD-5603

This restricts the view profile and removes the change links so only a super-user or someone from Admin team can change roles and teams. 

This PR is more about injecting in the permissions for following PR which will restrict  the change profile screen. 

At the moment i've decided to keep this as a lite form but the change user profile screen may be easier to convert to a crispy_form at that point I will consider moving both  views.

### DIFF
--- a/caseworker/core/rules.py
+++ b/caseworker/core/rules.py
@@ -12,6 +12,7 @@ from caseworker.advice.services import (
 )
 from caseworker.core.constants import (
     ADMIN_TEAM_ID,
+    SUPER_USER_ROLE_ID,
     TAU_TEAM_ID,
     LICENSING_UNIT_TEAM_ID,
     LICENSING_UNIT_SENIOR_MANAGER_ROLE_ID,
@@ -180,6 +181,21 @@ def is_organisation_active(request, organisation):
     return organisation["status"]["key"] == "active"
 
 
+@rules.predicate
+def is_super_user(request):
+    user = get_logged_in_caseworker(request)
+    return user.get("role", {}).get("id") == SUPER_USER_ROLE_ID
+
+
+@rules.predicate
+def check_user_is_not_logged_in_caseworker(request, user):
+    caseworker = get_logged_in_caseworker(request)
+    caseworker_user_id = caseworker["id"]
+    if not user:
+        return True
+    return user.get("user", {}).get("id") != caseworker_user_id
+
+
 rules.add_rule("can_user_allocate_case", is_case_caseworker_operable)
 rules.add_rule("can_user_change_case", is_user_allocated)
 rules.add_rule("can_user_move_case_forward", is_user_allocated)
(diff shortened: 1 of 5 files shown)


caseworker/core/rules.py
```diff
@@ -180,6 +181,19 @@ def is_organisation_active(request, organisation):
...
 
 
+@rules.predicate
+def is_super_user(request):
+    user = get_logged_in_caseworker(request)
+    return user.get("role", {}).get("id") == SUPER_USER_ROLE_ID
+
+
+@rules.predicate
+def user_is_not_logged_in_user(request, user):
```
kevincarrogan, line 191: The naming of this rule had me confused. 

When I first read this I thought it was a check to see if the user was logged in and not that it was a check to see if the user passed in is the same as the currently logged in user.

caseworker/core/rules.py
```diff
@@ -180,6 +181,19 @@ def is_organisation_active(request, organisation):
...
+    caseworker = get_logged_in_caseworker(request)
+    caseworker_user_id = caseworker["id"]
+    return True if not user else user.get("user", {}).get("id") != caseworker_user_id
```
kevincarrogan, line 194: I think this could be written in a way that’s much clearer to understand.

depsiatwal: updated

caseworker/core/rules.py
depsiatwal: is that name clearer ?

kevincarrogan (approved):

caseworker/core/rules.py
```diff
@@ -180,6 +181,21 @@ def is_organisation_active(request, organisation):
...
+    return user.get("role", {}).get("id") == SUPER_USER_ROLE_ID
+
+
+@rules.predicate
+def check_user_is_not_logged_in_caseworker(request, user):
+    caseworker = get_logged_in_caseworker(request)
+    caseworker_user_id = caseworker["id"]
+    if not user:
+        return True
+    return user.get("user", {}).get("id") != caseworker_user_id
```
currycoder, line 196: It seems like we are passing through a datastructure to this function that contains the user...  Could we not enforce that this function accepts user as `user` directly?

And do we really need `if not user: ...`?  Seems like we should always have a value and if not it's ok to hard error

currycoder (approved):
One suggestion, but otherwise looking good to me

caseworker/core/rules.py
depsiatwal: ahh I merged too fast will update in my next follow up PR which also touches this area. 

what do you mean by user directly a structured user object ?

depsiatwal: oh do you mean pass in 
user["user'] as opposed to  data which is what this dump lite view is doing

currycoder: Yeah just pass in the user as `user` rather than whatever other cruft.  Happy for it to follow - no worries.

PR #2230, opened 2024-11-05
[LTD-5611 add trigger list assessments to desnz countersign

currycoder:
### Aim


saruniitr (approved):

hnryjmes (approved):
happy with code changes. also I pulled the branch locally and went through a case that had trigger list goods and verified that the 3 screens do in fact show the new table.

PR #2237, opened 2024-11-14
LTD-5604-edit-user-screen 

depsiatwal:


core/forms/templates/forms/filter_select.html
```diff
@@ -0,0 +1,27 @@
+{% load static %}
+{% include "django/forms/widgets/select.html" %}
+
+<script nonce="">
```
markj0hnst0n, line 4: This should be something like:

`<script type="text/javascript" nonce="{{ request.csp_nonce }}">`

To avoid CSP blocking/reporting.

ui_tests/caseworker/conftest.py
```diff
@@ -116,6 +116,11 @@ def mock_sso_lu_senior_manager_user_caseworker_sign_in(driver, internal_url):  #
     mock_sso_signin_with_email(driver, internal_url, "luseniormanager@digital.trade.gov.uk")
 
 
+@given("I sign in as Enforcement Unit Manager")
+def mock_sso_en_senior_manager_user_caseworker_sign_in(driver, internal_url):  # noqa
+    mock_sso_signin_with_email(driver, internal_url, "enforcementmanager@digital.trade.gov.uk")
+
```
markj0hnst0n, line 122: I like that this is becoming more specific with these changes to UI tests

caseworker/core/rules.py
```diff
@@ -191,9 +191,7 @@ def is_super_user(request):
 def check_user_is_not_logged_in_caseworker(request, user):
     caseworker = get_logged_in_caseworker(request)
     caseworker_user_id = caseworker["id"]
-    if not user:
-        return True
-    return user.get("user", {}).get("id") != caseworker_user_id
+    return True if not user else user.get("user", {}).get("id") != caseworker_user_id
```
currycoder, line 194: I liked it better before and would go so far as...

```
    if not user:
        return True
    is_actioned_user_not_current_user = user.get("user", {}).get("id") != caseworker_user_id
    return is_actioned_user_not_current_user
```

caseworker/templates/users/profile.html
```diff
@@ -67,10 +67,13 @@ <h1 class="govuk-heading-l">
...
 			</dd>
 			<dd class="govuk-summary-list__actions">
-				<a id="link-edit-email" href="{% url 'users:edit' data.user.id %}#email" class="govuk-link govuk-link--no-visited-state">
-					{% lcs 'users.UserProfile.SummaryList.CHANGE' %}
-					<span class="govuk-visually-hidden">{% lcs 'users.UserProfile.SummaryList.EMAIL' %}</span>
-				</a>
+				{% test_rule 'can_caseworker_edit_user' request as show_edit_email %}
+				{% if show_edit_email %}
+					<a id="link-edit-email" href="{% url 'users:edit' data.user.id %}#email" class="govuk-link govuk-link--no-visited-state">
+						{% lcs 'users.UserProfile.SummaryList.CHANGE' %}
```
currycoder, line 73: Can we bin this `lcs` and replace with straight text?

caseworker/templates/users/profile.html
```diff
@@ -115,7 +118,7 @@ <h1 class="govuk-heading-l">
 				{{ data.user.default_queue.name }}
 			</dd>
 			<dd class="govuk-summary-list__actions">
-				<a id="link-edit-default-queue" href="{% url 'users:edit' data.user.id %}#default_queue" class="govuk-link govuk-link--no-visited-state">
+				<a id="link-edit-default-queue" href="{% url 'users:edit' data.user.id %}" class="govuk-link govuk-link--no-visited-state">
```
currycoder, line 121: Looks like lots of trouble taken above to ensure the edit link doesn't show - why not here?

caseworker/users/manage/forms.py
```diff
@@ -0,0 +1,82 @@
...
+        self._email_changed = False
+        if self.read_only:
+            for field in self._editable_fields:
+                self.fields[field].widget.attrs["disabled"] = True
+                self.fields[field].required = False
+
+        self.fields["team"].choices = [Choice(t["id"], t["name"]) for t in teams]
+        self.fields["role"].choices = [Choice(r["id"], r["name"]) for r in roles]
+
+        get_team_id = lambda q: q.get("team").get("id") if q.get("team") else None
```
currycoder, line 51: Could be cleaner;
```
lambda queue: queue["team"].get("id") if queue.get("team") else None
```

caseworker/users/manage/forms.py
```diff
@@ -0,0 +1,82 @@
...
+        label="Default Queue",
+        widget=FilterSelect(
+            parent_select_name="team",
+            attrs={"id": "default_queue", "class": "govuk-select"},
+        ),
+    )
+
+    def __init__(self, request, teams, roles, queues, can_caseworker_edit_user, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.read_only = not can_caseworker_edit_user
```
currycoder, line 40: I kind of feel like this is a bit backwards - we shouldn't pass something in to the form to tell it what fields it should make editable or not.  Instead can't we have a `EditCaseworkerUser` form and a `EditCaseworkerQueue` form which are declarative (and maybe inherit from a shared parent or something?) and then choose between one or the other in the View?

caseworker/users/manage/forms.py
```diff
@@ -0,0 +1,82 @@
...
+        self.request = request
+        self._email_changed = False
+        if self.read_only:
+            for field in self._editable_fields:
+                self.fields[field].widget.attrs["disabled"] = True
+                self.fields[field].required = False
```
currycoder, line 46: I *think* the overall field object should have `disabled = True` not just the widget, e.g. https://docs.djangoproject.com/en/5.1/ref/forms/fields/#disabled

caseworker/users/manage/views.py
```diff
@@ -0,0 +1,75 @@
...
+    @expect_status(
+        HTTPStatus.OK,
+        "Error editing user",
+        "Unexpected error editing user",
+    )
+    def edit_user(self, data):
+        # If user is updating their own default_queue, update the local user instance
+        if str(self.user_id) == self.request.session["lite_api_user_id"]:
+            self.request.session["default_queue"] = data.get("default_queue")
+        return update_gov_user(self.request, self.user_id, data)
```
currycoder, line 70: Shouldn't we perform the request first and only update the local session if it was successful?

caseworker/users/services.py
```diff
@@ -83,6 +84,18 @@ def put_gov_user(request, pk, json):
     return data.json(), data.status_code
 
 
+def update_gov_user(request, pk, json):
+    data = client.put(request, f"/caseworker/gov_users/{pk}/update", json)
+    return data.json(), data.status_code
+
+
+def get_gov_user_list(request, json):
```
currycoder, line 92: Could `json` be `filters` or something instead?  It's not really json but a dict, I think

caseworker/users/views/users.py
```diff
@@ -88,38 +86,6 @@ def get(self, request, **kwargs):
         return redirect(reverse_lazy("users:user", kwargs={"pk": request.session["lite_api_user_id"]}))
 
 
-class EditUser(SingleFormView):
```
currycoder, line 91: You love to see it

currycoder (commented):
Really love the direction of these changes.  Mostly minor comments aside from the forms comment

caseworker/users/services.py
depsiatwal: done

caseworker/users/manage/views.py
depsiatwal: yeah good point this is now after our update call

caseworker/core/rules.py
depsiatwal: done

caseworker/templates/users/profile.html
depsiatwal: done

caseworker/users/manage/forms.py
depsiatwal: yeah I got two versions of the form now and agree appears more slick

caseworker/users/manage/forms.py
currycoder: Are you missing a push? I still see the same

caseworker/users/manage/forms.py
```diff
@@ -0,0 +1,91 @@
...
+        self._email_changed = False
+        for field in self._editable_fields:
+            self.fields[field].disabled = True
+            self.fields[field].required = False
+
+    def clean(self):
+        cleaned_data = super().clean()
+        for field in self._editable_fields:
+            del cleaned_data[field]
+        return cleaned_data
```
currycoder, line 74: I *think* django does this for you if you have set `disabled = True` on the field.  Might be worth checking and then dropping this method if so

currycoder (approved):
Just one additional comment which I will leave with you

markj0hnst0n (approved):

caseworker/users/manage/forms.py
depsiatwal: not sure what happen to it lol fixed up

PR #2242, opened 2024-11-20
LTD-5659 Add simple health check

markj0hnst0n:
### Aim

DBT platform team needs to hit an endpoint and get a 200 to make sure that the django app is up and running

[LTD-5659](https://uktrade.atlassian.net/browse/LTD-5659)

[LTD-5659]: https://uktrade.atlassian.net/browse/LTD-5659?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


unit_tests/core/test_healthcheck.py
```diff
@@ -66,3 +67,21 @@ def test_healthcheck_down(backends):
         b"</status><response_time>0.23</response_time>"
         b"</pingdom_http_custom_check>\n"
     )
+
+
+def test_service_available_check_broken(client, backends):
+    backends.reset()
+    backends.register(HealthCheckBroken)
```
depsiatwal, line 74: why we registering the HealthCheckBroken then asserting everything is  OK

markj0hnst0n: To ensure that even if one of the health checks breaks we still get a 200 response form the endpoint

depsiatwal: ok that makes sense may be a comment to clarify why the test expects a 200

markj0hnst0n: Done!

core/health_check/views.py
```diff
@@ -8,3 +10,11 @@ def render_to_response(self, context, status):
...
         context["total_response_time"] = sum([plugin.time_taken for plugin in context["plugins"]])
         return super().render_to_response(context=context, status=status, content_type="text/xml")
+
+
+class ServiceAvailableHealthCheckView(MainView):
+    def get(self, request, *args, **kwargs):
+        return self.render_to_response()
+
+    def render_to_response(self):
+        return HttpResponse(status.HTTP_200_OK)
```
saruniitr, line 20: Is it worth checking to see if we can access db also or it doesn't make it simple anymore?

markj0hnst0n: Kevin had asked specifically that we don't do a db check and that we just check that the web service for the application is up

saruniitr (approved):

PR #2243, opened 2024-11-20
Ltd 5653 desnz multiple licence conditions

Tllew:
Converted GiveApprovalAdviceView to a sessionwizard
Added conditional checkboxes to the licence conditions
Added a DESNZeditview to handle the prepopulation of the sessionwizard
Renamed the old views to be -legacy to move us a step closer toward this flow being the default (because of this it may be difficult to read the file changes, I'd recommend checking the commit history for a simpler view of the changes)

This is hopefully the first step to converting this whole are to a sessionwizard and removing the factories

[LTD-5653](https://uktrade.atlassian.net/browse/LTD-5653)

[LTD-5653]: https://uktrade.atlassian.net/browse/LTD-5653?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


caseworker/advice/forms.py
```diff
@@ -484,3 +492,169 @@ def __init__(
...
+    proviso_radios = forms.MultipleChoiceField(
+        label="Add a licence condition (optional)",
+        required=False,
+        widget=forms.CheckboxSelectMultiple,
+        choices=(),
+    )
+
+    def clean(self):
+        cleaned_data = super().clean()
+        # only return proviso (text) for selected radios, nothing else matters, join by 2 newlines
```
currycoder, line 588: Could this comment be a little more explicit for new developers that have no idea what's happening?

caseworker/advice/forms.py
```diff
@@ -484,3 +492,169 @@ def __init__(
...
+    )
+    footnote_details = forms.CharField(
+        widget=forms.Textarea(attrs={"rows": 3, "class": "govuk-!-margin-top-4"}),
+        label="",
+        required=False,
+    )
+
+    def __init__(self, *args, **kwargs):
+        kwargs.pop("approval_reason")
+        kwargs.pop("proviso")
```
currycoder, line 647: Should this still be here?

caseworker/advice/services.py
```diff
@@ -321,12 +321,12 @@ def get_advice_subjects(case, countries=None):
...
-            "type": "proviso" if data["proviso"] else "approve",
+            "type": "proviso" if data.get("proviso", False) else "approve",
             "text": data["approval_reasons"],
-            "proviso": data["proviso"],
-            "note": data["instructions_to_exporter"],
-            "footnote_required": True if data["footnote_details"] else False,
-            "footnote": data["footnote_details"],
+            "proviso": data.get("proviso", ""),
+            "note": data.get("instructions_to_exporter", ""),
+            "footnote_required": True if data.get("footnote_details") else False,
```
currycoder, line 328: Any ideas what this is for?  I realise this is from old code but it seems a bit nuts

caseworker/advice/templates/advice/view_my_advice.html
```diff
@@ -33,16 +34,18 @@ <h2 class="govuk-error-summary__title" id="error-summary-title">
...
-                        Edit recommendation
-                    </a>
-                  {% endif %}
-                  {% if buttons.clear_recommendation %}
-                    <a role="button" draggable="false" class="govuk-button govuk-button--secondary" href="{% url 'cases:delete_advice' queue_pk case.id %}">
-                        Clear recommendation
-                    </a>
+                {% if buttons.edit_recommendation %}
+                  {% if caseworker|user_in_DESNZ_team%}
+                    <a role="button" draggable="false" class="govuk-button govuk-button--secondary" href="{% url 'cases:edit_advice_desnz' queue_pk case.id %}">Edit recommendation</a>
```
currycoder, line 39: Does this view handle both the approval and refusal advice?

currycoder: Not to say that the single view *should* handle both, but the net effect of hitting edit recommendation should probably be that refusal advice is edited when there is a refusal.  Is that the case?  Or was it before?

currycoder: Tried this locally and can see that I get to the edit approval flow when editing a refusal recommendation.

caseworker/advice/views.py
```diff
@@ -350,6 +356,69 @@ def get_context_data(self, **kwargs):
...
+        (AdviceView.DESNZ_RECOMMEND_APPROVAL, forms.DESNZRecommendAnApproval),
+        (AdviceView.LICENCE_CONDITIONS, forms.PicklistApprovalAdviceFormEdit),
+        (AdviceView.LICENCE_FOOTNOTES, forms.FootnotesApprovalAdviceForm),
+    ]
+
+    def get_form_initial(self, step):
+        my_advice = services.filter_current_user_advice(self.case.advice, self.caseworker_id)
+        advice = my_advice[0]
+
+        # default to other so that they're not likely to reselect, approval_radios isn't stored.
```
currycoder, line 410: This comment takes a lot of reading and cross-checking to fully understand.  Can this be made super explicit for somebody that has no knowledge of this?

caseworker/advice/views.py
```diff
@@ -350,6 +356,69 @@ def get_context_data(self, **kwargs):
         return context
 
 
+class GiveApprovalAdviceViewDESNZ(LoginRequiredMixin, CaseContextMixin, BaseSessionWizardView):
+    # on save combine everything into a single field
```
currycoder, line 360: I feel like this comment needs to be more targeted

caseworker/advice/views.py
```diff
@@ -350,6 +356,69 @@ def get_context_data(self, **kwargs):
...
+        return kwargs
+
+    def get_success_url(self):
+        return reverse("cases:view_my_advice", kwargs=self.kwargs)
+
+    def done(self, form_list, form_dict, **kwargs):
+        data = {}
+        for form_data in form_dict.values():
+            data.update(form_data.cleaned_data)
+        services.post_approval_advice(self.request, self.case, data)
```
currycoder, line 394: Could some status checking be put in?  Elsewhere we use a decorator which will produce an error message to the user I believe.

caseworker/advice/urls.py
```diff
@@ -7,9 +7,11 @@
     path("case-details/", views.CaseDetailView.as_view(), name="case_details"),
     path("select-advice/", views.SelectAdviceView.as_view(), name="select_advice"),
     path("approve-all/", views.GiveApprovalAdviceView.as_view(), name="approve_all"),
+    path("approve-all-desnz/", views.GiveApprovalAdviceViewDESNZ.as_view(), name="approve_all_desnz"),
```
currycoder, line 10: As per discussions on slack, can the `-desnz` bit be dropped?  Or do you plan to do this when this is widened for other teams later?

currycoder (commented):
Really great work - I think it's almost there, I just have a bunch of (mostly minor) enhancements to suggest.  The only real major is the comment about being put on the edit approval view when editing refusal advice.

caseworker/advice/views.py
Tllew:
```
        # When the form is prepopulated in the edit flow, 
        # the radio values are set to other because only the textfield values are stored it's not possible to replay the selected radio.
        ```

File: caseworker/advice/views.py (hunk 6 above)
Comment (Tllew) [reply caseworker/advice/views.py:360]: I think this was left in a s a part of my todo list, I don't think it was needed as the combination of form data is done in the form, I've added another commend in the done function to reflect the flattening of the dict_form though

File: caseworker/advice/services.py (hunk 3 above)
Comment (Tllew) [reply caseworker/advice/services.py:328]: this is just for formatting, it needs gets because the keys now won't exist if the steps have been skipped where previously they were present but empty strings.

in the next iteration I'd like to update this section, the refuse and the getting of advice so that we don't advice = my_advice[0] every time.

File: caseworker/advice/views.py [hunk 9]
```diff
@@ -350,6 +356,78 @@ def get_context_data(self, **kwargs):
         return context
 
 
+class GiveApprovalAdviceView(LoginRequiredMixin, CaseContextMixin, BaseSessionWizardView):
+
+    form_list = [
+        (AdviceView.DESNZ_RECOMMEND_APPROVAL, forms.DESNZRecommendAnApproval),
```
currycoder, line 362: Still some `DESNZ` mentions

caseworker/advice/views.py
```diff
@@ -350,6 +356,78 @@ def get_context_data(self, **kwargs):
...
+        (AdviceView.LICENCE_CONDITIONS, forms.LicenceConditionsForm),
+        (AdviceView.LICENCE_FOOTNOTES, forms.FootnotesApprovalAdviceForm),
+    ]
+
+    condition_dict = {
+        AdviceView.DESNZ_RECOMMEND_APPROVAL: C(is_desnz_team),
```
currycoder, line 368: Do we still need this condition?  I believe that the conditionality has been moved up to either being routed to this view or not, right?

caseworker/advice/views.py
```diff
@@ -350,6 +356,78 @@ def get_context_data(self, **kwargs):
...
+        AdviceView.LICENCE_CONDITIONS: C(form_add_licence_conditions("desnz_recommend_approval")),
+        AdviceView.LICENCE_FOOTNOTES: C(form_add_licence_conditions("desnz_recommend_approval")),
```
currycoder, line 370: More `desnz` to drop here I think

unit_tests/caseworker/advice/views/test_edit_advice.py
```diff
@@ -203,3 +208,153 @@ def test_edit_refuse_advice_get(
...
+    response = post_to_step(
+        AdviceView.DESNZ_RECOMMEND_APPROVAL,
+        {"approval_reasons": "reason updated", "add_licence_conditions": False},
+    )
+    assert response.status_code == 302
+    history = [item for item in requests_mock.request_history if user_advice_create_url in item.url]
+    assert len(history) == 1
+    history = history[0]
+    assert history.method == "POST"
+    assert history.json()[0] == {
```
currycoder, line 271: `history.json() == [{...}]` could be slightly simpler

caseworker/advice/forms.py
currycoder: Still wondering about this

caseworker/advice/forms.py
```diff
@@ -484,3 +492,169 @@ def __init__(
         self.application_pk = application_pk
         self.is_user_rfd = is_user_rfd
         self.organisation_documents = organisation_documents
+
+
+class DESNZRecommendAnApproval(PicklistAdviceForm, BaseForm):
```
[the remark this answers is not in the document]
currycoder: Still valid

caseworker/advice/services.py
currycoder: I meant the field in general - `footnote` and `footnote_required` seems odd

caseworker/advice/templates/advice/view_my_advice.html
currycoder: Might be worth creating/surfacing a follow-up ticket to the squad for this as I still think the code we have in the wild right now will take you to the "edit approval" flow when a refusal has been made

currycoder (approved):
Nice one.  Just a few minor comments/re-comments which I will leave with you

caseworker/advice/forms.py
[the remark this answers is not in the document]
Tllew: missed that, sorted

caseworker/advice/views.py
Tllew: I'd like to leave this in, just as it sets good context for adding further conditional forms for the next iteration of this work

caseworker/advice/templates/advice/view_my_advice.html
```diff
@@ -33,16 +34,18 @@ <h2 class="govuk-error-summary__title" id="error-summary-title">
...
+                    <a role="button" draggable="false" class="govuk-button govuk-button--secondary" href="{% url 'cases:edit_advice' queue_pk case.id %}">Edit recommendation</a>
+                  {% else %}
+                    <a role="button" draggable="false" class="govuk-button govuk-button--secondary" href="{% url 'cases:edit_advice_legacy' queue_pk case.id %}">Edit recommendation</a>
                   {% endif %}
+                {% endif %}
+                {% if buttons.clear_recommendation %}
+                  <a role="button" draggable="false" class="govuk-button govuk-button--secondary" href="{% url 'cases:delete_advice' queue_pk case.id %}">
+                      Clear recommendation
+                  </a>
+                {% endif %}
```
kevincarrogan, line 48: Is this an opportunity to wrap this using `django_rules`?

currycoder: Definitely up for this.  We have a bunch of follow-up work here which will involve switching other OGDs over to the new flows/getting rid of the old.  How would you feel about this enhancement accompanying those?  I'm a little aware that this PR is getting larger and keen to get it in front of people sooner.

caseworker/advice/views.py
```diff
@@ -387,7 +387,7 @@ def get_success_url(self):
         return reverse("cases:view_my_advice", kwargs=self.kwargs)
 
     @expect_status(
-        HTTPStatus.OK,
+        HTTPStatus.CREATED,
```
currycoder, line 390: I noticed that an exception was being thrown right at the end of the process when the form was submitted.  That seems to indicate that A) the unit tests aren't accurately mocking the API endpoint (which I'll fixed) and B) We probably need some E2E tests for this.

Could the E2E tests be added as a follow-up?

Tllew: Yeah e2e should definitely be a part of the follow up as everything will be converted to this flow, but again was hoping to do that as a part of the next step and just to reduce the size of this initial PR that targets DESNZ specifically

caseworker/advice/forms.py
kevincarrogan: I'm seeing this in all of the forms and it makes me wonder if there's a better way of solving this than just popping things to make them go away.

I also prefer `del` in cases where we are not using the returned value.

caseworker/advice/templates/advice/view_my_advice.html
```diff
@@ -1,5 +1,6 @@
 {% extends 'layouts/case.html' %}
 {% load crispy_forms_tags advice_tags %}
+{% load static advice_tags %}
```
kevincarrogan, line 3: `advice_tags` is being loaded twice and I can't see where `static` is being used.

caseworker/advice/views.py
```diff
@@ -153,7 +158,7 @@ def get_context_data(self, **kwargs):
         return {**context, "security_approvals_classified_display": self.security_approvals_classified_display}
 
 
-class GiveApprovalAdviceView(LoginRequiredMixin, CaseContextMixin, FormView):
+class GiveApprovalAdviceViewLegacy(LoginRequiredMixin, CaseContextMixin, FormView):
```
kevincarrogan, line 161: How long are we expecting to keep these legacy views around for?

If it's not going to be handled within this workstream then I think a comment on why it's legacy and what we expect to do with it in the future might be useful.

caseworker/advice/views.py
```diff
@@ -350,6 +356,78 @@ def get_context_data(self, **kwargs):
...
+    def post_approval_advice(self, data):
+        return services.post_approval_advice(self.request, self.case, data)
+
+    def done(self, form_list, form_dict, **kwargs):
+        data = {}
+        # flattens the form_dict
+        for form_data in form_dict.values():
+            data.update(form_data.cleaned_data)
+        self.post_approval_advice(data)
+        return redirect(self.get_success_url())
```
kevincarrogan, line 403: I feel that this is possibly what our payload builder is for.

https://github.com/uktrade/lite-frontend/blob/dev/core/wizard/payloads.py

unit_tests/caseworker/advice/views/test_edit_advice.py
```diff
@@ -203,3 +208,255 @@ def test_edit_refuse_advice_get(
...
+def url_desnz(data_queue, data_standard_case):
+    return reverse(f"cases:edit_advice", kwargs={"queue_pk": data_queue["id"], "pk": data_standard_case["case"]["id"]})
+
+
+@pytest.fixture
+def post_to_step(post_to_step_factory, url_desnz):
+    return post_to_step_factory(url_desnz)
+
+
+@mock.patch("caseworker.advice.views.get_gov_user")
```
kevincarrogan, line 223: We should use `mocker` instead of `@mock` directly.

unit_tests/caseworker/advice/views/test_edit_advice.py
```diff
@@ -203,3 +208,255 @@ def test_edit_refuse_advice_get(
...
+        AdviceView.RECOMMEND_APPROVAL,
+        {"approval_reasons": "reason updated", "add_licence_conditions": True},
+    )
+    assert response.status_code == 200
+    soup = beautiful_soup(response.content)
+    # redirected to next form
+    header = soup.find("h1")
+    assert header.text == "Add licence conditions, instructions to exporter or footnotes (optional)"
+
+    add_LC_response = post_to_step(
```
kevincarrogan, line 382: I think the abbreviation to `LC` here could just be `licence_condition`.

unit_tests/caseworker/advice/views/test_give_approval_advice_view.py
```diff
@@ -5,6 +5,7 @@
 from django.urls import reverse
 
 from caseworker.advice import services
+from caseworker.advice.constants import AdviceView
```
kevincarrogan, line 8: This confuses me. Why is a `View` class coming from the constants file?

unit_tests/caseworker/advice/views/test_select_advice_view.py
```diff
@@ -26,6 +29,25 @@ def test_select_advice_post(authorized_client, url, recommendation, redirect):
...
+                    "name": "DESNZ Chemical",
+                    "alias": services.DESNZ_CHEMICAL,
+                }
+            }
+        },
+        None,
+    )
+    response = authorized_client.post(url, data={"recommendation": "approve_all"})
+    assert response.status_code == 302
+    assert "approve-all" in response.url
```
kevincarrogan, line 48: Can't we just check that it's the actual URL instead of checking that a bit of it is correct?

caseworker/advice/forms.py
currycoder: I think this comes from over-simplifying `get_form_kwargs()` in the view.  Previously, there was a lot of if statements over there to ensure that forms got *only* the values they were interested in.  I think in trying to bin the if-statements, we have kind of coupled the forms a lot to those views.  I would propose that we ensure that forms only get what they are supposed to with something like;

```
        kwargs = super().get_form_kwargs(step)
        step_kwargs = {
            "recommend_approval": self.get_form_kwargs_recommend_approval,
            "licence_condition": self.get_form_kwargs_licence_condition,
            "licence_footnotes": self.get_form_kwargs_licence_footnotes,
        }
        return kwargs.update(step_kwargs[step]())
    
    def get_form_kwargs_licence_condition(self):
        proviso = get_picklists_list(
            self.request, type="proviso", disable_pagination=True, show_deactivated=False
        ) 
        return {"proviso": proviso}
```

Swapping out

caseworker/advice/forms.py
```diff
@@ -484,3 +492,167 @@ def __init__(
         self.application_pk = application_pk
         self.is_user_rfd = is_user_rfd
         self.organisation_documents = organisation_documents
+
+
+class RecommendAnApproval(PicklistAdviceForm, BaseForm):
+    DOCUMENT_TITLE = "Recommend approval for this case"
```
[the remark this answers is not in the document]
Tllew: looks to be remnance of the old form, removed all cases of this for children of BaseForm

caseworker/advice/conditionals.py
```diff
@@ -0,0 +1,15 @@
+from caseworker.advice import services
+
+
+def form_add_licence_conditions(step_name):
+    """Returns the boolean value for add_licence_conditions from the approval form"""
+
+    def _get_form_field_boolean(wizard):
+        cleaned_data = wizard.get_cleaned_data_for_step(step_name)
+        return cleaned_data.get("add_licence_conditions")
```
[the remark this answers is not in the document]
Tllew: `cleaned_data.get("add_licence_conditions",False)`

caseworker/advice/views.py
Tllew: I've added this, it looks ok but I think this pushes me over the edge to wanting to make a views folder and separating all the views into files to neaten things up a bit (views.py is over 1k lines now) I'd like to do this in the next step though.

unit_tests/caseworker/advice/views/test_give_approval_advice_view.py
Tllew: renamed to `AdviceSteps`

caseworker/advice/services.py
saruniitr: @currycoder @Tllew this is because these fields were already existing in the data model and Advice updated to work with them.
Footnote is an optional text box, initially UCD wanted to include a checkbox and when checked it reveals the textbox but that is too many clicks so it was decided to show the textbox by default and if there is any data then we set `footnote_required` accordingly.

kevincarrogan (approved):

caseworker/advice/views.py
Tllew: this workstream hopefully

PR #2248, opened 2024-11-25
LTD-5605-caseworker-add-user-view 

depsiatwal:
following on from this pattern 

https://github.com/uktrade/lite-frontend/pull/2237

This is the New Add form to add a new user, replacing the old lite forms.


caseworker/templates/teams/team.html
```diff
@@ -27,11 +27,14 @@ <h1 class="govuk-heading-l">
...
-			<div class="lite-tabs__controls">
-				<a id="button-add-user" href="{% url 'users:add' %}" class="govuk-button" draggable="false" role="button">
-					{% lcs 'teams.TeamPage.ADD_A_MEMBER_BUTTON' %}
-				</a>
-			</div>
+			{% test_rule 'can_caseworker_add_user' request as show_add_user %}
+			{% if show_add_user %}
+				<div class="lite-tabs__controls">
+					<a id="button-add-user" href="{% url 'users:add' %}" class="govuk-button" draggable="false" role="button">
+						{% lcs 'teams.TeamPage.ADD_A_MEMBER_BUTTON' %}
```
currycoder, line 34: Drop the `lcs`?

caseworker/templates/users/index.html
```diff
@@ -15,9 +15,12 @@ <h1 class="govuk-heading-l">
 			<a id="button-manage-roles" href="{% url 'users:roles' %}" class="govuk-button govuk-button--secondary" draggable="false" role="button">
 				{% lcs 'users.UsersPage.MANAGE_ROLES_BUTTON' %}
 			</a>
-			<a id="button-add-user" href="{% url 'users:add' %}" class="govuk-button" draggable="false" role="button">
-				{% lcs 'users.UsersPage.INVITE_BUTTON' %}
-			</a>
+			{% test_rule 'can_caseworker_add_user' request as show_add_user %}
+			{% if show_add_user %}
+				<a id="button-add-user" href="{% url 'users:add' %}" class="govuk-button" draggable="false" role="button">
+					{% lcs 'users.UsersPage.INVITE_BUTTON' %}
```
currycoder, line 21: And again!

caseworker/users/manage/views.py
```diff
@@ -78,3 +78,44 @@ def get_context_data(self, **kwargs):
...
+
+
+class AddCaseworkerUserView(LoginRequiredMixin, SuccessMessageMixin, FormView):
+    form_class = AddCaseworkerUser
+    template_name = "core/form.html"
+    success_message = "User added successfully"
+
+    def dispatch(self, *args, **kwargs):
+        if not rules.test_rule("can_caseworker_add_user", self.request):
+            raise Http404()
```
currycoder, line 90: I think this should probably be a 403

depsiatwal: done

caseworker/templates/teams/team.html
depsiatwal: ahh yeah replaced all in that file now.

caseworker/templates/users/index.html
depsiatwal: as above

caseworker/core/rules.py
```diff
@@ -224,3 +224,4 @@ def check_user_is_not_logged_in_caseworker(request, user):
 rules.add_rule("can_user_manage_organisation", is_user_manage_organisations_role & is_organisation_active)
 rules.add_rule("can_caseworker_deactivate", (is_super_user) & check_user_is_not_logged_in_caseworker)  # noqa
 rules.add_rule("can_caseworker_edit_user", (is_super_user))  # noqa
+rules.add_rule("can_caseworker_add_user", (is_super_user))  # noqa
```
markj0hnst0n, line 227: Do we need a new rule when it checks the same predicate as another rule?  Maybe a different name?

depsiatwal: can_caseworker_add_user  atm it's same check  is_super_user which I think is ok as it allow flexibility in the future. We may one day change is_super_user & is_admin_team for example and it would only impact add_user screen

caseworker/users/urls.py
```diff
@@ -8,7 +8,7 @@
 urlpatterns = [
     path("", users.UsersList.as_view(), name="users"),
     path("<uuid:pk>/", users.ViewUser.as_view(), name="user"),
-    path("add", users.AddUser.as_view(), name="add"),
+    path("add", views.AddCaseworkerUserView.as_view(), name="add"),
```
markj0hnst0n, line 11: Not sure if it matters but I think the convention is to have a / at the end

depsiatwal: yeah the others were from LITE forms these are new one which is more common other way

markj0hnst0n: I mean the url itself.  The first parameter in the path function.  `add/ `instead of `add`.  I like using it as it's useful for search.  Google recommends keeping the naming convention consistent across the app.

depsiatwal: ohh I see misunderstood the comment yeah I agree amended well spotted.

markj0hnst0n:
Just a more general comment.  The UI for the form isn't great with that massive e-mail input not sure if that's of huge importance on the caseworker side.

currycoder (approved):

PR #2250, opened 2024-11-29
LTD-5695 fix possibleURL redirection from remote source vulnerability

markj0hnst0n:
### Aim

Fix possible URL redirection from remote source vulnerability uncovered by GitHub codescan [here](https://github.com/uktrade/lite-frontend/security/code-scanning/40), [here](https://github.com/uktrade/lite-frontend/security/code-scanning/39) [and here](https://github.com/uktrade/lite-frontend/security/code-scanning/38)

Added checking for request.path in lite_forms SummaryListFormView to comply with the information outlined in the codescan links to resolve issue and remove unused submission variable from MultiFormView

[LTD-5695](https://uktrade.atlassian.net/browse/LTD-5695)

[LTD-5695]: https://uktrade.atlassian.net/browse/LTD-5695?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lite_forms/views.py
```diff
@@ -403,6 +417,7 @@ def get_next_form_page(self, form_pk, action, request, post_errors):
         if self.validate_only_until_final_submission:
             return self.generate_summary_list()
 
+        ensure_redirect_destination_relative(redirect(request.path).url)
```
currycoder, line 420: I think to avoid these warnings you might need to do something like 
```
sanitised_path = ensure_redirect_destination_relative(request.path)
return redirect(sanitised_path)
```

lite_forms/views.py
```diff
@@ -188,9 +205,6 @@ def post(self, request, **kwargs):
         self.init(request, **kwargs)
         submission = self.on_submission(request, **kwargs)  # noqa
 
-        if submission:
-            return redirect(submission)
-
```
markj0hnst0n, line 193: The `on_submission` method for the MultiFormView never returns a value in any of the code so it made this path difficult to tests.  As it wasn't used I removed it.

In exporter I've tested GenerateDocument, ExportDetails, ApplicationCopy and ExportLicenceQuestions.

In caseworker I've tested RegisterOrganisation and RegisterHMRC.

There doesn't appear to be any issues with the flow.

lite_forms/views.py
```diff
@@ -20,6 +22,21 @@
...
+class UnsafeRedirectDestination(SuspiciousOperation):
+    pass
+
+
+# Check that the redirect url is not an injected path taking the user somewhere outside of the application
+def ensure_redirect_destination_relative(destination):
+    destination_url_string = str(destination).replace("\\", "")
+    valid_url = not urlparse(destination_url_string).netloc and not urlparse(destination_url_string).scheme
+    if not valid_url:
+        raise UnsafeRedirectDestination(f"Redirect destination '{destination_url_string}' was not a relative URL")
```
kevincarrogan, line 37: Is there any real need to not just raise the suspicious operation exception here instead of making our own?

lite_forms/views.py
```diff
@@ -20,6 +22,21 @@
...
 from lite_forms.submitters import submit_paged_form
+from urllib.parse import urlparse
+
+
+class UnsafeRedirectDestination(SuspiciousOperation):
+    pass
+
+
+# Check that the redirect url is not an injected path taking the user somewhere outside of the application
+def ensure_redirect_destination_relative(destination):
```
kevincarrogan, line 33: I think Django has its own version of this that we could use. I can’t remember what it’s called but we’ve definitely used it in other parts of our code. 

Try to find something called `return_to` and it should get you to the answer.

lite_forms/views.py
currycoder: I thought it might be easier to track these in sentry this way, but otherwise no particular reason.

lite_forms/views.py
markj0hnst0n: removed to use different approach as discussed

lite_forms/views.py
markj0hnst0n: I've kept this in with a slightly amended format as I do enjoy a descriptive error.

kevincarrogan (approved):

lite_forms/views.py
```diff
@@ -340,7 +340,8 @@
         if self.validate_only_until_final_submission:
             return self.generate_summary_list()
 
-        return redirect(request.path)
+        sanitised_url = check_url(request, request.path)
+        return redirect(sanitised_url)
```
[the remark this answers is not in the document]
currycoder: I see that this is still triggering...  Do we need to put some sort of comment on this line to say to ignore it..?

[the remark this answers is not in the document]
kevincarrogan: If we're happy with our resolution and confident that it does solve this problem then we can dismiss this alert with an explanation.

I can do that once this is merged in.

currycoder (approved):

PR #2252, opened 2024-12-02
Asset compilation dbt platform

emileswarts:
### Aim

Compile static assets at build time (build pipeline) not container boot time (Procfile entrypoint).
Introduce a new entrypoint to be used on DBT Platform so as to not affect the service running on GOVPaas.

Don't add container IP to allowed hosts, this is now managed at the infrastructure level.

Correct path to assets on DBT Platform, traversing up the directories like on GOVPaas will end up on the root directory.

### DIFF
--- a/.copilot/image_build_run.sh
+++ b/.copilot/image_build_run.sh
@@ -1,3 +1,26 @@
 #!/usr/bin/env bash
 
 set -e
+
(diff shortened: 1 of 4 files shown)


depsiatwal (approved):
lgtm

PR #2254, opened 2024-12-03
Compile assets in build pipeline instead of in Procfile entrypoint

emileswarts:
###Aim

Compile static assets at build time (build pipeline) not container boot time (Procfile entrypoint).
Introduce a new entrypoint to be used on DBT Platform so as to not affect the service running on GOVPaas.

Don't add container IP to allowed hosts, this is now managed at the infrastructure level.

Correct path to assets on DBT Platform, traversing up the directories like on GOVPaas will end up on the root directory.

### DIFF
--- a/.copilot/image_build_run.sh
+++ b/.copilot/image_build_run.sh
@@ -1,3 +1,26 @@
 #!/usr/bin/env bash
 
 set -e
+
+export DJANGO_SECRET_KEY="example"
+export TOKEN_SESSION_KEY="example"
+export AUTHBROKER_URL="example"
+export AUTHBROKER_CLIENT_ID="example"
+export AUTHBROKER_CLIENT_SECRET="example"
+export AWS_STORAGE_BUCKET_NAME="example"
+export LITE_API_URL="example"
+export PERMISSIONS_FINDER_URL="example"
+export NOTIFY_FEEDBACK_TEMPLATE_ID="example"
+export NOTIFY_FEEDBACK_EMAIL="example"
+export FEEDBACK_URL="example"
+export INTERNAL_FRONTEND_URL="example"
+export LITE_EXPORTER_HAWK_KEY="example"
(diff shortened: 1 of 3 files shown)


depsiatwal (approved):
lgtm

.copilot/image_build_run.sh
```diff
@@ -1,3 +1,26 @@
...
+export LITE_EXPORTER_HAWK_KEY="example"
+export COPILOT_ENVIRONMENT_NAME="example"
+export LITE_INTERNAL_HAWK_KEY="example"
+
+declare -a settings_modules=(conf.exporter conf.caseworker)
+
+for setting_module in "${settings_modules[@]}"; do
+  export DJANGO_SETTINGS_MODULE=$setting_module
+  python manage.py collectstatic --ignore=*.scss,*.md,*.txt,*.json,LICENSE,license,CHANGES,changes
+done
```
currycoder, line 26: For my understanding, this now means that `collectstatic` has run and is part of the baked in app image?

emileswarts: Yeah it's either done at boot time or at build time, neither are ideal, but building it into the app image is the preferred option out of the two. A longer term solution is to ditch Whitenoise and to serve assets from S3.

Would reduce strain on your application servers as well.

conf/base.py
```diff
@@ -161,7 +160,10 @@
 
 LOGIN_URL = reverse_lazy("auth:login")
 
-DATA_DIR = os.path.dirname(BASE_DIR)
+if IS_ENV_DBT_PLATFORM:
+    DATA_DIR = BASE_DIR
```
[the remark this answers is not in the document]
emileswarts: The location for static files on DBT Platform will be:
/workspace/exporter/staticfiles
/workspace/caseworker/staticfiles

The change above stops from writing these static assets to /workspace/staticfiles

currycoder (approved):

PR #2263, opened 2024-12-05
Ltd 5714 other ogd sessionwizard

Tllew:
All other ogds can use sessionwizard, just need to address e2e tests and maybe bdd

[LTD-5714](https://uktrade.atlassian.net/browse/LTD-714)

### DIFF
--- a/caseworker/advice/conditionals.py
+++ b/caseworker/advice/conditionals.py
@@ -13,11 +13,3 @@ def form_add_licence_conditions(step_name):
 
 def is_desnz_team(wizard):
     return wizard.caseworker["team"]["alias"] in services.DESNZ_TEAMS
-
-
-def is_fcdo_team(wizard):
-    return wizard.caseworker["team"]["alias"] == services.FCDO_TEAM
-
-
-def default_form(wizard):
-    return not (is_fcdo_team(wizard) or is_desnz_team(wizard))
--- a/caseworker/advice/conditionals.py
+++ b/caseworker/advice/conditionals.py
@@ -11,5 +11,9 @@ def form_add_licence_conditions(step_name):
     return _get_form_field_boolean
 
 
-def is_desnz_team(wizard):
-    return wizard.caseworker["team"]["alias"] in services.DESNZ_TEAMS
+def is_fcdo_team(wizard):
+    return wizard.caseworker["team"]["alias"] == services.FCDO_TEAM
+
+
+def is_ogd_team(wizard):
+    return not is_fcdo_team(wizard)
--- a/caseworker/advice/forms/approval.py
+++ b/caseworker/advice/forms/approval.py
@@ -10,7 +10,6 @@ from core.forms.layouts import (
     ConditionalCheckboxesQuestion,
     RadioTextArea,
 )
-from core.forms.widgets import GridmultipleSelect
 
 
 class SelectAdviceForm(forms.Form):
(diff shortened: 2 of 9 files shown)


caseworker/advice/forms/approval.py
```diff
@@ -56,7 +56,7 @@ class Layout:
         TITLE = "Recommend an approval"
 
     approval_reasons = forms.CharField(
-        widget=forms.Textarea(attrs={"rows": 7, "class": "govuk-!-margin-top-4"}),
+        widget=forms.Textarea(attrs={"rows": 7, "class": "govuk-!-margin-top-4", "name": "approval_reasons"}),
```
currycoder, line 59: I'm surprised this `"name": "approval_reasons"` is necessary.  My understanding is that django injects that for you when this is rendered in the template.

caseworker/advice/views/approval.py
```diff
@@ -28,7 +47,7 @@ class GiveApprovalAdviceView(LoginRequiredMixin, CaseContextMixin, BaseSessionWi
     ]
 
     condition_dict = {
-        AdviceSteps.RECOMMEND_APPROVAL: C(is_desnz_team),
+        AdviceSteps.RECOMMEND_APPROVAL: C(is_ogd_team),
```
currycoder, line 50: I still don't get why this step is conditional.  Once you're in the new flow, surely this step **must** show.  We are handling whether or not to take users to the new or legacy flow outside of this view

currycoder (commented):
A few minor suggestions throughout to clean this up a bit.  Otherwise looking great

caseworker/advice/conditionals.py
```diff
@@ -11,5 +11,9 @@ def _get_form_field_boolean(wizard):
...
 
 
-def is_desnz_team(wizard):
-    return wizard.caseworker["team"]["alias"] in services.DESNZ_TEAMS
+def is_fcdo_team(wizard):
+    return wizard.caseworker["team"]["alias"] == services.FCDO_TEAM
+
+
+def is_ogd_team(wizard):
+    return not is_fcdo_team(wizard)
```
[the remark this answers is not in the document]
Tllew: because we're still excluding fcdo in this pr, i've changed the logic to be : ~C(is_fcdo_team)
and removed is_ogd_team to make this more obvious

caseworker/advice/forms/approval.py
Tllew: I believe this was me trying to force remove the form index from the approval reason, as the name here is formatted: formname-approval_reasons

ultimately this didn't work and I needed to change the ui tests to point to "recommend_approval-approval_reasons"

this is something the formwizard inserts so that you can have multiple fields with the same name in a single formwizard.

caseworker/advice/templates/advice/view_my_advice.html
```diff
@@ -34,10 +34,10 @@ <h2 class="govuk-error-summary__title" id="error-summary-title">
             <div class="govuk-grid-column-three-quarters">
               <h1 class="govuk-heading-xl">View recommendation</h1>
               {% if my_advice %}
-                {% test_rule 'can_user_make_desnz_edit' request case as can_user_make_desnz_edit %}
+                {% test_rule 'can_user_make_edit' request case as can_user_make_edit %}
 
                 {% if buttons.edit_recommendation %}
-                  {% if can_user_make_desnz_edit %}
+                  {% if can_user_make_edit %}
```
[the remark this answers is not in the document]
Tllew: I've looked into this a little, I do agree we could remove the `if buttons.` logic however there's some additional logic behind the edit_recommendation that I'm not confident we'd cover with the logic we have in the template, it seems like it could be a little out of scope at this stage in the LCDC project

caseworker/advice/views/approval.py
Tllew: this is still excluding fcdo to ensure if you were to manually navigate to the page, you'd see an error.

tests_common/constants.py
```diff
@@ -0,0 +1,5 @@
+class WebDriverDelay:
+    THIRTY = 30
+    FORTYFIVE = 45
+    TWENTY = 20
+    SIXTY = 60
```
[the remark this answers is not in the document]
kevincarrogan: What I was thinking was having a multiplier to increase these values (or even set it to 0) and that could be in the env.

So maybe a setting called something like `E2E_WAIT_MULTIPLIER` and change those constants so they are all `SIXTY = 60 * E2E_MULTIPLIER`.

caseworker/advice/forms/approval.py
currycoder: Ah yep, we have a helper for this kind of thing in the unit tests `post_to_step` maybe some patterns could be lifted for UI tests?

caseworker/advice/templates/advice/view_my_advice.html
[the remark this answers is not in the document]
currycoder: Happy for this to be something we come back to later - it sounds tricky to unpick currently.

currycoder (approved):

kevincarrogan (approved):

PR #2264, opened 2024-12-06
[LTD-5708] Pre-fill LU licence conditions

currycoder:
### Aim

LU and MOD-ECJU give advice through a "Review and combine" consolidation flow.  This means that they are collating advice for a number of teams at once (all OGDs for LU/all MOD teams for MOD-ECJU).

This change makes it so that the licence conditions textarea for the consolidation flow is pre-populated from conditions across advice from all "feeder" teams.

In terms of mechanical changes, this includes:
- Setting the `proviso` initial value as all feeder teams' provisos concatenated together.
- Retaining the ability for the consolidating team to add picklist values to the condition with a new `CannedSnippetTextArea` component.

[LTD-5708](https://uktrade.atlassian.net/browse/LTD-5708)

[LTD-5708]: https://uktrade.atlassian.net/browse/LTD-5708?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


caseworker/advice/forms.py
```diff
@@ -184,16 +186,46 @@ def __init__(self, *args, **kwargs):
         )
 
 
-class ConsolidateApprovalForm(GiveApprovalAdviceForm):
-    """Approval form minus some fields."""
+class ConsolidateApprovalForm(PicklistAdviceForm):
```
currycoder, line 189: Stop just inheriting from a form which has many more fields than needed and be explicit.

caseworker/advice/views/consolidate_advice.py
```diff
@@ -82,9 +86,17 @@ class ConsolidateApproveView(BaseConsolidationView):
...
+        """
+        Collate all provisos across all team advice in to a single string.
+        """
+        # Should be a set, but dict gives us consistent ordering
+        unique_provisos = {}
+        for team_advice in self.advice_to_consolidate:
+            for advice in team_advice:
+                if advice["proviso"]:
+                    unique_provisos[advice["proviso"]] = None
+        return "\n\n--------\n".join(unique_provisos.keys())
```
kevincarrogan, line 99: I can't quite work out how I feel about this. It makes sense but is slightly unintuitive.

https://pypi.org/project/orderedset/ or https://pypi.org/project/ordered-set/ maybe instead?

Where is the original ordering coming from and why does it need to be consistent out of interest?

caseworker/assets/javascripts/main.js
```diff
@@ -4,6 +4,7 @@ import { initAll } from "govuk-frontend";
 // our JS
 import "../../../core/assets/javascripts/definitions.js";
 import "../../../core/assets/javascripts/back-link.js";
+import { initCannedSnippetsTextArea } from "../../../core/assets/javascripts/canned-snippets-textarea.js";
 import initCookierBanner from "../../../core/assets/javascripts/cookies/banner.js";
 import { gaPushUserID } from "../../../core/assets/javascripts/ga-events.js";
```
kevincarrogan, line 9: All this `../../../` is getting a bit naff. I feel we could probably turn `core` into an alias somewhere.

core/assets/javascripts/__tests__/canned-snippets-textarea.test.js
```diff
@@ -0,0 +1,52 @@
...
+
+    const $textarea = document.querySelector("textarea");
+    const $div = document.querySelector("div");
+    const component = createComponent($div);
+
+    component.init();
+    const addSnippet1 = document.querySelector("a[data-snippet-key=snippet_1]");
+    const addSnippet2 = document.querySelector("a[data-snippet-key=snippet_2]");
+    await userEvent.click(addSnippet1);
+    expect($textarea.value).toBe(
```
kevincarrogan, line 35: We could use some of js-doms matchers here.

https://github.com/testing-library/jest-dom?tab=readme-ov-file#tohavedisplayvalue

core/forms/templates/gds/layout/canned_snippets.html
```diff
@@ -0,0 +1,22 @@
...
+      {% if legend_tag %}</{{ legend_tag }}>{% endif %}
+    </legend>
+  {% endif %}
+
+  {% include 'gds/layout/help_text_and_errors.html' %}
+
+  <div>
+    {% for choice in field.field.choices %}
+        <p>
+            {{ choice.1|unlocalize }} <a href="#" class="govuk-link govuk-link--no-visited-state" data-snippet-key="{{ choice.0|unlocalize }}">{{add_label}}</a>
```
kevincarrogan, line 17: These should be buttons instead of anchors as this isn't currently accessible.

caseworker/advice/views/consolidate_advice.py
currycoder: Yep I understand the weird feeling about this.  TBH this was a rough corner I haven't managed to round off so well from the prototype.
So the ordering consistency is to make it so that the conditions are sequenced in the same order as the summaries of team advice which appear above the form on the page.

This one looks recently updated; https://pypi.org/project/ordered-set/ will take a look

currycoder: Done

core/assets/javascripts/__tests__/canned-snippets-textarea.test.js
currycoder: Done

core/forms/templates/gds/layout/canned_snippets.html
currycoder: Done

kevincarrogan (approved):

saruniitr (approved):
looks good to me.
Noticed one issue when testing locally, I will add details to the thread in internal channel.

PR #2276, opened 2024-12-18
LTD-5734-fix-serial-number-update-error 

depsiatwal:
https://uktrade.atlassian.net/browse/LTD-5734

This Application had 1500 items for the same good entry. 
So when the exporter submits a serial number for 1500 entries the middleware throws an error 

"The number of GET/POST parameters exceeded settings.DATA_UPLOAD_MAX_NUMBER_FIELDS."

Hence have increased the amount so that exporter / caseworker are the same.


conf/base.py
```diff
@@ -19,6 +19,11 @@
 
 env = Env()
 
+# The maximum number of parameters that may be received via GET or POST
+# before a SuspiciousOperation (TooManyFields) is raised.
+# Increased due to potential of selecting all control list entries/ serial numbers for exporter
+DATA_UPLOAD_MAX_NUMBER_FIELDS = 3500
+
```
saruniitr, line 26: Only exporter provides these values so may be move it to `exporter.py`?

depsiatwal: no it was actually in caseworker originally hence why it's been moved to base.

kevincarrogan (approved):

conf/base.py
saruniitr: ok, just wondering if we need this in base instead of just in exporter.

saruniitr (approved):

PR #2278, opened 2024-12-18
[LTD-5728] Prevent 'other' radio showing as only option for advice picklist fields

currycoder:
### Aim

This change ensures that the "Other" option is suppressed when there are no picklist options to choose from for picklist fields in the advice flows.  In this situation, rather than seeing the "Other" radio input and an empty textbox, the user just sees the textbox.  This should be much clearer.

[LTD-5728](https://uktrade.atlassian.net/browse/LTD-5728)

[LTD-5728]: https://uktrade.atlassian.net/browse/LTD-5728?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


caseworker/advice/forms/approval.py
```diff
@@ -93,16 +77,10 @@ class Layout:
...
+        label="Licence condition",
         required=False,
     )
 
-    approval_radios = forms.ChoiceField(
-        label="What is your reason for approving?",
-        required=False,
-        widget=forms.RadioSelect,
-        choices=(),
-    )
```
saruniitr, line 105: oh this was included in this form by mistake before?

currycoder: Yes I believe so.  We do not need it at this stage in the wizard

saruniitr (approved):

caseworker/advice/forms/approval.py
```diff
@@ -137,10 +108,23 @@ def __init__(self, *args, **kwargs):
...
-
         return (ConditionalCheckboxes("proviso_checkboxes", *self.conditional_checkbox_choices),)
 
 
+class SimpleLicenceConditionsForm(BaseForm):
+    class Layout:
+        TITLE = "Add licence conditions (optional)"
+
+    proviso = forms.CharField(
+        widget=forms.Textarea(attrs={"rows": 7, "class": "govuk-!-margin-top-4"}),
```
kevincarrogan, line 119: I feel that I keep seeing this `govuk-!-margin-top-4` in forms. I wonder if this is a sign that we need a class for this somewhere.

Why do we keep adding it?

kevincarrogan (approved):

depsiatwal (approved):
all looks good to me

caseworker/advice/forms/approval.py
currycoder: You're right.  I've binned it all - it looks better without.

PR #2294, opened 2025-01-10
LTD-5769: Add option to approve multiple cases

saruniitr:
## Change description

Add button on cases queue view to approve multiple cases.
This is enabled selectively on specific teams queues and there is a rule to enforce this.

A new endpoint is added for this purpose and the relevant PR is https://github.com/uktrade/lite-api/pull/2362

[LTD-5769]: https://uktrade.atlassian.net/browse/LTD-5769?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


caseworker/advice/views/bulk_approval.py
```diff
@@ -0,0 +1,72 @@
...
+
+class BulkApprovalForm(forms.Form):
+    pass
+
+
+class BulkApprovalView(LoginRequiredMixin, SuccessMessageMixin, FormView):
+    """
+    Submit approval recommendation for the selected cases
+    """
+
```
depsiatwal, line 25: To protect the view I think we need to test user  permission here also similar to 

https://github.com/uktrade/lite-frontend/blob/f20b5fd9394b1a356328999d35d7be219f5fbe12/caseworker/cases/views/main.py#L515

depsiatwal (commented):
LGTM also assume can't be merged until API has been merged.

caseworker/advice/views/bulk_approval.py
```diff
@@ -0,0 +1,72 @@
...
+        HTTPStatus.CREATED,
+        "Error submitting bulk approval recommendation",
+        "Unexpected error submitting bulk approval recommendation",
+    )
+    def submit_bulk_approval_recommendation(self, queue_id, payload):
+        return post_bulk_approval_recommendation(self.request, queue_id, payload)
+
+    def form_valid(self, form):
+        queue_id = self.kwargs["pk"]
+        cases = self.request.POST.getlist("cases", [])
```
currycoder, line 51: Could it be possible to use the form validation and get this value from the form instead?  e.g. https://docs.djangoproject.com/en/5.1/ref/forms/fields/#multiplechoicefield

caseworker/advice/services.py
```diff
@@ -569,6 +569,12 @@ def unassessed_trigger_list_goods(case):
     ]
 
 
+def post_bulk_approval_recommendation(request, queue_id, data):
+    response = client.post(request, f"/caseworker/queues/{queue_id}/bulk-approval/", data)
+    response.raise_for_status()
```
currycoder, line 574: I think that with the `expect_status` decorator you are using below, this is probably unnecessary.  If a non-200 response came back this would raise and the nice error from `expect_status` would have no chance to return.

caseworker/queues/rules.py
```diff
@@ -0,0 +1,16 @@
+import rules
+
+BULK_APPROVE_ALLOWED_QUEUES = {
+    "MOD_CAPPROT": "93d1bc19-979d-4ba3-a57c-b0ce253c6237",  # /PS-IGNORE
+    "MOD_DI_INDIRECT": "0dd6c6f0-8f8b-4c03-b68f-0d8b04225369",  # /PS-IGNORE
+    "MOD_DSR": "a84d6556-782e-4002-abe2-8bc1e5c2b162",  # /PS-IGNORE
+    "MOD_DSTL": "1a5f47ee-ef5e-456b-914c-4fa629b4559c",  # /PS-IGNORE
+}
```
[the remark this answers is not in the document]
saruniitr: As we have consistent ids across I am trying to use id instead of alias.
Yeah I can add these as constants but will update advice code in a separate PR.

caseworker/advice/views/bulk_approval.py
saruniitr: I will have a look if this can be taken from the form itself.

caseworker/advice/services.py
saruniitr: ok, I will remove this.

caseworker/advice/views/bulk_approval.py
saruniitr: @currycoder I tried this and not able to get this working, not sure it is possible because the form is specified in the template using <form> and the button simply overrides form method and action in the template.
I tried defining multiple choice field for the empty form that I defined in the code but upon clicking the button it is taking to a page with a list of checkboxes.
Ok for this to be kept as is?

currycoder: I do think this aspect could be better..

Ideally this view would only handle the POST - right now it inherits from something which allows it to set a template and requires an empty form to be set.  If we aren't using the form or the template maybe a simpler view would be better?

In the interest of getting this through, this could always be a follow-up?

saruniitr: Actually `TemplateView` is working now and can see message banner. I probably missed something when I tried before.

currycoder: I think this should maybe just inherit from `django.views.View` instead - there should be no GET for this whatsoever right?  So the `forms.html` template will never render..

saruniitr: ok understood. Updated to just use `View` and also updated unit tests to check for messages.

currycoder (approved):
Nice one!

caseworker/advice/views/bulk_approval.py
currycoder: Great looks good - thanks for your patience!

caseworker/advice/views/bulk_approval.py
depsiatwal: You'll probably need a test for this now similar to 
https://github.com/uktrade/lite-frontend/blob/4219b25739f3f96671428f0f6ec5356137ce4d1f/unit_tests/caseworker/organisations/users/test_user_views.py#L260

depsiatwal (commented):
just require a view test

depsiatwal (approved):
Test include so LGTM

PR #2298, opened 2025-01-10
Update fieldsets to group related elements

hnryjmes:
### Aim

The forms listed in the ticket have been updated to make use of the Fieldset component: https://design-system.service.gov.uk/components/fieldset/ via the Fieldset component in Crispy Forms GDS: https://crispy-forms-gds.readthedocs.io/en/latest/components/fieldset.html

A Fieldset component with a Legend serving as the title is considered an accessible way to structure single question forms; see the ticket and link above for more information.

Using DevTools you can see that there is now a `<fieldset>` element which wraps the form inputs and is given a name through the `<legend>` element, this means that the accessibility requirements for WCAG to label form inputs and to structure the page to make relations clear to the user can both be met.

Mostly this is done through the use of a new class `FieldsetForm` that subclasses `BaseForm`. By having a separate class it means that we can easily roll out this new style across LITE on an individual form basis rather than changing everything that uses `BaseForm` at once.

[LTD-5170](https://uktrade.atlassian.net/browse/LTD-5170)

[LTD-5170]: https://uktrade.atlassian.net/browse/LTD-5170?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/exporter/applications/views/goods/goods.py
+++ b/exporter/applications/views/goods/goods.py
@@ -354,7 +354,10 @@ class AddGood(LoginRequiredMixin, BaseSessionWizardView):
 
     def get_context_data(self, form, **kwargs):
         context = super().get_context_data(form, **kwargs)
-        context["title"] = form.title
+        try:
+            context["title"] = form.Layout.TITLE
+        except AttributeError:
+            context["title"] = form.title
         # The back_link_url is used for the first form in the sequence. For subsequent forms,
         # the wizard automatically generates the back link to the previous form.
         context["back_link_url"] = reverse("applications:goods", kwargs={"pk": self.kwargs["pk"]})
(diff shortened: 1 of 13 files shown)


currycoder (approved):
Tested these forms locally and this seems all good

exporter/applications/views/goods/goods.py
```diff
@@ -354,7 +354,10 @@ def get_product_type(self):
 
     def get_context_data(self, form, **kwargs):
         context = super().get_context_data(form, **kwargs)
-        context["title"] = form.title
+        try:
+            context["title"] = form.Layout.TITLE
```
depsiatwal, line 358: why not just check if 
if hasattr(form.Layout, "TITLE")

probably more consistent with how we check in other areas

hnryjmes: updated

PR #2305, opened 2025-01-15
 LTD-5171-grouping-date-components

depsiatwal:
https://uktrade.atlassian.net/browse/LTD-5171

This is about grouping correctly of date fields for accessibility this shift the lite forms template to be compliant


exporter/applications/forms/export_details.py
```diff
@@ -135,7 +135,7 @@ def proposed_product_return_date_form():
         title=TemporaryExportDetails.PROPOSED_RETURN_DATE,
         questions=[
             DateInput(
-                title="",
+                title="Return to the UK Date",
```
depsiatwal, line 138: Title is now displayed as per template change this is the only date component that didn't have title set.

lite_forms/templates/components.html
```diff
@@ -7,7 +7,7 @@
 						govuk-form-group--error
 					{% endif %}
 			{% endif %}">
-	{% if question.title %}
+	{% if question.title and question.input_type != "date" %}
```
depsiatwal, line 10: removing from here as it's handled in the date template for group for accessibility

lite_forms/templates/components.html
```diff
@@ -17,7 +17,7 @@
 			{% endif %}
 		</label>
 	{% endif %}
-	{% if question.description %}
+	{% if question.description and question.input_type != "date" %}
```
depsiatwal, line 20: removing from here as it's handled in the date template for group for accessibility

lite_forms/templates/components/date.html
```diff
@@ -1,57 +1,65 @@
...
-						 
-				/>
-            </div>
-        </div>
-        <div class="govuk-date-input__item">
-            <div class="govuk-form-group">
-                <label class="govuk-label govuk-date-input__label {% if component.inline_title %}govuk-visually-hidden{% endif %}" for="{{ component.prefix }}month">
-                    Month
-                </label>
+<fieldset class="govuk-fieldset" role="group" aria-describedby="date_of_issue-hint">
```
kevincarrogan, line 1: I would guess that the `date_of_issue-hint` isn't always going to be the correct id here as this is a generic template.

lite_forms/templates/components/date.html
```diff
@@ -1,57 +1,65 @@
...
+	<legend class="govuk-fieldset__legend govuk-fieldset__legend–l">
+		<h1 class="govuk-fieldset__heading">
+			{{ component.title }}
+		</h1>
+	</legend>
+	<div id="return-date-hint" class="govuk-hint">
+		{{ component.description }}
+	</div>
+	<div class="govuk-date-input" id="return-date">
```
kevincarrogan, line 10: This refers to `return-date` and `return-date-hint` but is this always related to a return date?

lite_forms/templates/components.html
currycoder: Might be good to comment that this is why `input_type == date` is a special case in the template itself too.

currycoder (approved):
Tried this out locally and it seems fine - approved pending Kevin's comments

lite_forms/templates/components/date.html
depsiatwal: Yeah it's just I noticed that ID in the orginal template also and didn't want to change it in case something depends on it. 
https://github.com/uktrade/lite-frontend/blob/dev/lite_forms/templates/components/date.html#L2

lite_forms/templates/components.html
depsiatwal: Done

lite_forms/templates/components/date.html
depsiatwal: made more generic by adding name of the component

lite_forms/templates/components/date.html
depsiatwal: made more generic by adding name of the component

kevincarrogan (approved):

PR #2318, opened 2025-01-17
LTD-5822 remove end user from list of optional case sections

markj0hnst0n:
### Aim

End user shows as ‘Optional' but this has been made mandatory so UI should reflect this.  Should be 'Not started’ as per other mandatory sections as per ticket.

[LTD-5822](https://uktrade.atlassian.net/browse/LTD-5822)

[LTD-5822]: https://uktrade.atlassian.net/browse/LTD-5822?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


currycoder (commented):
This looks like it's bringing in loads of stuff from `dev` branch too.  Was there a need for that?  I think if we want to do that we should do it separately from a fixup PR like this.

markj0hnst0n:
Not quite sure how this happened but I've created a new PR for this one @currycoder

PR #2322, opened 2025-01-21
LTD-5812-file-upload-validation

depsiatwal:
https://uktrade.atlassian.net/browse/LTD-5812

The middleware mime check currently disables the form validation, as the error happens during the request, 

This has been disabled so that the form can handle the request and check the filetype. However this may present some issues as the file will still be uploaded and no mime check is bring done.


core/file_handler.py
```diff
@@ -47,24 +46,9 @@ def s3_client():
 
 class SafeS3FileUploadHandler(S3FileUploadHandler):
     """
-    S3FileUploadHandler with mime-type validation.
+    S3FileUploadHandler.
     """
 
-    ACCEPTED_FILE_UPLOAD_MIME_TYPES = settings.ACCEPTED_FILE_UPLOAD_MIME_TYPES
-
-    def receive_data_chunk(self, raw_data, start):
```
depsiatwal, line 55: debating if this is needed ?

If we want to absolutely make sure this is called we could replace all of this with a custom error message.

kevincarrogan:
What happens if you upload a file with a valid extension but an invalid mimetype?

core/file_handler.py
```diff
@@ -61,8 +61,9 @@ def receive_data_chunk(self, raw_data, start):
         if start == 0:
             mime = magic.from_buffer(raw_data, mime=True)
             if mime not in self.ACCEPTED_FILE_UPLOAD_MIME_TYPES:
-                self.abort()
-                raise UnacceptableMimeTypeError(f"Unsupported file type: {mime}")
+                # Return non this halts the upload and lets the Filefield handle the
+                # validation for user messages.
+                return
```
kevincarrogan, line 66: Does this really abort the file upload? Do you potentially still need to call `self.abort()`?

exporter/goods/forms/common.py
```diff
@@ -363,6 +363,8 @@ class Layout:
         error_messages={
             "required": "Select a document that shows what your product is designed to do",
         },
+        allow_empty_file=True,
+        validators=[FileExtensionValidator()],
```
kevincarrogan, line 367: Why are we now allowing an empty file everywhere?

depsiatwal: Our file handler returns none now when we upload a unsupported mimetype this halts uploading the file. 
The Filefield has validation in place and if it see's a file that none it raises Emtyfile validation which overrides all the existing validators and report 'Empty file error'

depsiatwal: however this means a 0KB file can be uploaded 
i'll see if there a way to prevent this

core/file_handler.py
depsiatwal: I was going by this 
https://github.com/django/django/blob/main/django/http/multipartparser.py#L330

https://docs.djangoproject.com/en/5.1/ref/files/uploads/

core/file_handler.py
```diff
@@ -62,7 +62,10 @@ def receive_data_chunk(self, raw_data, start):
             mime = magic.from_buffer(raw_data, mime=True)
             if mime not in self.ACCEPTED_FILE_UPLOAD_MIME_TYPES:
                 self.abort()
-                raise UnacceptableMimeTypeError(f"Unsupported file type: {mime}")
+                # Raise StopUpload signals the django multipartparser to exhaust the stream and stops the upload
+                exception = UnacceptableMimeTypeError(f"Unsupported file type: {mime}")
```
depsiatwal, line 66: StopUpload halts the upload which is good practice as it prevents non supported file types being uploaded to S3. 

As a result the calling form will not receive the file and sees an empty FileInput.

core/file_handler.py
```diff
@@ -62,7 +62,10 @@ def receive_data_chunk(self, raw_data, start):
...
+                logger.error(exception)
```
depsiatwal, line 67: We now log the error since the django multipartparser will swallow this exception.

PR #2327, opened 2025-01-21
LTD-5823-cta-notes-additional-docs 

depsiatwal:
https://uktrade.atlassian.net/browse/LTD-5823

CTA added as per ticket

### DIFF
--- a/exporter/templates/applications/additional-documents/additional-documents.html
+++ b/exporter/templates/applications/additional-documents/additional-documents.html
@@ -12,7 +12,7 @@
 			<h1 class="govuk-heading-l">{% block title %}{% lcs 'AdditionalDocuments.TITLE'%}{% endblock %}</h1>
 		</div>
 		<div class="lite-app-bar__controls">
-			<a href="{% url 'applications:attach_additional_document' application_id %}" class="govuk-button" draggable="false" role="button">
+			<a href="{% url 'applications:attach_additional_document' application_id %}" class="govuk-button govuk-button--secondary" draggable="false" role="button">
 				{% lcs 'AdditionalDocuments.ADD' %}
 			</a>
 		</div>
@@ -65,5 +65,9 @@
 				{% lcs 'AdditionalDocuments.NO_RESULTS' %}
 			</p>
 		</div>
+		<div class="lite-information-text__help"></div>
 	{% endif %}
+	<a class="govuk-button govuk-button--primary" href="{% url 'applications:task_list' application_id %}">
+        Save and continue
+    </a>
 {% endblock %}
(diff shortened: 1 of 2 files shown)


exporter/templates/applications/additional-documents/additional-documents.html
```diff
@@ -66,4 +66,7 @@ <h1 class="govuk-heading-l">{% block title %}{% lcs 'AdditionalDocuments.TITLE'%
 			</p>
 		</div>
 	{% endif %}
+	<a class="govuk-button govuk-button--primary" href="{% url 'applications:task_list' application_id %}">
```
currycoder, line 69: The content here is "save and continue" but it appears to be taking the user back to their task list.  Has this already been saved at this point..?

depsiatwal: I did wonder about this point as well, 

i'll put a note on the ticket and ask what the expectation is

depsiatwal: Concluded looks like UCD want to keep as "save and continue"

https://uktrade.atlassian.net/browse/LTD-5823

currycoder: Nice one.  Thanks for confirming that

currycoder (approved):

exporter/templates/applications/additional-documents/additional-documents.html
```diff
@@ -66,4 +66,7 @@ <h1 class="govuk-heading-l">{% block title %}{% lcs 'AdditionalDocuments.TITLE'%
 			</p>
 		</div>
```
hnryjmes, line 67: Screenshot 2025-01-23 at 10 20 25


the spacing doesn't look right when there are no additional documents added. I had a similar problem in https://github.com/uktrade/lite-frontend/pull/2320/ which I fixed by making an element appear with the css class `lite-information-text__help` as that has the margin bottom set which gives the right spacing for the button.

locally because I've got this branch pulled down I just tried adding an empty div `<div class="lite-information-text__help"></div>` right below the div with `lite-information-text` on but this is a bit hacky and you might be able to think of a more elegant way to do this.

hnryjmes (commented):
I can't make a line comment for some reason but at line 15 where the button is defined this should now be made grey as there should only be 1 CTA on the page. so where the class says `govuk-button` you can add `govuk-button--secondary` and that should do it.

https://github.com/uktrade/lite-frontend/pull/2327/files#diff-8f194965ed2d91978e538af3279b7d72d163fcb853862d1bc4c6e4606bac35e1R15

exporter/templates/applications/additional-documents/additional-documents.html
depsiatwal: fixed

hnryjmes (approved):
looks good

PR #2338, opened 2025-01-24
Add validation for invalid mime type file uploads

kevincarrogan:
### Aim


depsiatwal (commented):
Ignore this it may have been part of my testing as I had a O kb file  with .ods extension 
which I suppose will ok, had the wrong filetype but as it's 0KB it's not valid anyway

kevincarrogan:
> Ignore this it may have been part of my testing as I had a O kb file with .ods extension which I suppose will ok, had the wrong filetype but as it's 0KB it's not valid anyway

image

This is what you should see with an incorrect mimetype.

depsiatwal (approved):
All looks good to me and tested locally. 

I'm approving as I only add some additional tests.

Tllew (approved):
This is nice, still uses the file_handler, looks like you got around the error by returning an error message in file_complete?

PR #2345, opened 2025-01-30
LTD-5856 add F680 feature flag and initial exporter form flow framework

markj0hnst0n:
### Aim

On the exporter application:

-  To add a feature flag to enable us to display the F680 application path in the initial questions on hitting 'apply'
-  To build a reusable flow for questions and testing framework for the questions using form wizard
-  To show the initial application summary template with submission form (not currently hooked up to submit yet)

[LTD-5856](https://uktrade.atlassian.net/browse/LTD-5856)

[LTD-5856]: https://uktrade.atlassian.net/browse/LTD-5856?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


exporter/f680/views.py
```diff
@@ -0,0 +1,73 @@
...
+        )
+
+    def get_payload(self, form_dict):
+        return F680CreatePayloadBuilder().build(form_dict)  # /PS-IGNORE
+
+    def done(self, form_list, form_dict, **kwargs):
+        data = self.get_payload(form_dict)
+        response_data, _ = self.post_f680_application(data)  # PS-IGNORE
+        return redirect(self.get_success_url(response_data["id"]))
+
```
markj0hnst0n, line 55: While there is only one step in this flow it establishes the pattern for the wizard form flows which we will be using for each of the paths on the application both in the view and in the testing pattern.

exporter/f680/views.py
```diff
@@ -0,0 +1,73 @@
...
+        self.application = get_f680_application(request, kwargs["pk"])
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        context["application"] = self.application
+
+        return context
+
+    def get_success_url(self):
+        return reverse("f680:summary", kwargs={"pk": self.application["id"]})
```
markj0hnst0n, line 73: This just redirects to the application summary at the minute but will eventually submit the application

exporter/apply_for_a_licence/views.py
```diff
@@ -69,19 +73,6 @@ def get_success_url(self):
             return reverse_lazy("applications:task_list", kwargs={"pk": pk})
 
 
-class MODClearanceQuestions(LoginRequiredMixin, MultiFormView):
```
Tllew, line 72: was this just removing unused code?

exporter/f680/views.py
Tllew: Is this required before we can merge this PR?

Tllew (commented):
Looks really good just unsure about not being able to submit the application, maybe I'm missing some context

exporter/apply_for_a_licence/views.py
markj0hnst0n: Yeah that's right.  It was the previous F680/security clearance related path.

unit_tests/exporter/applications/views/test_f680s.py
```diff
@@ -0,0 +1,108 @@
...
+    assert "Name of the application" in soup.find("h1").text
+    assert isinstance(response.context["form"], ApplicationNameForm)
+
+    response = post_to_step(
+        ApplicationFormSteps.APPLICATION_NAME,
+        {"name": "F680 Test"},  # PS-IGNORE
+    )
+
+    assert response.status_code == 302
+    assert response.url == f680_summary_url_with_application
```
currycoder, line 82: I'd like for us to get to a place with unit tests where the tests reflect particular actions/conditions/outcomes with the unit under test.  Right now, it's hard for me to know exactly what view is under test here and the test covers a couple of different sequential actions (which means it's not isolated).

I would suggest we have a test class in here for each view class, and test methods within test making a single HTTP request to the URL with a given state and expecting a specific outcome.  Something like:

```
class TestApplicationCreateView:

    def test_GET_displays_form(..):
        response = authorized_client.get(f680_apply_url)  # PS-IGNORE
        assert response.context == {...}

    def test_GET_feature_flag_off_forbidden(..):
        ...
    
    def test_POST_invalid_form_data(..):
         # should POST the form with bad/missing data and assert a validation error response
         ...
    
    def test_POST_success(..):
        # should successfully POST the form and assert that the mocked API endpoint to create the application is called
        ...


class TestF680ApplicationSummaryView:

    def test_GET_application_not_found(..):
        ...
    
    def test_GET_feature_flag_off_forbidden(..):
        ...

    def test_GET_success(..):
        # Should set up a mock response and assert a templated outcome
        ...
```

currycoder (commented):
I think this is the right direction and it's very close, but I'd like to suggest some improvements to feature flagging and testing.

exporter/apply_for_a_licence/views.py
```diff
@@ -55,6 +54,11 @@ def get_success_url(self):
             return reverse_lazy("applications:task_list", kwargs={"pk": pk})
 
 
+class F680Questions(LoginRequiredMixin, RedirectView):  # /PS-IGNORE
+    def get_redirect_url(self, *args, **kwargs):
+        return reverse("f680:apply")  # /PS-IGNORE
```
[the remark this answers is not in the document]
markj0hnst0n: The `LicenceType` view dynamically creates the url path name for this view so it made sense to me just to use that rather than re-write some of the `apply-for-a-licence` form path.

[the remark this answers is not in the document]
currycoder: Ok fair enough

exporter/f680/views.py
markj0hnst0n: It's more just to have a pathway there so that we can use it when the time comes to add a submit.  I just wanted to get the test structure in place before I write any other views.

exporter/apply_for_a_licence/urls.py
```diff
@@ -14,7 +14,11 @@
 if not settings.FEATURE_FLAG_ONLY_ALLOW_SIEL:
     urlpatterns += [
         path("transhipment/", views.TranshipmentQuestions.as_view(), name="transhipment_questions"),
-        path("mod/", views.MODClearanceQuestions.as_view(), name="mod_questions"),
         path("<str:ogl>/", views.OpenGeneralLicenceQuestions.as_view(), name="ogl_questions"),
         path("<str:ogl>/<uuid:pk>/", views.OpenGeneralLicenceSubmit.as_view(), name="ogl_submit"),
     ]
+
+if settings.FEATURE_FLAG_ALLOW_F680:  # /PS-IGNORE
```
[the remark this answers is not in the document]
markj0hnst0n: I've added this in but I think it works slightly different to the example.  I had to explicitly call the `handle_no_permission` method in the dispatch of the form wizard.  I don't think the Mixin has a dispatch method to override.

It's possible I'm not using it right though.

unit_tests/exporter/applications/views/test_f680s.py
markj0hnst0n: Hopefully addressed this now

exporter/f680/services.py
```diff
@@ -0,0 +1,11 @@
...
+
+
+def post_f680_application(request, json):
+    data = client.post(request, "/exporter/f680/application/", json)
+    return data.json(), data.status_code
+
+
+def get_f680_application(request, application_id):
+    data = client.get(request, f"/exporter/f680/application/{application_id}/")
+    return data.json()
```
currycoder, line 11: Can this also return the status_code as above?

exporter/f680/views.py
```diff
@@ -0,0 +1,85 @@
...
+        return redirect(self.get_success_url(response_data["id"]))
+
+
+class F680ApplicationSummaryView(LoginRequiredMixin, F680FeatureRequiredMixin, FormView):  # PS-IGNORE
+    form_class = ApplicationSubmissionForm
+    template_name = "f680/summary.html"  # PS-IGNORE
+
+    def setup(self, request, *args, **kwargs):
+        super().setup(request, *args, **kwargs)
+        self.application = get_f680_application(request, kwargs["pk"])  # PS-IGNORE
```
currycoder, line 76: I think this will break when the F680 doesn't exist in the backend because there's no status code checking.  This endpoint should 404 in that situation rather than 500

unit_tests/exporter/applications/views/test_f680s.py
```diff
@@ -0,0 +1,179 @@
...
+        response = authorized_client.get(f680_summary_url_with_application)  # PS-IGNORE
+
+        assert isinstance(response.context["form"], ApplicationSubmissionForm)
+        assertTemplateUsed(response, "f680/summary.html")  # PS-IGNORE
+
+        content = BeautifulSoup(response.content, "html.parser")
+        heading_element = content.find("h1", class_="govuk-heading-l govuk-!-margin-bottom-2")
+        assert heading_element.string.strip() == "F680 Application"  # PS-IGNORE
+
+    def test_get_f680_summary_view_fail_with_feature_flag_off(
```
currycoder, line 137: There should be an extra test here which simulates the API returning a 404 for the application in question

currycoder (commented):
This is very close, I think there's just one 404 edge case left to cover off

exporter/apply_for_a_licence/forms/triage_questions.py
```diff
@@ -34,6 +34,16 @@ def opening_question():
...
         ),
+        Option(
+            key="f680",  # PS-IGNORE
+            value="Security Approval",
+            description=(
+                "Select if you need approval to give classified products or information to non-UK organisations, "
+                "governments and individuals. This includes F680 approval. You should apply for security approval"  # PS-IGNORE
+                " before you apply for a licence."
+            ),
+            disabled=not settings.FEATURE_FLAG_ALLOW_F680,  # PS-IGNORE
```
saruniitr, line 45: does it read better if it is `settings.FEATURE_FLAG_ALLOW_F680 is False`?

exporter/f680/urls.py
```diff
@@ -0,0 +1,12 @@
...
+
+from . import views
+
+
+app_name = "f680"
+
+urlpatterns = [
+    path("apply/", views.F680ApplicationCreateView.as_view(), name="apply"),  # PS-IGNORE
+    path("<uuid:pk>/apply/", views.F680ApplicationSummaryView.as_view(), name="summary"),  # PS-IGNORE
+    path("<uuid:pk>/apply/", views.F680ApplicationSummaryView.as_view(), name="submit"),  # PS-IGNORE
```
saruniitr, line 11: is this intentional?

markj0hnst0n: No.  It was from a stash of another branch and somehow got in there.  Removed.

exporter/apply_for_a_licence/forms/triage_questions.py
markj0hnst0n: `not settings.FEATURE_FLAG_ALLOW_F680` seemed to work better in my head but I think either is fine.  The default in the settings would have to be changed to True which felt wrong even if I left the name of the flag generic.

unit_tests/exporter/applications/views/test_f680s.py
```diff
@@ -134,6 +142,13 @@ def test_get_f680_summary_view_success(
         heading_element = content.find("h1", class_="govuk-heading-l govuk-!-margin-bottom-2")
         assert heading_element.string.strip() == "F680 Application"  # PS-IGNORE
 
+    def test_get_f680_summary_view_case_not_found(
+        self, authorized_client, set_f680_feature_flag, mock_f680_application_get_404
+    ):
+        response = authorized_client.get(str(uuid4))  # PS-IGNORE
```
currycoder, line 148: well of course this will 404!

exporter/f680/views.py
```diff
@@ -73,7 +73,8 @@ class F680ApplicationSummaryView(LoginRequiredMixin, F680FeatureRequiredMixin, F
 
     def setup(self, request, *args, **kwargs):
         super().setup(request, *args, **kwargs)
-        self.application = get_f680_application(request, kwargs["pk"])  # PS-IGNORE
+        application, _ = get_f680_application(request, kwargs["pk"])
+        self.application = application  # PS-IGNORE
```
currycoder, line 78: What I meant in my comment was, if the API call results in a 404, the exporter user should get a 404.  Otherwise, a non-200 status should raise an immediate 500 for the exporter user.
I think you can solve both with `@expect_status` and it's kwarg `reraise_404` - pretty much the same pattern you already have above with the `post_f680_application` method

markj0hnst0n: Added this in now

currycoder (approved):
Nice work!  Thanks for your patience

exporter/apply_for_a_licence/views.py
```diff
@@ -55,6 +55,11 @@ def get_success_url(self):
             return reverse_lazy("applications:task_list", kwargs={"pk": pk})
 
 
+class F680Questions(LoginRequiredMixin, RedirectView, F680FeatureRequiredMixin):  # /PS-IGNORE
+    def get_redirect_url(self, *args, **kwargs):
+        return reverse("f680:apply")  # /PS-IGNORE
```
saruniitr, line 60: I remember seeing a PR to ignore 'f680' so we don't have to add these 'PS-IGNORE', so can these be removed?

saruniitr (approved):
looks good to me

exporter/apply_for_a_licence/views.py
markj0hnst0n: It's in the next PR which branches off this one but I can remove them now in anticipation of that

PR #2350, opened 2025-02-04
LTD-5877 - Fix add existing end user link

marcuspp:
### Aim

To fix the add existing user to a new application view

[LTD-5877](https://uktrade.atlassian.net/browse/LTD-5877)


exporter/applications/views/parties/end_users.py
```diff
@@ -82,17 +82,17 @@ def get_context_data(self, **kwargs):
             {
                 "form_title": self.form_class.Layout.TITLE,
                 "back_link_text": "Back to application overview",
-                "back_link_url": reverse("applications:task_list", kwargs=self.kwargs),
+                "back_link_url": reverse("applications:task_list", kwargs={"pk": self.kwargs["pk"]}),
```
Tllew, line 85: I might be missing some context here, is it possible to get to this page without a kwargs["pk"]?

marcuspp: No, however it is possible more that the 'pk' is in self.kwargs. Hence being explicit fixes the issue.

Tllew (approved):
Looks good to me

PR #2359, opened 2025-02-12
LTD-5911-f680-drop-name 

depsiatwal:
https://uktrade.atlassian.net/browse/LTD-5911

Remove name from F680 creation

### DIFF
--- a/exporter/f680/views.py
+++ b/exporter/f680/views.py
@@ -4,22 +4,14 @@ from django.conf import settings
 from django.contrib.auth.mixins import AccessMixin
 from django.shortcuts import redirect
 from django.urls import reverse
-from django.views.generic import FormView
+from django.views.generic import FormView, TemplateView
 
 from core.auth.views import LoginRequiredMixin
 from core.decorators import expect_status
-from core.wizard.views import BaseSessionWizardView
 
-from .constants import (
-    ApplicationFormSteps,
-)
-from .forms import (
-    ApplicationNameForm,
-    ApplicationSubmissionForm,
-)
-from .payloads import (
-    F680CreatePayloadBuilder,
-)
+from .forms import ApplicationSubmissionForm
+from .payloads import F680CreatePayloadBase
+
 from .services import (
     post_f680_application,
     get_f680_application,
@@ -37,10 +29,7 @@ class F680FeatureRequiredMixin(AccessMixin):
         return super().dispatch(request, *args, **kwargs)
 
 
-class F680ApplicationCreateView(LoginRequiredMixin, F680FeatureRequiredMixin, BaseSessionWizardView):
-    form_list = [
-        (ApplicationFormSteps.APPLICATION_NAME, ApplicationNameForm),
-    ]
+class F680ApplicationCreateView(LoginRequiredMixin, F680FeatureRequiredMixin, TemplateView):
 
     @expect_status(
         HTTPStatus.CREATED,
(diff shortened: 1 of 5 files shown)


exporter/f680/views.py
```diff
@@ -58,11 +47,10 @@ def get_success_url(self, application_id):
             },
         )
 
-    def get_payload(self, form_dict):
-        return F680CreatePayloadBuilder().build(form_dict)
-
-    def done(self, form_list, form_dict, **kwargs):
-        data = self.get_payload(form_dict)
+    def dispatch(self, request, *args, **kwargs):
```
depsiatwal, line 50: it would be nicer to do the creation on a post but we've lost the post of the form in the liteforms view 
https://github.com/uktrade/lite-frontend/blob/dev/exporter/apply_for_a_licence/views.py#L24

Ideally we should rewrite this hub page to a Django form.

unit_tests/exporter/applications/views/test_f680s.py
```diff
@@ -68,13 +65,13 @@ def test_get_create_f680_view_success(
...
+        mock_application_post,
     ):
         response = authorized_client.get(f680_apply_url)
-
-        assert isinstance(response.context["form"], ApplicationNameForm)
-        soup = BeautifulSoup(response.content, "html.parser")
-        assert "Name of the application" in soup.find("h1").text
+        assert response.status_code == 302
+        assert response.url == f680_summary_url_with_application
+        assert mock_application_post.called_once
```
kevincarrogan, line 74: Can we check that it's actually called with what we expect?

exporter/f680/views.py
kevincarrogan: Then why haven't we?

exporter/f680/views.py
```diff
@@ -58,11 +47,10 @@ def get_success_url(self, application_id):
...
+        # Dispatch is overriden to create a dummpy application
+        super().dispatch(request, *args, **kwargs)
+        data = F680CreateBasePayload().build(application_data={})
         response_data, _ = self.post_f680_application(data)
         return redirect(self.get_success_url(response_data["id"]))
```
kevincarrogan, line 55: Don't override dispatch.

exporter/f680/views.py
```diff
@@ -58,11 +47,10 @@ def get_success_url(self, application_id):
...
 
-    def get_payload(self, form_dict):
-        return F680CreatePayloadBuilder().build(form_dict)
-
-    def done(self, form_list, form_dict, **kwargs):
-        data = self.get_payload(form_dict)
+    def dispatch(self, request, *args, **kwargs):
+        # Dispatch is overriden to create a dummpy application
+        super().dispatch(request, *args, **kwargs)
+        data = F680CreateBasePayload().build(application_data={})
```
kevincarrogan, line 53: You appear to be passing in an empty dictionary to a whole class to pretty much just get back an empty dictionary.

depsiatwal: The ideal here was to encapsulate the logic of creating a base application else where rather then hardcode  within the view. As we require {"Application", {}} as a minimum. 

The other alternative is let the backend handle this if we send an empty dict.

exporter/f680/views.py
depsiatwal: Fair point this would resolve overriding some of the methods as well 
I'll look into transitioning that view.

unit_tests/exporter/applications/views/test_f680s.py
depsiatwal: Yeah i'll change it , was basing it on other tests in this class which lack the insertion on the called parameters , was assuming this may be because this is still in early stages and code changes are quite liquid.

exporter/f680/views.py
kevincarrogan: I'm not suggesting that the backend handles it.

I'm just saying that this is a lot of faff just to send `{"application": {}}`.

currycoder (approved):
Had a go with this locally and looks good. Nice one

markj0hnst0n (approved):
LGTM.  Matches design requirements

PR #2373, opened 2025-02-21
LTD-5936 Enable 'Allocate to me' button for Caseworker on F680 application

markj0hnst0n:
### Aim

WIP

There is a corresponding [backend PR](https://github.com/uktrade/lite-api/pull/2431)

[LTD-5936](https://uktrade.atlassian.net/browse/LTD-5936)

[LTD-5936]: https://uktrade.atlassian.net/browse/LTD-5936?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/caseworker/cases/helpers/case.py
+++ b/caseworker/cases/helpers/case.py
@@ -262,11 +262,13 @@ class CaseView(CaseworkerMixin, TemplateView):
 
 
 def get_case_detail_url(case, queue_id):
-    url_names = {
-        "f680_clearance": "cases:f680:details",
-        "standard": "cases:case",
+    destinations = {
+        "f680_clearance": {"url": "cases:f680:details", "kwargs": {"queue_pk": queue_id, "pk": case.id}},
+        "standard": {"url": "cases:case", "kwargs": {"queue_pk": queue_id, "pk": case.id, "tab": "details"}},
     }
-    case_type = case.case_type["sub_type"]["key"]
-    url_name = url_names[case_type]
 
-    return reverse(url_name, kwargs={"queue_pk": queue_id, "pk": case.id})
+    case_type = case.case_type["sub_type"]["key"]
+    url_name = destinations[case_type]["url"]
+    kwargs = destinations[case_type]["kwargs"]
+
+    return reverse(url_name, kwargs=kwargs)
--- a/caseworker/cases/helpers/case.py
+++ b/caseworker/cases/helpers/case.py
@@ -261,13 +261,13 @@ class CaseView(CaseworkerMixin, TemplateView):
         return render(request, "case/case.html", self.get_context())
 
 
-def get_case_detail_url(case, queue_id):
+def get_case_detail_url(case_id, case_type, queue_id):
     destinations = {
-        "f680_clearance": {"url": "cases:f680:details", "kwargs": {"queue_pk": queue_id, "pk": case.id}},
-        "standard": {"url": "cases:case", "kwargs": {"queue_pk": queue_id, "pk": case.id, "tab": "details"}},
+        "f680_clearance": {"url": "cases:f680:details", "kwargs": {"queue_pk": queue_id, "pk": case_id}},
+        "standard": {"url": "cases:case", "kwargs": {"queue_pk": queue_id, "pk": case_id, "tab": "details"}},
     }
 
-    case_type = case.case_type["sub_type"]["key"]
+    case_type = case_type["sub_type"]["key"]
     url_name = destinations[case_type]["url"]
     kwargs = destinations[case_type]["kwargs"]
 
--- a/caseworker/cases/helpers/case.py
+++ b/caseworker/cases/helpers/case.py
@@ -267,8 +267,6 @@ def get_case_detail_url(case_id, case_type, queue_id):
         "standard": {"url": "cases:case", "kwargs": {"queue_pk": queue_id, "pk": case_id, "tab": "details"}},
     }
 
-    case_type = case_type["sub_type"]["key"]
     url_name = destinations[case_type]["url"]
     kwargs = destinations[case_type]["kwargs"]
-
     return reverse(url_name, kwargs=kwargs)
--- a/caseworker/cases/helpers/case.py
+++ b/caseworker/cases/helpers/case.py
@@ -259,3 +259,14 @@ class CaseView(CaseworkerMixin, TemplateView):
         else:
             getattr(self, "get_" + self.case.sub_type)()
         return render(request, "case/case.html", self.get_context())
+
+
+def get_case_detail_url(case, queue_id):
+    url_names = {
+        "f680_clearance": "cases:f680:details",
+        "standard": "cases:case",
+    }
+    case_type = case.case_type["sub_type"]["key"]
+    url_name = url_names[case_type]
+
+    return reverse(url_name, kwargs={"queue_pk": queue_id, "pk": case.id})
--- a/caseworker/cases/helpers/case.py
+++ b/caseworker/cases/helpers/case.py
@@ -264,7 +264,7 @@ class CaseView(CaseworkerMixin, TemplateView):
 def get_case_detail_url(case, queue_id):
     destinations = {
         "f680_clearance": {"url": "cases:f680:details", "kwargs": {"queue_pk": queue_id, "pk": case.id}},
-        "standard": {"url": "cases:case", "kwargs": {"queue_pk": queue_id, "pk": case.id}},
+        "standard": {"url": "cases:case", "kwargs": {"queue_pk": queue_id, "pk": case.id, "tab": "details"}},
     }
 
     case_type = case.case_type["sub_type"]["key"]
--- a/caseworker/cases/helpers/case.py
+++ b/caseworker/cases/helpers/case.py
@@ -168,9 +168,6 @@ class CaseView(CaseworkerMixin, TemplateView):
 
     def get_destination_countries(self):
         destination_countries = set()
-        application_type = self.case.case_type['sub_type']['key']
-        if application_type == 'f680_clearance':
-            return destination_countries
         all_parties = self.case.data["ultimate_end_users"] + self.case.data["third_parties"]
         if self.case.data["end_user"]:
             all_parties.append(self.case.data["end_user"])
--- a/caseworker/cases/helpers/case.py
+++ b/caseworker/cases/helpers/case.py
@@ -264,7 +264,7 @@ class CaseView(CaseworkerMixin, TemplateView):
 def get_case_detail_url(case, queue_id):
     destinations = {
         "f680_clearance": {"url": "cases:f680:details", "kwargs": {"queue_pk": queue_id, "pk": case.id}},
-        "standard": {"url": "cases:case", "kwargs": {"queue_pk": queue_id, "pk": case.id, "tab": "details"}},
+        "standard": {"url": "cases:case", "kwargs": {"queue_pk": queue_id, "pk": case.id}},
     }
 
     case_type = case.case_type["sub_type"]["key"]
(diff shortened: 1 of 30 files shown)


caseworker/templates/case/base-case-summary.html
```diff
@@ -67,11 +67,14 @@
...
         {% test_rule 'can_user_change_sub_status' request case as can_user_change_sub_status %}
-        {% if can_user_change_sub_status %}
-            <dd class="app-case__summary-list__actions">
-                <a id="link-case-sub-status-change" class="govuk-link govuk-link--no-visited-state" href="{% url 'cases:change_sub_status' queue.id case.id  %}">Change</a>
-            </dd>
+            {% if can_user_change_sub_status %}
+                <dd class="app-case__summary-list__actions">
+                    <a id="link-case-sub-status-change" class="govuk-link govuk-link--no-visited-state" href="{% url 'cases:change_sub_status' queue.id case.id  %}">Change</a>
+                </dd>
+            {% endif %}
```
currycoder, line 77: What problem is this solving?  If I remember correctly the sub-status endpoint was giving a 404 for this view - I would prefer to solve this at the endpoint level if we can.  I believe we can switch the endpoint to look for `BaseApplication` instead of `StandardApplication`

caseworker/queues/views/case_assignments.py
```diff
@@ -66,7 +67,13 @@ def allocate_to_me(self, data):
...
             self.request, data["queue_id"], [data["case_id"]], [data["user_id"]], "Allocated self to the case"
         )
-
+    
+    def get_success_url(self):
+        case = get_case(self.request, self.case_id)
+        if case.case_type['sub_type']['key'] == 'f680_clearance':
+            return reverse("cases:f680:details", kwargs={"queue_pk": self.queue_id, "pk": self.case_id})
+        
+        return reverse("cases:case", kwargs={"queue_pk": self.queue_id, "pk": self.case_id})
```
currycoder, line 76: I've thought about this a little bit..  For now I feel like the pragmatic thing would be to have a little helper function, perhaps `caseworker.cases.helpers.case.get_case_detail_url(case)`

This just returns the correct reversed URL for the case in question e.g.
```
def get_case_detail_url(case, queue_id):
    url_names = {
        "f680": "cases:f680:details",
        "standard": "cases:case",
    }
    case_type = case.case_type["sub_type"]["key"]
    url_name = url_names[case_type]
    return reverse(url_name, kwargs={"queue_pk": queue_id, "pk": case.id})
```

Eventually, we might want to go a little more object-oriented and have a manifest class for each application type like we do on the API.  But I feel like we need a few more examples of this first, so this simplest thing may be better for now

caseworker/queues/forms.py
```diff
@@ -266,4 +266,3 @@ class CaseAssignmentsAllocateToMeForm(forms.Form):
     queue_id = forms.CharField(widget=HiddenInput)
     user_id = forms.CharField(widget=HiddenInput)
     case_id = forms.CharField(widget=HiddenInput)
-    return_to = forms.CharField(widget=HiddenInput)
```
currycoder, line 269: I'm fairly sure that some flows depend on this return_to...  Can we find a way to avoid removing this?

conf/base.py
```diff
@@ -343,4 +343,4 @@ def show_toolbar(request):
 GTM_ID = env.str("GTM_ID", default="")
 
 GIT_COMMIT = env.str("GIT_COMMIT", default="")
-FEATURE_FLAG_ALLOW_F680 = env.bool("FEATURE_FLAG_ALLOW_F680", default=False)
+FEATURE_FLAG_ALLOW_F680 = env.bool("FEATURE_FLAG_ALLOW_F680", default=True)
```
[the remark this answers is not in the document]
markj0hnst0n: fixed it in isolation now so i don't forget!

caseworker/cases/helpers/case.py
```diff
@@ -168,6 +168,9 @@ def get_goods_summary(self):
 
     def get_destination_countries(self):
         destination_countries = set()
+        application_type = self.case.case_type['sub_type']['key']
+        if application_type == 'f680_clearance':
+            return destination_countries
```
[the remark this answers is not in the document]
markj0hnst0n: I've removed this part as it works without it now okay

caseworker/templates/case/base-case-summary.html
markj0hnst0n: fixed in the api side PR

caseworker/queues/views/case_assignments.py
markj0hnst0n: I've used something like this that takes into account the tab kwarg so it fits in with the current tests

exporter/f680/application_sections/conftest.py
```diff
@@ -142,10 +142,8 @@ def mock_f680_application_get_existing_data(requests_mock, data_f680_case):
                         "key": "controlled_info",
                         "answer": "It just is",
                         "raw_answer": "It just is",
-                        "question": (
-                            "Explain how the technology or information is controlled.Include countries classification levels and "
-                            "reference numbers. You can upload supporting documents later in your application"
-                        ),
+                        "question": "Explain how the technology or information is controlled.Include countries classification levels and reference"
+                        "numbers.  You can upload supporting documents later in your application",
```
saruniitr, line 146: The spaces are not consistent.
I think a space required before `Include` and there is an additional space before `You can upload ...`

unit_tests/caseworker/cases/views/test_case_assignments.py
```diff
@@ -98,10 +140,7 @@ def test_case_assignments_POST_remove_user_success(
     response = authorized_client.post(url, data={"assignment_id": str(data_assignment["id"])}, follow=True)
 
     assert response.status_code == 200
-    assert (
-        response.redirect_chain[-1][0]
-        == f"/queues/{data_queue['id']}/cases/{data_standard_case['case']['id']}/details/"
-    )
+    assert response.redirect_chain[-1][0] == f"/queues/{data_queue['id']}/cases/{data_standard_case['case']['id']}/"
     messages = [str(msg) for msg in response.context["messages"]]
```
markj0hnst0n, line 144: This just returns to the default tab now.  Allows use to have a uniform approach when another case type is added

unit_tests/caseworker/cases/views/test_case_assignments.py
```diff
@@ -135,10 +174,7 @@ def test_case_assignments_remove_user_POST_error(
     url = reverse("cases:remove-case-assignment", kwargs={"queue_pk": data_queue["id"], "pk": case["case"]["id"]})
     response = authorized_client.post(url, data={"assignment_id": str(data_assignment["id"])}, follow=True)
     assert response.status_code == 200
-    assert (
-        response.redirect_chain[-1][0]
-        == f"/queues/{data_queue['id']}/cases/{data_standard_case['case']['id']}/details/"
-    )
+    assert response.redirect_chain[-1][0] == f"/queues/{data_queue['id']}/cases/{data_standard_case['case']['id']}/"
```
markj0hnst0n, line 177: As above

caseworker/queues/views/case_assignments.py
```diff
@@ -67,6 +70,12 @@ def allocate_to_me(self, data):
             self.request, data["queue_id"], [data["case_id"]], [data["user_id"]], "Allocated self to the case"
         )
 
+    def get_success_url(self):
+        case = get_case(self.request, self.case_id)
+        url = get_case_detail_url(case, self.queue_id)
+
+        return url if self.success_url != url else self.success_url
```
currycoder, line 77: for readability, can I suggest;
```
if self.success_url:
    return self.success_url
case = get_case(self.request, self.case_id)
return get_case_detail_url(case, self.queue_id)
```

markj0hnst0n: I don't think this works as there will always be a `success_url` so the code below the if is unreachable.  I've done a version like this though

unit_tests/caseworker/cases/views/test_case_assignments.py
```diff
@@ -78,6 +86,40 @@ def test_case_assignments_remove_user_GET_404(authorized_client, data_queue, dat
...
+):
+
+    case = data_submitted_f680_case
+    url = reverse("cases:remove-case-assignment", kwargs={"queue_pk": data_queue["id"], "pk": case["case"]["id"]})
+    response = authorized_client.post(url, data={"assignment_id": str(data_f680_assignment["id"])}, follow=True)
+    assert response.status_code == 200
+    assert (
+        response.redirect_chain[-1][0]
+        == f"/queues/{data_queue['id']}/cases/{data_submitted_f680_case['case']['id']}/f680/"
+    )
```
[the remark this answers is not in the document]
markj0hnst0n: I did it this way just to keep consistency with the other tests and not add any further complexity.

caseworker/queues/views/case_assignments.py
markj0hnst0n: removed now 🙂

exporter/f680/application_sections/conftest.py
markj0hnst0n: Just reverted this back to the formatting in dev.  Not sure why it ever changed tbh

caseworker/queues/views/case_assignments.py
markj0hnst0n: Removed

caseworker/cases/helpers/case.py
```diff
@@ -168,10 +168,10 @@ def get_goods_summary(self):
...
     def get_destination_countries(self):
         destination_countries = set()
-        all_parties = self.case.data["ultimate_end_users"] + self.case.data["third_parties"]
-        if self.case.data["end_user"]:
+        all_parties = self.case.data.get("ultimate_end_users", {}) + self.case.data.get("third_parties", {})
+        if self.case.data.get("end_user", {}):
             all_parties.append(self.case.data["end_user"])
-        if self.case.data["consignee"]:
+        if self.case.data.get("consignee", {}):
             all_parties.append(self.case.data["consignee"])
```
markj0hnst0n, line 175: With F680 case this produced `KeyError`s so I've created a default for them.  Something resolve in future when thinking about parties on the case.

caseworker/cases/helpers/case.py
```diff
@@ -259,3 +259,16 @@ def get(self, request, *args, **kwargs):
...
+    destinations = {
+        "f680_clearance": {"url": "cases:f680:details", "kwargs": {"queue_pk": queue_id, "pk": case.id}},
+        "standard": {"url": "cases:case", "kwargs": {"queue_pk": queue_id, "pk": case.id, "tab": "details"}},
+    }
+
+    case_type = case.case_type["sub_type"]["key"]
+    url_name = destinations[case_type]["url"]
+    kwargs = destinations[case_type]["kwargs"]
+
+    return reverse(url_name, kwargs=kwargs)
```
currycoder, line 274: Could there be a unit test to prove this works as expected?  Would hope to see it in `caseworker/cases/helpers/tests/test_case.py`

markj0hnst0n: Added now 🙂

caseworker/cases/helpers/case.py
```diff
@@ -261,13 +261,13 @@ def get(self, request, *args, **kwargs):
         return render(request, "case/case.html", self.get_context())
 
 
-def get_case_detail_url(case, queue_id):
+def get_case_detail_url(case_id, case_type, queue_id):
```
currycoder, line 264: If you're going to change the contract a little bit - which seems fine - I would suggest `case_type` is just the string rather than a dictionary.  Let the caller get the correct key out of whatever it has and avoid leaking the API contract over to this function.

markj0hnst0n: Should be sorted now if I caught your drift right!

currycoder (approved):

PR #2378, opened 2025-02-26
[LTD-5949] F680 move case forward button

currycoder:
### Aim

This PR adds a rudimentary "Move case forward" button when a case is on a team queue.

This also introduces a success message with a link; in order to allow the user to see where the case went after they have moved it forward.

[LTD-5949](https://uktrade.atlassian.net/browse/LTD-5949)

[LTD-5949]: https://uktrade.atlassian.net/browse/LTD-5949?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/caseworker/f680/views.py
+++ b/caseworker/f680/views.py
@@ -41,7 +41,10 @@ class MoveCaseForward(LoginRequiredMixin, View):
         queue_pk = str(queue_pk)
         case_pk = str(pk)
         case = get_case(request, case_pk)
-        if not rules.test_rule("can_user_move_case_forward", self.request, case):
+        user_can_move_case = rules.test_rule("can_user_move_case_forward", self.request, case)
+        user_can_modify_f680 = rules.test_rule("can_user_modify_f680", self.request)
+        permission_granted = user_can_move_case and user_can_modify_f680
+        if not permission_granted:
             raise PermissionDenied("Cannot move case forward")
 
         move_case_forward(request, case_pk, queue_pk)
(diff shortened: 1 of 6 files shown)


caseworker/f680/views.py
```diff
@@ -25,3 +32,24 @@ def get_context_data(self, **kwargs):
             self.case["data"]["submitted_by"] = " ".join([submitted_by["first_name"], submitted_by["last_name"]])
         context_data["case"] = self.case
         return context_data
+
+
+class MoveCaseForward(LoginRequiredMixin, View):
```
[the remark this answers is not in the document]
currycoder: Yes our approach on the caseworker side right now is to not use a feature flag.  We may want to revisit this, but for now this approach seems OK.

caseworker/templates/includes/messages.html
```diff
@@ -2,7 +2,11 @@
 	<div class="govuk-width-container app-snackbar app-snackbar--success {% if message.tags %}app-snackbar--{{ message.tags }}{% endif %}">
 		<div class="app-snackbar__content">
 			<div class="app-snackbar__icon"></div>
+            {% if "safe" in message.tags %}
+			{{ message|safe }}
```
Tllew, line 6: what's the reasoning for this?

Tllew (approved):
approving assuming the mark_safe() is implemented

caseworker/templates/includes/messages.html
currycoder: As discussed on slack, have re-instated this and used `mark_safe` at the point of setting the success message

PR #2379, opened 2025-02-26
LTD-5948: Add option to make approval recommendation for F680 application

saruniitr:
## Change description

Add option to make approval recommendation for F680.
This is based on recommendation flow for SIEL where user can either approve/refuse all. Currently only approve option is presented.

When making recommendation we pull approval text (picklists) corresponding to the team. User can optionally enter provisos etc and submit recommendation. We then show the given recommendation details.

Currently any user can make a recommendation but this needs to be enclosed in a rule which will be done in subsequent PRs. This is mainly to limit the size of PR.

The option is add recommendation is not exposed to the user. It can only be accessed using the URL.

[LTD-5948](https://uktrade.atlassian.net/browse/LTD-5948)

[LTD-5948]: https://uktrade.atlassian.net/browse/LTD-5948?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/caseworker/core/rules.py
+++ b/caseworker/core/rules.py
@@ -35,6 +35,8 @@ def get_logged_in_caseworker(request):
 
 @rules.predicate
 def has_available_sub_statuses(request, case):
+    if case.reference != "siel":
+        return False
     return bool(get_case_sub_statuses(request, case["id"]))
 
 
--- a/caseworker/core/rules.py
+++ b/caseworker/core/rules.py
@@ -35,8 +35,6 @@ def get_logged_in_caseworker(request):
 
 @rules.predicate
 def has_available_sub_statuses(request, case):
-    if case.reference != "siel":
-        return False
     return bool(get_case_sub_statuses(request, case["id"]))
 
 
--- a/caseworker/f680/recommendation/forms/forms.py
+++ b/caseworker/f680/recommendation/forms/forms.py
@@ -1,7 +1,15 @@
 from django import forms
 
+from crispy_forms_gds.choices import Choice
 from crispy_forms_gds.helper import FormHelper
 from crispy_forms_gds.layout import Submit
+
+from core.common.forms import BaseForm
+from core.forms.layouts import (
+    ConditionalCheckboxes,
+    ConditionalCheckboxesQuestion,
+    RadioTextArea,
+)
 
 
 class SelectRecommendationTypeForm(forms.Form):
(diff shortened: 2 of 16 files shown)


caseworker/core/rules.py
```diff
@@ -35,6 +35,8 @@ def get_logged_in_caseworker(request):
 
 @rules.predicate
 def has_available_sub_statuses(request, case):
+    if case.reference != "siel":
+        return False
```
saruniitr, line 39: Temporary change till https://github.com/uktrade/lite-api/pull/2431 and the relevant FE change is merged

caseworker/f680/recommendation/forms/forms.py
```diff
@@ -0,0 +1,181 @@
...
+from crispy_forms_gds.layout import Submit
+
+from core.common.forms import BaseForm
+from core.forms.layouts import (
+    ConditionalCheckboxes,
+    ConditionalCheckboxesQuestion,
+    RadioTextArea,
+)
+
+
```
saruniitr, line 14: These forms are created separately for F680 with the idea they can be different following user feedback once they see initial version.

caseworker/f680/recommendation/services.py
```diff
@@ -0,0 +1,54 @@
...
+def get_current_user_recommendation(recommendation, caseworker, team_alias):
+    user_level_recommendation = filter_recommendation_by_level(recommendation, ["user"])
+    user_recommendation = filter_current_user_recommendation(user_level_recommendation, caseworker)
+    user_recommendation = filter_recommendation_by_team(user_recommendation, team_alias)
+    grouped_user_recommendation = group_recommendation_by_user(user_recommendation)
+    return grouped_user_recommendation
+
+
+def post_approval_recommendation(request, case, data, level="user-advice"):
+    json = [
```
saruniitr, line 41: Same here, some of these are same as advice function but unlike SIELs we are only going to create one `Advice` instance for F680 so some of these helper functions will be different.

caseworker/f680/recommendation/forms/forms.py
currycoder: My initial thought looking at this PR was that this was a lot of repetition from what we already have.  But I think I understand the intention here that we are trying to stand up a V1 which is what we already have, but allow us to rapidly and freely change things according to what we want to do for F680s.

I think there's probably a lot of opportunities for us to improve things as we go and perhaps chop away things like `MyRecommendationView`.

caseworker/f680/recommendation/mixins.py
```diff
@@ -0,0 +1,42 @@
...
+        return get_case(self.request, self.case_id)
+
+    @property
+    def caseworker_id(self):
+        return str(self.request.session["lite_api_user_id"])
+
+    @property
+    def caseworker(self):
+        data, _ = get_gov_user(self.request, self.caseworker_id)
+        return data["user"]
```
currycoder, line 28: Could all of these just be set once in setup as attributes on the class like below; https://github.com/uktrade/lite-frontend/pull/2379/files#diff-c611645ede8239940a522c62212e95f4392dacd3838a313892a1ae5b43f895edR32

caseworker/f680/recommendation/views.py
```diff
@@ -0,0 +1,133 @@
...
+)
+from caseworker.f680.recommendation.mixins import CaseContextMixin
+from caseworker.f680.recommendation.services import get_current_user_recommendation, post_approval_recommendation
+from caseworker.queues.services import get_queue
+from core.decorators import expect_status
+from core.wizard.conditionals import C
+from core.wizard.views import BaseSessionWizardView
+
+
+class CaseRecommendationView(LoginRequiredMixin, CaseworkerMixin, TemplateView):
```
currycoder, line 29: I think this could probably use your mixin

caseworker/f680/templates/f680/case/recommendation/recommendation_details.html
```diff
@@ -0,0 +1,60 @@
...
+                    <p class="govuk-body">{{ refusal_note.0.text|linebreaks }}</p>
+                {% else %}
+                    <h2 class="govuk-heading-m">Reason for refusing</h2>
+                    <p class="govuk-body">{{ recommendation.text|linebreaks }}</p>
+                {% endif %}
+
+              </div>
+            </div>
+        </div>
+        {% endif %}
```
currycoder, line 57: I'd like to think about this being a template include.  I think we will shortly be displaying recommendations from all teams on the base recommendation page so would be good to have a unified way to display them

caseworker/f680/templates/f680/case/recommendation/recommendation_details.html
```diff
@@ -0,0 +1,60 @@
...
+    {% if decision == 'approve' or decision == 'proviso' %}
+    <h2 class="govuk-!-font-weight-bold govuk-body govuk-!-padding-2 app-advice-recommendation-heading app-bg-colour--green">
+        {% if team %} Approved by {{ recommendation.team.name }} {% else %} Approved by {{ user|full_name }} {% endif %}
+		<span class="govuk-!-font-weight-regular app-advice-recommendation-heading__additional_details">{{ recommendation.created_at|parse_date|date:"d F Y" }}</span>
+    </h2>
+    {% elif decision == 'refuse' %}
+    <h2 class="govuk-heading-m app-bg-colour--red govuk-!-padding-2">
+        {% if team %} Refused by {{ recommendation.team.name }} {% else %} Refused by {{ user|full_name }} {% endif %}
+    </h2>
+    {% endif %}
```
currycoder, line 14: Using one template for all decision types feels quite messy, we could have..
```
{% include "f680/case/recommendation/recommendation_details.html" %}
```
recommendation_details.html:
```
{% if recommendation.decision == "approve" %}
    {% include "f680/case/recommendation/recommendation_detail_approve.html" %}
{% elif recommendation.decision == "refuse" %}
    {% include "f680/case/recommendation/recommendation_detail_refuse.html" %}
{% endif %}
```
This should make things a lot cleaner and easier to follow at the template level.

May be one to think more about when we come to use this to display all team recommendations on the base recommendation page.

currycoder (commented):
Some hopefully fairly quick things to change and some larger things to discuss

caseworker/f680/templates/f680/case/recommendation/recommendation_details.html
saruniitr: yeah that is the plan as I mainly want to get these changes in first as these are required for other tickets.

caseworker/f680/recommendation/mixins.py
saruniitr: this can be done but this needs repeating in multiple views. We could create a `BaseView` with that `setup()` but that will be same as mixin correct?

saruniitr: clarified internally

currycoder (approved):

markj0hnst0n (approved):

depsiatwal (approved):
agree with the other comments here I think the core thing is we can comfortably change things in isolation which this achieves.

PR #2384, opened 2025-02-28
LTD-5951: View all recommendation given for a F680 application

saruniitr:
## Change description

This PR updates the F680 landing page to show the details if the current user already made a recommendation otherwise we provide an option to make recommendation (assuming other rules are satisfied).

It also shows the recommendation given by other teams on this case.

Some of the template code is refactored so that it is easier to include in other templates.

[LTD-5951](https://uktrade.atlassian.net/browse/LTD-5951)

[LTD-5951]: https://uktrade.atlassian.net/browse/LTD-5951?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


caseworker/f680/recommendation/views.py
```diff
@@ -26,10 +29,25 @@
...
 
+    def setup(self, request, *args, **kwargs):
+        super().setup(request, *args, **kwargs)
+        self.title = (
+            f"View recommendation for this case - {self.case.reference_code} - {self.case.organisation['name']}"
+        )
+        self.recommendation = None
+        if recommendation := get_current_user_recommendation(
+            self.case.advice, self.caseworker_id, self.caseworker["team"]["alias"]
+        ):
```
saruniitr, line 40: This is either required in the view or in the template. Doing in the view we have more control and can add tests.
It also keeps template from get convoluted with too many changes.

caseworker/f680/templates/f680/case/recommendation/includes/recommendation_details_approve.html
```diff
@@ -1,23 +1,33 @@
...
-            <p class="govuk-body">{{ recommendation.note|linebreaks }}</p>
-            {% endif %}
-
-            {% if recommendation.footnote %}
-            <h2 class="govuk-heading-m">Reporting footnote</h2>
-            <p class="govuk-body">{{ recommendation.footnote|linebreaks }}</p>
-            {% endif %}
+<table id="table-licenceable-products" class="govuk-table advice-table">
+    <h2 class="govuk-!-font-weight-bold govuk-body govuk-!-padding-2 app-advice-recommendation-heading app-bg-colour--green">
+        {% if team %} Approved by {{ recommendation.team.name }} {% else %} Approved by {{ user|full_name }} {% endif %}
```
saruniitr, line 3: Most of the template changes are refactoring and not much of functional changes

caseworker/f680/recommendation/views.py
```diff
@@ -26,10 +29,25 @@
 class CaseRecommendationView(LoginRequiredMixin, CaseContextMixin, TemplateView):
```
depsiatwal, line 29: for these views do we not want to move to the new caseworker: namespace model 

Appreciate there more use here then we would like from old views. 
https://github.com/uktrade/lite-api/pull/2435

saruniitr: @depsiatwal I don't completely understand this.
This is already under `caseworker/f680` similar to other view.

caseworker/f680/tests/conftest.py
```diff
@@ -89,6 +89,17 @@ def f680_reference_code():
...
 
+@pytest.fixture
+def f680_cases_to_review():
+    return {
+        "id": "5641aa2b-09ca-47f6-adcf-682b0472bc93",
+        "alias": "F680 Cases to review",
+        "name": "F680 Cases to review",
+        "is_system_queue": True,
+        "countersigning_queue": None,
+    }
```
currycoder, line 100: Suggest this is called `queue_f680_cases_to_review`

caseworker/f680/templatetags/recommendation_tags.py
```diff
@@ -0,0 +1,34 @@
+from collections import defaultdict
+from django import template
+
+
+register = template.Library()
+
+
+@register.inclusion_tag("f680/case/recommendation/all-teams-recommendations.html", takes_context=True)
+def all_teams_recommendation(context):
```
currycoder, line 9: This seems a bit messy. For now, can't we just build a data-structure for this context in the `CaseRecommendationView`?

e.g.
```
team_recommendations = []
for recommendation in case.get("advice", []):
    team_recommendations.append({"team": recommendation.team, "recommendation": recommendation})
```

And then the template would consume `team_recommendations` using the all teams recommendations template you have

caseworker/f680/templatetags/recommendation_tags.py
```diff
@@ -0,0 +1,34 @@
...
+        decisions = {r["type"]["key"] for r in recommendations}
+        if decisions == {"approve"}:
+            decision_string = "has approved"
+        elif decisions == {"proviso"}:
+            decision_string = "has approved with licence conditions"
+
+        all_teams_recommendations.append(
+            {"team": item, "recommendations": teams_recommendations.get(item, []), "decision": decision_string}
+        )
+    context["teams_recommendations"] = all_teams_recommendations
```
depsiatwal, line 32: agree with above comments I think we can improve on this 

may be a follow up PR to tidy up.  As mentioned some kind of data structure to template tags to shift the complexity

depsiatwal (commented):
I think overall it's good few minor comments. but I feel some of this could come with follow up PRs.

caseworker/f680/recommendation/views.py
depsiatwal: yeah ignore I was getting confused with API

caseworker/f680/templatetags/recommendation_tags.py
saruniitr: we can, added this in preparation for future changes.
I will keep it simple for now.

caseworker/f680/rules.py
```diff
@@ -0,0 +1,25 @@
+import rules
+
+from caseworker.core.rules import is_user_allocated, get_logged_in_caseworker
+from caseworker.f680.recommendation.services import get_current_user_recommendation
+
+
+@rules.predicate
+def has_user_made_recommendation(case, user):
```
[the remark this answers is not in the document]
saruniitr: I think this is good as predicate only, this cannot be a rule as this doesn't represent any action by itself.
This is used in the `can_user_make_f680_recommendation`.

I kept it separate because we might use it other places in future. For now I will just combine.

currycoder (approved):

depsiatwal (approved):
lgtm

PR #2386, opened 2025-03-04
UAT release

kevincarrogan:
UAT release


hnryjmes (commented):
initially I didn't like that there was a [commit](https://github.com/uktrade/lite-frontend/pull/2386/commits/1372672b363cd3faba03ec66ee93d87d7a0cd5b0) which switched on a feature flag by default, but that change seems to have been undone by a later commit

hnryjmes (approved):

PR #2387, opened 2025-03-04
[LTD-5956] F680 navigation

currycoder:
### Aim

Add tab navigation for F680s - "Details" and "Recommendations" for now.  This change also includes adding further rules around the "Make recommendation" and "Generate approval letter" buttons so that these show appropriately when the case is in the correct state.

There's also a couple of refactors around using the common `F680CaseworkerMixin`

[LTD-5956](https://uktrade.atlassian.net/browse/LTD-5956)

[LTD-5956]: https://uktrade.atlassian.net/browse/LTD-5956?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #2388, opened 2025-03-05
LTD-5959: Record final level recommendation on F680

saruniitr:
## Change description

The recommendation given on MOD-ECJU F680 cases under final review is recorded as final level advice.

[LTD-5959](https://uktrade.atlassian.net/browse/LTD-5959)

[LTD-5959]: https://uktrade.atlassian.net/browse/LTD-5959?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


depsiatwal (approved):
good for first iteration, understanding a lot of this is reusing exiting advice. 

I feel though the filter can be improved heavily and refactored without this level of complexity

caseworker/f680/recommendation/services.py
```diff
@@ -4,11 +4,16 @@
 from core import client
 
 
-def filter_current_user_recommendation(all_recommendation, user_id):
+def is_f680_finalise_queue(queue_id, caseworker):
+    team_id = caseworker["team"]["id"]
+    return team_id == constants.MOD_ECJU and queue_id == constants.MOD_ECJU_F680_CASES_UNDER_FINAL_REVIEW
```
currycoder, line 9: I think this is a good situation for a django rule.

I'm trying to get us away from queue-specific rules and towards case status specific ones at the moment, see: https://github.com/uktrade/lite-frontend/pull/2387/files#diff-02569cf32e78bf016fd24a078c83e1988ce518d3d7de5ee416ec44a3dbe05dd9R29-R35

It may be that a further dimension comes in to these rules of current team, but keeping this simple for now.

Could this be a django rule which interrogates case status?

caseworker/f680/recommendation/services.py
```diff
@@ -29,12 +34,18 @@ def group_recommendation_by_user(recommendation):
     return result
 
 
-def get_current_user_recommendation(recommendation, caseworker_id, team_alias):
-    user_level_recommendation = filter_recommendation_by_level(recommendation, ["user"])
-    user_recommendation = filter_current_user_recommendation(user_level_recommendation, caseworker_id)
-    user_recommendation = filter_recommendation_by_team(user_recommendation, team_alias)
-    grouped_user_recommendation = group_recommendation_by_user(user_recommendation)
-    return grouped_user_recommendation.get(caseworker_id)
+def get_current_user_recommendation(queue_id, all_recommendations, caseworker):
```
currycoder, line 37: I'm wondering if "current user recommendation" really make sense?

My understanding is that this drives what is shown at the top of advice landing page.  Could it be better to show the current team's recommendation there?

caseworker/f680/recommendation/services.py
```diff
@@ -29,12 +34,18 @@ def group_recommendation_by_user(recommendation):
...
+    level = ["user"]
+    team_id = caseworker["team"]["id"]
+    caseworker_id = caseworker["id"]
+    if is_f680_finalise_queue(queue_id, caseworker):
+        level = ["final"]
```
currycoder, line 42: I think we should just pass `level` in to this explicitly if we continue using it.

caseworker/f680/recommendation/views.py
```diff
@@ -35,9 +36,7 @@ def setup(self, request, *args, **kwargs):
             f"View recommendation for this case - {self.case.reference_code} - {self.case.organisation['name']}"
         )
         self.recommendation = None
-        if recommendation := get_current_user_recommendation(
-            self.case.advice, self.caseworker_id, self.caseworker["team"]["alias"]
-        ):
+        if recommendation := get_current_user_recommendation(str(self.queue_id), self.case.advice, self.caseworker):
```
currycoder, line 39: As mentioned in slack, this has a lot of overlap with my PR https://github.com/uktrade/lite-frontend/pull/2387

caseworker/f680/recommendation/services.py
saruniitr: I can rename this to 'current user recommendation'.
I am not following about showing current team's recommendation because there can be multiple users in a team and we only show one user recommendation of that team.
This is used on the landing page as well as details page after a user submits their recommendation.
On the landing page we already show all teams recommendation.

caseworker/f680/recommendation/services.py
saruniitr: ok. We can pass the `level` in but which level value depends on the current queue, team. This condition simply moves to caller (view).

caseworker/f680/recommendation/views.py
saruniitr: correct, I will resolve conflicts once that is merged in.

caseworker/f680/recommendation/mixins.py
```diff
@@ -0,0 +1,40 @@
+from caseworker.advice.constants import AdviceLevel
```
currycoder, line 1: I think this can be deleted - I believe this is all covered in `F680CaseworkerMixin`

caseworker/f680/rules.py
```diff
@@ -46,18 +46,27 @@ def f680_case_ready_for_move(request, case):
...
         # TODO: Remove this once we get stop the case going to MOD-ECJU Review and combine
-        if team == "MOD_ECJU":
+        if team["alias"] == "MOD_ECJU":
             return True
 
     return False
 
 
+@rules.predicate
+def case_ready_for_outcome(request, case):
```
currycoder, line 62: This looks duplicated with a function above

caseworker/f680/views.py
```diff
@@ -29,6 +31,9 @@ def dispatch(self, request, *args, **kwargs):
         data, _ = get_gov_user(self.request, self.caseworker_id)
         self.caseworker = data["user"]
 
+        self.case_ready_for_outcome = case_ready_for_outcome(request, self.case)
```
currycoder, line 34: Would suggest dropping this and using `self.recommendation_level` only - it seems like that can be done everywhere this is used.

caseworker/f680/views.py
```diff
@@ -29,6 +31,9 @@ def dispatch(self, request, *args, **kwargs):
...
+        self.recommendation_level = AdviceLevel.FINAL if self.case_ready_for_outcome else AdviceLevel.USER
```
currycoder, line 35: Would recommend re-using the statuses buckets here instead e.g.

```
self.recommendation_level = self.get_recommendation_level(self.case)

...
get_recommendation_level(self, case):
    if case.status in RECOMMENDATION_STATUSES:
        return AdviceLevel.USER
    if case.status in OUTCOME_STATUSES:
        return AdviceLevel.FINAL
    return None
```

saruniitr: ok, for now this is sufficient but we may want to check few more things hence I used the rule.
I will change it and we can use the rule if required in future.

caseworker/f680/recommendation/mixins.py
```diff
@@ -0,0 +1,35 @@
+from caseworker.cases.services import get_case
+from caseworker.queues.services import get_queue
+from caseworker.users.services import get_gov_user
+
+
+class CaseContextMixin:
```
currycoder, line 6: I think this still needs removing as it's been superseded

currycoder (approved):
One straggling comment but otherwise looking good!  Thanks for your patience Arun

caseworker/f680/recommendation/mixins.py
saruniitr: removed file

PR #2395, opened 2025-03-07
LTD-5964 Remove unnecessary options on security grading screen in Entity/Party flow

markj0hnst0n:
### Aim

Security grading form in add entity flow had some unnecessary information recorded so removed this.  I made 'other' as a conditional textbox and took the opportunity to add validation to the conditional textboxes on the assembly/manufacture screen also.

[LTD-5964](https://uktrade.atlassian.net/browse/LTD-5964)

[LTD-5964]: https://uktrade.atlassian.net/browse/LTD-5964?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


exporter/f680/application_sections/user_information/forms.py
```diff
@@ -135,57 +135,30 @@ def get_layout_fields(self):
...
-            Choice("confidential", "Confidential"),
-            Choice("secret", "Secret"),
-            Choice("top-secret", "Top Secret", divider="Or"),
-            Choice("other", "Other"),
-        ),
+        choices=SecurityGrading.security_release_choices,
         label="Select security classification",
         widget=forms.RadioSelect,
     )
-    other_security_classification = forms.CharField(
```
currycoder, line 159: "Other" still seems to be an option here in the above choicefield. But my understanding is that this is the field where they explain what "Other" is.  Do we need this still or should the "Other" choice be dropped?

saruniitr: I think removing this not correct.
Is it that when the user selects 'Other' then this text field is shown conditionally? so it is removed in that sense by default and shown only when other is selected?

markj0hnst0n: This was an oversight on my part.  The designs actually have it as a conditional question so I've added back in

exporter/f680/constants.py
```diff
@@ -13,23 +16,20 @@ class SecurityGrading:
...
-        (SECRET, "Secret"),
-        (TOP_SECRET, "Top-secret"),
-        (OTHER, "Other"),
+        Choice(UNCLASSIFIED, "Unclassified"),
+        Choice(OFFICIAL, "Official"),
+        Choice(OFFICIAL_SENSITIVE, "Official - sensitive"),
+        Choice(RESTRICTED, "Restricted"),
+        Choice(CONFIDENTIAL, "Confidential"),
+        Choice(SECRET, "Secret"),
+        Choice(TOP_SECRET, "Top-secret"),
```
depsiatwal, line 25: the copy for Top-secret , Official - sensitive

doesn't match the copy on the screen which is 
Top Secret / Official-Sensitive

exporter/f680/application_sections/user_information/forms.py
```diff
@@ -140,48 +140,39 @@ def get_layout_fields(self):
...
-        help_text="For example, 27 3 2025",
-    )
 
     def get_layout_fields(self):
         return (
             "prefix",
-            "security_classification",
-            "other_security_classification",
+            ConditionalRadios(
+                "security_classification",
```
depsiatwal, line 168: any reason why we have the choice filed security_classification
and then the below values ?

exporter/f680/constants.py
markj0hnst0n: Amended this copy

exporter/f680/application_sections/user_information/forms.py
markj0hnst0n: Changed this now to add the choices for the ConditionalRadio recursively and use constants for the choices

depsiatwal (approved):
lgtm tested locally

PR #2403, opened 2025-03-11
LTD-5965 Change checkboxes for radio buttons

markj0hnst0n:
### Aim

In the user information section when asked if assembly or manufacture is required change the current checkboxes for radio buttons

[LTD-5965](https://uktrade.atlassian.net/browse/LTD-5965)

[LTD-5965]: https://uktrade.atlassian.net/browse/LTD-5965?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


depsiatwal (commented):
I noticed no tests require updating here

We probably at the stage where we probably need to enhance our tests where we assert the eventual payload to the API.

currycoder:
I think Gurdeep is on to something; https://github.com/uktrade/lite-frontend/blob/dev/exporter/f680/application_sections/user_information/tests/test_views.py#L406-L413

The tests here *do* check the payload sent to the API - and this test is showing that multiple values are still accepted for this radio button input - which is wrong!

That's because only the widget was changed, not the form field itself - which is still multiple choice; https://github.com/uktrade/lite-frontend/blob/dev/exporter/f680/application_sections/user_information/forms.py#L218-L221

depsiatwal:
> @depsiatwal I don't think that's in the remit of this PR but maybe another ticket?

Fair enough but wanted to highlight as it became evident in this ticket. 
Perhaps a separate ticket to remind us to go back to tests and make sure we mock assert the API calls. ???

markj0hnst0n:
@depsiatwal @currycoder  I've updated this one so that the field is no longer multiple choice as I think that was a hole in this one.  As Brendan says there mock API calls are there and have been amended now to reflect the new data type.

currycoder (approved):

PR #2404, opened 2025-03-11
LTD-5967-supporting-documents 

depsiatwal:
https://uktrade.atlassian.net/browse/LTD-5967

Still some WIP 

we still need to complete JSON with doc IDs 
and also allow ability to delete which will require enhancing the endpoint 

New endpoint for API is here 
https://github.com/uktrade/lite-api/pull/2448


exporter/f680/application_sections/supporting_documents/views.py
```diff
@@ -0,0 +1,80 @@
...
+        context_data = super().get_context_data(**kwargs)
+        context_data["back_link_url"] = self.get_success_url()
+        return context_data
+
+    @expect_status(
+        HTTPStatus.OK,
+        "Error updating F680 supporting documents",
+        "Unexpected error updating F680 supporting documents",
+    )
+    def update_supporting_application_documents(self):
```
depsiatwal, line 70: This function is to update the ID in the supporting documents incase we need to reference them in a future date. 
I did consider an incremental update but thought this is easiest to keep them in sync incase another process deletes the document(s). 

Also I have left the payload intact so it's consistent with the other sections. i.e label, type but technically this isn't needed. 

Also another function this will have to to check and confirm that 'supporting documents' have been completed else we could have to make another API call just to check if we have documents

markj0hnst0n (approved):
looks grand

exporter/f680/application_sections/supporting_documents/views.py
currycoder: Yep I like this approach.  Get all the documents from the API and then put them in to the JSON in one shot.

currycoder: One other thing we could do as a follow-up is to raise a validation error of some kind on the application submit API endpoint if we find a disparity between `ApplicationDocument` objects and the JSON here.  What do you think?

depsiatwal: kind of know what you mean but not sure I 100% follow you mean at the end of the submission ?
just trying to figure the scenario where this might be out of sync

currycoder: Reason I suggest application submission is that is the point that the application becomes immutable from an exporter perspective.  So I had it in my head that for things like this where we create an object alongside the JSON, it could be a good time to check that the objects and JSON tally up and fail loudly if not

exporter/f680/application_sections/supporting_documents/views.py
```diff
@@ -0,0 +1,80 @@
...
+        self.supporting_documents, _ = self.get_f680_supporting_documents(self.application_id)
+        section_payload = {
+            "label": self.section_label,
+            "items": [{"id": doc["id"]} for doc in self.supporting_documents["results"]],
```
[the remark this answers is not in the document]
currycoder: Looks like you've added these as flat attributes, I meant more mimicking the pattern that we already have for fields e.g.
https://github.com/uktrade/lite-frontend/blob/dev/exporter/f680/payloads.py#L30-L38

So it could be...

```
"items": [
    {
        "id": "...",
        "fields": [
            {
                    "key": "description",
                    "answer": "my document description",
                    "raw_answer": "my document description",
                    "question": form.get_field_label("description"),
                    "datatype": "string",
            },
            {
                    "key": "file",
                    "answer": file_name,
                    "raw_answer": file_name,
                    "question": form.get_field_label("file"),
                    "datatype": "string",
            }
        ]
    }, ...
]
```

exporter/f680/payloads.py
```diff
@@ -83,3 +83,35 @@ def build(self, section, section_label, application_data, form_dict, item_id=Non
             application_data["sections"] = {section: section_payload}
 
         return {"application": application_data}
+
+
+class F680DictPayloadBuilder(F680PatchPayloadBuilder):
+    def build(self, section, section_label, application_data, dict_data):
```
currycoder, line 89: This looks like a good idea - we have one place for application payload building still, which is really nice.

exporter/templates/f680/supporting_documents/supporting-documents.html
```diff
@@ -0,0 +1,70 @@
...
+				{% endfor %}
+			</tbody>
+		</table>
+
+	{% else %}
+		<div class="lite-information-text">
+			<span class="lite-information-text__icon" aria-hidden="true">!</span>
+			<p class="lite-information-text__text">
+				<span class="govuk-visually-hidden">Information</span>
+				NO Docs
```
currycoder, line 52: Feels a bit https://media.tenor.com/hkPsx1W6v_QAAAAC/indiana-jones-harrison-ford.gif

Could I suggest "No supporting documents"

exporter/templates/f680/supporting_documents/supporting-documents.html
```diff
@@ -0,0 +1,70 @@
+{% extends 'layouts/base.html' %}
+
+{% load additional_documents svg %}
+
+{% block back_link %}
+	<a href="{{back_link_url}}">Back</a>
```
currycoder, line 6: This seems to render a bit weirdly, I think we are missing some govuk link classes?

currycoder (commented):
A couple of minor suggestions, once coverage and tests are passing this looks good to go

exporter/templates/f680/supporting_documents/supporting-documents.html
depsiatwal: done

exporter/templates/f680/supporting_documents/supporting-documents.html
depsiatwal: fixed up

currycoder (approved):

PR #2406, opened 2025-03-12
Delegate out processing of licence type questions to separate apps

kevincarrogan:
### Aim

Delegate out the processing of the application type answer.

This is so that the logic about what to do with the application is in the django app of that application type.

[LTD-5994](https://uktrade.atlassian.net/browse/LTD-5994)

[LTD-5994]: https://uktrade.atlassian.net/browse/LTD-5994?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/exporter/f680/licence_type_processors.py
+++ b/exporter/f680/licence_type_processors.py
@@ -1,0 +1,25 @@
+from django.shortcuts import redirect
+
+from http import HTTPStatus
+
+from core.decorators import expect_status
+
+from exporter.f680.services import post_f680_application
+
+
+class F680LicenceLicenceTypeProcessor:
+    def __init__(self, request):
+        self.request = request
+
+    @expect_status(
+        HTTPStatus.CREATED,
+        "Error creating F680 application",
+        "Unexpected error creating F680 application",
+    )
+    def post_f680_application(self, data):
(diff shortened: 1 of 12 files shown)


exporter/f680/licence_type_processors.py
```diff
@@ -0,0 +1,25 @@
+from django.shortcuts import redirect
+
+from http import HTTPStatus
+
+from core.decorators import expect_status
+
+from exporter.f680.services import post_f680_application
+
+
+class F680LicenceLicenceTypeProcessor:
```
saruniitr, line 10: OGD users mentioned in their feedback to remove all reference to licence in F680 so best to rename this as well.

kevincarrogan: It's called that at the moment as that's what the form refers to it as so I've kept that for now.

At some point I intend to rename that whole form section, and as a result the name of this, so that we no longer call it a "licence type'.

saruniitr (approved):

currycoder (approved):

PR #2418, opened 2025-03-19
LTD-6012 Update exporter journey error messages

markj0hnst0n:
### Aim

Update copy on error messages for exporter journey as there was mostly placeholder copy.  [Required error copy doc here.](https://uktrade.atlassian.net/wiki/spaces/LITE/pages/4696342550/F680+error+messages?force_transition=4b07deb5-7c98-425b-ba8e-bda7092ab9bd) 
 Some copy not required as discussed with UCD so this appears as crossed out on the doc.

Had to add ability to use bespoke error messages for text boxes which appear conditionally in certain steps so updated the `add_required_to_conditional_text_field` method to accept a dictionary which will accommodate this.

[LTD-6012](https://uktrade.atlassian.net/browse/LTD-6012)

[LTD-6012]: https://uktrade.atlassian.net/browse/LTD-6012?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


exporter/f680/application_sections/user_information/forms.py
```diff
@@ -197,52 +208,8 @@ class Layout:
...
-        help_text="",
+        error_messages={"required": "Enter how the end-user will use the item"},
     )
 
     def get_layout_fields(self):
         return ("end_user_intended_end_use",)
-
-
-class EndUserAssembleManufactureForm(BaseForm):
-    class Layout:
```
markj0hnst0n, line 208: The form removed here wasn't used as the step was removed in [this](https://uktrade.atlassian.net/browse/LTD-6007) ticket

exporter/f680/application_sections/approval_details/forms.py
```diff
@@ -76,9 +79,23 @@ class ApprovalTypeChoices(TextChoices):
...
+            "required": "Enter details about what you're seeking approval to do",
+        },
     )
 
+    def clean(self):
+        required_fields = ["demonstration_in_uk", "demonstration_overseas"]
+        cleaned_data = super().clean()
+        approval_choices = cleaned_data.get("approval_choices", [])
+
+        for choice in approval_choices:
```
currycoder, line 92: To make this kind to a future developer maintaining this, could you perhaps put a comment explaining that we can't put `required=True` directly on these fields as the field only becomes required when the paired checkbox is ticked

currycoder: What stops us using `add_required_to_conditional_text_field()`?

exporter/f680/application_sections/user_information/forms.py
currycoder: Nice spot!

exporter/f680/application_sections/approval_details/forms.py
markj0hnst0n: I've added in more comments to try and make things a bit clearer in terms of the use of `required=False`.

`add_required_to_conditional_text_field()` isn't used because the `parent_field_response` parameter only takes a single answer and the response in this case is a multiple choice field and hence is sent as a list.  It would mean having to iterate over `required_fields` and the `parent_field_response` so I thought this was cleaner.

core/common/forms.py
```diff
@@ -51,12 +51,19 @@ def __init__(self, *args, **kwargs):
     def get_title(self):
         return self.Layout.TITLE
 
-    def add_required_to_conditional_text_field(self, parent_field, parent_field_response, required_field):
+    def add_required_to_conditional_text_field(self, field_data):
```
kevincarrogan, line 54: Is there a reason to have swapped this out for a dict instead of using keyword arguments with defaults?

markj0hnst0n: It probably would have made more sense to use kwargs but this is just a function of how the method evolved.  I've changed it now.  I've only used a default for the error message as the other field information is integral to the point of using ti.

kevincarrogan (approved):

PR #2419, opened 2025-03-19
[LTD-5993] Change F680 JSON contract for fields to be dict

currycoder:
### Aim

Change the F680 application JSON contract so that `fields` is a dictionary as opposed to an array.  This allows us to have a much easier time with DRF serializer validation and additionally makes lookups much easier.

Companion API PR: https://github.com/uktrade/lite-api/pull/2462

[LTD-5993](https://uktrade.atlassian.net/browse/LTD-5993)

[LTD-5993]: https://uktrade.atlassian.net/browse/LTD-5993?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/exporter/f680/application_sections/conftest.py
+++ b/exporter/f680/application_sections/conftest.py
@@ -204,7 +204,10 @@ def mock_f680_application_get_existing_data(requests_mock, data_f680_case):
                         "key": "controlled_info",
                         "answer": "It just is",
                         "raw_answer": "It just is",
-                        "question": "Explain how the technology or information is controlled.Include countries classification levels and reference numbers. You can upload supporting documents later in your application",
+                        "question": (
+                            "Explain how the technology or information is controlled."
+                            "Include countries classification levels and reference numbers. You can upload supporting documents later in your application"
+                        ),
                         "datatype": "string",
                     },
                     "controlled_information": {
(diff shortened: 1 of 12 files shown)


marcuspp (approved):

exporter/f680/application_sections/conftest.py
```diff
@@ -84,219 +84,253 @@ def mock_f680_application_get_existing_data(requests_mock, data_f680_case):
...
                         "key": "used_by_uk_armed_forces_info",
                         "answer": "",
                         "raw_answer": "",
                         "question": "Explain how it will be used",
                         "datatype": "string",
                     },
-                ],
+                },
                 "type": "single",
+                "fields_sequence": [
```
depsiatwal, line 306: what's the purpose of fields_sequence ? and where is it defined

currycoder: Good question.  Fields sequence is needed because JSON objects/dictionaries don't guarantee ordering.  So before we had a list of fields; the order of that list reflected the order that the questions were asked in the forms.  Now we need this additional datastructure to tell us the ordering - we then use that on the caseworker side when displaying these questions/answers on the case detail page.

depsiatwal (approved):
lgtm

PR #2420, opened 2025-03-20
Allow the exporter licence type question to be skipped based on feature flag

kevincarrogan:
### Aim

Allow the export licence type question to be skipped for certain organisations.

This doesn't affect the end payload in that it will still post a payload that represents a SIEL but this prepares for when we start removing that.

[LTD-6029](https://uktrade.atlassian.net/browse/LTD-6029)

[LTD-6029]: https://uktrade.atlassian.net/browse/LTD-6029?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


exporter/applications/views/conditionals.py
```diff
@@ -0,0 +1,5 @@
+from django.conf import settings
+
+
+def is_indeterminate_export_licence_type_allowed(wizard):
+    return wizard.get_organisation() in settings.FEATURE_FLAG_INDETERMINATE_EXPORT_LICENCE_TYPE_ALLOWED_ORGANISATIONS
```
markj0hnst0n, line 5: Is there a reason for using this instead of Django rules?

kevincarrogan: Good call. I have now changed this.

saruniitr (approved):

conf/exporter.py
```diff
@@ -132,3 +132,6 @@
 CSP_REPORT_URI = env.tuple("EXPORTER_CSP_REPORT_URI", default=("",))
 E2E_WAIT_MULTIPLIER = env.int("E2E_WAIT_MULTIPLIER", default=1)
 FEATURE_FLAG_F680_ALLOWED_ORGANISATIONS = env.list("FEATURE_FLAG_F680_ALLOWED_ORGANISATIONS", default=[])
+FEATURE_FLAG_INDETERMINATE_EXPORT_LICENCE_TYPE_ALLOWED_ORGANISATIONS = env.list(
+    "FEATURE_FLAG_INDETERMINATE_EXPORT_LICENCE_TYPE_ALLOWED_ORGANISATIONS", default=[]
+)
```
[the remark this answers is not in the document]
kevincarrogan: I'm going to keep it as it is for now.

The reason I wasn't using "skip" or "ignore" is that I didn't want it tied specifically to being about skipping that one question and wanted to convey it's more about being able to apply for a licence that isn't specific.

I'll have a think about better wording and come up with something that aligns to how we are describing what we're doing overall.

PR #2427, opened 2025-03-24
LTD-6051: Add option to clear recommendations

saruniitr:
## Change description

Once user makes recommendation then they have an option to clear it.

Add side panel with case details.

[LTD-6051]: https://uktrade.atlassian.net/browse/LTD-6051?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/caseworker/f680/recommendation/services.py
+++ b/caseworker/f680/recommendation/services.py
@@ -81,5 +81,9 @@ def post_recommendation(request, case, data):
         for item in data
     ]
     response = client.post(request, f"/caseworker/f680/{case['id']}/recommendation/", json)
-    response.raise_for_status()
     return response.json(), response.status_code
+
+
+def clear_recommendation(request, case):
+    response = client.delete(request, f"/caseworker/f680/{case['id']}/recommendation/")
+    return None, response.status_code
--- a/caseworker/f680/recommendation/services.py
+++ b/caseworker/f680/recommendation/services.py
@@ -51,7 +51,7 @@ def group_recommendations_by_team_and_users(case_recommendations):
     return grouped_result
 
 
-def current_user_recommendations(request, case, caseworker):
+def recommendations_by_current_user(request, case, caseworker):
     team_id = caseworker["team"]["id"]
     caseworker_id = caseworker["id"]
 
--- a/caseworker/f680/rules.py
+++ b/caseworker/f680/rules.py
@@ -4,7 +4,7 @@ from core.constants import CaseStatusEnum
 
 from caseworker.core.rules import is_user_allocated, get_logged_in_caseworker
 from caseworker.f680.recommendation.services import (
-    current_user_recommendations,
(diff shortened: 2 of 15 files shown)


caseworker/f680/rules.py
```diff
@@ -28,6 +28,15 @@ def can_user_make_f680_recommendation(request, case):
     return case["data"]["status"]["key"] in RECOMMENDATION_STATUSES
 
 
+@rules.predicate
+def can_user_clear_f680_recommendation(request, case):
+    user = get_logged_in_caseworker(request)
+    if not user:
+        return False
+
+    return bool(current_user_recommendations(request, case, user))
```
depsiatwal, line 37: what is current_user_recommendations is it the recommendations of the current user ?

caseworker/f680/recommendation/services.py
```diff
@@ -89,5 +89,9 @@ def post_recommendation(request, case, data):
         for item in data
     ]
     response = client.post(request, f"/caseworker/f680/{case['id']}/recommendation/", json)
-    response.raise_for_status()
     return response.json(), response.status_code
+
+
+def clear_recommendation(request, case):
+    response = client.delete(request, f"/caseworker/f680/{case['id']}/recommendation/")
+    return None, response.status_code
```
depsiatwal, line 97: if we always return none 
and the function call is checking for content.none 
then how will we know it's failed ?

Possible to return a suncessfull message from API or may even help if we do _raise_for_status()

saruniitr: This is used with `expect_status()` decorator which checks for the expected status and raises a helpful error message that we provide if that is not the case. `raise_for_status()` interfers with that flow hence not used that here.

caseworker/f680/rules.py
depsiatwal: if this is like a permission possible to call it 
recommendations_owned_by_current_user ? to make it more clear

saruniitr: to clarify, you mean rename `current_user_recommendations ` to `recommendations_owned_by_current_user `?
Actually `current_user_recommendations()` is a service function used in other places also to get list of recommendations of current user.

or you mean to rename `can_user_clear_f680_recommendation` to `recommendations_owned_by_current_user`?
Because we name this based on the action it is like this. Recommendations owned by current user doesn't indicate any action.

marcuspp (approved):

caseworker/f680/rules.py
depsiatwal: I was actually thinking rename current_user_recommendations  to recommendations_owned_by_current_user

depsiatwal (approved):
lgtm

PR #2429, opened 2025-03-25
LTD: Hide recommendations tab if the case status is not appropriate

saruniitr:
## Change description

Users are only allowed to make recommendations if the case is in the correct status otherwise switching to this tab raises an exception.

Recommendation tab is hidden if it is not applicable for the given case status. Add unit tests for this.
Group rules unit tests under different classes.


depsiatwal (approved):
lgtm

markj0hnst0n (approved):

PR #2431, opened 2025-03-26
LTD-6015 Add ability to remove entity in user information summary page

markj0hnst0n:
### Aim

Add ability to remove entity in user information summary page

[LTD-6015](https://uktrade.atlassian.net/browse/LTD-6015)

[LTD-6015]: https://uktrade.atlassian.net/browse/LTD-6015?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/exporter/f680/application_sections/user_information/tests/test_views.py
+++ b/exporter/f680/application_sections/user_information/tests/test_views.py
@@ -853,9 +853,29 @@ class TestUserInformationRemoveEntityView:
         response = authorized_client.get(missing_f680_user_information_remove_entity_url)
         assert response.status_code == 404
 
-    def test_GET_user_information_remove_entity(
+    def test_GET_user_information_remove_entity_not_has_user_entities(
         self,
         authorized_client,
+        mocker,
+        f680_user_information_remove_entity_url,
+        mock_f680_application_get_existing_data,
+        mock_patch_f680_application,
+    ):
+        mocker.patch(
+            "exporter.f680.application_sections.user_information.views.UserInformationRemoveEntityView.has_user_entities",
+            return_value=False,
+        )
+        response = authorized_client.get(f680_user_information_remove_entity_url)
+        assert (
+            mock_patch_f680_application.last_request.json()["application"]["sections"]["user_information"]["items"]
+            == []
+        )
+        assert response.status_code == 302
+
+    def test_GET_user_information_remove_entity_has_user_entities(
+        self,
+        authorized_client,
+        mocker,
         f680_user_information_remove_entity_url,
         mock_f680_application_get_existing_data,
         mock_patch_f680_application,
--- a/exporter/f680/application_sections/user_information/tests/test_views.py
+++ b/exporter/f680/application_sections/user_information/tests/test_views.py
@@ -880,7 +880,7 @@ class TestUserInformationRemoveEntityView:
         data_f680_case,
     ):
         mocker.patch(
-            "exporter.f680.application_sections.user_information.views.UserInformationRemoveEntityView.has_user_entities",
+            "exporter.f680.application_sections.user_information.views.UserInformationRemoveEntityView.has_user_entities_remaining",
             return_value=False,
         )
         response = authorized_client.get(f680_user_information_remove_entity_url)
--- a/exporter/f680/application_sections/user_information/tests/test_views.py
+++ b/exporter/f680/application_sections/user_information/tests/test_views.py
@@ -894,3 +894,4 @@ class TestUserInformationRemoveEntityView:
 
         assert response.status_code == 302
         assert response.url == reverse("f680:summary", kwargs={"pk": data_f680_case["id"]})
+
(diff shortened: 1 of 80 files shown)


exporter/templates/f680/user_information/summary.html
```diff
@@ -14,7 +14,8 @@ <h1 class="govuk-heading-l govuk-!-margin-bottom-2">Tell us about the entity</h1
 				<tr class="govuk-table__row">
 					<th scope="col" class="govuk-table__header">Name</th>
 					<th scope="col" class="govuk-table__header">Entity type</th>
-					<th scope="col" class="govuk-table__header">Actions</th>
+					<th scope="col" class="govuk-table__header"></th>
+					<th scope="col" class="govuk-table__header"></th>
 				</tr>
```
markj0hnst0n, line 19: Adding these table headers maintained the table structure, specifically the break line between the headers and the table data

marcuspp (approved):

exporter/f680/application_sections/user_information/views.py
```diff
@@ -102,3 +102,54 @@ def get_context_data(self, pk, **kwargs):
...
+            application_data["application"].get("sections", {}).get("user_information", {}).get("items", [])
+        )
+        for item in user_information_items:
+            if item["id"] == entity_to_remove_id:
+                user_information_items.remove(item)
+                self.patch_f680_application(application_data)
+
+    def has_user_entities_remaining(self):
+        application, _ = self.get_f680_application(self.application["id"])
+        return application.get("application", {}).get("sections", {}).get("user_information", {}).get("items", [])
```
currycoder, line 142: Funnily enough I did something similar recently on the caseworker side for the outcome flow.  Could I suggest instead of doing a further call to get the application (which is difficult to mock, I see that you mocked this method itself), the patch call should return the updated application so I think you could update `self.application` from that and have it available here.  So you wouldn't need to do the follow-up GET f680 application call.

As for making this happen in the `mock_patch_application` fixture, this pattern should help https://stackoverflow.com/a/67732303

markj0hnst0n: Removed this extra call and changed the tests to accommodate

exporter/f680/application_sections/user_information/views.py
```diff
@@ -102,3 +102,55 @@ def get_context_data(self, pk, **kwargs):
...
+    def remove_application_entity(self, entity_to_remove_id):
+        application_data = self.application
+        user_information_items = (
+            application_data["application"].get("sections", {}).get("user_information", {}).get("items", [])
+        )
+        for item in user_information_items:
+            if item["id"] == entity_to_remove_id:
+                user_information_items.remove(item)
+                application_data, _ = self.patch_f680_application(application_data)
+                self.application = application_data
```
saruniitr, line 140: Do we really need to patch as we iterate?
Is it ok to remove all ids from json first and then call patch at the end? is that an option?

exporter/f680/application_sections/user_information/views.py
```diff
@@ -102,3 +102,55 @@ def get_context_data(self, pk, **kwargs):
...
+        if has_user_entities_remaining:
+            return redirect(
+                reverse(
+                    "f680:user_information:summary",
+                    kwargs={
+                        "pk": self.application_id,
+                    },
+                )
+            )
+        return redirect(reverse("f680:summary", kwargs={"pk": self.application_id}))
```
saruniitr, line 156: Could this be moved the `get_success_url()`?

saruniitr (commented):
added few simple comments

exporter/f680/application_sections/user_information/views.py
markj0hnst0n: Good spot I'll just call the patch at the end after the loop.  We need the whole json object and not just the id to patch so I had to do it that way

exporter/f680/application_sections/user_information/views.py
markj0hnst0n: Done.  Feels much better that way!

exporter/f680/application_sections/user_information/tests/test_views.py
```diff
@@ -742,3 +742,155 @@ def test_POST_to_step_with_required_conditional_validation_error(
...
+        mock_patch_f680_application,
+        data_f680_case,
+    ):
+        response = authorized_client.get(f680_user_information_remove_entity_url)
+
+        assert (
+            mock_patch_f680_application.last_request.json()["application"]["sections"]["user_information"]["items"]
+            == []
+        )
+        assert response.status_code == 302
```
saruniitr, line 874: these two need to be flipped.

exporter/f680/application_sections/user_information/tests/test_views.py
```diff
@@ -742,3 +742,155 @@ def test_POST_to_step_with_required_conditional_validation_error(
...
+        response = authorized_client.get(f680_user_information_remove_entity_url)
+
+        assert (
+            mock_patch_f680_application_no_user_information_items.last_request.json()["application"]["sections"][
+                "user_information"
+            ]["items"]
+            == []
+        )
+
+        assert response.status_code == 302
```
saruniitr, line 895: same here

saruniitr (approved):

PR #2437, opened 2025-03-28
Add ECJU queries to F680s

marcuspp:
### Aim


caseworker/cases/views/queries.py
```diff
@@ -3,32 +3,36 @@
...
         return self.prefix
 
     def form_valid(self, form):
-        data = {
-            "response": form.cleaned_data["reason_for_closing_query"],
-            "responded_by_user": self.lite_user["id"],
-        }
-        put_ecju_query(request=self.request, pk=self.kwargs["pk"], query_pk=self.kwargs["query_pk"], json=data)
-
+        put_ecju_query(
```
marcuspp, line 17: Kept the behaviour the same was whats currently in prod. Not sure if the status code should be checked and an form error raised if it fails

caseworker/f680/templates/f680/case/queries/list.html
```diff
@@ -0,0 +1,101 @@
+{% extends 'f680/case/base.html' %}
+{% load rules svg %}
+{% load crispy_forms_tags crispy_forms_gds %}
+
+{% block details %}
+  <div class="govuk-grid-row">
+  {% test_rule 'can_user_add_an_ejcu_query' request case as can_user_add_an_ejcu_query %}
```
marcuspp, line 7: Is just adding this rule around the send query button sufficient for handling who can add queries. Or should additional checks be added to the view?

caseworker/f680/templates/f680/case/queries/list.html
```diff
@@ -0,0 +1,101 @@
+{% extends 'f680/case/base.html' %}
```
currycoder, line 1: Could it be possible to factor out some of the commonality in this template in to an include which the SIELs template use as well?  Just thinking about keeping this as DRY as is practical.

I realise there are F680-specific URL names in here, but perhaps those could be passed in as context?

caseworker/cases/views/queries.py
currycoder: Yes I think that's a good idea.  Note we have a general pattern for this; https://github.com/uktrade/lite-frontend/blob/cc36a4df0077a9e96fc2fdc25c2addc762977688/caseworker/f680/recommendation/views.py#L106-L112

caseworker/f680/templates/f680/case/queries/list.html
currycoder: It's probably a good idea to reflect the same rule in the view - I think we should start doing that where we can.

caseworker/f680/templates/f680/case/queries/list.html
```diff
@@ -0,0 +1,101 @@
...
+									</span>
+								</div>
+								<div class="app-documents__item-details">
+									<a {% if document.safe == True %}href="{% url 'cases:document' queue_pk=queue.id  pk=case.id file_pk=document.id %}"
+										{% endif %} class="govuk-link--no-visited-state">{{ document.name }}</a>
+									{% if document.description %}
+									<p class="govuk-body">{{ document.description }}</p>
+									{% endif %}
+								</div>
+							</li>
```
saruniitr, line 88: Could some of these be moved to includes? eg this one.

caseworker/f680/tests/test_views.py
```diff
@@ -283,3 +301,101 @@ def test_POST_success(
...
+        self.queue = data_queue
+        self.query = data_ecju_queries_gov_serializer["ecju_queries"][0]
+        self.url = reverse("cases:f680:ecju_queries", kwargs={"queue_pk": self.queue["id"], "pk": self.f680_case_id})
+        self.new_url = reverse(
+            "cases:f680:new_ecju_query", kwargs={"queue_pk": self.queue["id"], "pk": self.f680_case_id}
+        )
+        self.close_url = reverse(
+            "cases:f680:close_ecju_query",
+            kwargs={"queue_pk": self.queue["id"], "pk": self.f680_case_id, "query_pk": self.query["id"]},
+        )
```
saruniitr, line 322: Is this necessary or not possible with existing fixtures?

marcuspp: This is just so each test does not need to call all the fixtures.

caseworker/f680/views.py
```diff
@@ -7,27 +7,37 @@
...
+from caseworker.cases.views.queries import CloseQueryMixin
 
 
-class F680CaseworkerMixin(CaseworkerMixin):
+class F680CaseworkerMixin(UserPassesTestMixin, CaseworkerMixin):
     current_tab = None
 
-    def dispatch(self, request, *args, **kwargs):
+    def test_func(self):
+        return rules.test_rule("can_user_modify_f680", self.request)
```
marcuspp, line 35: This check is now run on all F680 caseworker views. Do we think it should?

caseworker/f680/views.py
```diff
@@ -46,6 +56,7 @@ def get_recommendation_level(self, case):
 
     def get_context_data(self, **kwargs):
         context_data = super().get_context_data(**kwargs)
+        context_data["is_f680"] = True
```
marcuspp, line 59: Not ideal, but part of the the bigger issue with urls and identifying if a case is an f680 or not. Had to go  bool route as used in a loop.

caseworker/f680/views.py
```diff
@@ -117,3 +121,43 @@ class NotesAndTimelineView(LoginRequiredMixin, F680CaseworkerMixin, NotesAndTime
...
+    template_name = "f680/case/queries/new.html"
+    form_class = NewECJUQueryForm
+
+    def form_valid(self, form):
+        data = {"question": form.cleaned_data["question"], "query_type": "ecju_query"}
+        post_ecju_query(self.request, self.case_id, data)
+        return super().form_valid(form)
+
+    def get_success_url(self):
+        return reverse("cases:f680:ecju_queries", kwargs={"pk": self.case_id, "queue_pk": self.queue_id})
```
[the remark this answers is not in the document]
marcuspp: https://uktrade.atlassian.net/browse/LTD-6084

caseworker/f680/views.py
currycoder: I think this is OK for now.  We can review this when we get closer to switching this on for everybody

caseworker/f680/views.py
currycoder: Do you think the redirect pattern we have discussed (ticket https://uktrade.atlassian.net/jira/software/c/projects/LTD/boards/204?selectedIssue=LTD-6075) will negate the need for this?

If so, maybe this could be another jira-linked TODO

currycoder (approved):
A couple of minor questions inline.  Looks good!

saruniitr (approved):

PR #2442, opened 2025-04-02
LTD-6074-f680-approval-letter-generation

depsiatwal:
https://uktrade.atlassian.net/browse/LTD-6074

some minor changes required here as the serializer has changed


currycoder (changes requested):
These changes are unnecessary now, right?  My understanding is that the contract is no longer changing

depsiatwal:
> These changes are unnecessary now, right? My understanding is that the contract is no longer changing

correct all changes revered just logo change now

currycoder:
This is superseded

PR #2445, opened 2025-04-03
Handle additional edge cases for address formatting

kevincarrogan:
### Aim

Stop errors when someone has put the word "country" in the address field itself.

This was failing as there was an `in` lookup that assumed that the variable was always a dict but could have just been a string with the word "country" in it.

This resolves this and any other edge cases I could think of.

Ideally we should have a single format for addresses that everything adheres to instead of having a function that tries to handle a multitude of cases, but for this specific fix I chose just to fix it in this particular template tag.

[LTD-6082](https://uktrade.atlassian.net/browse/LTD-6082)

[LTD-6082]: https://uktrade.atlassian.net/browse/LTD-6082?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core/builtins/custom_tags.py
+++ b/core/builtins/custom_tags.py
@@ -232,37 +232,49 @@ def get_address(data):
     from {'address': {'address_line_1': '10 Downing St', ...}
     or {'address': '10 Downing St ...', 'country': {'name': United Kingdom'}}
     """
-    if data and "address" in data:
-        address = data["address"]
-        country = data.get("country")
+    if not data:
+        return ""
 
-        if "country" in address:
-            country = address.get("country")
+    if not isinstance(data, dict):
+        return ""
 
-        if isinstance(address, str):
-            if country:
-                return address + ", " + country["name"]
-            else:
-                return address
+    if "address" not in data:
+        return ""
 
-        if "address_line_1" in address:
-            address = [
-                address["address_line_1"],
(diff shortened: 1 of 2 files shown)


saruniitr (approved):
looks good to me

core/builtins/custom_tags.py
```diff
@@ -232,37 +232,49 @@ def get_address(data):
...
-    if data and "address" in data:
-        address = data["address"]
-        country = data.get("country")
+    if not data:
+        return ""
 
-        if "country" in address:
-            country = address.get("country")
+    if not isinstance(data, dict):
+        return ""
```
currycoder, line 239: Am I right in thinking that previously this filter could operate on a string or dict and return some output?  It seems like now if the filter operates on a string it returns nothing

kevincarrogan: Not quite.

`data` has always been a dict. I've just done an additional check here just to make sure that if a non-dict is passed through it won't break completely.

What may be a string or a dict is the value that's assigned to the `address` key.

currycoder: Thanks, that clears it up for me

currycoder (approved):

PR #2449, opened 2025-04-07
LTD-6094: Display recommendations tab for all case statuses

saruniitr:
## Change description

It was hidden when the status is submitted as they don't really make recommendations but this behaviour is inconsistent.

Update case details in side panel to include `intended_use`

[LTD-6094](https://uktrade.atlassian.net/browse/LTD-6094)

[LTD-6094]: https://uktrade.atlassian.net/browse/LTD-6094?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/caseworker/f680/templates/f680/case/recommendation/includes/case_details.html
+++ b/caseworker/f680/templates/f680/case/recommendation/includes/case_details.html
@@ -1,63 +1,91 @@
-{% load static advice_tags %}
-
-    {% with product=case.data.product %}
-        <h2 class="govuk-heading-m">Case details</h2>
-        <details class="govuk-details" data-module="govuk-details">
-          <summary class="govuk-details__summary">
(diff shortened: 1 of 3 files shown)


caseworker/f680/templates/f680/case/recommendation/includes/case_details.html
```diff
@@ -1,63 +1,91 @@
-{% load static advice_tags %}
-
-    {% with product=case.data.product %}
-        <h2 class="govuk-heading-m">Case details</h2>
```
saruniitr, line 4: Added intended use and updated the spacing to be 4 spaces to be consistent.

markj0hnst0n (approved):

currycoder (approved):

PR #2451, opened 2025-04-07
[LTD-6070] Add ability to clear F680 outcome records

currycoder:
### Aim

This change adds a "Clear" button to each row of the outcome table - so that MOD-ECJU are able to clear an outcome grouping after they have created it.

[LTD-6070](https://uktrade.atlassian.net/browse/LTD-6070)

[LTD-6070]: https://uktrade.atlassian.net/browse/LTD-6070?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/caseworker/f680/outcome/services.py
+++ b/caseworker/f680/outcome/services.py
@@ -1,3 +1,5 @@
+from json.decoder import JSONDecodeError
+
 from core import client
 
 
@@ -9,6 +11,15 @@ def post_outcome(request, case_id, data):
 def get_outcomes(request, case_id):
     response = client.get(request, f"/caseworker/f680/{case_id}/outcome/")
     return response.json(), response.status_code
+
+
+def delete_outcome(request, case_id, outcome_id):
+    response = client.delete(request, f"/caseworker/f680/{case_id}/outcome/{outcome_id}/")
+    try:
+        body = response.json()
+    except JSONDecodeError:
+        body = None
+    return body, response.status_code
 
 
 def get_hydrated_outcomes(request, case):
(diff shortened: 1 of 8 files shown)


caseworker/f680/outcome/services.py
```diff
@@ -11,6 +13,15 @@ def get_outcomes(request, case_id):
     return response.json(), response.status_code
 
 
+def delete_outcome(request, case_id, outcome_id):
+    response = client.delete(request, f"/caseworker/f680/{case_id}/outcome/{outcome_id}/")
+    try:
+        body = response.json()
+    except JSONDecodeError:
+        body = None
+    return body, response.status_code
```
saruniitr, line 22: Could this just be `return None, response.status_code`?
because API doesn't return any content in the destroy case correct? or is there a case where we get content?

currycoder: I believe we will still get content on an error which is what this is trying to cover.  It did start to persuade me that to make things simpler for our service functions perhaps our API should always respond with a JSON body, but that's for another time.

saruniitr: ok, even in that status code will be different and `expect_status` catches it.

saruniitr (approved):

caseworker/f680/outcome/services.py
markj0hnst0n: This pattern works fine but would it be neater even to get the destroy method in the API to return an empty object?

markj0hnst0n (approved):
Looks good.  Just one comment but it's not a blocker

caseworker/f680/outcome/services.py
currycoder: Yes, I think we should just make all our endpoints return content to avoid these sorts of conditionals.  For now though - since the feeling is this is minor - this will go in as-is.

PR #2453, opened 2025-04-08
LTD-6106-show-leter-generate-view 

depsiatwal:
https://uktrade.atlassian.net/browse/LTD-6106

Ability to generate documents in the UI


caseworker/f680/rules.py
```diff
@@ -96,10 +99,54 @@ def releases_without_outcome(request, case):
     return len(releases_without_outcome) > 0
 
 
+@rules.predicate
+def all_releases_decided(request, case):
+    user = get_logged_in_caseworker(request)
+    if not user:
```
depsiatwal, line 105: since all rules follow this pattern think we need to shift this to decorator now else we will need to create the same test for all new rules.

caseworker/f680/conftest.py
```diff
@@ -103,6 +155,13 @@ def mock_f680_case(f680_case_id, requests_mock, data_submitted_f680_case):
     return requests_mock.get(url=url, json=data_submitted_f680_case)
 
 
+@pytest.fixture
+def mock_f680_case_under_review(f680_case_id, requests_mock, data_submitted_f680_case):
```
currycoder, line 159: Could we call this `mock_f680_case_under_final_review` to avoid ambiguity?

caseworker/f680/document/views.py
```diff
@@ -1,25 +1,31 @@
...
-class AllDocuments(LoginRequiredMixin, F680CaseworkerMixin, FormView):
-    form_class = DocumentGenerationForm
-    template_name = "f680/document/all_documents.html"
-    current_tab = "recommendations"
+class F680DocumentMixin(F680CaseworkerMixin):
+
+    def dispatch(self, *args, **kwargs):
+        dispatch = super().dispatch(*args, **kwargs)
+        if not rules.test_rule("can_user_make_f680_outcome_letter", self.request, self.case):
+            raise Http404()
```
currycoder, line 27: Should this be a 403 instead?

caseworker/f680/document/views.py
```diff
@@ -29,11 +35,35 @@ class AllDocuments(LoginRequiredMixin, F680CaseworkerMixin, FormView):
...
+        f680_letter_templates = self.get_case_letter_templates()
+        if rules.test_rule("can_user_make_approval_f680_outcome_letter", self.request, self.case):
+            allowed_letter_generation.append(OutcomeType.APPROVE)
+        if rules.test_rule("can_user_make_refusal_f680_outcome_letter", self.request, self.case):
+            allowed_letter_generation.append(OutcomeType.REFUSE)
+
+        for lt in f680_letter_templates:
+            for decision in lt["decisions"]:
+                if decision["name"]["key"] in allowed_letter_generation:
+                    filtered_f680_letter_templates.append(lt)
```
currycoder, line 55: Does the endpoint not allow filtering on decision?  I had it in my head we were using django filters

currycoder (commented):
A few comments inline.  Looks very close though

caseworker/f680/document/views.py
saruniitr: Yeah there is filtering on decision but currently it is only allowing filtering by one value.
I will update API to accept multiple values.

saruniitr: updated in https://github.com/uktrade/lite-api/pull/2505

currycoder (approved):

caseworker/f680/document/views.py
```diff
@@ -1,25 +1,33 @@
...
 from .forms import GenerateDocumentForm, DocumentGenerationForm
 
 
-class AllDocuments(LoginRequiredMixin, F680CaseworkerMixin, FormView):
-    form_class = DocumentGenerationForm
-    template_name = "f680/document/all_documents.html"
-    current_tab = "recommendations"
+class F680DocumentMixin(F680CaseworkerMixin):
+
+    def dispatch(self, *args, **kwargs):
```
marcuspp, line 25: Update this to use the `test_func` pattern found in the `F680CaseworkerMixin`  ie. The `MoveCaseForward` view specifies an additional rule

caseworker/f680/document/views.py
```diff
@@ -77,6 +97,15 @@ def generate_document(self, template_id, text):
             },
         )
 
+    def dispatch(self, request, *args, **kwargs):
```
marcuspp, line 100: Avoid using dispatch if possible as this is ran prior to authentication. Move into the `def post` maybe?

caseworker/f680/document/views.py
saruniitr: updated

caseworker/f680/document/views.py
saruniitr: I think this can totally be removed.
API returns the templates that correspond to the outcomes of this case. Also the `template_id` being checked here is actually from the results returned by the func `self.get_case_letter_templates()` itself in base class.
So we are checking its own result against it and it never fails.

caseworker/f680/document/views.py
```diff
@@ -1,25 +1,33 @@
...
 
 
-class AllDocuments(LoginRequiredMixin, F680CaseworkerMixin, FormView):
-    form_class = DocumentGenerationForm
-    template_name = "f680/document/all_documents.html"
-    current_tab = "recommendations"
+class F680DocumentMixin(F680CaseworkerMixin):
+
+    def test_func(self):
+        return rules.test_rule("can_user_make_f680_outcome_letter", self.request, self.case)
```
marcuspp, line 27: This needs to call the parent also

``` 
def test_func(self):
        return all([super().test_func(), rules.test_rule("can_user_make_f680_outcome_letter", self.request, self.case)])
```

marcuspp (approved):

PR #2458, opened 2025-04-09
LTD-6119: Add clear recommendation option on the landing page

saruniitr:
## Change description

Currently the option to clear recommendation is only available on ‘View my recommendation’ page. This is the page user is taken to after making a recommendation. If user navigates away then this option is not available.

[LTD-6119](https://uktrade.atlassian.net/browse/LTD-6119)

[LTD-6119]: https://uktrade.atlassian.net/browse/LTD-6119?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/caseworker/f680/templates/f680/case/recommendation/recommendation.html
+++ b/caseworker/f680/templates/f680/case/recommendation/recommendation.html
@@ -26,6 +26,14 @@
 
         {% if user_recommendations and not pending_recommendations %}
             <h1 class="govuk-heading-xl">View recommendation</h1>
+
+            {% test_rule 'can_user_clear_f680_recommendation' request case as can_user_clear_f680_recommendation %}
+            {% if can_user_clear_f680_recommendation %}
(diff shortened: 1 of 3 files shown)


caseworker/f680/templates/f680/case/recommendation/recommendation.html
```diff
@@ -26,6 +26,14 @@ <h1 class="govuk-heading-xl">Outcomes</h1>
 
         {% if user_recommendations and not pending_recommendations %}
             <h1 class="govuk-heading-xl">View recommendation</h1>
+
+            {% test_rule 'can_user_clear_f680_recommendation' request case as can_user_clear_f680_recommendation %}
```
currycoder, line 30: Looking at the django rule for this, am I right in thinking this only works if a user has existing recommendations and it only clears those existing recommendations?  i.e. if MOD-DI user1 made recommendations on a case, MOD-DI user2 would not be able to clear them?

currycoder (approved):

markj0hnst0n (approved):
Looks grand

PR #2462, opened 2025-04-14
[LTD-6123] F680 finalise flow

currycoder:
### Aim

This change adds a "Finalise and publish to exporter" button to the "all documents" view.  This is wired up with the finalise endpoint (adjusted in PR https://github.com/uktrade/lite-api/pull/2507)

[LTD-6123](https://uktrade.atlassian.net/browse/LTD-6123)

[LTD-6123]: https://uktrade.atlassian.net/browse/LTD-6123?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/caseworker/f680/document/views.py
+++ b/caseworker/f680/document/views.py
@@ -13,14 +13,14 @@ from django.http import Http404
 from core.auth.views import LoginRequiredMixin
 from core.decorators import expect_status
 
-from caseworker.cases.services import get_generated_document_preview, post_generated_document
+from caseworker.cases.services import get_generated_document_preview, post_generated_document, finalise_case
 from caseworker.core.views import handler403
 from caseworker.letter_templates.services import get_letter_templates_list
 
 from caseworker.f680.outcome.services import get_outcomes
 from caseworker.f680.views import F680CaseworkerMixin
 
-from .forms import GenerateDocumentForm, DocumentGenerationForm
+from .forms import GenerateDocumentForm, FinaliseForm
 
 
 class F680DocumentMixin(F680CaseworkerMixin):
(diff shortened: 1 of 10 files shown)


caseworker/f680/document/views.py
```diff
@@ -45,19 +47,33 @@ def get_case_letter_templates(self):
...
-        context_data["back_link_url"] = self.get_success_url()
+        context_data["back_link_url"] = reverse(
+            "cases:f680:recommendation", kwargs={"pk": self.case_id, "queue_pk": self.queue_id}
+        )
         return context_data
 
+    def form_valid(self, form):
+        self.finalise()
+        success_message = "F680 finalised successfully"
+        messages.success(self.request, success_message)
```
saruniitr, line 73: This can be a follow-up, once document is generated we are displaying a message but not updating the document table that document is generated. Should we update the text to 're-generate' or something? or show a link to generated document indicating that it is generated?

saruniitr (approved):
looks straightforward, checked locally and I can see finalise and publish option after generating letter

caseworker/f680/document/views.py
```diff
@@ -94,9 +115,19 @@ def generate_document(self, template_id, text):
...
         )
 
+    def extra_setup(self, request):
+        super().extra_setup(request)
+        template_id = str(self.kwargs["template_id"])
+        # Check to see if the letter template being generated is permissible
+        try:
+            self.template = [t for t in self.get_case_letter_templates() if t["id"] == template_id][0]
+        except IndexError:
+            raise Http404
```
saruniitr, line 129: I think we can remove this because API returns templates based on outcome decisions.
`template_id` that we get here is actually from the results of `self.get_case_letter_templates()` so it will always pass.

caseworker/f680/document/views.py
currycoder: Yes I think we should do both.  Would you mind making this as a follow-up?  I'm not sure what remaining tickets we still have for letter generation

saruniitr: yes, can be a follow-up as mentioned above

exporter/templates/f680/application_detail.html
```diff
@@ -88,7 +88,7 @@ <h1 class="govuk-heading-l">
     {% elif type == 'ecju-queries' %}
         {% include "includes/ecju-queries.html" with object_type="application" case_id=application.id %}
     {% elif type == 'generated-documents' %}
-        {% include "core/ecju-generated-documents.html" %}
+        {% include "core/ecju-generated-documents.html" with case_id=application.id %}
     {% elif type == 'activity' %}
```
markj0hnst0n, line 92: 🤝

markj0hnst0n (approved):

caseworker/f680/document/views.py
currycoder: I'll leave this in for now as I believe it may be covering (a very unlikely) edge case

PR #2464, opened 2025-04-15
[LTD-6144] - Application manifests

marcuspp:
### Aim

The introduction of Application manifests to support different urls needed for each application type. 

[LTD-6144](https://uktrade.atlassian.net/browse/LTD-6144)

[LTD-6144]: https://uktrade.atlassian.net/browse/LTD-6144?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


caseworker/core/services.py
```diff
@@ -89,7 +89,7 @@ def get_permissible_statuses(request, case):
 
     # TODO: Make this list of dis-allowed caseworker-settable statuses driven
     # by the API
-    if case_type == CaseType.APPLICATION.value:
+    if case_type in [CaseType.APPLICATION.value, "security_clearance"]:
```
currycoder: Could we use the enum for `security_clearance` too?

core/application_manifests/helpers.py
```diff
@@ -14,3 +14,9 @@ def get_caseworker_manifest_for_application_reference(reference_code):
     if reference_code and reference_code.startswith("F680"):
         return application_manifest_registry.get_manifest_by_type(ManifestType.CASEWORKER_F680.value)
     return application_manifest_registry.get_manifest_by_type(ManifestType.CASEWORKER_STANDARD_APPLICATION.value)
+
+
+def get_exporter_manifest_for_application_reference(reference_code):
```
currycoder, line 19: could these convenience functions have some big TODO comments on them to explain that this is meant to be temporary and that where we use them, we should ideally have a handle on the case type instead.

currycoder (approved):
Neat.  I see a couple of CI failures but this is otherwise looking good.

core/application_manifests/helpers.py
marcuspp: I considered this but not sure its wise. It could be temporary for a long time. Making the case type available in every call might not be simple. For example the template tag used for the notes and timelines is for `CaseNoteMentions`.

markj0hnst0n (approved):

PR #2471, opened 2025-04-23
LTD-6168-drop-customisation-text 

depsiatwal:
Removes the ability to customise and repeat preview of letter.
Now this allows review initially then the user needs to generate the letter


caseworker/f680/templates/f680/document/preview.html
```diff
@@ -3,12 +3,12 @@
 
 {% block details %}
 	<div class="app-letter-preview__container">
-
-        {% crispy form %}
-
+		<h1 class="govuk-heading-l">{% block title %}Generate document{% endblock %}</h1>
+		<p id="subtitle" class="govuk-body"><strong>This is a preview of the letter.  Check the preview is correct and then hit “Generate”</strong></p>
```
markj0hnst0n, line 7: You could use `TITLE` and `SUBTITLE` in the `BaseForm` Layout for this.

markj0hnst0n: I've updated my comment to include info about the preview too as you can just inject that bit of HTML into the layout.

depsiatwal: > def get_layout_fields(self):
>         return (HTML('<div class="app-letter-preview__page" id="preview">{{ preview|safe }}</div><br>'),)

ok fair enough i've refactored the template to accommodate, agree it's much neater now.

markj0hnst0n (approved):
One comment but not a blocker

saruniitr (approved):

PR #2472, opened 2025-04-23
LTD-6169 add generated documents to caseworker docs tab

markj0hnst0n:
### Aim

WIP

[LTD-6169](https://uktrade.atlassian.net/browse/LTD-6169)

### DIFF
--- a/caseworker/f680/tests/test_views.py
+++ b/caseworker/f680/tests/test_views.py
@@ -514,6 +514,12 @@ def mock_f680_generated_documents_get(requests_mock, f680_case_id, generated_doc
 
 
 @pytest.fixture
+def mock_f680_generated_documents_get_fail(requests_mock, f680_case_id):
+    url = client._build_absolute_uri(f"/cases/{f680_case_id}/documents/")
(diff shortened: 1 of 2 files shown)


caseworker/f680/tests/test_views.py
```diff
@@ -456,6 +457,22 @@ def document_data(data_f680_case):
     ]
 
 
+@pytest.fixture
+def generated_document_data(data_f680_case):
+    return [
+        {
+            "id": str(uuid.uuid4()),  # /PS-IGNORE
```
Tllew, line 464: this ID and the above document_data.id will be different values, I assume that ok right?

if not maybe another helper function to generate the uuid to keep it consistent would work

Tllew (approved):
minor comment, looks good

caseworker/f680/tests/test_views.py
markj0hnst0n: They should be different documents in this case as one is an exporter uploaded doc and the other is an outcome letter generated by the caseworker.

marcuspp (approved):

PR #2477, opened 2025-04-25
LTD-6166-outcome-letter-flow

depsiatwal:
https://uktrade.atlassian.net/browse/LTD-6166

Updates to outcome document generation flow now includes
- Validation when not all letters are generated 
- ability to view generated docs.


caseworker/f680/document/views.py
```diff
@@ -36,13 +36,26 @@ def handle_no_permission(self):
...
     def get_case_letter_templates(self):
         """Return letter templates that correspond to the case outcomes"""
         outcomes, _ = get_outcomes(self.request, self.case.id)
         decisions = list({item["outcome"] for item in outcomes})
         filters = {"case_type": self.case.case_type["sub_type"]["key"], "decision": decisions}
         f680_letter_templates, _ = self.get_letter_templates_list(filters)
-        return f680_letter_templates["results"]
+        return f680_letter_templates
+
+    def get_outcome_templated_documents(self):
```
currycoder, line 55: I think the naming is overly complicated and it's really hard to follow from that.

If I understand correctly, we are getting all the applicable templates for this case; `required_outcome_documents`

caseworker/f680/outcome/services.py
```diff
@@ -52,3 +52,17 @@ def get_releases_with_no_outcome(request, existing_outcomes, case):
...
         remaining_request_ids_without_outcome.add(release_request["id"])
     return remaining_requests_without_outcome, remaining_request_ids_without_outcome
+
+
+def get_outcome_documents(request, case_id):
+    response = client.get(request, f"/caseworker/f680/{case_id}/outcome_document/")
+    return response.json(), response.status_code
+
+
+def get_outcome_documents_templated_list(letter_templates, outcome_docs):
```
currycoder, line 62: This could be called `get_required_outcome_documents()`

caseworker/f680/outcome/services.py
```diff
@@ -52,3 +52,17 @@ def get_releases_with_no_outcome(request, existing_outcomes, case):
...
+    outcome_documents_templated_list = letter_templates.copy()
+    for outcome_documents_template in outcome_documents_templated_list:
+        for outcome_doc in outcome_docs:
+            if outcome_documents_template["id"] == outcome_doc["template"]:
+                outcome_documents_template["document"] = outcome_doc
```
currycoder, line 67: Maybe this could be called `required_outcome_document["generated_document"]`

caseworker/f680/templates/f680/document/all_documents.html
```diff
@@ -10,16 +10,48 @@ <h1 class="govuk-heading-l">{% block title %}Outcome Documents{% endblock %}</h1
...
+                                    Sent
+                                {% elif  outcome_template_document.document %}
+                                    Generated
+                                {% else %}
+                                    Ready
+                                {% endif %}
+                            </td>
                                 <td class="govuk-table__cell">
-                                    <a class="govuk-button govuk-button--primary" href="{% url "cases:f680:document:generate" queue_pk=queue_id pk=case.id template_id=letter_template.id %}">Generate</a>
+                                    {% if not outcome_template_document.document.visable_to_exporter %}
```
currycoder, line 50: `visible_to_exporter` - has the other side of this if statement been tested?

caseworker/f680/outcome/services.py
depsiatwal: same as above

caseworker/f680/outcome/services.py
depsiatwal: done

caseworker/f680/document/views.py
```diff
@@ -60,13 +73,21 @@ def finalise(self):
...
 
     def form_valid(self, form):
+        if self.get_required_outcome_documents_with_no_documents():
+            form.add_error(
+                None,
+                [
+                    "Click generate for all letters. Finalise and publish to exporter can only be done once all letters have been generated."
+                ],
+            )
+            return super().form_invalid(form)
```
saruniitr, line 90: Is this not possible to do this in the `clean()` method?

depsiatwal: yeah I though about that but that would be in the form hence seemed more suitable in the view as we call service / API etc. and all the logic is in the mixin

saruniitr: Actually its just this bit that needs to be in the clean method,
```
        if self.get_required_outcome_documents_with_no_documents():
            form.add_error(
                None,
                [
                    "Click generate for all letters. Finalise and publish to exporter can only be done once all letters have been generated."
                ],
            )
```

Once you add error then `form_valid()` won't be called so below finalise() is not executed.

saruniitr: we discussed internally. Moving to form is not possible as the return value changes of this function changes as the user generates documents. It is just simpler to keep it in the view.

caseworker/f680/templates/f680/document/all_documents.html
currycoder: Ah I see - it has been tested but the mock API response reflects the typo too!  `visable` should be `visible`.  I think only an end to end test would be able to surface this issue.

depsiatwal: fixed typo

PR #2478, opened 2025-04-28
LTD-6162: Convert refusal reasons field to autocomplete

saruniitr:
## Change description

F680 refusal reasons checkboxes are replaced with autocomplete component.
Users enter the criteria code and we save the combined text as `refusal_reasons`.
They also have an option to enter additional text which was possible earlier so the functionality is maintained.

[LTD-6162](https://uktrade.atlassian.net/browse/LTD-6162)

[LTD-6162]: https://uktrade.atlassian.net/browse/LTD-6162?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


caseworker/assets/javascripts/refusal-review-consolidate.js
```diff
@@ -1,10 +1,11 @@
...
+const initDenialReasons = (elementId) => {
+  const denialReasonField = document.getElementById(elementId);
   if (!denialReasonField) return;
   const multiSelector = new MultiSelector(denialReasonField);
   multiSelector.init();
 };
 
-initDenialReasons();
+initDenialReasons("id_denial_reasons");
+initDenialReasons("id_release_request_refusal_reasons-refusal_reasons");
```
currycoder, line 11: Could we do this based on a generic HTML class name instead of tying the autocomplete to specific components via HTML ID?

saruniitr: ok, I will check and update

saruniitr: It is not working with class name for some reason.
By default the widget is coming up with `selectmultiple` class name so I tried with that first. I have also added a custom class name in the form using the widget `attrs` but in both cases it is not working.
Only if we use the `id` of the parent div it is working.

This is how I modified,
```
import MultiSelector from "core/multi-selector";

const initDenialReasons = (className) => {
  const denialReasonFields = document.getElementsByClassName(className);
  if (!denialReasonFields.length == 0) return;

  for (let field of denialReasonFields) {
    const multiSelector = new MultiSelector(field);
    multiSelector.init();
  }
};

initDenialReasons("selectmultiple");

```

saruniitr:
> I'm getting an error when I use `WMD`, `MEND` as a refusal reason. I haven't checked all the refusals but it might just the ones composed of letters due to casing?

@markj0hnst0n @marcuspp fixed the key error issue when the criteria code is different from numbers

caseworker/assets/javascripts/refusal-review-consolidate.js
currycoder: This worked fine for me Arun;
```
import MultiSelector from "core/multi-selector";

const initDenialReasons = (className) => {
  const denialReasonFields = document.getElementsByClassName(className);
  if (denialReasonFields.length == 0) return;

  for (let field of denialReasonFields) {
    let fieldParent = field.parent;
    const multiSelector = new MultiSelector(field);
    multiSelector.init();
  }
};

initDenialReasons("lite-refusal-reasons-autocomplete");
```

Ofc it required setting the CSS class name on the widgets that use this javascript.

saruniitr: I see you are taking the parent, may be thats why it is not working for me.
I will try this.

caseworker/f680/recommendation/tests/test_forms.py
```diff
@@ -113,8 +113,9 @@ def test_entity_conditions_form_valid(data, valid_status, errors):
     ),
 )
 def test_entity_refusal_reasons_form_valid(data, valid_status, errors):
+    denial_reasons = {"1": [("1", "1"), "1a", "1a"], "2": [("2a", "2a")]}
     reasons = [{"id": "1", "display_value": "1", "description": "does not meet criteria"}]
-    form = EntityRefusalReasonsForm(data=data, refusal_reasons=reasons)
+    form = EntityRefusalReasonsForm(data=data, denial_reasons_choices=denial_reasons.items(), refusal_reasons=reasons)
     assert form.is_valid() == valid_status
```
markj0hnst0n, line 119: Might be good to see some of the refusals without integers used in the test data just to give added confidence

markj0hnst0n (approved):
Just one comment but not a blocker

currycoder (approved):

PR #2482, opened 2025-04-28
[LTD-6164] Make validity_start_date initial value evaluate at runtime instead of module-load

currycoder:
### Aim
Make validity_start_date initial value evaluate at runtime instead of module-load

[LTD-6164](https://uktrade.atlassian.net/browse/LTD-6164)

[LTD-6164]: https://uktrade.atlassian.net/browse/LTD-6164?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/caseworker/f680/outcome/forms.py
+++ b/caseworker/f680/outcome/forms.py
@@ -88,7 +88,7 @@ class ApproveOutcomeForm(BaseForm):
         label="Validity start date",
         help_text="For example 25 2 2025",
         require_all_fields=False,
-        initial=timezone.now().date(),
+        initial=timezone.now,
         error_messages={
             "required": "Enter the validity start date",
             "incomplete": "Enter the validity start date",


caseworker/f680/outcome/forms.py
```diff
@@ -88,7 +88,7 @@ class Layout:
         label="Validity start date",
         help_text="For example 25 2 2025",
         require_all_fields=False,
-        initial=timezone.now().date(),
+        initial=timezone.now,
```
depsiatwal, line 91: does this not just change it from datetime Now to just date ?

saruniitr (approved):

caseworker/f680/outcome/forms.py
currycoder: It makes it a callable so that it's evaluated when the form is actually loaded on to a page rather than once when we deploy the application.

depsiatwal: that makes sense

depsiatwal: though the fact we didn't have to change a test lets me question our coverage.

depsiatwal (approved):
lgtm

caseworker/f680/outcome/forms.py
saruniitr: tests use freeze_time so it won't change

PR #2492, opened 2025-05-02
Ltd 6200 allow export licence exporter journey

kevincarrogan:
### Aim


exporter/applications/forms/third_party.py
```diff
@@ -32,10 +31,8 @@ def _third_party_role_form(application, title, button, options, back_url):
     )
 
 
-def third_party_forms(request, application, strings, back_url, clearance_options=None):
+def third_party_forms(request, application, strings, back_url):
     form_options = role_option_list.copy()
-    if application["case_type"]["sub_type"]["key"] != F680:
```
Tllew, line 37: was this just not required?

exporter/applications/helpers/task_lists.py
```diff
@@ -17,57 +13,27 @@
...
-
 def get_application_task_list(request, application, errors=None):
     user_permissions = get_user_permissions(request)
     additional_documents, _ = get_additional_documents(request, application["id"])
-    application_type = application.sub_type
     is_editing, edit_type = get_edit_type(application)
 
     context = {
-        "strings": _get_strings(application_type),
+        "strings": applications.StandardApplicationTaskList,
```
Tllew, line 32: this stuff is where I'm thinking we could do with almost another implementation of an application manifest so that given a new application type we could maybe just implement a tasklist and some tabs and be on the right direction

exporter/applications/views/end_use_details.py
```diff
@@ -27,34 +26,6 @@ def post(self, request, **kwargs):
         return form_view.post(request, **kwargs)
 
     def _get_form_view(self, request, **kwargs):
-        if self.application.sub_type == F680:
```
Tllew, line 30: unused logic?

Tllew (commented):
Really good progress to what we want I think

exporter/applications/helpers/task_lists.py
kevincarrogan: I actually think we'd just get rid of this task list stuff completely.

F680s doesn't use it at all and so I think we'll eventually phase it out on our side.

This is sort of why I've done this to encourage us not to use this pattern but I don't really think we have the time right now to completely refactor this.

This change allows it to keep working but only for SIELs.

exporter/applications/views/end_use_details.py
kevincarrogan: Pretty much everything related to F680s has been re-written so any of these `if` statements related to it aren't used anymore.

exporter/applications/templatetags/show_application_link.py
```diff
@@ -8,14 +8,16 @@
 
 @register.simple_tag
 def show_application_link(application, template="show_application_link.html"):
```
depsiatwal, line 10: Probably sill used in some templates but I remember something about links for application specific being moved to application manifest.

exporter/templates/applications/application.html
```diff
@@ -29,38 +29,36 @@ <h1 class="govuk-heading-l">
...
-					{% endif %}
-					{% test_rule 'can_user_appeal_case' request application as can_user_appeal_case %}
-					{% if can_user_appeal_case %}
-						<a href="{% url 'applications:appeal' application.id %}" id="button-appeal-refusal" class="govuk-button govuk-button--secondary">
-							Appeal refusal decision
-						</a>
-					{% endif %}
+				{%  if not status_is_terminal %}
+					<a href="{% url 'applications:withdraw' application.id %}" id="button-withdraw-application" class="govuk-button govuk-button--secondary">
+						{% lcs 'applications.ApplicationSummaryPage.Buttons.WITHDRAW_ACCESS_BUTTON' %}
```
depsiatwal, line 34: hate lcs would be nice to remove at some point

depsiatwal (approved):
Been through this and all and looks sensible 
Also pulled and tested locally and was manage to submit application successfully.

saruniitr (approved):
looks good to me, takes out lot of unused code.

exporter/applications/rules.py
```diff
@@ -63,6 +63,9 @@ def can_invoke_major_editable(request, application):
 
 @rules.predicate
 def is_organisation_in_indeterminate_export_licence_type_cohort(request):
+    if "*" in settings.FEATURE_FLAG_INDETERMINATE_EXPORT_LICENCE_TYPE_ALLOWED_ORGANISATIONS:
+        return True
```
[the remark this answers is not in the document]
kevincarrogan: This is kind of the same thing that was done for F680s where, for certain environments, you may want to turn it on wholesale but for other environments you just want to specify it for a particular organisation.

Instead of having two different flags I just combined it into one and chose to use the same pattern that Django already uses for things like `ALLOWED_HOSTS`.

Once this should be turned on for everyone this flag would then be removed.

exporter/applications/helpers/task_lists.py
```diff
@@ -17,57 +13,27 @@
...
         "application": application,
-        "application_type": application_type,
         "is_editing": is_editing,
         "edit_type": edit_type,
-        "licence_type": get_application_type_string(application),
+        "licence_type": applications.ApplicationPage.Summary.Licence.STANDARD,
```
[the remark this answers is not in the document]
kevincarrogan: I was really just trying to limit the number of changes I was making so just removed the function and just use the return of that function directly.

PR #2498, opened 2025-05-08
Fix deleting applications that don't have a case sub_type

kevincarrogan:
### Aim

Fix withdraw, surrender and delete views for case that don't have a sub type.

[LTD-6210](https://uktrade.atlassian.net/browse/LTD-6210)

[LTD-6210]: https://uktrade.atlassian.net/browse/LTD-6210?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/exporter/applications/helpers/summaries.py
+++ b/exporter/applications/helpers/summaries.py
@@ -9,7 +9,7 @@ def draft_summary(draft):
     return Summary(
         values={
             strings.applications.ApplicationSummaryPage.REFERENCE_NAME: draft["name"],
-            strings.applications.ApplicationSummaryPage.TYPE: draft.sub_type_value,
+            strings.applications.ApplicationSummaryPage.TYPE: draft.type_reference_value,
             strings.applications.ApplicationSummaryPage.CREATED_AT: draft.created_at,
         },
         classes=["govuk-summary-list--no-border"],
--- a/exporter/applications/helpers/summaries.py
+++ b/exporter/applications/helpers/summaries.py
@@ -24,7 +24,7 @@ def application_summary(application):
         values={
             strings.applications.ApplicationSummaryPage.REFERENCE_NAME: application["name"],
             strings.applications.ApplicationSummaryPage.REFERENCE_CODE: application["reference_code"],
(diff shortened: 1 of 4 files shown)


Tllew (approved):

exporter/applications/helpers/summaries.py
```diff
@@ -9,7 +9,7 @@ def draft_summary(draft):
     return Summary(
         values={
             strings.applications.ApplicationSummaryPage.REFERENCE_NAME: draft["name"],
-            strings.applications.ApplicationSummaryPage.TYPE: draft.sub_type_value,
+            strings.applications.ApplicationSummaryPage.TYPE: draft.type_reference_value,
```
depsiatwal, line 12: really tempted to say can we not just remove these constants

unit_tests/exporter/applications/views/test_surrender_application.py
```diff
@@ -0,0 +1,84 @@
...
+
+from django.urls import reverse
+
+
+@pytest.fixture()
+def surrender_url(application_id):
+    return reverse("applications:surrender", kwargs={"pk": application_id})
+
+
+def parse_summary(summary_element):
```
depsiatwal, line 11: like this wondering if at some point worth shifting to some test helper classes.

depsiatwal (approved):
lgtm

unit_tests/exporter/applications/views/test_surrender_application.py
kevincarrogan: Why helper classes?

depsiatwal: as in a test helper file especially as we seem to be incorporating more and more rendered content (soup) into our tests.

kevincarrogan: I don't think we should have test helper files as we're using pytest and fixtures are literally that.

PR #2499, opened 2025-05-08
Remove checking of case type for final advice permissions

kevincarrogan:
### Aim

Fix issues that were found in testing for the new Export Licence type on the caseworker side.

[LTD-6211](https://uktrade.atlassian.net/browse/LTD-6211)

[LTD-6211]: https://uktrade.atlassian.net/browse/LTD-6211?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Tllew (approved):
Looks good

caseworker/templates/includes/generic-case-row.html
```diff
@@ -1,4 +1,4 @@
-{% if case.case_type.reference.key in 'siel,oiel' %}
+{% if case.case_type.type.key == 'application' %}
```
depsiatwal, line 1: was a bit worried what this value was for F680 but checked and confirmed it's security_clearance as expected.

caseworker/cases/views/main.py
```diff
@@ -180,9 +180,7 @@ def get_advice_additional_context(self):
 
         final_advice = filter_advice_by_level(self.case["advice"], "final")
 
-        if final_advice and advice_helpers.check_user_permitted_to_give_final_advice(
-            self.case.data["case_type"]["sub_type"]["key"], self.permissions
-        ):
+        if final_advice and advice_helpers.check_user_permitted_to_give_final_advice(self.permissions):
             current_advice_level += ["final"]
```
saruniitr, line 184: pretty sure this is related to Advice1.0, may be best to remove this whole function.
git blame also show this is not touched but I am not sure how this is affecting in the current code.

PR #2500, opened 2025-05-08
LTD-6208-f680-documents-tab-incorrect-user 

depsiatwal:
https://uktrade.atlassian.net/browse/LTD-6208

Exporter Supporting documents don't support capturing the user since it uses ApplicationDocument this is currently consistent with SIELS as well. 

For Auto Generated Documents We only highlight that it's auto generated and if it's been sent to exporter similar to Siels. 

https://github.com/uktrade/lite-frontend/blob/dev/caseworker/templates/case/tabs/documents.html


markj0hnst0n (approved):
👌

PR #2501, opened 2025-05-09
LTD-6205: Show unique countries for F680 in queue view

saruniitr:
## Change description

If we have multiple entities in the same country then we are showing them multiple times now, update this to show unique countries sorted by name.

Intended use is also updated to use 'show more' if the text is longer than 200 chars. In case of SIELs it is not working as expected so update to have same behaviour for all case types.

[LTD-6205](https://uktrade.atlassian.net/browse/LTD-6205)

[LTD-6205]: https://uktrade.atlassian.net/browse/LTD-6205?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


caseworker/queues/views/cases.py
```diff
@@ -284,13 +284,26 @@ def _transform_goods(self, case):
             goods_summary["total_value"] += Decimal(good["value"])
         return goods_summary
 
+    def transform_f680_data(self, case):
```
saruniitr, line 287: All of this view assumes only SIEL case types. To avoid if-else in each function this is defined so that we transform all of F680 data in this function.

caseworker/queues/views/cases.py
```diff
@@ -284,13 +284,26 @@ def _transform_goods(self, case):
...
+        if case["case_type"]["reference"]["key"] != "f680":
```
depsiatwal, line 288: make sense for now if we get may more I would suggest a refactor or some mapper function.

caseworker/templates/includes/siel-case-row.html
```diff
@@ -127,13 +127,8 @@
...
 	<td class="govuk-table__cell" data-customiser-key="intended_end_use">
-        <span class="app-intended-end-use">
-        {% if case.intended_end_use %}
-            {{case.intended_end_use|truncatechars:200}}
-        {% endif %}
-        {% if case.intended_end_use|length > 200 %}
-            <a href="{% url 'cases:case' queue.id case.id %}details/#end-use-details" class="govuk-link govuk-link--no-visited-state">See more</a>
-        {% endif %}
+        <span data-max-length="200" class="app-intended-end-use">
+            {{ case.intended_end_use }}
```
depsiatwal, line 131: what gonna happen now when intended_end_use is very long ? 
do we no longer want to truncate ?

saruniitr: if it is longer than 200 chars it will show a 'More...' link, clicking on it expands to show complete text.
Previously it was taking to a different page where we show case details.

depsiatwal (approved):
lgtm

PR #2511, opened 2025-05-15
LTD-6222: F680 specific changes to HCSAT survey

saruniitr:
## Change description

Main change is to include case_type as part of the payload. Content changes include changing the title and help text as per case_type.

Relevant API PR is https://github.com/uktrade/lite-api/pull/2567

[LTD-6222]: https://uktrade.atlassian.net/browse/LTD-6222?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


exporter/applications/views/hcsat.py
```diff
@@ -23,13 +24,30 @@ def get_application_url(self):
...
+        title = "Overall, how would you rate your experience with the 'apply for a standard individual export licence (SIEL)' service today?"
+
+        if self.application["case_type"]["reference"]["key"] == "f680":
+            title = (
+                "Overall, how would you rate your experience with the 'apply for F680 security approval' service today?"
+            )
+
+        kwargs["title"] = title
+
+        return kwargs
```
currycoder, line 49: It seems like we are repeating this method across a couple of classes.

It got me thinking, could we use the new application manifest to return the `service_name`? https://github.com/uktrade/lite-frontend/blob/e8b74d558b4414a813a3ba7a50c3b8f2c4a9cde2/exporter/f680/application_manifest.py#L7

It seems like that is the real thing that varies and it could be passed in to the form to construct this string AND used in the `form_valid` method below to avoid more `if case_type` conditionals

saruniitr: yeah that is better alternative

exporter/applications/application_manifest.py
```diff
@@ -6,4 +6,5 @@
 
 @application_manifest_registry.register(ManifestType.EXPORTER_STANDARD_APPLICATION.value)
 class StandardApplicationExporterManifest(BaseManifest):
+    service_name = "apply for a standard individual export licence (SIEL)"
```
marcuspp, line 9: I think for the api we have been setting a default of None in the `BaseManifest` not sure if we want to do the same here.

